PR #1, opened 2019-05-22
Adds initial CircleCI config

Delerme:
Adds an initial CircleCI config that builds the bundles and deploys them to the
proper s3 location, doing a dry run for branches other than master.


.circleci/config.yml
```diff
@@ -1,15 +1,93 @@
...
-  # Declare a dependency on the welcome-orb
-  welcome: circleci/welcome-orb@0.3.1
-
-# Orchestrate or schedule a set of jobs, see https://circleci.com/docs/2.0/workflows/
+  aws-s3: circleci/aws-s3@1.0.9
+jobs:
+  # generate bundles
+  build:
+    docker:
+      - image: circleci/node:7.10
```
b1lly, line 9: Lets add an item to the backlog to upgrade our node environment, and make sure we circle back to this.

.circleci/config.yml
```diff
@@ -1,15 +1,93 @@
...
+    steps:
+      - checkout
+      # Download and cache dependencies
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+      - run: npm install
+      - save_cache:
```
b1lly, line 20: For my understanding, whats the value in the caching of dependencies? Is this just caching the JSON file or is this caching the actual artifacts that are downloaded?

.circleci/config.yml
```diff
@@ -1,15 +1,93 @@
...
+          at: ~/answers
+      - run:
+          name: Configure AWS
+          command: |
+            mkdir ~/.aws
+            cp .circleci/aws_config ~/.aws/config
+            echo -e $AWS_PROD_PROFILE >> ~/.aws/config
+      - aws-s3/copy:
+          from: dist
+          to: '${AWS_DEPLOY_BUCKET}'
```
b1lly, line 67: Is this going to have name collision issues with existing production S3 artifacts? Aren't there a few buckets? One for production, one for development? or is that not true anymore?

.circleci/config.yml
```diff
@@ -1,15 +1,93 @@
...
+  version: 2
+  build_and_deploy:
     jobs:
-      - welcome/run
\ No newline at end of file
+      - build
+      - hold:
+          type: approval
+          requires:
+            - build
```
b1lly, line 81: Just for education bits since CircleCI is new, can you explain to me what this whole "hold" situation does in regards to the deploy? What does it stop, when does it stop? What does the workflow look like?

For instance, is it the case that it will build the artifacts, but require a manual button to be clicked in order to trigger the upload to S3?

b1lly (commented):
Nice job! Just some clarifying questions, since circleci is new to me as well.

.circleci/config.yml
Delerme: Yea it's caching node_modules using the package file as the key, so the cache is purged when the package file changes. Greatly speeds up npm_install

.circleci/config.yml
Delerme: Yea you have to manually approve the deployment step, similar to teamcity. I figured that made the most sense for now, at least until we have some tests running. Here's a screenshot of what it looks like: https://drive.google.com/open?id=1GFT5oqweQIJJapvTE4MZbuFbhG4-_5S-

Before I approved, the "hold" job has a pause button, and when you click on it, it asks if you want to approve or not.

.circleci/config.yml
Delerme: We only have access to a single folder in a single bucket at the moment, though we could get other buckets and use contexts to deploy to different ones depending on the branch. Not shown here, but the variable in question includes the path as well, so it's <bucket>/answers, so it's not writing to anything we don't own.

.circleci/config.yml
Delerme: Yea for sure, added.

.circleci/config.yml
b1lly: So would the dryrun deploy collide with the production artifact? Thus overriding it? That's not really what we want, right? Like, we wouldn't be able to have a development artifact and a production artifact with this workflow, right? Unless we renamed the files and created explicit build files for a certain branch that were named differently? Or maybe i'm misunderstanding dryrun.

.circleci/config.yml
b1lly: Ok, perfect. And we'd add the tests steps before that as well. Sounds good to me!

.circleci/config.yml
Delerme: Dryrun doesn't actually write the files to S3, it does all the steps up to it though, so when the commit is eventually in master, the real deploy would succeed, and the real files would get overwritten with the newest version. In the future we can also add a step that actually deploys the files from non-master branches into a subfolder for testing purposes, but for now dryrun doesn't upload any artifacts, only commits on master do

b1lly: Oh, so what's the value in a dry-run? Is there an artifact to test after? Or is this just to see if a "branch" passes the same tests as `master` before promoting it? Was that your thinking?

Delerme: Right, to catch any sort of errors with that step before it happens in master. This might be configuration changes (if we mess with what files we copy or something) or issues with aws or credentials or something. Basically the deploy step in master should not fail, and we should be able to test any changes we want to make to it outside of master (notably I haven't run the real deploy step yet because this isn't in master, but I've been able to test my changes against the dryrun)

b1lly: OK, i'm good with this for now, but expect it to change once we figure out the development environment situation and s3 buckets, and branching patterns.

b1lly (approved):

.circleci/config.yml
```diff
@@ -1,15 +1,93 @@
...
+            branches:
+              only: master
+          requires:
+            - hold
+      - deploy_dryrun:
+          filters:
+            branches:
+              ignore: master
+          requires:
+            - hold
```
kevincaffrey, line 93: i think hold should only be on master? no need to hold for PRs, since they currently aren't deployed anywhere, and later if we do deploy them somewhere, it'll be a unique location per PR that we probably want to happen automatically anyway

Delerme: True, working on another PR right now for deploying dev assets and I'll be sure to remove the hold there

PR #2, opened 2019-05-23
Adjust deploy setup

Delerme:
Replace deply & deploy dryrun with deploy_latest, deploy_branch, and
deploy_version. Deploy_latest updates the existing latest version in S3
from master and is cached for an hour, since the library is expected to change
relatively frequently. Deploy_branch deploys development branches with no cache
in subfolders in S3 with the branches name, which will be necessary for testing
PRs against the static test site. Deploy_version uses git version tags to deploy
a specific version to S3 with long-lived cache headers. Versions are considered
final, so we may only utilize them for code freezes in the short term, but will
become more useful in the future.


.circleci/config.yml
```diff
@@ -46,11 +46,11 @@ jobs:
           to: '${AWS_DEPLOY_BUCKET}'
           arguments: |
             --acl public-read \
-            --cache-control "max-age=86400" \
+            --cache-control "max-age=3600" \
```
kevincaffrey, line 49: I'm not sure about this part. I thought we were going to rely on the CDN for caching and trigger cache purges as a stage in CI. If we do that, I imagine the 1 hour cache at the S3 level might conflict in some rare cases where the CDN had just been refreshed in <1 hour, we purge the CDN cache, but then the CDN gets a cached copy from S3. Or does the CDN cache bust?

.circleci/config.yml
```diff
@@ -65,30 +65,66 @@ jobs:
             echo -e $AWS_PROD_PROFILE >> ~/.aws/config
       - aws-s3/copy:
           from: dist
-          to: '${AWS_DEPLOY_BUCKET}'
+          to: '${AWS_DEPLOY_BUCKET}/dev/${CIRCLE_BRANCH}'
+          arguments: |
+            --acl public-read \
+            --cache-control "no-cache" \
+            --recursive \
+            --profile production \
```
kevincaffrey, line 73: what does this profile indicate?

.circleci/config.yml
```diff
@@ -65,30 +65,66 @@ jobs:
...
 workflows:
   version: 2
   build_and_deploy:
     jobs:
-      - build
+      - build:
+          filters:
+            tags:
+              only: /^v.*/
```
kevincaffrey, line 103: I may be misunderstanding what this does, but it looks like we'll only try to build if if the commit is tagged? Won't this break for PR branches?

.circleci/config.yml
```diff
@@ -65,30 +65,66 @@ jobs:
...
           filters:
             branches:
               only: master
           requires:
             - hold
-      - deploy_dryrun:
+      - deploy_branch:
           filters:
             branches:
               ignore: master
```
kevincaffrey, line 120: I imagine this should also ignore tagged versions, which may very well be branched off of master (in the case of point release hotfixes for old versions - though maybe we'll never need that?)

.circleci/config.yml
```diff
@@ -65,30 +65,66 @@ jobs:
...
+          requires:
+            - build
+      - deploy_version:
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
```
kevincaffrey, line 128: maybe i need to read the docs on these filters, since this regex seems like it would include the master branch in ignores?

.circleci/config.yml
```diff
@@ -65,30 +65,66 @@ jobs:
             echo -e $AWS_PROD_PROFILE >> ~/.aws/config
       - aws-s3/copy:
           from: dist
-          to: '${AWS_DEPLOY_BUCKET}'
+          to: '${AWS_DEPLOY_BUCKET}/dev/${CIRCLE_BRANCH}'
```
kevincaffrey, line 68: is there an environment variable to get commit hash instead? i imagine branch names won't always be unique, so there may be some surprising behavior by using branch name here.

.circleci/config.yml
Delerme: This cache header is actually more about local browser caching, which we cannot purge, even through cloudflare. From what I understand, cloudflare also uses this value for its cache, but that's less of a problem because we could always manually purge that. I don't think s3 cares about this value at all, besides sending it along as a header?

.circleci/config.yml
Delerme: Our AWS config isn't a simple user with permissions, instead it's a user with access to roles which in turn have permissions. To use this setup, you have to make a profile (which I called production) that in turn specifies the role id (ARN) to use for this command. It's populated in the "configure aws" command above which gets the ARN from environment variables. So basically we need that line for any command we want to run

.circleci/config.yml
Delerme: Yea it's super misleading, but CircleCI treats branches and tags completely separately. By default jobs are run for all branches and no tags. This filter makes it run for all tags, but does not stop it from running for all branches/commits, so it does both. Down below in deploy_version I have to specifically filter out all branches to have it run for just tags

.circleci/config.yml
Delerme: Similar to above, it's super misleading but jobs don't run for tags by default, you have to add a filter for them if you do, so deploy_branch won't have this step for tags

.circleci/config.yml
Delerme: yup, that's actually intentional. If it included master it would trigger two builds, one for the tag and one for master. Notably it would trigger one for every commit on master, and what we instead want is one for each tag.

.circleci/config.yml
Delerme: There is one for that, but I figured using the branch would be better so that if I'm developing on some branch like feature-33-newcomponent, whenever I push that's the url I can test on, and it updates automatically. Similarly for pull requests if any changes are made and another commit is pushed the PR can maintain its test url. Seems a little nicer to use from the developer standpoint but yea things will definitely conflict if we're not careful. On the bright side I don't think github allows duplicate remote branches so we have some protection there?

.circleci/config.yml
Delerme: I added a section to the CI doc "Deployments and Lifecycles" that might better explain my ideas here

kevincaffrey (approved):

PR #5, opened 2019-05-24
Dist

b1lly:
Remove distribution files from repository.


b1lly:
Wait, i'm confused. If I have two pull requests should they both have been branched from master? I branched this one from my latest branch (domready) which is why its showing the other two commits. I think in the future I probably always want to branch from master.

PR #6, opened 2019-05-28
Adds Jest configuration and test structure

Delerme:
Adds Jest dev depdency for testing, as well as a simple config for it.
Includes a simple example test to demonstrate the desired test structure.
Updated the circleci workflow to run after all buildsteps.


tests/core/storage/storage.js
```diff
@@ -0,0 +1,71 @@
...
+  });
+});
+
+describe('publishing and subscribing', () => {
+  it('publishes to subscribers', () => {
+    const spy = jest.fn();
+    storage.on('update', 'module1', spy);
+    const data = {
+      module1: { key: 'module1value' },
+    }
```
b1lly, line 56: There's a few use cases for pub/sub.

1. Future subscribes (listening on a key that doesn't exist)
2. Subscribe to existing key 
3. Many listeners to same storage key
4. Remove subscribe

I believe you tested both 1 and 4.

b1lly (commented):
This looks great! Just a little note about some test cases to consider. 

Thoughts on adding for the multi subscribers?

tests/core/storage/storage.js
Delerme: added some more tests, let me know if I should add any other ones

tests/core/storage/storage.js
```diff
@@ -0,0 +1,120 @@
...
+    storage.on('update', 'module2', module2Spy);
+    const newData = {
+      module1: { key: 'newmodule1value' },
+    }
+    storage.insert(newData);
+    expect(module1Spy).toHaveBeenCalled();
+    expect(module2Spy).not.toHaveBeenCalled();
+  });
+
+  it('can be unsubscribed from', () => {
```
b1lly: this is subscribing/unsubscribing for futures.

we should also have a use case for sub/unsub from existing keys.

b1lly (commented):
Just one more use case and LGTM!

tests/core/storage/storage.js
Delerme: added

.circleci/config.yml
```diff
@@ -27,6 +27,25 @@ jobs:
           paths:
             - dist
             - .circleci
+  # run the jest unit tests
+  unit_test:
+    docker:
+      - image: circleci/node:7.10
+    working_directory: ~/answers
+    steps:
+      - checkout
```
kevincaffrey, line 36: it would be nice to define everything before npm test as a shared command, since its identical to the steps in build.

something like:

```
commands:
  checkout_and_install_deps:
    steps:
      - checkout
      - restore_cache
      .....
```

Then both the unit_test and build jobs could use that command as their first step.

You could probably also do similar treatment for the various deploy jobs, with some parameters.

https://discuss.circleci.com/t/circleci-2-1-config-overview/26057

Delerme: good catch, made the updates here: https://github.com/yext/answers/pull/7

PR #9, opened 2019-06-03
ui/components: refactor map component slightly for demo cases; suppor…

b1lly:
…t interactive mapbox

Keep in mind this is fast/for demo. I plan on refactoring the interface and some of the implementations at a later time.


Delerme (approved):
We should probably get that linter set up

b1lly:
> We should probably get that linter set up

For sure. 

https://yexttest.atlassian.net/secure/RapidBoard.jspa?rapidView=281&projectKey=WAT&view=planning&selectedIssue=WAT-203&epics=visible

PR #11, opened 2019-06-05
ui/maps: support more configuration options for google maps

b1lly:
Tomorrow i'm going to clean up the data/modeling part. Also share the configuration and create a static map component. So there's a bit of refactoring to do but this is just the first iteration.


Delerme (approved):
sweet

PR #12, opened 2019-06-07
Continued refactor of maps

b1lly:
Maps should be generated from components, instead of template helpers.

To do so, I had to make some slight updates to the rendering of a component. Because of that, I added some basic tests to cover some of the main use cases for rendering.

This change starts to cleanup some of the map stuff, but there's definitely a lot of work to do in regards to the Template Component syntax.

Also note, this is a half-way-there solution, because I still need to refactor the MapBox implementation of the template helper.

This implementation is the beginning of google. Along the way, I learned a bunch of stuff, so I'll continue the refactor.


b1lly:
Bump @Delerme

Delerme (approved):
seems fine for now

PR #14, opened 2019-06-10
Adds initial custom errors and reporter

Delerme:
Adds a small set of custom errors as well the initial error reporter. Use the
error reporter in the core methods.

(WIP, looking for feedback while I finish the backend)


src/core/errors/errors.js
```diff
@@ -0,0 +1,56 @@
...
+    this.boundary = boundary;
+    this.causedBy = causedBy;
+  }
+
+  toJson() {
+    return JSON.stringify(this);
+  }
+}
+
+const errorsMap = {
```
b1lly: Style NIT: We should always declare const at top of scope (e.g. in this case top of file).

src/core/errors/errors.js
```diff
@@ -0,0 +1,56 @@
+export class BaseError extends Error {
```
b1lly: AnswersBaseError ?

src/core/errors/errors.js
b1lly: I think you should add a comment in here about the classification groups and how you think about grouping them between 100-400. That way, if other people read the source they can just know how to easily extend it. Not everyone is going to have a reference to your doc always.

src/core/errors/errors.js
```diff
@@ -0,0 +1,56 @@
...
+let errorsObject = {};
+
+Object.getOwnPropertyNames(errorsMap)
+  .forEach(e => errorsObject[`${e}Error`] = function(message, boundary, causedBy) {
+      return new BaseError(errorsMap[e], message, boundary, causedBy);
+    });
+
+export default errorsObject;
+
+export function Translate(builtinError, boundary) {
```
b1lly: This is somewhat of a factory helper for the errors object, right?

Would it be more convenient to add a static from method on the AnswersBaseError class?

```AnswersBaseError.from(error, ...);```

File: src/core/errors/errors.js [hunk 4]
```diff
@@ -0,0 +1,56 @@
...
+  if (builtinError instanceof EvalError) {
+    return new errorsObject.EvalError(builtinError.message, boundary);
```
b1lly: Wait, isn't this going to lose the call stack of the actual native error? Don't we want to be reporting that?

src/core/errors/errors.js
```diff
@@ -0,0 +1,56 @@
...
+  'Internal': 102,
+  'Range': 103,
+  'Reference': 104,
+  'Syntax': 105,
+  'Type': 106,
+  'URI': 107,
+  'Rendering': 200,
+  'Network': 300,
+  'Storage': 400,
+};
```
b1lly: Let's try to keep it loose to begin. We can always add more codes in the future. Lets start with just the groupings (100, 200, 300, 400).

For each of the groupings, I feel like we should have an explicit class definition for those errors. 

```throw new AnswersRenderError(....);```

Then, at our boundary, we'd have an implementation like so:
```
import { ERRORS } from './errors/errors'
import AnswersErrorReporter from './errors/errorreporter';

catch (e) {
  if (e instanceof ERRORS.AnswersBaseError) {
   AnswersErrorReporter.report(e);
  } else {
    AnswersErrorReporter.report(
        ERRORS.AnswersUnexpectedError.from(e)
     )
  } finally { 
    if (debug) { // Maybe this is uneeded.
      // rethrow native?
    }
  }
}
```
But if an unexpected exception happens, I think it should still report it as a native thing.

File: src/core/errors/errors.js [hunk 6]
```diff
@@ -0,0 +1,56 @@
...
+
+let errorsObject = {};
+
+Object.getOwnPropertyNames(errorsMap)
+  .forEach(e => errorsObject[`${e}Error`] = function(message, boundary, causedBy) {
+      return new BaseError(errorsMap[e], message, boundary, causedBy);
+    });
+
```
b1lly, line 40: This is neat, but I think we should be a little more explicit about class definition. I think it'd help for readability/search ability, and if/when we start adding variety. 

See explicit class definition comments above.

src/core/errors/errorreporter.js
```diff
@@ -0,0 +1,60 @@
...
+  report(err) {
+    let error = err;
+    if (!(error instanceof BaseError)) {
+      error = Translate(error);
+    }
+
+    if (this._isDev) {
+      console.error(error);
+      return;
+    }
```
b1lly, line 15: I don't think he reporter should be the one logging errors. I think whatever is using the reporter should decide what to do with it.

src/core/errors/errorreporter.js
```diff
@@ -0,0 +1,60 @@
...
+      params: {
+        'error': error.toJson(),
+        'version': this.libVersion,
+        'answersKey': this.answersKey,
+      },
+    });
+
+    request.get()
+      // TODO(jdelerme): swallow error instead?
+      .catch(console.err);
```
b1lly, line 32: I'd inspect the error here and see if we want to retry. I'd consider the dev-mode case to potentially throw an AnswersNetworkError(...)

src/core/errors/errorreporter.js
```diff
@@ -0,0 +1,60 @@
...
+  return instance;
+}
+
+export function GetReporter() {
+  if (!instance) {
+    throw new Errors.ReferenceError('Error reporter instance is not initialized!');
+  }
+
+  return instance;
+}
```
b1lly, line 36: Are you trying to accomplish a singleton pattern here because i'm not sure if this is correct?

b1lly (commented):
Nice job dude, just some suggestions.

src/core/errors/errors.js
Delerme: Done

src/core/errors/errors.js
Delerme: Done

src/core/errors/errors.js
Delerme: Done

src/core/errors/errors.js
Delerme: Done

src/core/errors/errors.js
Delerme: Fixed

src/core/errors/errors.js
Delerme: Done

src/core/errors/errors.js
Delerme: Yea I was kind of worried about making a ton of classes but I don't think it matters too much, at least in the short term. Made them into classes.

src/core/errors/errorreporter.js
Delerme: I think it makes more sense for modules to report all errors to the reporter and for the reporter to decide what makes sense based on mode, bundle, etc. Otherwise, every place that handles errors will have to decide based on these factors themselves, which will result in a lot of duplicate logic

src/core/errors/errorreporter.js
Delerme: Yea that probably makes sense. Punting for now though

src/core/errors/errorreporter.js
Delerme: I sort of was, I was mainly trying to avoid every place that uses the reporter needing to know the api and answers key but I think that's a fine requirement for now. Removed.

src/core/errors/errors.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  static from(builtinError, boundary) {
+    const error = new AnswersBasicError(builtinError.message, boundary);
+    error.stack = builtinError.stack;
+    return error;
+  }
+}
+
+export class AnswersBasicError extends AnswersBaserror {
```
b1lly: Typo

AnswersBaseError (here and throughout file)

src/core/core.js
```diff
@@ -66,7 +70,7 @@ export default class Core {
       .then(data => {
         this.storage.insert(data);
       })
-      .catch(error => console.error(error));
+      .catch(e => this._errorReporter.report(e));
```
b1lly: I'm not sure this is how I envisioned it working, more specifically for boundaries. It'd be unfortunate to have to catch and report every error. CERTAINLY, that use case should be supported if we want to be explicit, but I guess we have to make a decision on our error handling practices in the library.

1. Should we choose to always re-throw errors?
2. Or, not even catch unless there is something explicit we want to do?

I guess I was envisioning all the internals would throw errors, and not catch and log them. If they needed to explicitly handle and log them for whatever reason, the practice would be to rethrow. If they can't rethrow because they want to protect the stack for whatever reason, then they should explicitly report it.

That way, at the top level of our app (one level up from the core), we'd have our first boundary that handles reporting exceptions without having to care about that internally.

src/core/errors/errorreporter.js
b1lly: In my mind, the error reporting wasn't about handling them or logging them. It's about reporting them to the server.

To your point, I think that's something we need to decide (see my comment above). What practices do we want to follow? Do we make it part of our "style" or "practice" to always re-throw and always report to catch and report them at the top most level? 

Consider this, an implementer had their own reporting systems and wanted to capture errors included from our library, or wanted to handle the errors from our library a little more deliberately. With the pattern that we have now, doesn't seem possible.

b1lly (commented):
Nice man! Much better ;)

src/core/errors/errorreporter.js
Delerme: Added reporter to window listener, and also return the error from report in case somewhere lower down wants to report and rethrow (will likely be needed for promises, since those errors don't trigger the window error listener

src/core/errors/errors.js
Delerme: Fixed

src/core/core.js
Delerme: Removed for now, though something like this will probably still be needed since the window error listener doesn't catch errors inside promises

src/answers-umd.js
```diff
@@ -77,6 +83,12 @@ class Answers {
       this._onReady.call(this);
     });
 
+    if (!this._isDev && !opts.suppressErrorReports) {
+      this._errorReporter = new ErrorReporter(opts.apiKey, opts.answersKey);
+      window.addEventListener('error', e => this._errorReporter.report(e.error));
+      window.addEventListener('unhandledrejection', e => this._errorReporter.report(e.error));
+    }
```
b1lly, line 90: do you want to use

DOM.on(...) ? 

Probably not necessary though. 

Might be worth using the abstraction incase we ever want to support older browsers. Then we can handle the addEventListener case more easily.

src/core/errors/errors.js
```diff
@@ -0,0 +1,51 @@
...
+ * 4XX errors: Core errors
+ */
+export class AnswersBaseError extends Error {
+  constructor(errorCode, message, boundary, causedBy) {
+    super(message);
+    this.errorCode = errorCode;
+    this.errorMessage = message;
+    this.boundary = boundary;
+    this.causedBy = causedBy;
+    this.reported = false;
```
b1lly, line 15: Is there a reason why we keep track of whether or not it was reported. Do we do stuff based on this status?

b1lly (commented):
Otherwise LGTM :)

src/answers-umd.js
Delerme: What's the incompatibility? The event listener stuff seems to be well supported in older browsers

src/core/errors/errors.js
Delerme: To avoid duplicate reports. Not an issue right now, but if we decide to report lower down and rethrow we don't want to report it again.

b1lly:
LGTM

src/answers-umd.js
b1lly: For IE9 it doesn't support all the events properly, but not worth worrying about it. I just think we should be consistent with our DOM abstraction so that when/if we make changes in the future about how event handlers work (namespacing, removale, etc) it's all contained behind a single interface.

For this case its just interacting with window so I don't see a big deal but in general we should probably always use the DOM abstraction.

Either way though, not a big deal here. I'd heir on consistency here, though.

Whatever you decide.

Delerme: Yea agree in general, I think I'll leave this here since it's a little more clear without the abstraction, but we can always revisit

PR #15, opened 2019-06-13
Adds filter bar component

Delerme:
WIP, waiting on filter query endpoint to work to test E2E, but it's worth taking a look now


src/core/core.js
```diff
@@ -59,9 +59,9 @@ export default class Core {
     });
   }
 
-  verticalSearch(queryString, verticalKey) {
+  verticalSearch(queryString, verticalKey, filter) {
```
b1lly, line 62: Note:

Usually when we start going more than 2 params in a signature, my spidey senses start tingling to start thinking about using an Object as the interface instead. Even if you don't do it now, it's worth starting to think about for the next iteration. Along that train of thinking, I also consider the difference between required arguments vs optional ones. Once theres a combination of those things, I tend to start itching for a refactor.

src/core/search/autocompletedatatransformer.js
```diff
@@ -39,6 +39,29 @@ export default class AutoCompleteDataTransformer {
...
+      highlightedSections.push({
+        label: section.label,
+        results: highlightedResults,
+      });
+    }
+
+    return AutoCompleteDataTransformer.clean('autocomplete.input', {
+      'sections': highlightedSections
+    });
+  }
```
b1lly: We might want to start thinking about a view model here, similar to what I started for the search transformers.

src/core/search/search.js
```diff
@@ -13,7 +13,7 @@ export default class Search {
      * @type {string}
      * @private
      */
-    this._baseUrl = this._isLocal ? 'http://' + window.location.hostname : 'https://liveapi.yext.com';
+    this._baseUrl = !isLocal ? 'http://' + window.location.hostname : 'https://liveapi.yext.com';
```
b1lly: Is this logic right?

src/core/search/autocompletedatatransformer.js
```diff
@@ -39,6 +39,29 @@ export default class AutoCompleteDataTransformer {
...
+          result,
+          AutoCompleteDataTransformer.highlight(value, subStrings)));
+      }
+      highlightedSections.push({
+        label: section.label,
+        results: highlightedResults,
+      });
+    }
+
+    return AutoCompleteDataTransformer.clean('autocomplete.input', {
```
b1lly: Shouldn't the moduleId be dynamic? You might need to bring in the bar key into the transformation as input. The reason being, if you were to append multiple 'FilterPicker' components on the page, the data should be scoped to the FilterPicker bar key provided.

Originally, that was the "quick fix", if you know of a better way that would be great too.

src/core/search/autocompletedatatransformer.js
```diff
@@ -39,6 +39,29 @@ export default class AutoCompleteDataTransformer {
...
+
+        highlightedResults.push(Object.assign(
+          result,
+          AutoCompleteDataTransformer.highlight(value, subStrings)));
+      }
+      highlightedSections.push({
+        label: section.label,
+        results: highlightedResults,
+      });
+    }
```
b1lly: It might be worth creating an actual model for the FilterPicker data. I know it will probably be simple to start, but what do you think about establishing a practice that we have concrete models for all of this data that comes back.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -25,6 +25,11 @@ export default class AutoCompleteComponent extends Component {
   constructor(opts = {}) {
     super(opts)
 
+    /**
+     * Whether autocomplete is simple or filter
+     */
+    this.filterSearch = opts.filterSearch || false
```
b1lly: isFilterPicker? Not sure I like this name.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -25,6 +25,11 @@ export default class AutoCompleteComponent extends Component {
   constructor(opts = {}) {
     super(opts)
 
+    /**
+     * Whether autocomplete is simple or filter
+     */
```
b1lly, line 31: @type {boolean}

src/ui/components/search/autocompletecomponent.js
```diff
@@ -290,5 +303,17 @@ export default class AutoCompleteComponent extends Component {
...
+    if (key === Keys.ENTER) {
+      e.preventDefault();
+
+      const filter = JSON.stringify(sections[this._sectionIndex].results[this._resultIndex].filter);
+
+      this.updateQuery(value);
+      this._onSubmit(value, filter);
+      this.reset();
+      this.close();
+    }
```
b1lly, line 327: I'm not sure we want the keyboard navigation to handle submission. You can remove this here.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -242,10 +251,14 @@ export default class AutoCompleteComponent extends Component {
...
-    this.core.autoComplete(value, this._experienceKey, this._barKey);
+    if (this.filterSearch) {
+      this.core.autoCompleteFilter(value, this._experienceKey, this._barKey);
+    } else {
+      this.core.autoComplete(value, this._experienceKey, this._barKey);
+    }
   }
 
-  handleNavigateResults(key, e) {
+  handleNavigateResults(key, value, e) {
```
b1lly: Navigation doesn't need to handle submission so you can remove this as an input arg.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -200,7 +209,7 @@ export default class AutoCompleteComponent extends Component {
     if (opt_value === undefined) {
       let sections = this._state.get('sections'),
           results = sections[this._sectionIndex].results;
-      opt_value = results[this._resultIndex].shortValue;
+      opt_value = results[this._resultIndex].shortValue || results[this._resultIndex].value;
```
b1lly: The model (or state) thats provided to this component should handle this type of logic. We should make the autocomplete operate off one model. I mentioned creating a concrete model in the data transformation layer; if we did that, this or case would be avoided for instance.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -145,22 +150,26 @@ export default class AutoCompleteComponent extends Component {
...
     DOM.on(queryInput, 'keydown', (e) => {
-     this.handleNavigateResults(e.keyCode, e);
+      this.handleNavigateResults(e.keyCode, queryInput.value, e);
     });
 
     // Allow the user to select a result with the mouse
     DOM.delegate(this._container, '.js-yext-autocomplete-option', 'mousedown', (evt, target) => {
       let data = target.dataset,
-          val = data.value,
+          val = data.short || data.value,
```
b1lly: See model comment below.

src/ui/components/search/filterpicker.js
```diff
@@ -0,0 +1,185 @@
+import Component from '../component';
+import DOM from '../../dom/dom';
+
+export default class FilterPicker extends Component {
```
b1lly, line 4: Do you think its worth having a common base class that this extends?

src/ui/components/search/filterpicker.js
```diff
@@ -0,0 +1,185 @@
...
+     * @type {string} CSS selector
+     */
+    this._inputEl = opts.inputEl || '.js-yext-query';
+
+    /**
+     * The title used, provided to the template as a data point
+     * Optionally provided.
+     * @type {string}
+     */
+    this.title = opts.title || 'Answers Universal Search';
```
b1lly: This probably isn't needed.

src/ui/components/search/filterpicker.js
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * The filter string to use for the provided query
+     * Optionally provided
+     * @type {string}
+     */
+    this.filter = opts.filter || this.getUrlParams().get('filter') || '';
+  }
+
+  static get type() {
+    return 'FilterBar';
```
b1lly: I think FilterPicker is probably a fine name. Might not need to override this.

b1lly (commented):
Main comments are around creating a concrete data model in the data transformation layer, for the FilterPicker component / AutoComplete, and stylistic nits. Otherwise nice job!

b1lly (commented):
Main comments are around creating a concrete data model in the data transformation layer, for the FilterPicker component / AutoComplete, and stylistic nits. Otherwise nice job!

src/core/core.js
Delerme: Yea as we start refactoring core I think it makes sense to have immutable request objects with required and optional keys. Will wait until refactor for this case

src/core/search/autocompletedatatransformer.js
Delerme: Done

src/core/search/autocompletedatatransformer.js
Delerme: Yup, done

src/core/search/autocompletedatatransformer.js
Delerme: Done

src/core/search/search.js
Delerme: No, I was using it for testing so I didn't have to run the api locally. Fixed

src/ui/components/search/autocompletecomponent.js
Delerme: done

src/ui/components/search/autocompletecomponent.js
Delerme: changed

src/ui/components/search/autocompletecomponent.js
Delerme: this case was actually to fix SearchBar, for whatever reason value is no longer set with the old autocomplete endpoint. Will revisit this later, as I think SearchBar needs a data model now

src/ui/components/search/autocompletecomponent.js
Delerme: Done

src/ui/components/search/filterpicker.js
Delerme: Removed

src/ui/components/search/filterpicker.js
Delerme: It seems a little confusing, since a picker is generally a single or multi select dropdown component. FilterBar sound like SearchBar which I think is good because they are very similar behavior-wise

src/ui/components/search/filterpicker.js
Delerme: Not sure, it doesn't seem like we'd get a whole lot from it, most of the shared code here is just boilerplate that many components have, so there's probably a better solution than inheritance (compose a config object or something?)

src/ui/components/search/autocompletecomponent.js
Delerme: I think it makes more sense, otherwise it's super confusing as to why enter even works. Also, this function has access to the data needed whereas putting it in filterpicker.js would require changing the form to send the necessary data.

src/ui/components/search/autocompletecomponent.js
Delerme: As above, I think submission on enter makes more sense here than anywhere else

src/core/models/filtersdata.js
```diff
@@ -0,0 +1,69 @@
+export default class FiltersData {
```
b1lly: Is this FiltersData or AutoCompleteData? 'Cause isn't the highlighting chunk also used for the normal auto complete that isn't Filters?

src/ui/components/search/filterpicker.js
b1lly: It's possible that we'll expand it into both single and multi in the future.

src/ui/components/search/autocompletecomponent.js
b1lly: If that's the case, I still think there should be a separate handler for submission. It shouldn't be part of navigate.

See comment in the event binding.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -145,22 +151,26 @@ export default class AutoCompleteComponent extends Component {
...
+        this.core.autoCompleteFilter(queryInput.value, this._experienceKey, this._barKey);
+      } else {
+        this.core.autoComplete(queryInput.value, this._experienceKey, this._barKey);
+      }
     });
 
     // Allow the user to navigate between the results using the keyboard
     DOM.on(queryInput, 'keydown', (e) => {
-     this.handleNavigateResults(e.keyCode, e);
+      this.handleNavigateResults(e.keyCode, queryInput.value, e);
```
b1lly: this.handleNavigateResults(e.keyCode, e);
this.handleSubmitResult(e.keyCode, queryInput.value);

It's possible to even check the keycode here, instead of within the handlers but I don't feel strongly.

src/ui/components/search/autocompletecomponent.js
b1lly: Whatever is creating the data object that's being rendered into the markup should handle this, no?

src/ui/components/search/autocompletecomponent.js
b1lly: I'd make it a separate handler. See comment above.

src/ui/components/search/filterpicker.js
b1lly: Yeah, but I also think some of the behavior is common between the both, not just the state. 

Like for instance, the browser history logic, and most of the search logic. I think we want to figure out a way to share the code in this regard, whichever direction we decide.

src/ui/components/search/filterpicker.js
```diff
@@ -0,0 +1,185 @@
...
+     * @type {string} CSS selector
+     */
+    this._inputEl = opts.inputEl || '.js-yext-query';
+
+    /**
+     * The title used, provided to the template as a data point
+     * Optionally provided.
+     * @type {string}
+     */
+    this.title = opts.title;
```
b1lly, line 45: Not sure this is used.

src/ui/components/search/filterpicker.js
Delerme: Yea, I guess it's more of the dropdown select vs text input bar with autocomplete. At the very least, I think associating this component with SearchBar is best to avoid confusion, so FilterBar seems better

src/ui/components/search/filterpicker.js
Delerme: Gets rendered above the search bar

src/core/models/filtersdata.js
Delerme: Just filters, but they are similar enough it can be both. Made it a little more generic so now it can be used for both

src/ui/components/search/autocompletecomponent.js
Delerme: done

src/ui/components/search/autocompletecomponent.js
Delerme: yea done

src/ui/components/search/autocompletecomponent.js
Delerme: done

src/ui/components/search/autocompletecomponent.js
Delerme: done

src/core/models/autocompletedata.js
```diff
@@ -0,0 +1,74 @@
...
+      if (j === matchedSubstrings.length - 1 && end < value.length) {
+        highlightedValue += value.slice(end);
+      }
+
+      nextStart = end;
+    }
+
+    return highlightedValue;
+  }
+}
```
b1lly, line 74: Great! We might want to add tests for this. You can probably do it separate commit, if you want.

src/ui/components/search/filterpicker.js
b1lly: Per our discussion, lets leave it as FilterPicker or FilterSelector for now. And we can revisit it later.

b1lly (commented):
LGTM minor comments ;) 

Good shit man.

src/core/models/autocompletedata.js
Delerme: Yea will do separately

Delerme:
Switched the name to FilterPicker. Fun fact, you still need to override even if you want it to default, because the default is actually "Component" (since the type is set in Component's constructor). May want to fix that later, but for now doing the simple override

b1lly:
LGTM

PR #16, opened 2019-06-14
Pass a beta flag from URL params to the API

kevincaffrey:
The beta flag will be used in the short term to preview changes in
search algorithms before they are live for everyone. The backend API may
change behavior when a beta flag is present. By letting the beta flag be
specified in URL params, we can easily share links to let non-technical
users test algorithm changes side by side with the currently live
behavior.

This will be deprecated in the future in favor of more fully featured
experimentation hooks.


Delerme (approved):
nice

PR #30, opened 2019-06-24
Multiple Components

Delerme:
Refactors storage and various other things to support multiple filter components


src/ui/components/componentmanager.js
```diff
@@ -93,12 +95,18 @@ export default class ComponentManager {
       componentManager: this
     }, opts);
 
+    let componentClass = this._componentRegistry[componentType];
+
+    if (!componentClass.duplicatesAllowed) {
```
b1lly: Two nits:

1. Probably should be a function instead of a property
2. Yext style for naming variables/functions is typically:  [verb][noun]

```
if (componentClass.allowDuplicates()) {

}
```

I have mixed feelings about "public properties" in Javascript. They should be used sparingly.

src/ui/components/componentmanager.js
```diff
@@ -93,12 +95,18 @@ export default class ComponentManager {
...
+      this._checkName(componentType, opts.name);
```
b1lly: The name here doesn't describe what it's checking. As someone reading this code, I need to dig into the implementation to see whats going one. As a general coding hygiene, you want to try to make your code a bit more prosaic so that other people reading your code don't need to dive into implementation details. 

Additionally, I think this methods contract/behavior seems unintuitive. The caller of the method should decide whether or not to throw an exception, and the function should just return true or false.

Probably better to do something like this:

```
if (this._isDuplicateName(componentType, opts.name)) {
  // throw error here
}
```

And your implementation:

```
isDuplicateName(componentType, name) {
  if (//duplicate) { 
    return true;
  }
  return false;
}
```

Then you can test it easily, its name and what it does is obvious. The caller decides how to handle the response, and the contract is clear and concise from the top level without having to read the details. The method name describes exactly what its checking for, and the result of the function is expected.

src/ui/components/componentmanager.js
```diff
@@ -117,7 +125,15 @@ export default class ComponentManager {
...
+      throw new AnswersComponentError(
+        `Another component with name ${name} already exists`,
+        componentType);
+    }
+  }
+
   getActiveComponent (type) {
-    return this._activeComponents[type];
+    return this._activeComponents.find(c => c.type === type);
   }
```
b1lly, line 135: Did you test SearchComponent to make sure its working properly? I believe its using this method.

Side question: What if there are two components of the same type? Does this interface break? We might want to consider that use case. I don't think it would be something that happens often but if it is, this certainly wouldn't work properly here.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -153,9 +149,9 @@ export default class AutoCompleteComponent extends Component {
     DOM.on(queryInput, 'focus', () => {
       this.reset();
       if (this.isFilterSearch) {
-        this.core.autoCompleteFilter(queryInput.value, this._verticalKey, this._barKey);
+        this.core.autoCompleteFilter(queryInput.value, this._verticalKey, this._barKey, this.name);
```
b1lly: Maybe we should take in a configuration object now as the interface. That's a lot of arguments.

I'd also like to add tests for these methods if possible.

src/core/core.js
```diff
@@ -73,23 +75,26 @@ export default class Core {
       .query(queryString)
       .then(response => SearchDataTransformer.transform(response, urls))
       .then(data => {
-        this.storage.insert(data);
+        this.storage.insert(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
+        this.storage.insert(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
+        this.storage.insert(StorageKeys.UNIVERSAL_RESULTS, data[StorageKeys.UNIVERSAL_RESULTS]);
```
b1lly, line 81: I mentioned this in an earlier comment, not sure if there was a response. It might have gotten lost since the comment wasn't on a PR.

But, do we want to make a bulk insert interface?

e.g.
```
this.storage.insert({
  [StorageKeys.NAVIGATION]: data[StorageKeys.NAVIGATION],
  [StorageKeys.DIRECT_ANSWER]: data[StorageKeys.DIRECT_ANSWER],
  ....
})
```

And also support a single insert

```
this.storage.insert(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION])
```

I think it's more convenient, and you also get performance gains since object resolution in Javascript is generally slow.

src/core/storage/storage.js
```diff
@@ -1,35 +1,18 @@
...
 
-  init () {
-
-  }
-
-  insert (data) {
-    if (data === undefined) {
-      return;
+  insert (key, data) {
+    if (this._moduleDataContainer[key] === undefined) {
```
b1lly, line 18: If someone doesn't provide a key, or data, how do you want to handle that here?

tests/core/storage/storage.js
```diff
@@ -1,135 +1,90 @@
...
-    storage.insert(newData);
-    expect(storage.getState('module1').key).toBe('module1value');
-    expect(storage.getState('module2').key).toBe('module2value');
-    expect(storage.getState('module3').key).toBe('module3value');
+    storage.insert(StorageKeys.AUTOCOMPLETE, { key: StorageKeys.AUTOCOMPLETE });
+    storage.insert(StorageKeys.NAVIGATION, { key: StorageKeys.NAVIGATION });
+    storage.insert(StorageKeys.DIRECT_ANSWER, { key: StorageKeys.DIRECT_ANSWER });
+    expect(storage.getState(StorageKeys.AUTOCOMPLETE).key).toBe(StorageKeys.AUTOCOMPLETE);
+    expect(storage.getState(StorageKeys.NAVIGATION).key).toBe(StorageKeys.NAVIGATION);
+    expect(storage.getState(StorageKeys.DIRECT_ANSWER).key).toBe(StorageKeys.DIRECT_ANSWER);
```
b1lly: Isn't this the same test as stores data by the provided key?

tests/core/storage/storage.js
```diff
@@ -1,135 +1,90 @@
...
-    };
-    storage.insert(data);
-    storage.on('update', 'module1', spy);
-    const newData = {
-      module1: { key: 'newmodule1value' }
-    };
-    storage.insert(newData);
+    storage.insert(StorageKeys.AUTOCOMPLETE, { key: StorageKeys.AUTOCOMPLETE });
+    storage.on('update', StorageKeys.AUTOCOMPLETE, spy);
+    storage.insert(StorageKeys.AUTOCOMPLETE, { key: StorageKeys.NAVIGATION });
```
b1lly: Just to confirm my understanding, whats the purpose of the key in data? It doesn't have one right? its just data to be tested right? The fact that the data is key: [keyVal] doesn't matter, right? That was a little confusing.

tests/core/storage/storage.js
```diff
@@ -1,135 +1,90 @@
...
-      module1: { key: 'newmodule1value' }
-    };
-    storage.insert(newData);
+    storage.insert(StorageKeys.AUTOCOMPLETE, { key: StorageKeys.AUTOCOMPLETE });
+    storage.insert(StorageKeys.UNIVERSAL_RESULTS, { key: StorageKeys.UNIVERSAL_RESULTS });
+    storage.on('update', StorageKeys.AUTOCOMPLETE, module1Spy);
+    storage.on('update', StorageKeys.UNIVERSAL_RESULTS, module2Spy);
+    storage.insert(StorageKeys.AUTOCOMPLETE, { key: StorageKeys.NAVIGATION });
     expect(module1Spy).toHaveBeenCalled();
     expect(module2Spy).not.toHaveBeenCalled();
```
b1lly: do we want to update these names from module to something different?

README.md
```diff
@@ -79,6 +79,7 @@ Every component has the same base configuration options.
 
 |  option   | type       | description                               | required      |
 |-----------|------------|-------------------------------------------|---------------|
+| name          | string     | a uniqure name, if using multiple components of the same type  | optional      |
```
b1lly: unique

src/core/core.js
```diff
@@ -81,23 +82,34 @@ export default class Core {
       });
   }
 
-  autoComplete (queryString, verticalKey, barKey) {
-    const storageKey = barKey ? `${StorageKeys.AUTOCOMPLETE}.${barKey}` : StorageKeys.AUTOCOMPLETE;
+  autoComplete (queryString, verticalKey, barKey, autoCompleteName) {
```
b1lly: probably time to make this a configuration object. I think 4 arguments is definitely pushing the fold. barKey is optional, right?

src/core/core.js
```diff
@@ -81,23 +82,34 @@ export default class Core {
...
     return this._autoComplete
       .query(queryString, verticalKey, barKey)
       .then(data => {
-        this.storage.insert(storageKey, data);
+        this.storage.insert(`${StorageKeys.AUTOCOMPLETE}.${autoCompleteName}`, data);
       });
   }
 
-  autoCompleteFilter (input, verticalKey, barKey) {
+  autoCompleteFilter (input, verticalKey, barKey, autoCompleteName) {
```
b1lly: same as above.

src/core/core.js
b1lly: You also might want to JSDoc the arguments and purpose. I'm not sure its clear to me what autoCompleteName is from looking at the signature.

src/core/core.js
```diff
@@ -81,23 +82,34 @@ export default class Core {
...
     return this._autoComplete
       .queryFilter(input, verticalKey, barKey)
       .then(data => {
-        this.storage.insert(`${StorageKeys.AUTOCOMPLETE}.${barKey}`, data);
+        this.storage.insert(`${StorageKeys.AUTOCOMPLETE}.${autoCompleteName}`, data);
       });
   }
 
+  applyFilter (pickerName, verticalKey, filter) {
```
b1lly: I actually don't think the core should care about "picker" as a thing. The picker is an implementation detail of the component ui library and not the core.

In other words, i'd do something more like:

```
applyFilter(namespace, verticalKey, filter)
```

Also see my structural message.

src/ui/components/componentmanager.js
```diff
@@ -117,7 +125,15 @@ export default class ComponentManager {
     return component;
   }
 
+  _checkName (componentType, name) {
```
b1lly: See above comment.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -258,9 +254,9 @@ export default class AutoCompleteComponent extends Component {
 
     this.reset();
     if (this.isFilterSearch) {
-      this.core.autoCompleteFilter(value, this._verticalKey, this._barKey);
+      this.core.autoCompleteFilter(value, this._verticalKey, this._barKey, this.name);
     } else {
-      this.core.autoComplete(value, this._verticalKey, this._barKey);
+      this.core.autoComplete(value, this._verticalKey, this._barKey, this.name);
     }
```
b1lly, line 259: This looks like the same exact logic as above in the DOM focus, can we encapsulate this somewhere and just use it in both places?

src/ui/components/search/filterpickercomponent.js
```diff
@@ -153,7 +154,7 @@ export default class FilterPickerComponent extends Component {
       filter: filter
     }, query, '?' + params.toString());
 
-    this.core.verticalSearch('', this._verticalKey, filter);
+    this.core.applyFilter(this.name, this._verticalKey, filter);
```
b1lly: OK, I think we just introduced some dependencies to the search interface, and some wonky side effects that I don't think we actually want here.

The search interface should be simple. 

1. It takes input (query and filters),
2. issues a network request. 

Having it magically insert into storage and pull data out of storage to build the request probably isn't what we want to do here. We want to do that stuff before it gets here and the provide it as input.

src/ui/components/search/filterpickercomponent.js
```diff
@@ -172,14 +173,14 @@ export default class FilterPickerComponent extends Component {
 
   bindBrowserHistory () {
     DOM.on(window, 'popstate', () => {
-      this.query = this.getUrlParams().get('query');
-      this.filter = this.getUrlParams().get('filter');
+      this.query = this.getUrlParams().get(`${this.name}.query`);
```
b1lly, line 185: How come the query is name spaced? Shouldn't there only ever be one query?

b1lly (commented):
Nice job dude.

A few minor style things and there's one main structural thing that I think needs to be addressed.

The structural thing is how the search interface and implementation actually works. We can talk offline if there's any confusion but I think what you have here is confusing and error prone. I don't think the search flow should have a dependency on storage at all. I think the information it needs to issue a network request should be provided to it as input. Doing some magic inside the search flow to set and get the state of filters in order to build the request doesn't seem like the right thing to do.  It's unnecessary complexity and can definitely have some weird side effects and race conditions in the future.

I think all that complexity of updating state and retrieving state of filters should be handled outside of the search path (before it), and instead should be provided to it.

src/ui/components/componentmanager.js
Delerme: done

src/ui/components/componentmanager.js
Delerme: done

src/ui/components/componentmanager.js
Delerme: yea that's why I kept this function at all. Does not work with two components of the same type, but neither did the original implementation. I notice you had a TODO for removing the need for this function, so I think we should close out that TODO and delete this function entirely

src/ui/components/search/autocompletecomponent.js
Delerme: I split out each autocomplete to make it easier to see what's going on (the component now contains the logic for deciding which core method to call). As discussed a config object doesn't help a lot in this case because all of the arguments are required, so the config object makes it harder to reason about the function and see exactly what it needs. Barkey will be deprecated soon which will bring these methods back down to three args as well

src/core/core.js
Delerme: I considered it but I prefer the insertions to be explicit. It's much easier to see what data is getting inserted under which key, instead of hunting down all the keys that have been put on some object. It also opens up room for bugs like accidentally modifying the object and putting data we don't want stored, etc. I can investigate some nicer ways of handling this if it comes up again

src/core/storage/storage.js
Delerme: Added preconditions with error throwing

tests/core/storage/storage.js
Delerme: yea I guess so, removed

tests/core/storage/storage.js
Delerme: Yea I just copied over the tests you had. For clarity though I changed all the test data to avoid this confusion in the future

tests/core/storage/storage.js
Delerme: yea probably, done

README.md
Delerme: done

src/core/core.js
Delerme: see above, added better docs

src/core/core.js
Delerme: done

src/core/core.js
Delerme: done

src/ui/components/componentmanager.js
Delerme: done

src/ui/components/search/autocompletecomponent.js
Delerme: done

src/ui/components/search/filterpickercomponent.js
Delerme: done

src/ui/components/search/filterpickercomponent.js
Delerme: nope, since there can be several filter pickers there can be several text inputs, and each one has a separate query. I.e. box1 has 'software engineer' and box2 has 'tysons, virginia'. We want each one stored individually so if I link to the page each box gets populated with the correct value

src/ui/components/search/autocompletecomponent.js
Delerme: Also in terms of tests for this component, I have to figure out
- loading handlebar templates into jest from source files (you have a todo for this already)
- abstracting out the test environment so we don't need as much setup, and can render our components like normal (will probably use enzyme for this)
- ensuring the full component works with mocked out interfaces

So I'll be working on that now but won't include it in this commit, as there are a lot of unknowns and there's no need for that to block this

b1lly: Testing the component would be ideal, but I agree theres a ton of complexity there. I think there's a ton of logic the component depends on that we can start testing already without having to figure out a lot of those things. For instance, we don't have to necessarily test the component in order to add tests to the core methods (e.g. autoCompleteFilter). I think those would be a good starting point.

src/ui/components/componentmanager.js
b1lly: FWIW I thought having a function that encapsulated the meaning of the expression was a cool thing to have. I was just saying the name could've been improved, and the contract in/out could be adjusted.

Either way thought works.

Delerme: ack

src/ui/components/search/autocompletecomponent.js
Delerme: yea core just calls out to storage which I added tests for, and autocomplete which I will add tests to when I figure out a good way to mock out the api request stuff

src/core/core.js
b1lly: is setFilter or addFilter a better name?

Delerme: already changed to setFilter, unless my push got messed up

b1lly (approved):
LGTM as long as you manually tested both universal, vertical cases (search and auto complete).

PR #32, opened 2019-06-25
ensure pinConfig function doesn't get re-written as pinConfig object

aginsberg:
This is to fix an issue with customizing map pins. When a function was provided for the 'pinConfig' arg, it would get called and set as the new value of 'pinConfig', meaning that in subsequent iterations of the 'markers.forEach' loop, 'pinConfig' would be the object created from the first passthrough. This resulted in every marker having the same attributes, including the numeric label.

This CR is to change it to call the 'pinConfig' function on each iteration of the loop.


benmcginnis (approved):
LGTM, thoughts @b1lly ?

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -161,35 +161,38 @@ class GoogleMapMarkerConfig {
...
+      let pinConfigObj = {};
       if (typeof pinConfig === 'function') {
-        pinConfig = pinConfig(
+        pinConfigObj = pinConfig(
           marker.item,
           MapProvider.DEFAULT_PIN_CONFIG,
           marker);
+      } else {
+        pinConfigObj = pinConfig;
       }
```
b1lly, line 173: NIT: You can drop the else block and just do:
```
let pinConfigObj = pinConfig;
if (typeof pinConfig === 'function') {
  pinConfigObj = pinConfig(...);
}
```

b1lly (commented):
Thanks for identifying and correcting this! Nice job.

One caveat. For the short term, we stopped developing directly off master and instead are developing off a branch called 'latest'. Is it possible to send a PR to latest instead?

Also, would you mind adding a unit test for the GoogleMapMarkerConfig class? We have recently added JEST as a our unit testing framework and started beefing up some test coverage, and as part of that effort we decided that all bugs that are "fixed" should be accompanied by unit tests.

src/ui/components/map/providers/googlemapprovider.js
aginsberg: Got it, I will make this fix, add the test, and open a new PR to the 'latest' branch. Closing this pr

PR #33, opened 2019-06-26
added SearchParams as shim for URLSearchParams and URL searchParams f…

creotutar:
…or ie11


creotutar:
wip

src/ui/dom/SearchParams.js
```diff
@@ -0,0 +1,80 @@
+export default class SearchParams {
```
b1lly: A little nit.

Our code style for file names are always lowercase. So `SearchParams.js` => `searchparams.js`.

Secondly, I know its a work in progress, but would you mind accompanying this class with some unit tests and JS doc comments at the class / function level?

src/ui/dom/SearchParams.js
```diff
@@ -0,0 +1,80 @@
...
+  constructor (url) {
+    this.useNative = false;
+    if (window.URLSearchParams) {
+      this.useNative = true;
+      this.params = new URLSearchParams(url);
```
b1lly: I'm not sure how I feel about the native fallback. Reason being, if it's polyfilled by the users browser already, then we can't guarantee their implementation is what we'd expect.

Aside, in general, storing the *native* params and keeping track of whether or not its native so that every method can behavior two different ways seems weird and excessive. The alternative is to just return the new URLSearchParams(url) here from the constructor. I'm not sure I see value in wrapping the native behavior and relying on the implementation, unless the interface was going to be different. 

But, I also think that's where unit tests will save us.

```
constructor(url) {
  if (window.URLSearchParams) {
    return new URLSearchParams(url);
  }

  // otherwise, custom implementation
}
```

That way all of the methods don't need to be implemented in two ways.

src/core/search/search.js
```diff
@@ -1,9 +1,10 @@
 import HttpRequester from '../http/httprequester';
 import ApiRequest from '../http/apirequest';
+import { SearchParams } from '../../ui/index';
```
b1lly: WOOPS, I totally overlooked this.

Seeing the "UI" as a dependency in the "Core" is a little bit of a red flag. Maybe thats not the best place for our class.  Maybe it makes more sense for the SearchParams to live in the Core package instead of the DOM package? 

Thoughts? @Delerme

b1lly (commented):
Nice start man! I think it's a good step in the right direction.

I definitely want to see some unit tests, and some better documentation, though.

I know its a work in progress, but wanted to make sure we were thinking about those as part of the implementation :).

Delerme:
sweet, can you rebase this onto v0.7.1?

creotutar:
@Delerme done

src/ui/dom/searchparams.js
```diff
@@ -0,0 +1,121 @@
...
+  constructor (url) {
+    /**
+     * Mapping of all query parameters in the given url, query param -> value
+     * Only used if URLSearchParams does not exist in the window
+     * @type {Object}
+     * @private
+     */
+    this._params = {};
+
+    if (window.URLSearchParams) {
```
b1lly: if (window && window.URLSearchParams) {

}

Just to be safe, because it might be possible for this to run in test suites that don't have a reference to the WINDOW object.

src/ui/dom/searchparams.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {string} url The url
+   * @returns {Object} mapping from query param -> value
+   */
+  parse (url) {
+    let mapping = {};
+    let search = url;
+    if (url.indexOf('?') > -1) {
+      search = url.slice(url.indexOf('?') + 1);
+    }
+    if (search === '') {
```
b1lly: Neat! 

No need to change based on my suggestion but wanted to just add some insight to some stylistic for code formatting/organization.

Notice the new lines and grouping of functional codes below. Spacing makes things more readable and easier to follow along. When writing code try to group things into logical operations and add spacing between different sequences in order to emphasis the readability. 

Also note, I haven't tested the code below.

```
let params = {};

// No url params are present in the required format
if (url.indexOf('?') === -1) {
  return params;
}

// Otherwise, assume it's there and parse it out
const encodedQuery = url.slice(url.indexOf('?') +1);
const encodedParams = encodedQuery.split('&');
for (let i = 0; i < encodedParams.length; i ++) {
  const keyVal = encodedParams[i].split('=');
  // Invalid format for this key/val pair
  if (kevVal.length =< 1) {
    continue;
  }

  params[kevVal[0]] = SearchParams.decode(keyVal[1]);
}

return params;
```

b1lly (commented):
Seems good to me! Some minor nits and we're good to go!

b1lly (approved):
LGTM!

PR #35, opened 2019-06-27
 Add tests for autocomplete search 

Delerme:


tests/ui/components/search/autocomplete.js
```diff
@@ -0,0 +1,140 @@
...
+      answersKey
+    });
+  });
+
+  describe('queryFilter', () => {
+    const input = 'test';
+    const expectedUrl = `${baseUrl}/answers/filtersearch?v=20190101&api_key=${apiKey}&input=${input}&answersKey=${answersKey}&experienceKey=${verticalKey}&inputKey=${barKey}`;
+
+    it('creates a proper GET request for filters', () => {
+      global.fetch = jest.fn(() => Promise.resolve({ json: () => Promise.resolve({}) }));
```
b1lly: I guess I was thinking we'd mock out the "HTTPRequester" instead of the global fetch method. Reason being, we could change the implementation details of fetch in the future. Right now, we basically depend directly on it, but I created the HTTPRequester abstraction to allow us to change the implementation details in the future (e.g. its not very well supported).

We might need to "provide" the HTTPRequester to the APIRequest object as a dependency, or, change the interface for how HTTPRequester works (perhaps it takes in a request object) instead of the Request containing the requestor. 

All depends on what we decide, I guess. Either way is probably good, but the important thing is that the HTTPRequestor would be the interface that returns a promise; the HTTPRequestor would be the thing that is mocked out. And the HTTPRequestor would probably be either provided as a dependency or used differently.

I think providing as a dependency and mocking it out might be a reasonable first step without a huge refactor for this test.

Thoughts?

src/core/search/autocomplete.js
```diff
@@ -51,15 +52,17 @@ export default class AutoComplete {
       params: {
         'input': input,
         'answersKey': this._answersKey,
-        'experiencekey': verticalKey,
+        'experienceKey': verticalKey,
```
b1lly, line 57: Is this supposed to be vertical key? I thought we had a change that deprecated this?

b1lly (commented):
Pretty much spot on for what I was thinking. The only question is whether or not we want to change the "promise" mocking from fetch to the HTTPRequestor that I mentioned in the below comment.

tests/ui/components/search/autocomplete.js
Delerme: done

src/core/search/autocomplete.js
Delerme: I changed all the usages in the library but didn't want to mess with the server request. Could change it and test in a future commit

tests/core/http/httprequester.js
```diff
@@ -0,0 +1,177 @@
...
+
+  beforeEach(() => {
+    requester = new HttpRequester();
+  });
+
+  describe('get', () => {
+    const expectedUrl = `${baseUrl}/answers/query?apiKey=${apiKey}&input=${input}&answersKey=${answersKey}&verticalKey=${verticalKey}&barKey=${barKey}`;
+
+    it('creates a proper GET request with params', () => {
+      global.fetch = jest.fn(() => Promise.resolve({ json: () => Promise.resolve({}) }));
```
b1lly, line 22: if your using the http requester then this shouldnt be needed? I guess the idea would be to create a mock HttpRequester that stubs out each of the interface methods, and then use it.

Or does that seem wrong?

b1lly (commented):
Minor thing. I don't feel super strongly but I think the idea of overriding the global fetch might be weird.

tests/core/http/httprequester.js
Delerme: That's what I did in the autocomplete tests. This file is testing http requester itself so we can't mock it, we have to mock fetch.

Autocomplete, which uses http requester, does use a mock requester.

b1lly (approved):

PR #38, opened 2019-07-02
Have setFilter in core replace the data in storage

Delerme:
By default, storage merges the old and new data which causes issues with
filters, which doesn't maintain any high level shape.


src/core/core.js
```diff
@@ -66,8 +66,8 @@ export default class Core {
       .verticalQuery(queryString, verticalKey, filter)
       .then(response => SearchDataTransformer.transformVertical(response))
       .then(data => {
-        this.storage.insert(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
-        this.storage.insert(StorageKeys.VERTICAL_RESULTS, data[StorageKeys.VERTICAL_RESULTS]);
+        this.storage.merge(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
+        this.storage.merge(StorageKeys.VERTICAL_RESULTS, data[StorageKeys.VERTICAL_RESULTS]);
```
b1lly: I don't think this is correct at all.

Why would we be merging? State coming from the server should be final? Otherwise you're going to have results persisting across queries?

But maybe i'm confused with the name?

src/core/core.js
```diff
@@ -76,14 +76,14 @@ export default class Core {
       .query(queryString)
       .then(response => SearchDataTransformer.transform(response, urls))
       .then(data => {
-        this.storage.insert(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
-        this.storage.insert(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
-        this.storage.insert(StorageKeys.UNIVERSAL_RESULTS, data[StorageKeys.UNIVERSAL_RESULTS]);
+        this.storage.merge(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
+        this.storage.merge(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
+        this.storage.merge(StorageKeys.UNIVERSAL_RESULTS, data[StorageKeys.UNIVERSAL_RESULTS]);
```
b1lly: Same comment everywhere?

src/core/storage/moduledata.js
```diff
@@ -10,7 +10,29 @@ export default class ModuleData extends EventEmitter {
...
+
+    this.capturePrevious();
+
+    for (const key of Object.keys(newData)) {
+      if (this._data[key] !== newData[key]) {
+        this._data = newData;
+        this.emit('update', this._data);
+        return this;
+      }
+    }
```
b1lly, line 30: I'm pretty confused by the logic here.

You're checking if the old data set has a key, from the new object. And if it does, replace the entire object? Why even do the check at all?

src/core/storage/storage.js
```diff
@@ -7,7 +7,30 @@ export default class Storage {
...
+   */
+  set (key, data) {
+    this._initDataContainer(key, data);
+    this._moduleDataContainer[key].set(data);
+  }
+
+  _initDataContainer (key, data) {
     if (key === undefined || key === null || typeof key !== 'string') {
       throw new AnswersStorageError('Invalid storage key provided', key, data);
     }
```
b1lly, line 26: What do you think about pushing the key and data checking down to the ModuleData class?

src/core/core.js
Delerme: yea that's what the behavior of the original method was, it looked intentional. If it wasn't I can just delete the merge method.

I needed the storage method to completely replace because like you said results persisting across queries, but the insert method was implemented as a merge and it looked intentional, so I made a new method and renamed insert to reflect its behavior, if you didn't mean to implement it as a merge I can delete it and use set everywhere, just wanted to maintain the existing behavior

src/core/core.js
Delerme: see above, it was merge to begin with, I renamed it to match the behavior because it threw me off

src/core/storage/moduledata.js
Delerme: to avoid emitting an update as a result of a noop. I.e. if you keep calling set() with the same object, should each call result in an update? I don't think so, so this is just making sure something actually changed before triggering all the listeners

src/core/storage/storage.js
Delerme: I think storage is more concerned about the key, since it's a storage key, so it makes sense to check it here, and for the data I think it's better to check preconditions up front so we always know we have what we need to avoid any bugs in the future

src/core/core.js
b1lly: It's not so much a merge, because merge would combine data for each key with the previous.

It's more of an update with dirty checking. I only wanted to emit updates if there were changes.

src/core/storage/moduledata.js
b1lly: OK, can you just add a comment above the code that explains the reasoning.

Otherwise LGTM

src/core/core.js
Delerme: As discussed, the merge behavior of the original function doesn't seem to be intentional, so deleted it

src/core/storage/moduledata.js
Delerme: done

b1lly (approved):
LGTM

PR #39, opened 2019-07-02
Adds component analytics support

Delerme:
Adds the analytics reporter and event types, as well as the template hook
functionality so events can be registered via the templates.


src/core/analytics/analyticsevent.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Model for the analytics event type
+ */
+export default class AnalyticsEvent {
+  constructor (type, label) {
+    this.event_type = type.toUpperCase();
```
b1lly: eventType

src/core/analytics/analyticsevent.js
```diff
@@ -0,0 +1,13 @@
...
+    this.label = label;
+  }
+
+  addPayload (payload) {
+    Object.assign(this, payload);
```
b1lly: What exactly is the payload in this context? I have no idea what payload is supposed to look like?

I'd recommend defining it explicitly as a model. Or through documentation. Additionally, might be worth encapsulating it as an explicit property if its user defined so that variables don't get overridden by external consumers of the interface that we dont want to be overridden (in the future if the model expands)

src/core/analytics/analyticsreporter.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Class for reporting analytics events to the server
+ */
+export default class AnalyticsReporter {
+  constructor (apiKey, answersKey) {
+    this._apiKey = apiKey;
+    this._answersKey = answersKey;
+
+    let params = new URL(window.location.toString()).searchParams;
+    this._isLocal = params.get('local');
```
b1lly, line 14: Is there a better way to handle this, thats easy? If so, can we establish the pattern here and then plan to migrate the other classes?

src/core/analytics/analyticsreporter.js
```diff
@@ -0,0 +1,54 @@
...
+
+    // Temporary workaround for getting internal business ID for the analytics endpoint
+    // To be removed when the endpoint is moved behind liveapi
+    const businessIdReq = new ApiRequest({
+      baseUrl: this._isLocal ? 'http://' + window.location.hostname : 'https://liveapi.yext.com',
```
b1lly: can the base url be defined in the constructor?

src/core/analytics/analyticsreporter.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+export default class AnalyticsReporter {
+  constructor (apiKey, answersKey) {
+    this._apiKey = apiKey;
+    this._answersKey = answersKey;
+
+    let params = new URL(window.location.toString()).searchParams;
+    this._isLocal = params.get('local');
+
+    // Temporary workaround for getting internal business ID for the analytics endpoint
```
b1lly: // TODO(jdelerme) Temporary workaround [etc..]

b1lly (commented):
Beastmode!

src/core/analytics/analyticsevent.js
Delerme: the backend needs "event_type"

src/core/analytics/analyticsevent.js
Delerme: it has no structure, it's just adding additional data to the event, i.e. some events require things like user name or email and others want query ID, so this is how you add generic options. Renamed to addOptions to make that more clear

src/core/analytics/analyticsreporter.js
Delerme: I've been thinking about it, haven't come up with a good solution yet so leaving this here for now

src/core/analytics/analyticsreporter.js
Delerme: done

src/core/analytics/analyticsreporter.js
Delerme: done

src/core/analytics/analyticsevent.js
b1lly: I wouldn't couple this directly to a backend model. This should be a front-end model, right?
Perhaps make a transformation step that gets called on the model before its sent that formats it into the serverside model?

src/core/analytics/analyticsevent.js
b1lly: Sounds good. I'd add some JSDoc though as well to explain its usage/behavior, if it hasn't been added already.

src/core/analytics/analyticsreporter.js
b1lly: I think taking the search params (or request object) at the top level of the app and making it a global property on the core (or in the storage) is probably the best approach. We can have an implementation to refactor, but the more classes we add accessing this the more we'll have to refactor in the future.

src/core/analytics/analyticsevent.js
Delerme: done

src/core/analytics/analyticsevent.js
Delerme: done

src/core/analytics/analyticsreporter.js
Delerme: yea I considered that but that would tie together a lot of properties, like local templates and local search backend; what if I want local templates but live search or vice versa? I think instead we'll want to remove the concept of local from everywhere and instead pass in alternative URLs for each thing, and have that set in the umd? Either way, not doing in this commit, I don't think 1 more usage to ctrl-F later is a serious concern in this case

b1lly (approved):
LGTM!

PR #40, opened 2019-07-03
Add JSDoc

Delerme:
Adds JSDoc dev dependency and npm script. Prepend every file with @module so
jsdoc properly classifies them. Check in the first generation.

To be used with GitHub pages for automatic updating and hosting


b1lly (commented):
Can you check in the JS changes separate from the published/generated HTML?

Delerme:
Done, force pushed two commits, one with the js changes and one with the docs

PR #41, opened 2019-07-08
Adds initial CONTRIBUTING file

Delerme:


b1lly (approved):
LGTM. Might want to add information about install + setup

e.g. npm install and then the other npm commands to dev/build etc.

PR #42, opened 2019-07-08
Add filterbox and filteroptions component

Delerme:
Adds the FilterBox component which manages a set of filters, searching with them
when applied, and the static filter component FilterOptions which renders a set
of radio buttons or checkboxes which each correspond to a filter.


Delerme:
Working on tests but wanted to get eyes on for feedback

src/core/models/filter.js
```diff
@@ -6,19 +6,123 @@ export default class Filter {
     Object.freeze(this);
   }
 
+  /**
+   * Parse a JSON format filter returned from the server into a Filter
+   * @param {*} responseFilter A filter in JSON format returned from the backend
+   */
```
b1lly, line 17: What are your thoughts on also adding the @return {Filter} annotation for each of this?

src/core/models/filter.js
```diff
@@ -6,19 +6,123 @@ export default class Filter {
...
     return new Filter({
       '$and': filters
     });
   }
+
+  /**
+   * OR filters with the same keys, then AND the resulting groups
+   * @param  {...Filter} filters The filters to group
+   */
+  static group (...filters) {
```
b1lly, line 49: in ES6, destructing these things is actually pretty slow. The ES6 implementation did change how this works to make it faster, but I can't remember how many browsers adopted it, which is why I avoided it.

Anyway, its much faster to access the arguments object (which actually is slow itself, inherently) but better than this.

If destructing isn't a problem anymore its worth keeping, because the syntax is beautiful. But just worth considering.

src/core/models/filter.js
```diff
@@ -6,19 +6,123 @@ export default class Filter {
     Object.freeze(this);
   }
```
b1lly, line 12: Lets add some unit tests for these class.

src/core/models/filter.js
```diff
@@ -6,19 +6,123 @@ export default class Filter {
...
+  }
+
+  /**
+   * Create a new filter with the given matcher
+   * @private
+   * @param {string} field The subject field of the filter
+   * @param {string} matcher The matcher for the filer
+   * @param {*} value The value for the filter
+   */
+  static _makeFilter (field, matcher, value) {
```
b1lly: our factory "pattern" is typically a static method named "from (...arguments)".

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+import DOM from '../../dom/dom';
+import * as StorageKeys from '../../../core/storage/storagekeys';
+import Filter from '../../../core/models/filter';
+
+/** @module FilterBoxComponent */
+
+/**
+ * Renders a set of filters, and searches with them when applied.
+ * Multiple FilterBox components will AND together but will not share state.
+ */
```
b1lly, line 13: just add the

@module FilterBoxComponent as part of the top level class documentation instead of a separate multi-line-comment.
```
/**
  * Describe behavior
  * @module FilterBoxComponent
  * @class
  */
```
It's probably worth always adding both, but I don't feel strongly.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+        'FilterBox');
+    }
+
+    this._filterConfigs = opts.filters;
+    this._verticalKey = opts.verticalKey || null;
+    this._filterComponents = [];
+    this._filters = [];
+    this._templateName = `filters/filterbox`;
+    this._searchOnChange = opts.searchOnChange || false;
+  }
```
b1lly, line 72: Can we annotate all of these properties and values so that they show up in the jsdoc properly as well?

```
/**
  * Usage
  * @type {Object}
  * @default 'Value'
  */
```

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,97 @@
+import Component from '../component';
+import { AnswersComponentError } from '../../../core/errors/errors';
+import Filter from '../../../core/models/filter';
+import DOM from '../../dom/dom';
+
+/** @module FilterOptionsComponent */
```
b1lly: same comment

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,97 @@
...
+
+const controls = {
```
b1lly: Style nit: 

CONTROLS

All constants in the "global scope" should be capitalized.

That way when im reading code below I know its not a local variable.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,97 @@
...
+      throw new AnswersComponentError(
+        'FilterOptions component requires options to be provided',
+        'FilterOptions');
+    }
+
+    this._control = controls[opts.control];
+    this._options = opts.options;
+    this._templateName = `controls/${opts.control}`;
+    this._storeOnChange = opts.storeOnChange || false;
+    this._onChange = opts.onChange || function () {};
```
b1lly: jsdoc

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,97 @@
...
+
+    if (!opts.control || controls[opts.control] === undefined) {
+      throw new AnswersComponentError(
+        'FilterOptions requires a valid "control" to be provided',
+        'FilterOptions');
+    }
+
+    if (!opts.options) {
+      throw new AnswersComponentError(
+        'FilterOptions component requires options to be provided',
```
b1lly, line 32: usually `options` are not "required" . Maybe name it config?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,97 @@
...
+/** @module FilterOptionsComponent */
+
+const controls = {
+  'singleoption': {
+    inputs: '.js-yext-radio'
+  },
+  'multioption': {
+    inputs: '.js-yext-checkbox'
+  }
+};
```
b1lly: should the selectors be configurable? 

Perhaps you can give me some sample usages of your interface so I can look at it, at a high level.

b1lly (commented):
Savage mode. Mostly stylistic nits. I do have some minor functional thoughts but i'll do that in a separate review. Nothing major so the design/approach shouldn't change much.

src/core/models/filter.js
Delerme: done

src/core/models/filter.js
Delerme: we use babel so all es6 features are just for the syntax, none of it makes it to the browser

src/core/models/filter.js
Delerme: done

src/core/models/filter.js
Delerme: done

src/ui/components/filters/filterboxcomponent.js
Delerme: That doesn't seem to work. Based on the jsdoc docs the current format is correct
https://jsdoc.app/howto-es2015-modules.html
https://jsdoc.app/howto-es2015-classes.html

src/ui/components/filters/filterboxcomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: see above

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: yea, though this is the FilterOptions component so passing in `options` seems to make more sense, and I think it's more clear than `config`

src/ui/components/filters/filteroptionscomponent.js
Delerme: yup done

Delerme:
I tried to get some tests for the components but ran into trouble with the rendering and mounting, and think it's probably not worth blocking this. I think we should figure that out before scalars though

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+      this._applyFilters();
+    }
+  }
+
+  /**
+   * Save current filters to storage and trigger a search with all filters
+   * in storage
+   * @private
+   */
+  _applyFilters () {
```
b1lly: Minor nit.

Might be worth separating this method into two.

1. persistFilters to storage.
2. trigger a search, using filters.

that'd make it easier to test in the future as well.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,130 @@
...
+import Component from '../component';
+import { AnswersComponentError } from '../../../core/errors/errors';
+import DOM from '../../dom/dom';
+import * as StorageKeys from '../../../core/storage/storagekeys';
+import Filter from '../../../core/models/filter';
+
+/**
+ * Renders a set of filters, and searches with them when applied.
+ * Multiple FilterBox components will AND together but will not share state.
+ */
```
b1lly, line 13: @extends Component

src/core/models/filter.js
```diff
@@ -6,19 +6,134 @@ export default class Filter {
     Object.freeze(this);
   }
```
b1lly, line 12: Can we give an example in documentation with this filter object structure might look like? Or is it possible to set all the defaults to null?
```
this.$or = null;

this.$and = null;
```

Just so that when im reading through this code I have a reference of what some of these filter objects can look like without having to dig through the backend.

src/core/models/filter.js
```diff
@@ -6,19 +6,134 @@ export default class Filter {
...
+
+  /**
+   * OR filters with the same keys, then AND the resulting groups
+   * @param  {...Filter} filters The filters to group
+   * @returns {Filter}
+   */
+  static group (...filters) {
+    const groups = {};
+    for (const filter of filters) {
+      const key = Object.keys(filter)[0];
```
b1lly, line 52: Just a question. Is it possible for a filter to have an `and` and an `or`? or is it only ever going to be one or the other? If not, the Filter object probably needs some better JSDoc describing what it represents. See above comment.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * If true, trigger a search on each change to a filter
+     * @type {boolean}
+     * @private
+     */
+    this._searchOnChange = config.searchOnChange || false;
+
+    /**
+     * The components created for each filter config
+     * @type {Component[]}
```
b1lly, line 54: Is this new sytax? It used to be `{Array.<Component>}`

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,130 @@
...
+            this.onFilterChange(i, f);
+          }
+        }));
+      component.mount();
+      this._filterComponents.push(component);
+    }
+
+    // Initialize apply button
+    if (!this._searchOnChange) {
+      const button = DOM.query(this._container, '.js-yext-filterbox-apply');
```
b1lly: we might want to make this selector configurable as well.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,130 @@
...
+    for (let i = 0; i < this._filterConfigs.length; i++) {
+      const config = this._filterConfigs[i];
+      const component = this.componentManager.create(config.type, Object.assign({},
+        config,
+        {
+          parent: this,
+          name: `${this.name}.filter${i}`,
+          storeOnChange: false,
+          container: `.js-yext-filterbox-filter${i}`,
+          onChange: (f) => {
```
b1lly, line 96: filter

I know golang code style likes to use 1 variable names, but lets let the minified handle that and keep our stuff readable.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Save current filters to storage and trigger a search with all filters
+   * in storage
+   * @private
+   */
+  _applyFilters () {
+    const validFilters = this._filters.filter(f => f !== undefined && f !== null);
+    const combinedFilter = validFilters.length > 1 ? Filter.and(...validFilters) : validFilters[0];
```
b1lly: ternaries are great but they hurt readability. I think we should heir on using these sparingly. Can we update this to be its expanded form? Here and below?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,122 @@
...
+import Filter from '../../../core/models/filter';
+import DOM from '../../dom/dom';
+
+const SUPPORTED_CONTROLS = [
+  'singleoption',
+  'multioption'
+];
+
+/**
+ * Renders a set of options, each one representing a filter in a search.
```
b1lly, line 18: What do you think about defining an example configuration in the JSdoc for all of our components moving forward? That way when people open the class they know how to use it. I can take the time to update all the existing components JSDoc if you think it'd be good to have.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,122 @@
+/** @module FilterOptionsComponent */
+
+import Component from '../component';
+import { AnswersComponentError } from '../../../core/errors/errors';
+import Filter from '../../../core/models/filter';
+import DOM from '../../dom/dom';
+
+const SUPPORTED_CONTROLS = [
```
b1lly, line 12:
```
/**
 * @type {Array.<string>}
 */
```

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!config.options) {
+      throw new AnswersComponentError(
+        'FilterOptions component requires options to be provided',
+        'FilterOptions');
+    }
+
+    /**
+     * The list of filter options to display
+     * @type {object[]}
```
b1lly, line 38: Unless this is a new JSDoc thing can you update it everywhere? Otherwise ignore my comment.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,122 @@
...
+  onMount () {
+    const elements = DOM.queryAll(this._container, this._optionSelector);
+
+    if (elements.length === 0) {
+      throw new AnswersComponentError(
+        `Could not initialize FilterOptions. Can not find inputs ${this._optionSelector}`,
+        'FilterOptions');
+    }
+
+    elements.forEach(e => DOM.on(e, 'change', () => this._applyFilter()));
```
b1lly: You might be able to just do all this using event delegation instead of binding to each one.

```
DOM.delegate(this._container, this._optionSelector, 'change', () => {

});
```

Also, we can update the `DOM.on(...)` method if we want to support multiple selectors to make usage and the interface easier, if you prefer.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -0,0 +1,122 @@
...
+    elements.forEach(e => e.setAttribute('checked', 'false'));
+    this._applyFilter();
+  }
+
+  /**
+   * Get the current option values and alert the parent of the change, optionally
+   * committing to storage.
+   * @private
+   */
+  _applyFilter () {
```
b1lly: I think applyFilter is a bad name here. I also think this method could be broken into two separate unit testable operations if it were to stay this way.

Separately, I think there's a way to provide the values from the onChange handler instead of having to requery the DOM to parse the values. That way this would be unit testable without having to depending on the DOM.

For instance,

One approach would be to render the DOM based on the local state. Whenever the DOM changes, update the local state. And when you ready to use the state, you pull it from local state instead of the DOM.

So overall, the goal would be to make the component be the source of truth for the state of the component options. When the DOM changes, it updates the source of truth. Then all these DOM dependencies go away and its easier to unit test.

We can talk offline if this is a little confusing.

b1lly (changes requested):
Dude super great!

The only thing I would suggest is reducing the dependency of the component on the DOM. The Component should be the source of truth for the state, which handles the rendering. The DOM integration (e.g. change event handling), should then update the local state to reflect the state of the DOM.

Then the component could work off the local state, instead of having to query the DOM for the state whenever it needs to do things (e.g. persist to local storage, or bubble up the state to external code).

It's also much easier to test state then the DOM, and also easier to work and reason about state in memory than state in DOM.

src/ui/components/filters/filterboxcomponent.js
Delerme: done

src/core/models/filter.js
Delerme: it's kinda arbitrary and the spec for it is kind of intense, I guess I'll just link to our api docs which has more details then we could ever provide

src/core/models/filter.js
Delerme: yea that's what this function does, creates a single filter representing the combination of other filters. As mentioned above I just linked to our api docs which better specifies the structure of a filter

src/ui/components/filters/filterboxcomponent.js
Delerme: maybe, I've always used it 
https://jsdoc.app/tags-type.html

src/ui/components/filters/filterboxcomponent.js
Delerme: done

src/ui/components/filters/filterboxcomponent.js
Delerme: yup done

src/ui/components/filters/filterboxcomponent.js
Delerme: what if I do them on separate lines? I agree the single line is hard to read, but I think the multiline form is easier to comprehend than a full if statement because it doesn't require a variable reassignment. Check now and if you still don't like it I'll make a full if

src/ui/components/filters/filterboxcomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: don't we do that already in the README? Went ahead and added these components there

src/ui/components/filters/filteroptionscomponent.js
Delerme: Don't know if it's new, I've always used it. Seems more readable than the alternative so updated it everywhere

src/ui/components/filters/filteroptionscomponent.js
Delerme: yup meant to switch to delegate after POC, done.

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/components/filters/filterboxcomponent.js
b1lly: Ah, ok. I've always down the former way. Whatever we decide should be consistent across the library though. So if you haven't already done a lot of the Component[] syntax maybe just update it here? Or we can switch all the other docs to follow this pattern at another point.

src/ui/components/filters/filteroptionscomponent.js
b1lly: We do. I was saying as a developer sometimes its much easier to just look at the top of the file to see its usage instead of having to open a separate doc. It's just a convenience thing. And the README will become big and hairy over time, which is the only reason I was suggesting it.

Either way, SGTM, was more of a suggestion.

Delerme: alright I'll start thinking about something and punt to another commit

src/ui/components/filters/filterboxcomponent.js
Delerme: I already switched all Array.<x> usages to x[] in this commit

src/ui/components/filters/filterboxcomponent.js
b1lly: Was it done?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -0,0 +1,149 @@
...
+      ? Filter.and(...validFilters)
+      : validFilters[0];
+    this.core.setFilter(this.name, combinedFilter || {});
+  }
+
+  /**
+   * Trigger a search with all filters in storage
+   */
+  _search () {
+    const allFilters = this.core.storage.getAll(StorageKeys.FILTER);
```
b1lly, line 142: might be better for this to be provided as input for unit testing purposes.

But otherwise LGTM

b1lly (commented):
Savage. Last comment.

Otherwise LGTM.

src/ui/components/filters/filterboxcomponent.js
Delerme: eh it'll have to be mocked at some level, may as well be here

src/ui/components/filters/filterboxcomponent.js
Delerme: hmm, must have got wrecked by my rebase, try now

src/ui/components/filters/filterboxcomponent.js
b1lly: In this case you'll need two mocks and need to set state in the storage.

By just providing the filters as input, all you need to do is mock the verticalSearch and provide the input. The behavior to test is the search with filters, not the storage.

Either way nbd.

b1lly (approved):

src/ui/components/filters/filterboxcomponent.js
Delerme: punting for now

PR #44, opened 2019-07-09
Storage keys uses enum

b1lly:


src/core/storage/storagekeys.js
```diff
@@ -1,8 +1,9 @@
 /** @module */
-
-export const NAVIGATION = 'navigation';
-export const UNIVERSAL_RESULTS = 'universal-results';
-export const VERTICAL_RESULTS = 'vertical-results';
-export const AUTOCOMPLETE = 'autocomplete';
-export const DIRECT_ANSWER = 'direct-answer';
-export const FILTER = 'filter';
+export const StorageKeys = {
```
Delerme: jsdoc and maybe export default?

src/ui/components/search/filterpickercomponent.js
```diff
@@ -13,7 +13,7 @@ export default class FilterPickerComponent extends Component {
      * The template name to use for rendering with handlebars
      * @type {string}
      */
-    this._templateName = 'search/search';
+    this._templateName = 'search/filterpicker';
```
Delerme, line 16: this is actually meant to be search, and I was supposed to delete filterpicker.hbs, feel free to delete in this commit

Delerme (changes requested):

src/ui/components/search/filterpickercomponent.js
b1lly: I didn't want the filter picker and the search templates to be shared. It's possible they diverge, and we cant guarantee that someone in the future is going to want to customize one and not the other.

src/core/storage/storagekeys.js
b1lly: done. it's possible that we move the keys into the storage module as a named export in the future instead of having two files, though.

src/ui/components/search/filterpickercomponent.js
Delerme: cool, that worked out then

Delerme (approved):

src/core/storage/storagekeys.js
Delerme: yea makes sense

PR #46, opened 2019-07-09
Remove instances of functions modifying params

Delerme:
Remove a few instances of functions modifying their params and other
data accidentally with Object.assign


b1lly (approved):
LGTM. Although, we might want to create an internal merge method in the future. Instead of doing Object.assign everywhere.

PR #47, opened 2019-07-09
Add locale parameter to configuration & requests in Js library

vijay267:
WAT-244

The configuration now takes an optional "locale" parameter on initialization and passes that parameter value to all of its api requests as a "locale" request parameter (query, vertical query, autocomplete, vertical autocomplete, and autocomplete filter). Currently the default for locale is 'en'.

E.g. Requests will now look like GET /v2/accounts/me/answers/query?locale=[locale]&api_key=[api_key]...

TEST=manual
Tested by manually testing the various API requests and made sure the correct locale appeared (either specified or default if not specified). Updated appropriate tests.


src/core/core.js
```diff
@@ -33,6 +33,18 @@ export default class Core {
      */
     this._answersKey = opts.answersKey;
 
+    /**
+     * A reference to the client locale used for all requests. If not specified, defaults to "en" (for
+     * backwards compatibility).
+     * @type {string}
+     * @private
+     */
+    if (opts.locale != null) {
```
Delerme: you can replace this with 
`this._locale = opts.locale || 'en';

src/core/core.js
```diff
@@ -48,18 +60,20 @@ export default class Core {
...
-      answersKey: this._answersKey
+      answersKey: this._answersKey,
+      locale: this._locale
     });
 
     /**
      * An abstraction containing the integration with the RESTful autocomplete API
-     * For both vertical and universal search
-     * @type {Search}
+     * For autocomplete filter, vertical autocomplete, and universal autocomplete
```
Delerme: I think we're now calling autocomplete filter "filter search"

src/core/search/autocomplete.js
```diff
@@ -39,6 +39,13 @@ export default class AutoComplete {
      * @private
      */
     this._version = opts.version || 20190101 || 20190301;
+
+    /**
+     * The locale to use for the request
+     * @type {string}
+     * @private
+     */
+    this._locale = opts.locale || null;
```
Delerme: is this a required param? if so don't fallback to null and instead chech that opts.locale is provided and throw if now

if (!opts.locale) {
 throw new AnswersBasicError(...);
}

src/core/search/search.js
```diff
@@ -37,6 +37,13 @@ export default class Search {
      * @private
      */
     this._version = opts.version || 20190101 || 20190301;
+
+    /**
+     * A local reference to the locale to use for the request
+     * @type {string}
+     * @private
+     */
+    this._locale = opts.locale || null;
```
Delerme: ditto

Delerme (changes requested):

src/core/core.js
vijay267: Done

src/core/core.js
vijay267: True, was originally trying to match other code, but might as well use the current name.

Delerme (commented):
lgtm, to @b1lly

b1lly (dismissed):
Yeah, I think this is fine for now.

I think there's a bigger refactor coming where:

1. There is a singleton HTTPRequester
2. The HTTPRequestor is available through the core, to `search` and `autocomplete`
3. The HTTPRequestor has some `global properties` that can be configured
4. HTTPRequestor acts on APIRequest, which contains data about the request, and can inherit `global` state from the HTTPRequestor or override them if defined
5. The HTTPRequestor contains things like api key, answers key, and local as some of the global options

That way when we make changes like this in the future we can just do it in one place.

@Delerme 

Otherwise, LGTM.

Delerme (dismissed):

Delerme (approved):

PR #49, opened 2019-07-10
Remove isLocal param checks

Delerme:
Place the default base urls for the api, templates, and analytics in the
constants file, instead of each file checking for the isLocal param. For now
these values can be changed directly for local development, and in the future
these constants could be populated by environment variables or other.


src/core/search/searchapi.js
```diff
@@ -41,11 +29,12 @@ export default class SearchApi {
      * @private
      */
     this._version = opts.version || 20190101 || 20190301;
+
+    this._requester = new HttpRequester();
```
b1lly: Is this used?

Is the idea for the HttpRequester to take in an ApiRequest as an object in the future? I originally baked it into the ApiRequest, but maybe it makes sense for the requester to operate on the request?

src/core/analytics/analyticsreporter.js
```diff
@@ -42,7 +38,7 @@ export default class AnalyticsReporter {
     event.addOptions({ answers_key: this._answersKey });
 
     const request = new ApiRequest({
-      baseUrl: this._analyticsUrl,
+      baseUrl: ANALYTICS_BASE_URL,
```
b1lly, line 41: am I missing something here?

Shouldn't this be a configurable thing? In the future, what if we wanted to point our code at our own staging/dev environments for internal testing?

src/core/search/autocompleteapi.js
```diff
@@ -48,7 +38,6 @@ export default class AutoCompleteApi {
    */
   queryFilter (input, verticalKey, barKey) {
     let request = new ApiRequest({
-      baseUrl: this._baseUrl,
       endpoint: '/v2/accounts/me/answers/filtersearch',
```
b1lly, line 41: Yeah, I think we messed up something here. You basically eliminated the ability to test our JS against our internal environments (e.g. if we were running the backend services locally, or in a staging or office environment) it would be almost impossible to make this work without having to update code in various places.

I think there's two use cases we want to support:

1. As Yext developers, we want the ability to be able to configure the library to communicate with our internal services. Basically, telling the library to use these backends (e.g. baseUrl configuration) at the top level. In your case, I'd have to update code in various places to make it work. 

2. As a general developer who doesn't have access to Yext services, and who doesn't care about backends, I want to easily point the library i'm using at local templates for development. I don't really care that the search and error reporter are hitting production endpoints.

b1lly (changes requested):

src/core/analytics/analyticsreporter.js
Delerme: Change the constants file, I don't think it makes sense to expose the option on the library level if it's meant to be internal only. Now instead of chasing down this variable when testing locally, you can just change it in the constants file for testing

src/core/search/autocompleteapi.js
Delerme: Beforehand we had to update the code in a variety of places, because they each had their own isLocal checks that all had the same logic. Now it's read from the constants file so for developing locally you just change that single file. Unlinke before, this also allows you to change it to any url (useful if you're serving your page separately from the api you want to test)

So you can still test against internal environments, and more so this makes it both easier and better

For point 2) the template url is already configurable on the library level

src/core/search/searchapi.js
Delerme: Huh yea it's not, may as well remove it while I'm here (I just moved it down a few lines for clarity but it was here before)

Done

b1lly (approved):
OK, seem fine then.

It might be unfortunate to have to keep rewriting the constants on every branch and stashing the changes before committing them.

Have you thought about the development flow for that yet? What are your thoughts around there?

Delerme:
Yea as mentioned in the commit message I think we can populate these constants from environment variables or maybe grunt tasks in the future, but this is an important first step

PR #51, opened 2019-07-11
Save query to storage, and use in all searches

Delerme:
When a query is entered in the search bar, persist it to storage and use it in
the search methods of filterbox and filtersearch. Similarly, include the
currently applied filters to searches with the searchbar


src/ui/components/search/filtersearchcomponent.js
```diff
@@ -142,37 +143,51 @@ export default class FilterSearchComponent extends Component {
...
 
     this.core.setFilter(this.name, Filter.fromResponse(filter));
+  }
+
+  /**
+   * Perform the vertical search with all saved filters and query,
+   * optionally redirecting based on config
+   */
+  search () {
+    // If we have a redirectUrl, we want the params to be
```
b1lly: I feel like the interface should just take in query and filters as arguments instead of depending on the state. I think it'd be easier to unit test in the future in that case.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -126,6 +126,7 @@ export default class FilterBoxComponent extends Component {
 
   /**
    * Save current filters to storage to be used in the next search
+   * @private
    */
   _saveFiltersToStorage () {
```
b1lly, line 131: Probably want to provide the query instead of depending on the storage to look it up?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -142,37 +143,51 @@ export default class FilterSearchComponent extends Component {
...
     params.set(`${this.name}.query`, query);
     params.set(`${this.name}.filter`, filter);
 
-    // If we have a redirectUrl, we want the params to be
-    // serialized and submitted.
-    if (typeof this.redirectUrl === 'string') {
-      window.location.href = this.redirectUrl + '?' + params.toString();
-      return false;
-    }
-
```
b1lly, line 162: I'm not sure we want to push state when there's a redirect?

b1lly (changes requested):
Nice

src/ui/components/search/filtersearchcomponent.js
Delerme: it's a little confusing because both are named query, but one is a autocomplete query for filter search (which is not used in the real search) whereas the one from storage is a search query, so don't need query here

As for passing in the filter, we'd still to grab the others from storage otherwise so it seems better to just put our filter into storage to prevent weird states, and to avoid having to manually combine the filters again

src/ui/components/filters/filterboxcomponent.js
Delerme: there's no storage lookup here

src/ui/components/search/filtersearchcomponent.js
Delerme: fixed

src/ui/components/search/filtersearchcomponent.js
b1lly: The idea of providing input and running operations against input instead of operations against state is that unit testing is harder to setup and test.

But if you instead did this from the caller:

`search(this.core.storage.getState(StorageKeys.QUERY))` 

You're effectively accomplishing the same thing. However, by providing it the `input` as a dependency instead of creating a hard dependency on the storage state within the function, you're making it easier to test. If it's just input, when you go to unit test this function, you never have to setup the storage or rely on its state to execute it because you can just provide it the input as 'foo'.

Search shouldn't care about the implementation of 'what to search'. You should tell it what to search on by giving it the information as input. By making it depend on the 'what', you're mixing the implementation of what and how.

I've found that avoiding stateful dependencies, and instead providing their value, is something that has made writing unit tests / testable code easier for me.

Anyway, just food for thought.

b1lly (approved):

PR #54, opened 2019-07-15
Navigation should support query params as config 

b1lly:
Test coming separately. Did a little refactor and cleanup to support better. Still not in the best state, but I think its better.


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -3,81 +3,109 @@
...
- * NavigationComponent exposes an interface for building a dynamic
- * navigation that is powered by universal search updates.
- * @extends Component
+ * The Tab is a model that is used to power the Navigation tabs in the view.
+ * It's initialized through the configuration provided to the component.
  */
-export default class NavigationComponent extends Component {
-  constructor (opts = {}) {
-    super(opts);
+export class Tab {
```
Delerme, line 10: should all these properties be optional with those defaults? It seems like some are required (or at least useless as empty), probably worth removing the fallbacks and making the types {!string}

b1lly: Done. Added tests.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -1,83 +1,118 @@
...
+  constructor (config) {
+    /**
+     * The name of the tab that is exposed for the link
+     * @type {string}
+     */
+    this.label = config.label;
+    if (typeof this.label !== 'string') {
+      throw new AnswersComponentError('NavigationComponent: label is a required configuration option for tab.');
```
Delerme: constructor format is (message, component) so change to ('label is a...', 'NavigationComponent')

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -1,83 +1,118 @@
...
-    return new URLSearchParams(window.location.search.substring(1));
+    /**
+     * The data storage key
+     * @type {string}
+     */
+    this.moduleId = 'navigation';
+
+    /**
+     * The handlebars template to use
+     * @type {string}
```
Delerme, line 96: @private for these vars

Delerme:
looks fine, you'll have to rebase onto latest for the merge conflicts

Delerme (approved):

PR #56, opened 2019-07-15
Cut v0.7

Delerme:
# Version 0.7.0

### FilterSearch Component

Used to search and select filters configured in the Yext platform. The select filter applies to the current search.

### FilterBox Component

A control used to display and manage a set of static filters (see below). The selected filters affect the current search when applied.

### Static Filter Components

The first of our static filter components has been added: FilterOptions. FilterOptions creates a list of options, either checkboxes or radio buttons, with each representing a search filter. These components may be used alone or within a FilterBox. More static filter components are on the way.

### JSDocs

Generate JSDocs will now be checked in under the docs/ directory and updated with each version. These will be hosted in the future.

### Component Analytics

Interface for creating and sending analytic events. Data attributes can be added to templates to automatically create and send click events with the provided metadata.

## Minor Changes & Fixes

- Storage refactored to use StorageKeys for indexing data, with an optional namespace.

- Multiple components of the same type can be used and work together

- Add a CONTRIBUTING file

- NavigationComponent now supports persisting tab ordering through query params

- Minor fixes for maps

- Remove checks for local development. Test urls can be configured in the constants file, additional support coming soon

- Add the locale to search requests, defaulted to 'en'


b1lly (approved):
LGTM

PR #57, opened 2019-07-15
Make circle aws deploy commands idempotent

Delerme:
merging to master since this is necessary for fuzzy versioning


b1lly (dismissed):

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ commands:
       - run:
           name: Configure AWS
           command: |
-            mkdir ~/.aws
+            mkdir -p ~/.aws
             cp .circleci/aws_config ~/.aws/config
             echo -e $AWS_PROD_PROFILE >> ~/.aws/config
```
kevincaffrey, line 38: won't this now be slightly non-idempotent? I guess it's fine to have "$AWS_PROD_PROFILE" a bunch of times at the end of the aws config? 

it does make me wonder if there is some way to refactor the config such that the "configure aws" step just gets run once, rather than trying to make it safe to run a bunch of times. is it even really necessary to use the aws/config thing? can those options just be passed directly to the s3 copy command? eg, adding "--output json" to the arguments, and same for whatever AWS_PROD_PROFILE contains (it could be split up into a few variables as needed?)

kevincaffrey (dismissed):
Seems fine, but perhaps could be improved. My 2c provided in the comment inside the review (which you could ignore if you feel like its not worth it)

.circleci/config.yml
Delerme: ah yea fixed. I think the config thing protects some config magic but I can revisit later. Also hoping to use redirects in the future instead of this double copy

Delerme: (actually the file gets overwritten in the cp step previous)

Issue #64, opened 2019-07-16
verticalURL Lost


rosiegrant:
I'm displaying a "view all" link in one of my templates based on a verticalURL in the universal search, it appears it's stopped appearing, which would indicate the template is no longer seeing verticalURL

`{{#if verticalURL}}<a class="view-more-link" href="{{verticalURL}}">View All <i class="fal fa-chevron-right"></i></a>{{/if}}`

b1lly:
Should be fixed.

PR #65, opened 2019-07-16
core/search: urls should be part of the results

b1lly:
This was an oversight in your refactor.


b1lly:
This fixes https://github.com/yext/answers/issues/64

Delerme (approved):

PR #66, opened 2019-07-16
ui/results: map options for results were not piping through properly

b1lly:
immediate fix, may need to investigate further.


Delerme (changes requested):
causes an error in vertical results because 'data' is frozen, so the object can't be extended

b1lly:
Manually Tested:

Components:
-Navigation
-Map
-Search (Vertical and Universal)
-Results
-UniversalResults

Delerme (approved):

PR #67, opened 2019-07-17
adding custom pin support for mapbox

aginsberg:


src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -75,15 +75,30 @@ export default class MapBoxMapProvider extends MapProvider {
       center: [mapData.mapCenter.longitude, mapData.mapCenter.latitude]
     });
 
+    let pinConfig = this._pinConfig;
```
Delerme, line 78: looks like this is never reassigned so you can use const

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -75,15 +75,30 @@ export default class MapBoxMapProvider extends MapProvider {
...
+    let pinConfigObj,
+      wrapper;
+
     this._map.on('load', () => {
       if (mapData && mapData.mapMarkers.length > 0) {
         let markers = mapData.mapMarkers;
         for (let i = 0; i < markers.length; i++) {
+          if (pinConfig) {
+            if (typeof pinConfig === 'function') {
```
Delerme, line 87: can combine these two if statements, if (pinConfig && typeof pinConfig === 'function') {}
it'll short circuit

Delerme (commented):
also sorry the CONTRIBUTING file is out of date, we'll actually be using branches named by the next version, instead of the 'latest' branch, so can you rebase onto the v0.7.2 branch?

PR #68, opened 2019-07-17
Use HTTPS for analytics endpoint

Delerme:


src/core/http/httprequester.js
```diff
@@ -39,7 +39,8 @@ export default class HttpRequester {
       Methods.POST,
       url,
       Object.assign({
-        body: JSON.stringify(data)
+        body: JSON.stringify(data),
+        credentials: undefined
```
b1lly, line 43: I think this should be configured from the caller and not explicit here.
Reason being, credentials is part of cookie tracking, and so there are probably some post requests in the future that would want this information provided in the request

b1lly (changes requested):

src/core/http/httprequester.js
Delerme: done (api request wasn't passing through options)

b1lly (approved):

PR #70, opened 2019-07-17
custom map pin support for mapbox

aginsberg:


Delerme (commented):
looks fine to me, will let billy take a look and approve

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -75,15 +75,28 @@ export default class MapBoxMapProvider extends MapProvider {
       center: [mapData.mapCenter.longitude, mapData.mapCenter.latitude]
     });
 
+    const pinConfig = this._pinConfig;
+    let pinConfigObj,
+      wrapper;
+
     this._map.on('load', () => {
       if (mapData && mapData.mapMarkers.length > 0) {
```
b1lly: I know you didn't write this, but this check shouldn't be needed anymore (since it already gets short circuited up at the top of init).

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -75,15 +75,28 @@ export default class MapBoxMapProvider extends MapProvider {
...
         let markers = mapData.mapMarkers;
         for (let i = 0; i < markers.length; i++) {
+          if (pinConfig && typeof pinConfig === 'function') {
```
b1lly: Do you want to pull the configuration stuff out into a separate class, similar to what we did for google?

Basically, create a MapBoxConfig model that represents the configuration for MapBox, using our configuration format as input?

Then we can use those objects to create the map using the mapbox api, as well as add some tests.

Thoughts?

b1lly (changes requested):
Nice. The implementation looks sound, but i'd like to clean up the code a little bit by separating some of the concerns while we're here, and add some tests if possible. 

I think by encapsulating the configuration logic into a configuration model it'll make the implementation cleaner and easier to extend in the future.

Also, important for tests ;)

src/ui/components/map/providers/mapboxmapprovider.js
aginsberg: Sounds good to me! I'll try and get something out later today/tomorrow

b1lly (approved):
Savage! LGTM

PR #79, opened 2019-07-22
templates: Add new AccordionResult template

benmcginnis:
Adds new AccordionResult template and updates UniversalResults to respond to
a new "useAccrdion" VerticalResult config property, which will cause the
Vertical to use the AccordionResult component instead of the standard
Results component.


Delerme:
Sweet, can you add JSDoc comments to the AccordionResultsComponent class and all of its methods

src/ui/templates/results/resultsaccordion.hbs
```diff
@@ -0,0 +1,26 @@
+{{#if results}}
+    <section class="yxt-Accordion yxt-Results yxt-Results--{{vertical}} js-yxt-Accordion">
+        {{> results/resultssectionheader}}
+        <ul class="yxt-Accordion-list yext-result-items"
+            data-prop="results">
+            {{#each results}}
+                <li class="yxt-Accordion-item js-yxt-Accordion-item yxt-AccordionResult{{#if collapsed}} is-collapsed{{/if}}{{#if resultModifier}} yxt-AccordionResult--{{resultModifier}}{{/if}}">
```
b1lly: Where does the template get all these states from? Is it provided by the component itself? 'Cause I don't see a setState override anywhere. So not exactly sure where `collapsed` and `resultModifier` etc are coming from.

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,64 @@
+/** @module AccordionResultsComponent */
+import ResultsComponent from './resultscomponent.js';
+
+const selectorBase = '.js-yxt-Accordion';
```
b1lly: I'd bake this into the constructor so that it could be configurable.

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,64 @@
...
+
+export default class AccordionResultsComponent extends ResultsComponent {
+  constructor (opts = {}) {
+    super(opts);
+    this._templateName = opts.templateName || 'results/resultsaccordion';
```
b1lly: we started adding JS doc to all of our new classes (and ones we modify)

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,64 @@
...
+  onMount () {
+    this.bindAccordionToggles();
+
+    return super.onMount();
+  }
+
+  bindAccordionToggles () {
+    for (const accordion of document.querySelectorAll(selectorBase)) {
+      for (const item of accordion.querySelectorAll(this.itemSelector())) {
+        this.setupAnimations(accordion, item);
```
b1lly: accordianEl, itemEl

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+  }
+
+  setupAnimations (accordion, item) {
```
b1lly: ^

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,64 @@
...
+    // TODO: (bmcginnis) decide if we need to collapse other results
+    const toggle = item.querySelector(this.toggleSelector());
+    const body = item.querySelector(this.bodySelector());
+    item.isCollapsed = item.classList.contains('is-collapsed');
```
b1lly: shouldn't this state be part of the result items data instead of the state being attached/retrieved from the DOM? If we triggered a re-render, this wouldn't work.

I think the idea is that the state powers the renderable state.

State -> Render (should know if an item is expanded or collapsed).
Click Handler -> Updated State -> Triggers Re-render.

The way that i'd approach this:
1. During setState, augment each`result` item object with the views stateful 'isCollapsed' property
2. Each `Result` should have an `identifier` that can be referenced. When the template is rendered, the DOM should have this reference somewhere (e.g. data-result-id="1"). The template should also include the class for `collapsed` based on the result state.
3. On click of DOM, it grabs the reference above, uses `getState` and updates the proper `Result` data item based on the reference with the proper collapsed state.
4. Trigger a new setState with the updated data, which will trigger a re-render.

b1lly (changes requested):
A few minor styling nits, and then a structural comment about using local state instead of the DOM for the collapsed state.

src/ui/components/results/accordionresultscomponent.js
benmcginnis: As we've discussed, I'm concerned about the accessibility impact of destroying and re-rendering the elements on click.  I'd like to move forward with this architecture and perhaps look at migrating to the state based implementation in the future when we have time for thorough screen-reader testing.

src/ui/templates/results/resultsaccordion.hbs
```diff
@@ -0,0 +1,42 @@
...
+            aria-controls="#{{id}}">
+        <h3 class="yxt-AccordionResult-title">
+            {{title}}
+        </h3>
+        <div class="yxt-AccordionResult-indicator" aria-hidden="true">
+        </div>
+    </button>
+    <div class="yxt-AccordionResult-body js-yxt-AccordionResult-body" id="{{id}}">
+        <div class="yxt-AccordionResult-content">
+          <div class="yxt-AccordionResult-details">
```
creotutar: can we have details on the same line

creotutar: for when we do `white-space: pre-line` so we don't have that beginning newline but preserve the newlines in the text*

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,159 @@
+/** @module AccordionResultsComponent */
+import ResultsItemComponent from './resultsitemcomponent';
+
+export default class AccordionResultsComponent extends ResultsItemComponent {
+  constructor (opts = {}) {
```
b1lly: we use `config` as the object now, `opts` is legacy pattern.

src/ui/components/results/accordionresultscomponent.js
```diff
@@ -0,0 +1,142 @@
...
+  static defaultTemplateName (config) {
+    return 'results/resultsaccordion';
+  }
+
+  /**
+   * overrides _onMount to add bindings to change the height on click
+   * @returns {AccordionResultsComponent}
+   * @override
+   */
+  _onMount () {
```
benmcginnis, line 51: onMount / mount are not currently being called on children components, so I had to override _onMount

b1lly: gotcha. I think @Delerme change in the 0.9.0 version actually resolves this.

src/ui/components/results/accordionresultscomponent.js
b1lly: Just need to address this and we're good.

b1lly (approved):

PR #80, opened 2019-07-22
templates: Update DirectAnswers, Results & Result templates

benmcginnis:
Updates DirectAnswers, Results and ResultsItem templates based on updated
designs. Also adds a new resultssectionheader template to encapsulate
new heading designs.


b1lly (commented):
I think you need the accompany these templates with the proper data models transforms internally. You can look at the core/models for some of this stuff. The goal would be to refactor those to be able to power these templates directly.

Additionally, also, want to talk to you a little bit about the various types of `resultitem` cards and how you see that being powered by configuration. Certainly, we'd have to refactor the component JS to accept the proper configuration, but I'd like to understand what your thinking for this if possible.

Aside, the markup looks good with the minor nit:

Is it possible to switch all the HTML to `2 space tab indent`, instead of 4? This goes for all commits. We should probably add HTML linting to the library builds @Delerme

benmcginnis:
@b1lly added the basic models here, as we discussed.  Works out of the box for profiles that have a description field as well as GCSE results.

src/ui/templates/results/directanswer.hbs
```diff
@@ -1,14 +1,59 @@
...
+        /
+      </span>
+      <span class="yxt-DirectAnswer-fieldName">
+        {{{answer.fieldName}}}
+      </span>
+    </h2>
+  </div>
+  <div class="yxt-DirectAnswer-content">
+    <div class="yxt-DirectAnswer-fieldValue">
+      {{{answer.value}}}
```
creotutar, line 24: can we have some answerValue field rather than use the answer.value? that way if a client wants to have a custom value for a specific `answer.fieldApiName`, they won't have to override answer.value and rather can pass it in as a separate attribute? unless we are thinking they shouldn't be spread adding data and instead passing in their own fields

e.g. not
```
transformData {
return {
 ...data
}
}
```

but rather
```
transformData {
return {
 answer: {
   entityName: x
   fieldName: y
   value: something
}
}
}
```

creotutar: silly me! you can just do, feel free to ignore

```
transformData {
  return { 
    ...data,
    answer: {...data.answer, value: 'new val'},
  }
}
```

benmcginnis: I think your first point is actually right though.  if you see the Results model we move the profile into a _raw field to avoid overwriting any profile fields.  I think we probably want to do that in DirectAnswers too, I just implemented this first and didn't update it in response to the change in Results.js

src/core/models/result.js
```diff
@@ -1,8 +1,16 @@
...
+    this._raw = data;
+    // data.htmlTitle is from GSCE, but we want to strip html from it
+    // before displaying, by default.
+    this.title = data.name || data.htmlTitle.replace(/(<([^>]+)>)/ig, '') || null;
+    // data.htmlSnippet is from GSCE
+    this.details = data.description || data.htmlSnippet || null;
+    // data.link is from GCSE;
+    this.link = data.website || data.link || null;
+    this.ordinal = index + 1 || null;
   }
```
b1lly: This is a great start! I think what we're going to want to do here is define our model, and separate the transformation logic so that we can scale the number of incoming models easily.

So:

1. Create the Result data model (this._raw, this.title, this.details, etc) like you did.
2. Using a factory pattern, inspect data and call different transformers to return the Result.

So for instance:

```
ResultFactory {
  /**
    * @returns {Result}
    */
  static from(data) {
    // .. inspect the data and choose different transformers
    if (data.htmlSnippet) {
      return ResultFactory.fromGSCE(data)
    }
  }

  static fromGSCE(data) {
    return Result({ ... })
  }

 static fromLocation(data) {
   return Result({ ... })
  }

 static fromEvent(data) {
   return Result({ ... })
  }
}
```

And then the caller would just do

```
ResultFactory.from(data)
```

That way we can have unit tests for all the transformations in the factory, and scale the variations of input models really easily.

b1lly (changes requested):

src/core/models/result.js
b1lly: @Delerme Just to collaborate on the thinking here.

benmcginnis:
@ndawg we also need to rebase this and handle the conflicts

src/core/models/result.js
benmcginnis: @b1lly should we go ahead and delete the LocationResult, EventResult, etc. ?  We can do it in this pr and replace with the result factory.

benmcginnis: I think right now, we probably can only really do the GSCE result / base result, but we can setup the factory

src/core/models/resultfactory.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default class ResultFactory {
+  /**
+   * @returns {Result}
+   */
+  static from(data, index) {
+    if (data.htmlSnippet) {
+      return ResultFactory.fromGCSE(data, index)
+    } else {
+      return ResultFactory.fromGeneric(data, index)
```
benmcginnis: you don't need the else

src/core/models/result.js
```diff
@@ -1,8 +1,13 @@
 /** @module Result */
 
 export default class Result {
-  constructor (data = {}) {
-    Object.assign(this, data);
+  constructor (data = {}, index) {
```
benmcginnis: i think, with the factory update you can revert this change

src/core/models/result.js
```diff
@@ -15,7 +20,7 @@ export default class Result {
       // transform resultData.data into html-friendly strings that highlight values.
 
       // Check for new data format, otherwise fallback to legacy
-      results.push(new Result(resultsData[i].data || resultsData[i]));
+      results.push(new Result(resultsData[i].data || resultsData[i], i));
```
benmcginnis: i think you need this to be `ResultFactory.from(resultsData[i].data || resultsData[i], i)`

benmcginnis: or we would need to delete this method

src/ui/components/results/locationresultsitemcomponent.js
```diff
@@ -5,15 +5,37 @@ import ResultsItemComponent from './resultsitemcomponent';
...
   static get type () {
     return 'LocationResultsItemComponent';
   }
 
+  /**
+   * The template to render
+   * @returns {string}
+   * @override
+   */
+  static defaultTemplateName () {
```
benmcginnis: this is on a separate PR

benmcginnis:
Can you also run `npm run fix` on the latest commit?

src/core/models/result.js
b1lly: @benmcginnis Yeah, I think the idea is to delete the individual `entity templates/components` like you suggest, and use a default one based off the core data model (e.g. title, subtitle, etc). This will require us to change some of the logic in the ResultsComponent too.

I think the goal is for the `entity-specific` components/templates to change from being `entity-specific`, to being more 'Layout-specific`.

src/core/models/result.js
ndawg: I didn't see `Result.from` being used anywhere, so this method could probably be deleted

ndawg:
@b1lly I believe all your requested changes have been addressed whenever you're ready to take a look!

src/core/models/result.js
```diff
@@ -4,20 +4,4 @@ export default class Result {
   constructor (data = {}) {
     Object.assign(this, data);
```
b1lly: instead of using Object.assign(this, data)... can you explicitly define all the core properties about the model here?

For instance,

```
this.title = data.title || 'defaultValue'; // or null if it's default should be null

this.details = data.details || null;

// ... etc
```

src/ui/templates/results/eventresultsitem.hbs
```diff
@@ -1,19 +1,58 @@
...
+      {{/if}}
+
+      {{#if details}}
+        <div class="yxt-Result-details">
+          {{{details}}}
+        </div>
+      {{/if}}
+    </div>
+
+    {{#if callsToAction}}
```
b1lly: I don't see this being set as part of the model in any of the factory transformers. Can you please define it in the Result model as well in regard to the other comment I had?

src/ui/templates/results/eventresultsitem.hbs
```diff
@@ -1,19 +1,58 @@
...
+      <h3 class="yxt-Result-title">
+        {{#if link}}
+          <a class="yxt-Result-titleLink" href="{{link}}">
+        {{/if}}
+
+        {{{title}}}
+
+        {{#if link}}
+          </a>
+        {{/if}}
```
b1lly: aka

```
{{#if link}}
  <a class="yxt-Result-titleLink" href="{{link}}">{{title}}</a>
{{else}}
  {{title}}
{{/if}} 
```

src/ui/templates/results/eventresultsitem.hbs
```diff
@@ -1,19 +1,58 @@
...
-    {{#each photoGallery}}
-      <li class="yext-event-photo">
-        <img src="{{image.url}}" width=50 height=50>
-      </li>
-    {{/each}}
-    </ul>
+<li class="yxt-Result yxt-Result--event yxt-Result-item">
+  <div class="yxt-Result-wrapper">
+    {{! // TODO(bmcginnis): add date formatting in model }}
+    {{#if startDate.day}}
```
b1lly: I also don't see this as part of the core model/factory defined above.

src/ui/templates/results/resultsitem.hbs
```diff
@@ -1,11 +1,75 @@
...
+    {{#if bigDate.day}}
+      <div class="yxt-Result-dateWrapper">
+        <div class="yxt-Result-dateInnerWrapper">
+          <span class="yxt-Result-day">{{bigDate.day}}</span>
+          <span class="yxt-Result-month">{{bigDate.month}}</span>
+        </div>
+      </div>
+    {{/if}}
+
+    {{#if image.url}}
```
b1lly, line 20: Same comment about all these properties (ordinal, bigDate, image, etc) They all seem to be missing from the core model definition.

b1lly (changes requested):

src/core/models/result.js
```diff
@@ -2,22 +2,12 @@
...
-    }
-
-    return results;
+    this.title = data.title || null;
+    this.details = data.details || null;
+    this.link = data.link || null;
+    this.id = data.id || null;
+    this.startDate = data.startDate || null;
+    this.subtitle = data.subtitle || null;
+    this.callsToAction = data.callsToAction || null;
```
benmcginnis: i think we also want to have the raw profile data here.  So I think in here we should have `this._profile` or `this._raw` like I had before.

src/core/models/resultfactory.js
```diff
@@ -0,0 +1,33 @@
...
+
+    return ResultFactory.fromGeneric(data);
+  }
+
+  static fromGeneric (data) {
+    return Result({
+      title: data.name,
+      details: data.description,
+      link: data.website,
+      id: data.id
```
benmcginnis, line 44: and here you would pass `profile: data` or `raw: data`

src/ui/components/results/directanswercomponent.js
```diff
@@ -8,18 +8,56 @@ export default class DirectAnswerComponent extends Component {
...
     return 'DirectAnswer';
   }
+
+  /**
+   * The template to render
+   * @returns {string}
+   * @override
+   */
+  static defaultTemplateName () {
+    return 'results/directanswer';
```
benmcginnis: this is on another pr, please remove

src/ui/templates/results/eventresultsitem.hbs
benmcginnis: @ndawg you can actually remove the changes to the eventresultsitem.hbs, it's not actually ready for prime time.

src/ui/templates/results/eventresultsitem.hbs
benmcginnis: I think right now there is no way for us to generically define a call to action for every profile or even a default profile, other than "View Details" or something similar.  @ndawg can add the `this.callsToaction = null` field for completeness, but I think it's probably a lot more to think about for what the call to action model should look like and how implementors should be able to add them to a result as it's incredibly custom implementation to implementation and entity to entity

src/ui/templates/results/locationresultsitem.hbs
```diff
@@ -1,14 +1,48 @@
...
-  <div class="yext-location-result-item-info">
-    <h3 class="yext-location-name">{{name}}</h3>
-
-    <div class="yext-location-info">
-      <span class="yext-address">
-        {{address.line1}}, {{address.city}}, {{address.region}}, {{address.postalCode}}
-      </span>
-      <div>
-        <span class="yext-phone">Phone: {{mainPhone}}</span>
+<li class="yxt-Result{{#if modifier}} yxt-Result--{{modifier}}{{/if}} yxt-Result-item">
```
benmcginnis: @ndawg you can also reset the changes to this template, it's not ready for prime time.

src/ui/templates/results/resultsitem.hbs
benmcginnis: @ndawg added them though we still need to add `this.image = null`, `this.showOrdinal = false` and `this.callsToAction = []`

src/core/models/resultfactory.js
benmcginnis: @b1lly are you still good with this mechanism for preserving access to the raw profile?

src/ui/templates/results/resultsitem.hbs
benmcginnis: @b1lly this should be resolved.

src/ui/templates/results/resultsitem.hbs
```diff
@@ -1,11 +1,71 @@
...
-      <br>
-      {{displayLink}}
+    {{#if callsToAction}}
+      <div class="yxt-Result-ctasWrapper">
+        <div class="yxt-Result-ctas">
+          {{#each callsToAction}}
+            {{#if this.url}}
+              <a class="yxt-Result-cta" href="{{this.url}}"
+                {{#if eventType}} data-eventtype="{{eventType}}"{{/if}}
+                {{#if eventOptions}} data-eventoptions='{{eventOptions}}'{{/if}}>
```
benmcginnis: i think these actually need to be `this.eventType` and `this.eventOptions` in order to have the context of the iteration

src/core/models/result.js
```diff
@@ -2,22 +2,16 @@
...
-
-    return results;
+    this._raw = data.raw || null;
+    this.title = data.title || null;
+    this.details = data.details || null;
+    this.link = data.link || null;
+    this.id = data.id || null;
+    this.subtitle = data.subtitle || null;
+    this.modifier = data.modifier || null;
+    this.showOrdinal = data.showOrdinal || false;
```
b1lly: I think `showOrdinal` is a configuration property and not a data property (it's provided from the component).

You probably want `ordinal` here though, right?

Your template actually should use ``{{#if _config.showOrdinal}}`` I think? @benmcginnis 


Also, would you mind adding JS doc to each of the properties?

src/core/models/resultfactory.js
b1lly: Yes, raw is great :)

src/ui/sass/modules/_Result.scss
```diff
@@ -0,0 +1,140 @@
...
+  &-cta
+  {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    cursor: pointer;
+    padding-top: 8px;
+    padding-left: 8px;
+    padding-right: 8px;
+    padding-bottom: 8px;
```
b1lly, line 83: aka

`padding: 8px;`

Can you adjust this EVERYWHERE that you're setting padding?

I know this is a little bit of a nit and a micro optimization, but using short hands as much as possible is a good practice for file size optimizations. Over a pretty large code base they can add up.

src/ui/sass/modules/_Result.scss
```diff
@@ -0,0 +1,140 @@
...
+      margin-left: 16px;
+    }
+  }
+
+  &-ctaIconWrapper
+  {
+    padding-top: 10px;
+    padding-left: 10px;
+    padding-right: 10px;
+    padding-bottom: 10px;
```
b1lly, line 99: aka `padding: 10px;`

b1lly (changes requested):
SUPER close!

Some minor nits, but really liking the outcome :)

src/ui/sass/modules/_Result.scss
benmcginnis: So, generally we follow this practice in consulting: https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

We also use cssnano, which I forgot to include in the build tools and one of the things cssnano does is collapse these automatically but it will do the minimum shorthand thereby ensuring we don't accidentally clobber properties by using the short hand.

I've added css nano and the output css is using the short hand.

src/core/models/resultfactory.js
```diff
@@ -0,0 +1,62 @@
...
+   * @param data
+   * @returns {Result}
+   */
+  static fromGeneric (data) {
+    return new Result({
+      raw: data,
+      title: data.name,
+      details: data.description,
+      link: data.website,
+      id: data.id
```
b1lly, line 44: I think this makes sense for generic/default. Is the long term vision to support other entities for the other fields that aren't currently being utilized?

e.g. bigDate, image, etc.

I think they make sense for `events` or `people`. And I think those are the three entities we're planning to launch with. Perhaps we can utilize the `data.type` to know which entity it is to power these fields properly?

src/core/models/result.js
```diff
@@ -2,22 +2,15 @@
...
+    this._raw = data.raw || null;
+    this.title = data.title || null;
+    this.details = data.details || null;
+    this.link = data.link || null;
+    this.id = data.id || null;
+    this.subtitle = data.subtitle || null;
+    this.modifier = data.modifier || null;
+    this.bigDate = data.bigDate || null;
+    this.image = data.image || null;
+    this.callsToAction = data.callsToAction || [];
```
b1lly, line 66: just need jsdoc and we're good!

b1lly (changes requested):
Just need JSDoc for the main model and we're pretty good ;)

b1lly (approved):

PR #81, opened 2019-07-22
templates: update Search & Nav

benmcginnis:
Incorporate SearchBar & Nav templates and functionality based on
updated designs.


src/ui/sass/modules/_Nav.scss
```diff
@@ -0,0 +1,67 @@
+/** @define Nav */
+
+.yxt-Nav
```
aginsberg, line 3: Out of curiosity, why do we need to prepend all class names with 'yxt'? Is it to avoid conflicts in the event a client needs to iframe Answers?

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,5 +1,9 @@
-<nav class="yext-nav">
-{{#each tabs}}
-  <a class="yext-nav-item {{#if isActive}}active{{/if}}" href="{{url}}">{{label}}</a>
-{{/each}}
-</nav>
\ No newline at end of file
+<div class="yxt-Nav-modal js-yxt-navModal"></div>
+<nav class="yxt-Nav-container js-yxt-navContainer">
+    {{#each tabs}}
```
aginsberg: Looks like the consulting templates being brought over are using 4 spaces instead of 2. Probs not an issue, just wanted to make sure that's not an issue

aginsberg (commented):
LGTM

src/ui/sass/modules/_Nav.scss
benmcginnis: it's actually to prevent conflicts in case they _don't_ use an iframe. the iframe itself prevents conflicts.

src/ui/components/search/searchcomponent.js
```diff
@@ -217,7 +230,9 @@ export default class SearchComponent extends Component {
   setState (data) {
     return super.setState(Object.assign({
       title: this.title,
-      searchText: this.searchText,
+      labelText: this.labelText,
+      submitIcon: this.submitIcon,
+      submitText: this.submitText,
       query: this.query
     }, data));
```
b1lly, line 237: Seems fine. I have a longer term vision of separating model data from configuration data.

I haven't fully thought it through, but my initial thinking was in the BASE classes `setState` to always provide `this._opts` as a config object. 

`this._opts` is currently an object that contains all configuration provided from the public interface.

For instance, the Component base class would be something like:

```
setState(data) {
  return Object.assign(data, { config: this._opts });
}
```

That way all templates would always have access to all of the config options that are provided to the component, and templates would reference them via `{{config.labelText}}` as apposed to `{{labelText}}`.

That's not something you have to do in this change, but wanted to just throw it out there as an idea that we can do in future versions if it's something you think would be helpful.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +224,61 @@ export default class NavigationComponent extends Component {
...
+      item.classList.remove('yxt-Nav-item');
+    }
+
+    // Click listener to display/hide dropwdown, will be removed by DOM.empty in unMount()
+    moreButton.addEventListener('click', () => {
+      modal.classList.toggle('is-active');
+    });
+
+    // Close the modal when clicking outside of it, will be removed by DOM.empty in unMount()
+    window.addEventListener('click', (e) => {
```
b1lly: In this case, we need to manage the life cycle of these click handlers. This is going to leak memory and potentially cause wonky side effects after a few re-renders/mounts.

We should remove the existing handler before adding a new one because `window` is outside of the component DOM and is more of a global object.

For events that are bound inside the component, they matter less because the DOM gets torn down and rebuilt before its mounted, which'll automatically clean up the events that were bound to the old DOM.

Additionally, do we want to support the `escape` key for exiting the drop down as well?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +224,61 @@ export default class NavigationComponent extends Component {
...
+    for (const item of navItems) {
+      let childPos = item.offsetLeft;
+      let childWidth = item.offsetWidth;
+
+      // Determines if the nav item is overflowing and needs to be added to the dropdown
+      // Offsets by the width of the more button to make sure there is enough space to add it
+      if (childPos + childWidth > parentWidth - moreButtonWidth) {
+        toAppend.push(item);
+      }
+    }
```
b1lly: I think there's a big open question with product/ux right now about what behavior we actually want to support with this natively.

One thread is that only `navItems` that come back in response of a search should be shown at the top level, and the others should be under `more`. For instance, if you search, and two verticals come back `locations` and `events`, it should display `locations` and `events` primary links and all the other links would show up under `more`.

Now, that above doesn't support vertical search navigations being too long, or initial page loads being to long. So maybe some combination of both?

Your implementation in strictly based on the container width to hide excessive links. I'd be hesitant to include this directly into the library as core behavior atm until they finalize that decision. Let me try to get an answer today.

If the decision is what I mentioned above, i'll provide two different variables to the template for the navigation (tabs, moreTabs). Which should make a lot of the logic for constructing the nav just be part of the template rendering. And then the only work we'd need to do in javascript would be to show and hide the more.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +224,61 @@ export default class NavigationComponent extends Component {
...
+   * dropdown menu to display tabs least relevant
+   * to the query.
+   *
+   * @override
+   */
+  onMount () {
+    const parent = DOM.query(this._container, '.js-yxt-navContainer');
+    const navItems = DOM.queryAll(this._container, '.js-yxt-navItem');
+    const modal = DOM.query(this._container, '.js-yxt-navModal');
+    const moreButton = DOM.query(this._container, '.js-yxt-navMore');
```
b1lly: as a code style nit:

I like to have a suffix on {HTMLElement} objects with `el` for readability. It helps as people are scanning through the code to know that they have access to the DOM API without having to scroll to the declarations. It helps to distinguish the difference between DOM objects and what is an actual primitive object from memory.

```
const parentEl = DOM.query(...);
const navItemsEl = DOM.queryAll(...);
// etc
```

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +224,61 @@ export default class NavigationComponent extends Component {
...
+
+    // Shows the 'More' button if there are verticals to display in the dropdown
+    if (toAppend.length) {
+      moreButton.classList.remove('yxt-Nav-item--more');
+    }
+
+    // Adds cutoff elements to the dropdown
+    for (const item of toAppend) {
+      modal.appendChild(item);
+      item.classList.add('yxt-Nav-dropDownItem');
```
b1lly: NOTE:

DOM.addClass(node, 'class'), but we don't have one for remove or toggle. Feel free to use, add, or ignore!

b1lly (commented):
The search stuff looks good. The Navigation stuff i'm less set on because of product stuff. Perhaps two separate PRs?

src/ui/components/navigation/navigationcomponent.js
benmcginnis: I'll separate out the more logic from the core template update, so it can be addressed separately.

benmcginnis:
@b1lly I removed the controversial stuff, ready for you to take another look.

b1lly (approved):

PR #82, opened 2019-07-22
templates: allow implementors to set config.templateName

benmcginnis:
config.templateName was defined but not used by any of the Components.  This
changes each component to respect config.templateName if it's defined and if
not, fallback to the default templateName for the Component.  For Components
with multiple allowable templates, this will allow implementors to choose
a variation without overriding the template parameter.


Delerme:
seems fine to me, to billy

b1lly (commented):
I think this makes sense.

Just an idea: 

One alternative approach could be to update the base `Component` class to do something like:

```
this._templateName = config.templateName || this.constructor.defaultTemplateName() ;
```

And then, the implementation pattern wouldn't be to deal with the configuration for assignment in the constructor. Instead, each subclass would just have to create a static method override like:

```
static defaultTemplateName() {
  return 'search/filtersearch';
}
```

It's the same thing, but I think it abstracts away the `configuration` stuff to that is supposed to be part of the core configuration into the base class, and creates an 'override' pattern to follow. 

Either way works for me, though. Just wanted to see your thoughts here.

benmcginnis:
That's a lot nicer, updating.

benmcginnis:
@b1lly updated based on your comment.

src/ui/components/component.js
```diff
@@ -137,7 +137,7 @@ export default class Component {
      * This is only used if _template is empty.
      * @type {string}
      */
-    this._templateName = opts.templateName || 'default';
+    this._templateName = opts.templateName || this.constructor.defaultTemplateName();
```
b1lly: I just realized you might have to provide `opts` to this.constructor.defaultTemplateName(opts);

There's only one use case thats going to need it (see below).

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -67,19 +67,21 @@ export default class FilterOptionsComponent extends Component {
...
   static get type () {
     return 'FilterOptions';
   }
 
+  /**
+   * The template to render, based on the control
+   * @returns {string}
+   * @override
+   */
+  static defaultTemplateName () {
```
b1lly: Was this tested, does this work? If it does, i'd be surprised, since `this` should be pointing at the class object and not the instance object... But if its magically working then great,  can ignore what I say below.


If its not working we need to do what I suggested above and:

```
static defaultTemplateName (config) {
  return `controls/${config.control}`
}
```

Unless you got a better idea?

b1lly (changes requested):

b1lly (approved):

PR #83, opened 2019-07-22
Add postcss configuration

benmcginnis:
Add postcss to handle browser prefixing and conversion of pixels to rems with
a base font size of 16 pixels.  So a pixel measurement of 20px would be
converted to  1.25rem in the output answers.css.  Additionally, convert line-
height measurements to unitless measurements.

This allows ensures Answers Lib's compliance with WCAG 2.1 Success Criteria 1.4.4.


b1lly (approved):
LGTM

aginsberg (approved):

PR #84, opened 2019-07-23
Static mapbox custom markers

aginsberg:


tests/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -53,6 +53,25 @@ describe('serizalize an array of MapBoxMarker configs', () => {
...
+      map: map,
+      position: { latitude: 44, longitude: 45 },
+      label: '1',
+      staticMapPin: 'https%3A%2F%2Fimg.png'
+    }),
+    new MapBoxMarkerConfig({
+      map: map,
+      position: { latitude: 30, longitude: 31 },
+      label: '2'
+    })];
```
b1lly: Is this not the same as:

```
const markers = MapBoxMarkerConfig.from([{
  map: map,
  position: { ... }
  // ... etc
},
{
  map: map,
  position: { ... }
}]
```

b1lly (changes requested):
Nice job!

tests/ui/components/map/providers/mapboxmapprovider.js
aginsberg: Not sure I understand, in the case of the first marker, it doesn't need the label since it's using a url for the marker image. In the second one, it's using the marker's label as part of the encoded marker, i.e. the expected output from this test is 'url-https%3A%2F%2Fimg.png(45,44),pin-s-2(31,30)', the 'pin-s-2' is using the label for the 2.

b1lly: I'm saying the way that you're constructing your list of MapBoxMarkerConfigs can be done more simply by using the .from() method and passing it the array of object literals.

aginsberg: Ah I see, I'll change this

b1lly (approved):

PR #86, opened 2019-07-24
adding filter reset option to FiltersBoxComponent

aginsberg:


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -136,6 +149,25 @@ export default class FilterBoxComponent extends Component {
...
+  /**
+   * Clear filers in storage and uncheck all selected filters
+   * @private
+   */
+  _clearFilters () {
+    this._filters = [];
+    for (const filterComponent of this._filterComponents) {
+      if (filterComponent._type === 'FilterOptionsComponent') {
+        filterComponent.clear();
+        for (const option of filterComponent._options) {
```
Delerme: this shouldn't access filterComponent's private properties, instead have filterOption.clear() do all the necessary clearing it should do, so move this there

Delerme (changes requested):
any template changes needed?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -136,6 +149,22 @@ export default class FilterBoxComponent extends Component {
...
+  _clearFilters () {
+    this._filters = [];
+    for (const filterComponent of this._filterComponents) {
+      if (filterComponent._type === 'FilterOptionsComponent') {
+        filterComponent.clear();
+      }
+    }
+
+    this._saveFiltersToStorage();
+    this._search();
```
b1lly: I don't think clearFilters should be automatically triggering a search. I think that should be a decision made by the code that's calling to clear filters. For instance, what if someone wanted to clear the filters, select a few other things, and then search? By separating them, you're able to reuse this method and just re-implement a new caller,.

This should be part of the resetButton handler. I feel LESS strongly about decoupling the saving of filters with the clear action, but ultimately i'm leaning towards also making that the responsibility of the caller.

For instance, the code should be more:

```
DOM.on(resetButton, 'click', () => {
  this._clearFilters();
  this._saveFiltersToStorage();
  this._search();
});

File: src/ui/templates/filters/filterbox.hbs [hunk 3]
```diff
@@ -1,10 +1,34 @@
...
-      <button type="button" class="js-yext-filterbox-apply">Apply</button>
+<div class="FilterBox{{#if mod}} FilterBox--{{mod}}{{/if}} js-filterbox">
+  <div class="FilterBox-titleWrapper">
+    {{#if toggle}}
+      <button class="FilterBox-toggle js-filterBoxToggle">
+        {{#if title}}
+          {{title}}
+        {{else}}
+          Filters
+        {{/if}}
```
b1lly, line 9: @Delerme do we want to figure out a pattern for defaulting data values? Right now we'd just override setState, so that we wouldn't need to do this type of checking in the template. I think if ``{{title}}`` defaulted to a `Filters in the component, that probably would be a better approach to this?

That would also give the component the flexibility of a configuration option to override it.

b1lly (changes requested):

b1lly (changes requested):
Please rebase this properly so the review doesn't include other peoples commits.

aginsberg:
@b1lly Is there anything else I can do here? I believe I addressed all the requested changes, but please let me know if I missed anything.

b1lly:
What is the deal with this? is this something that is still needed? Is it dead? can I close it? Should we figure out how to merge it in?

b1lly:
I'm going to close this, if no response @benmcginnis @aginsberg

benmcginnis:
Hey @b1lly sorry this slipped my mind.  We probably do need a solution to this in the library for WCAG reasons, but we can probably close this PR, I think it needs more thinking from UX.

Issue #89, opened 2019-07-24
`businessId` should be supported as initialization interface


b1lly:
Instead of analytics depending on an initial server request, let it depend on businessId provided from the interface.

In a future version we'll make it a required field.

PR #91, opened 2019-07-24
Make businessId optional argument

Delerme:
Instead of automatically fetching the businessId from the server, have it as an
option in the config. Currently optional, and analytics will be disabled without
it. Closes #89


b1lly (approved):

b1lly:
Just let consulting know if this change.

PR #92, opened 2019-07-24
adding ability to show number of applied filters

aginsberg:


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -148,5 +162,18 @@ export default class FilterBoxComponent extends Component {
...
+    if (this._showNumFilters && numElement) {
+      const filters = DOM.queryAll(this._container, this._inputSelector);
+      const numAppliedFilters = filters.filter(e => e.checked).length;
+
+      if (numAppliedFilters > 0) {
+        numElement.innerHTML = numAppliedFilters;
+      } else {
+        numElement.innerHTML = '';
+      }
+    }
```
b1lly, line 177: This is not how the problem should be solved. If we want to provide the num of applied filters to the template, it should be done through the data (setState), not through directly injecting it into the DOM.

Additionally, if there are features/enhancements you need/want to build. We should certainly have conversations around them first before just submitting pull requests for them. You can do this through opening a issue in github, or posting in #answers-js. Once we have consensus on the feature/enhancement and the approach, then it would be ok for either team (consulting or engineering) to implement.

Regardless, any feature enhancement should be accompanied by a GitHub issue at the very least first.

I don't think this one is that big of a deal, but in general I think ANY new feature or enhancement should have conversations about before just contributing code.

@Delerme

b1lly (changes requested):

src/ui/components/filters/filterboxcomponent.js
aginsberg: Sorry about that, I'll be sure to make issues in the future

PR #93, opened 2019-07-25
results: section config updates

benmcginnis:
Modify UnversalResultsComponent to pass section specific configuration down to
the Section model so it can be rendered by the resultssectionheader.hbs
template.

Additionally, modify the Section model to provide a default sectionTitle param
to resultssectionheader.hbs by deriving it from the verticalConfigId by default.


src/ui/components/results/universalresultscomponent.js
```diff
@@ -37,4 +37,16 @@ export default class UniversalResultsComponent extends Component {
...
+    if (data.sections) {
+      data.sections = data.sections.map(section => {
+        if (this._opts && this._opts.config && this._opts.config[section.verticalConfigId]) {
+          return Object.assign(section, this._opts.config[section.verticalConfigId]);
+        }
+        return section;
+      });
+    }
+    return super.setState(data);
+  }
```
b1lly, line 61: I think what you want to do is manage this configuration as part of the configuration in the `ResultsComponent`, classes constructor. And within the ResultsComponent class, use the `setState` there instead to pipe these config options into its local state.

I know `ResultComponent` is ugly, all the options that are available here should also already be available in `this._opts` from the `ResultsComponent` (without having to access the section.verticalConfigId). If you want to be explicit about which configuration options you're maintaining within that class (for maintainability), you should explicitly add `this.[optionname] = opts.whatever` so that people reading the component know which config options are available.

b1lly (changes requested):

benmcginnis:
I think the updates on v0.8.0 make this obsolete, i'm going to close this pr

PR #94, opened 2019-07-25
Prevent duplicate searches in core

Delerme:


src/core/core.js
```diff
@@ -74,9 +74,27 @@ export default class Core {
...
+     * back-to-back
+     * @type {object}
+     * @private
+     */
+    this._lastSearch = {
+      query: '',
+      urls: {},
+      filter: ''
+    };
   }
```
b1lly: I actually think that this behavior should be prevented in the UI library instead of the core.

src/core/core.js
```diff
@@ -104,6 +128,11 @@ export default class Core {
    * @param {string} namespace the namespace to use for the storage key
    */
   autoCompleteUniversal (input, namespace) {
+    if (input === this._lastSearch.query) {
```
b1lly: As a note, I'm not sure we care about restricting the auto complete behavior. We certainly can, but I don't think its really necessary.

b1lly (changes requested):

src/core/core.js
Delerme: Yea figured it didn't really matter either way, but may as well

src/core/core.js
Delerme: hmm it seems better to prevent it at the core level, I don't think we want ill-configured components or templates to ever spam the api with duplicate requests

src/ui/components/search/searchcomponent.js
```diff
@@ -108,6 +108,14 @@ export default class SearchComponent extends Component {
...
      */
     this.query = opts.query || this.getUrlParams().get('query') || '';
+
+    /**
+     * The minimum time allowed in milliseconds between searches to prevent
+     * many duplicate searches back-to-back
+     * @type {number}
+     * @private
+     */
+    this._searchCooldown = opts.searchCooldown || 1000;
```
b1lly: do we want to rename opts = config while we're here? Also, I think 1 second is too long. Lets start with like 200-300ms

b1lly (changes requested):

src/ui/components/search/searchcomponent.js
Delerme: yea may as well. Changed to 300

b1lly (approved):
Might need to update the readme too, to say its default is 300ms not 1s

Otherwise LGTM

PR #96, opened 2019-07-25
Filter options

aginsberg:


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -68,6 +68,13 @@ export default class FilterOptionsComponent extends Component {
      */
     this._onChange = config.onChange || function () {};
 
+    /**
+     * The label to br used in the legend
```
Delerme: typo

Delerme (changes requested):
any template changes needed?

src/ui/components/filters/filteroptionscomponent.js
aginsberg: oops, will fix
I can include the template change in this cr as well

src/ui/templates/controls/multioption.hbs
```diff
@@ -1,10 +1,21 @@
...
-    value="{{value}}"
-    data-index="{{@index}}"
-    {{#if checked}}checked{{/if}}>
-  <label for="js-yext-checkbox-{{../name}}-{{@index}}">{{label}}</label>
-{{/each}}
+<fieldset class="Multioption-fieldset">
+  <legend class="Multioption-legend">
+    {{#if label}}
+      {{label}}
+    {{else}}
```
Delerme: I'd probably remove the if/else logic if label is always expected to be defined

Delerme (changes requested):

src/ui/templates/controls/multioption.hbs
aginsberg: Good point, will remove

Delerme:
lgtm, to billy

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -20,6 +20,7 @@ const SUPPORTED_CONTROLS = [
 export default class FilterOptionsComponent extends Component {
   constructor (config = {}) {
     super(config);
+    console.log(config)
```
b1lly: woops

b1lly (changes requested):

src/ui/components/filters/filteroptionscomponent.js
aginsberg: This has been removed!

b1lly (approved):
nice job.

PR #101, opened 2019-07-26
Return null instead of empty object for undefined state

Delerme:
Confirmed the only two places that use it expect null


b1lly (approved):
LGIT

PR #105, opened 2019-07-29
Adds dynamic filters component

Delerme:


src/core/core.js
```diff
@@ -76,13 +76,16 @@ export default class Core {
     });
   }
 
-  verticalSearch (queryString, verticalKey, filter) {
+  verticalSearch (searchOptions) {
     return this._searcher
-      .verticalQuery(queryString, verticalKey, filter)
+      .verticalQuery(Object.assign({}, searchOptions, { isDynamicFiltersEnabled: this._isDynamicFiltersEnabled }))
```
b1lly, line 84: why is dynamicFiltersEnabed part of the core? Seems like it should be part of the UI component

src/core/models/dynamicfilters.js
```diff
@@ -0,0 +1,32 @@
+/** @module DynamicFilters */
+
+/**
+ * Model representing a set of dynamic filters
+ */
+export default class DynamicFilters {
+  constructor (data) {
+    this.filters = data.filters || [];
```
b1lly, line 12: jsdoc?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -83,6 +92,12 @@ export default class FilterBoxComponent extends Component {
   }
 
   onMount () {
+    if (this._filterComponents.length) {
+      this._filterComponents.forEach(c => c.remove());
+      this._filterComponents = [];
+      this._filters = [];
+    }
```
b1lly, line 99: I'd consider making a filters model that is a wrapper around the `array` filterComponents (e.g. `FitlerComponentCollection`) that expose an interface for add, remove, clear, get, etc.

then you can just do
```
this._filterComponents = new FilterComponentCollection();
```
```
this._filterComponents.clear();
this._filterComponents.remove(c)
this._filterComponents.add(c) // or push if you want to be consistent
```

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -124,16 +140,30 @@ export default class FilterBoxComponent extends Component {
...
    * @private
    */
   _saveFiltersToStorage () {
-    const validFilters = this._filters.filter(f => f !== undefined && f !== null);
-    const combinedFilter = validFilters.length > 1
-      ? Filter.and(...validFilters)
-      : validFilters[0];
-    this.core.setFilter(this.name, combinedFilter || {});
+    const validFilters = this._filters
+      .filter(f => f !== undefined && f !== null && Object.keys(f).length > 0);
```
b1lly: neat! Can you expand this logic out so its more legible.

b1lly (changes requested):

b1lly:
Is it possible to add tests for some of the model/transformation stuff? Or no?

src/core/core.js
Delerme: `dynamicFiltersEnabled` determines whether or not api requests include `retrieveFacets` or not. We don't want to default to always requesting because retrieving facets is a significant performance hit. 

This property just happens to be controlled by the presence of a UI component, but is totally detached. The only purpose it serves is to optionally fetch more data from the backend, and the ui lib sets it to true in some circumstances

There may be a better name for this property if you have some ideas

src/core/models/dynamicfilters.js
Delerme: done

src/ui/components/filters/filterboxcomponent.js
Delerme: hmm maybe. Main issue here is with clearing the old components and I don't think `this._filterComponents.clear()` is that much more clear than `this._filterComponents = []`. Punting for now

src/ui/components/filters/filterboxcomponent.js
Delerme: Done

src/ui/components/filters/dynamicfilterscomponent.js
```diff
@@ -0,0 +1,102 @@
...
+      Object.assign({}, {
+        parent: this,
+        name: `${this.name}.filterbox`,
+        container: '.js-yext-dynamic-filters',
+        searchOnChange: this._searchOnChange,
+        verticalKey: this._verticalKey,
+        isDynamic: true,
+        filters
+      })
+    );
```
b1lly, line 98: If someone changes the template, would this break pretty easily? Would using markup syntax make this a little less fragile?

src/ui/components/filters/filterboxcomponent.js
b1lly: SGTM.

Just to clarify, I was showing the higher level interface it would expose.

`filterComponents.clear()` would internally iterate on each component, call `c.remove()` and set the value to empty `[]`.

Probably not the best name.... but was just elaborating.

src/core/core.js
b1lly: Yeah, but why is it referencing `this._isDynamicFiltersEnabled`?

Shouldn't it just be provided by searchOptions, which would default to false if not provided?

Yeah, I don't feel strongly about the name but some alternatives: `fetchDynamicFilters` or `retrieveDynamicFilters`.

b1lly (approved):

src/core/core.js
Delerme: once dynamic filters is enabled, it should be disabled, i.e. it should be enabled once at the start if appropriate and then sent with every request. If it were included with searchOptions, you'd have to guarantee every caller of verticalSearch properly set it, which isn't feasible

src/ui/components/filters/dynamicfilterscomponent.js
Delerme: not sure what you mean but can do in a different commit

src/ui/components/filters/filterboxcomponent.js
Delerme: yea seems like premature abstraction but if we add any more manipulations I'll wrap it

PR #107, opened 2019-07-29
templates: update NavigationComponent to add "more" dropdown

benmcginnis:
When there are too many links to fit in the NavigationComponent's container,
move items that are too long into a "more" dropdown.


benmcginnis:
@b1lly I know the requirements for this are in flux, but we need this landed for the existing implementations.  Once Jeremy et al nail down how they want this to behave, we're happy to update.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -114,6 +115,13 @@ export default class NavigationComponent extends Component {
      * @private
      */
     this._tabOrder = this.getDefaultTabOrder(config.tabs, this.getUrlParams());
+
+    /**
+     * Handles to the close modal click listeners, to be removed in unMount()
+     * @type {array}
```
Delerme, line 114: can you specify the item types too? can do something like @type {object[]} if it's a generic item type

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +231,73 @@ export default class NavigationComponent extends Component {
...
   }
+
+  /**
+   * If there are too many tabs, create a "more"
+   * dropdown menu to display tabs least relevant
+   * to the query.
+   *
+   * @override
+   */
+  onMount () {
```
Delerme, line 244: I haven't seen the mocks so this is a little hard to visualize, but I think it can done with css instead of all this imperative logic?

benmcginnis: It looks like this: Screen Shot 2019-07-29 at 3 44 39 PM

I don't think we can really do it with css.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -223,4 +231,73 @@ export default class NavigationComponent extends Component {
...
+    const closeModalClickListener = (e) => {
+      // TODO (bmcginnis): this is not robust if they have an image in the button or modal
+      if (e.target === modalEl || e.target === moreButtonEl) {
+        return;
+      }
+      modalEl.classList.remove('is-active');
+    };
+
+    window.addEventListener('click', closeModalClickListener);
+    this._listenersToTearDown.push({ type: 'click', fn: closeModalClickListener });
```
b1lly, line 292: This is neat! In the future, I want to bake this logic into the the DOM library with a notion of `namespacing`. That way we can provide namespaces to handlers we're binding or removing.

e.g. imagine `NavComponent` as a string:

```
DOM.on(window, 'click', 'NavComponent', () => { });
```
```
DOM.off(window, 'click', 'NavComponent');
```

Or in JQUERY world:

```
DOM.on(window, 'click.NavComponent', () => {})
DOM.off(window, 'click.NavComponent');
```

b1lly (commented):
This seems fine to me. But, on any other day I'd say this `shouldn't` be in the library. You should be able to accomplish the same thing without it being part of the core. This is very specific to the implementation cases. And we don't know what the best implementation would be yet.

I'm fine with adding it though, as long as your OK with the fact that we can rip this behavior out and change it at any point, and your customers wont be mad with it magically changing or breaking then this LGTM.

I certainly don't want to be blocked by existing integrations for changing this code.

In the future, i'd like to avoid adding things to the library until we're certain of the behavior we want to support. I want to avoid others integrating or depending on behavior that could change at any point in time.

benmcginnis:
@b1lly yeah I'm ok with this changing whenever, to be clear this design came from Jeremy / Product not consulting.  We just implemented it because we had to have it for the Yext Answers implementation, but they want this for all clients.  

What about it makes you feel like it shouldn't be in the core component library?

src/ui/components/navigation/navigationcomponent.js
benmcginnis: Along those lines, perhaps something to consider would be for the core Component should expose an 
`addListener(fn, type = 'click', target = this._container)` method that automatically keeps track of listeners that need to be removed in teardown.  That way subclasses wouldn't have to worry about it?  The DOM library could also have this.

b1lly:
> @b1lly yeah I'm ok with this changing whenever, to be clear this design came from Jeremy / Product not consulting. We just implemented it because we had to have it for the Yext Answers implementation, but they want this for all clients.
> 
> What about it makes you feel like it shouldn't be in the core component library?

Yup, I understand the mocks you're getting are from Product and Jeremy. I think the disconnect there is that a lot of what we're doing in these early integrations are much more "UX Experiments" than final vision things. We have a lot to learn!

What you guys are building for the early iterations doesn't necessarily mean its "finalized" in terms of product or ux vision. In a lot of ways, we don't know what the best practice or behavior is that we want to give to others. Some of the implementations you're doing are at the cost of "trade-offs", time-constraints, and limitations of the software, while others are because the clients push back for various things. 

Again, I think it's important to use the first 5-10 integrations to really understand what we want. So in my mind, the decision we made for the navigation behavior are "Consulting specific". It may be the case that we end up wanting it for the core library, but its highly likely that its not the end result. I anticipate a lot of iteration on the navigation behavior (and other things) over the next month or so. 

I want to avoid people expecting or depending on a component working a particular way, if we're not sure how its going to work a revision after. And because of that, I want to be more deliberate about the behaviors we include into the library.

src/ui/components/navigation/navigationcomponent.js
b1lly: Great idea! Made an item #115

PR #110, opened 2019-07-30
adding alt tags to static map images

aginsberg:
issue #110


src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -51,7 +51,7 @@ export default class GoogleMapProvider extends MapProvider {
 
     let encodedMarkers = GoogleMapMarkerConfig.serialize(googleMapMarkerConfigs);
     return `
-      <img src="//maps.googleapis.com/maps/api/staticmap?${encodedMarkers}&size=${this._width}x${this._height}&${this.generateCredentials()}">`;
+      <img src="//maps.googleapis.com/maps/api/staticmap?${encodedMarkers}&size=${this._width}x${this._height}&${this.generateCredentials()}" alt="">`;
```
aginsberg, line 54: @benmcginnis Do you think it makes sense for the alt tags to be empty here?

benmcginnis: I think we need to think about this, can you read these resources and lmk what you think?

https://webaim.org/techniques/alttext/
https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html

b1lly: @aginsberg @benmcginnis i'm not sure empty alt tags add any value? But, i'm also not an accessibility expert so any type of advice/direction here you guys can provide would be great.

aginsberg: Most screen readers would read out the url of the image if there is no alt tag, so an empty alt tag would allow us to avoid that. We typically do this for images that are purely decorative and don't contain content that is important for the user to know about. 
The issue with the static map is that it's difficult to provide alt text this is descriptive enough to convey to the user what is on the map, i.e. where the pins are, what they represent

benmcginnis:
It's one of the reasons we prefer interactive maps as they're easier to
ensure we're fully WCAG compliant.  Alex my thoughts are two-fold.  I think
we probably either want to make this a <figure> element where the <legend>
contains all of the results OR use an aria-describedby attribute
referencing the results list.  I'm not sure which is the best approach and
we should probably test the experience on Voice Over, Chrome Vox and NVDA
at a minimum.

What I want to do as we get closer to launching is a full WCAG audit on the
library itself with the screen readers above and JAWS (we need to buy a
license).

src/ui/components/map/providers/googlemapprovider.js
b1lly: Are we not able to do something generic for this e.g. alt="Displaying entities as pins on a map"?

benmcginnis: Sorry, in thinking about this, it occured to me that we can't use static map images at all if we want to be WCAG compliant. The main reason is that the font size, color, and background color of the text inside the map image has to be configurable by the user's browser font size settings.  See: https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-text-presentation.html for details.

aginsberg:
Closing as static maps can't be WCAG compliant

PR #111, opened 2019-07-30
Add search states

Delerme:


src/core/models/verticalresults.js
```diff
@@ -1,17 +1,26 @@
 /** @module VerticalResults */
 
 import Section from './section';
+import SearchStates from '../storage/searchstates';
 
 export default class VerticalResults {
   constructor (data = {}) {
-    Object.assign(this, data);
+    Object.assign(this, { searchState: SearchStates.RESULTS }, data);
```
b1lly: what if the data contains no results?

b1lly (changes requested):
This implementation seems fine, but as a discussion thing.
Would it be better if the templates could act on booleans instead of having to do string matching?

For instance, I think think the ideal case would be something like:

```
{{#if isPreLoad}}

{{/if}}
```
```
{{#if isSearchLoading}}

{{/#if}}
```
```
{{#if hasResults}}

{{#else}}
   // has no results
{{/if}}
```

Thoughts? I'm open to either but the above seems like the most ideal if possible.

src/core/models/verticalresults.js
Delerme: the `results` property can be checked for no results, notably vertical results already does this. So RESULTS signifies that the search is complete, not necessarily that results exist. Worth renaming?

Delerme:
the separate booleans kind of implies multiple of them can be true, which is not the case, but I guess as long as we make that clear it doesn't matter, I'll switch it over

Delerme:
added the boolean props and renamed RESULTS to SEARCH_COMPLETE

src/ui/components/results/resultscomponent.js
```diff
@@ -68,11 +69,11 @@ export default class ResultsComponent extends Component {
...
   setState (data, val) {
-    if (Object.keys(data).length === 0) {
-      return this;
-    }
-
+    const searchState = data.searchState || SearchStates.PRE_SEARCH;
     return super.setState(Object.assign({}, data, {
+      isPreSearch: searchState === SearchStates.PRE_SEARCH,
+      isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
+      isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE,
```
b1lly, line 76: we can also create a SearchState(searchstate) class/helper; that returns this object. so that it can be shared amongst the results components.

b1lly (approved):
nice.

src/ui/components/results/resultscomponent.js
Delerme: true, will keep that in mind if we have to duplicate this later

Issue #104, opened 2019-07-29
IE11: hitting enter to search does not work when selecting autocomplete option


aginsberg:
Video here shows autocomplete not working properly on ie11 https://drive.google.com/file/d/1Q0fWZRMq6iB-U992Bv2C72RDSk37a5PH/view

aginsberg:
@b1lly @Delerme Is it okay for Consulting to work on this?

Delerme:
Is this for search bar or filter search?

aginsberg:
This is for search bar autocomplete

Delerme:
Seems fine to me, I'd try to apply the same fix to filter search when you find it too

PR #113, opened 2019-07-30
#104: fixing search and autocomplete functionality in IE

aginsberg:
Removing Object.assign for browser compatibility: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility

Removing Event constructor for browser compatibility: https://developer.mozilla.org/en-US/docs/Web/API/Event/Event#Browser_compatibility

Using correct version of matches depending on browser: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill


src/ui/dom/dom.js
```diff
@@ -175,6 +188,10 @@ export default class DOM {
 
   static delegate (ctxt, selector, evt, handler) {
     let el = DOM.query(ctxt);
+    if (!Element.prototype.matches) {
```
aginsberg, line 191: This doesn't seem like the best place to put this, any ideas on how to improve this?

Delerme:
I don't know if we've had an official discussion yet, but I think we're moving toward the lib not supporting IE by default, and instead having users polyfill when necessary, similar to other npm packages. We could optionally polyfill it ourselves and have a separate IE bundle, but either way I don't think we'll be removing usages of things that aren't supported in IE for those reasons.

@b1lly

aginsberg:
> I don't know if we've had an official discussion yet, but I think we're moving toward the lib not supporting IE by default, and instead having users polyfill when necessary, similar to other npm packages. We could optionally polyfill it ourselves and have a separate IE bundle, but either way I don't think we'll be removing usages of things that aren't supported in IE for those reasons.
> 
> @b1lly

@Delerme @b1lly This can all be handled on the Pages side, if we don't want to be handling it in the lib. I will close this CR as well as the issue

PR #116, opened 2019-07-30
Send query ID with every analytic event

Delerme:


src/answers-umd.js
```diff
@@ -54,16 +55,19 @@ class Answers {
...
+    });
+    this.components.setCore(core)
       .setRenderer(this.renderer);
 
     if (config.businessId) {
+      const reporter = new AnalyticsReporter(config.answersKey, config.businessId, config.analyticsOptions);
       this.components
-        .setAnalyticsReporter(new AnalyticsReporter(config.answersKey, config.businessId, config.analyticsOptions));
+        .setAnalyticsReporter(reporter);
+      core.storage.on('update', storagekeys.QUERY_ID, id => reporter.setQueryId(id));
```
b1lly: SORRY! 

I changed my mind. Lets have AnalyticsReporter depend on the core, and let it handle these implementation details. This seems more like a behavior of the reporter than the caller.

b1lly (changes requested):

src/answers-umd.js
Delerme: done

src/core/analytics/analyticsreporter.js
```diff
@@ -4,12 +4,13 @@ import ApiRequest from '../http/apirequest';
...
 import { AnswersAnalyticsError } from '../errors/errors';
 import { ANALYTICS_BASE_URL } from '../constants';
+import StorageKeys from '../storage/storagekeys';
 
 /**
  * Class for reporting analytics events to the server
  */
 export default class AnalyticsReporter {
-  constructor (answersKey, businessId, globalOptions = {}) {
+  constructor (core, answersKey, businessId, globalOptions = {}) {
```
b1lly, line 13: we going to probably want to make this configuration an object in the future

b1lly (approved):

src/core/analytics/analyticsreporter.js
```diff
@@ -18,6 +19,9 @@ export default class AnalyticsReporter {
      * @private
      */
     this._globalOptions = Object.assign({}, globalOptions, { answersKey });
+
+    // listen to query id updates
+    core.storage.on('update', StorageKeys.QUERY_ID, id => this.setQueryId(id));
```
b1lly, line 24: you probably want to store a reference to the core.

this._core = core;

b1lly (approved):

Issue #114, opened 2019-07-30
Autocomplete updates


aginsberg:
Some bundled changes for global visual QA:

1. Left align autocomplete options
2. Remove borders between autocomplete options
3. Update template to use SUIT class naming conventions

PR #119, opened 2019-07-31
Updating autocomplete class naming and styling (#114)

aginsberg:


Delerme:
It seems like we should be using either hyphen or camelCase for classes, but not both. Doesn't really matter but seems kind of strange to me. Any thoughts?

aginsberg:
> It seems like we should be using either hyphen or camelCase for classes, but not both. Doesn't really matter but seems kind of strange to me. Any thoughts?

We try to follow SUIT naming conventions https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md, so .yxt-AutoComplete is the base component, anything followed by a '-' is a descendant, and '--' indicates a modifier. I am trying to use the same naming convention that was used here https://github.com/yext/answers/pull/81/files#diff-9156501bba9bf6edadf5b90e9d3398b1R4

benmcginnis:
@Delerme yeah, the only thing about this that's weird / doesn't really fit is the name spacing that @b1lly wants the library to have so we potentially can avoid style collisions with clients' css.  But it works.  The convention is basically `.namespace-Module-child` and `.namespace-Module--modifier` . The benefit of adopting this naming convention is that it's very clear what's a module and what's an element, and you can easily separate concerns between modules.

If you're familiar at all with [BEM](http://getbem.com/introduction/), SUIT is basically the same thing, but is easier to type and read.

src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -0,0 +1,41 @@
...
+  overflow: hidden;
+
+  &-option
+  {
+    padding: 12px;
+
+    &.yxt-selected
+    {
+      color: white;
+      background-color: black;
```
b1lly: do we want to use variables for colors? perhaps creating a `_colors.scss` file that gets imported for use? @benmcginnis

src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -0,0 +1,41 @@
...
+    {
+      cursor: pointer;
+
+      &:hover
+      {
+        background-color: gray;
+      }
+    }
+  }
+}
```
b1lly, line 46: Do you want to remove the dead autocomplete scss code from answers.scss?

src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -0,0 +1,41 @@
...
+    &.yxt-selected
+    {
+      color: white;
+      background-color: black;
+    }
+
+    strong
+    {
+      font-weight: 700;
+    }
```
b1lly, line 34: should there be some base file for `fonts`?

b1lly (changes requested):
A few questions about long term organization of code.

src/ui/sass/modules/_AutoComplete.scss
aginsberg: For Pages, we have a single 'variables.scss' file, where we store things like colors, font-weights, etc. Would it make sense for me to add that?

src/ui/sass/modules/_AutoComplete.scss
aginsberg: @b1lly It looks like answers.scss only contains file imports now. Is there a different file you're referring to?

src/ui/sass/modules/_AutoComplete.scss
aginsberg: By base file, do you mean someplace to store variables like this?

src/ui/sass/modules/_AutoComplete.scss
aginsberg: @b1lly I just pushed a commit that shows what it might look like with a variables.scss file. If we like this way I can refactor the Nav and Searchbar scss files to use this as well

src/ui/sass/modules/_AutoComplete.scss
benmcginnis: Yeah I removed the styles in a previous pr.

src/ui/sass/modules/_AutoComplete.scss
benmcginnis: @b1lly yes, I was thinking about doing this I just hadn't written my thoughts down yet.  I think we want to handle font styles & colors in a similar way by setting defaults.

src/ui/sass/_variables.scss
```diff
@@ -0,0 +1,12 @@
+// Font Variables
+$font-weight-bold: 700;
```
benmcginnis: I think you want to mark these with `!default` so implementers can override with their own settings and they will take precedence

src/ui/sass/modules/_AutoComplete.scss
b1lly: Oh, woops!

src/ui/sass/_variables.scss
```diff
@@ -0,0 +1,12 @@
...
+$font-weight-semibold: 600;
+$font-weight-normal: 400;
+$font-weight-light: 300;
+
+// AutoComplete Variables
+$autocomplete-background-color: white;
+$autocomplete-box-shadow: 0 4px 6px 0 rgba(32, 33, 36, 0.28);
+$autocomplete-selected-color: white;
+$autocomplete-selected-background-color: black;
+$autocomplete-option-hover-background-color: gray;
```
b1lly: If they're autocomplete specific, I don't see harm in defining them at the header of the autocomplete scss. Thoughts @benmcginnis ?

src/ui/sass/modules/_AutoComplete.scss
b1lly: Yup, I was thinking longer term there'd probably be things for like:

```
_fonts.scss
_colors.scss
_layouts.scss
```
etc. etc.

b1lly (changes requested):

src/ui/sass/_variables.scss
benmcginnis: I generally like the approach of putting things in their corresponding modules, but I think bootstrap goes with a global variable file for a reason: it's easier for implementors to see all of the variables that they might override.

I don't really feel strongly, though.

benmcginnis:
@aginsberg I think you need to rebase this on v0.8.0

b1lly (approved):
Nice :)

Issue #118, opened 2019-07-31
Unexpected cmd + click behavior for result card links


creotutar:
Context: After upgrading to v0.7.4 (was previously at b98060dd1fba75ef09ac19e9dd00d18ebe8cf263). Clicking on a link with analytics for data-eventtype and data-eventoptions in the attributes.

Current Behavior: On a `cmd + click` on a call-to-action link on a result card, the webpage opens in the same window.

Expecting: The webpage to the open in a different window.

PR #121, opened 2019-07-31
Send analytics events with sendBeacon

Delerme:
Switch to sendBeacon instead of standard post requests so events are still sent
even when the page is unloaded. This allows us to remove the custom link logic,
which prevents breaking native built-in behavior like cmd+clicking links. Fixes #118


b1lly (commented):
Are there polyfills to this approach? Even though we're not supporting IE11 within the library, we need to make sure theres a clear path for how it can be supported.

Also, anything that impacts IE11 as a "no-support" thing is tracked and noted to @benmcginnis so they aren't hit with surprised issues.

Delerme:
Yea there are, the polyfill is a blocking request (so the page won't redirect until the request finishes, similar to our original approach). Okay I'll let @benmcginnis know once merged

b1lly (approved):

Issue #120, opened 2019-07-31
Create 'Locator' Component


aginsberg:
Because of the frequent need for a vertical results column next to a map with corresponding map pins, it would be useful to have a 'Locator' component that will add both to the page with the standard Consulting Locator styling (somewhat similar to this http://cobalt.yextdemos.com.yextpages.net/search/ace)

aginsberg:
@b1lly I believe you've already talked to Ben about this, just wanted to check if you have any concerns with what I've described above.

b1lly:
There are higher order component stories in the product backlog, which includes this! Higher order components will be coming in future releases.

PR #123, opened 2019-08-01
PC-57301: Create Locator component (#120)

aginsberg:
After talking with Billy, this will not be the finalized way of how we handle higher level components, but this is fine for consulting's purposes


src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+import Component from '../component';
+
+/**
+ * LocatorComponent creates a Map next to a Results component,
+ * used for vertical results pages with map.
+ *
+ * @extends Component
+ */
+export default class LocatorComponent extends Component {
+  constructor (opts = {}) {
```
b1lly: we're using the word `config` instead of opts now. slowly updating old code, but new code should follow this pattern

src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+    super(opts);
+    this._resultsParams = {};
+    this._mapParams = {};
```
b1lly: jsdoc for all of this (which should include their purpose and sample structure of object)

Also, I think resultsConfig or mapConfig is a better name.

src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Reset LocatorComponent config options assigned via the Component constructor
+     * since all options should be passed to the lower level components
+     */
+    this.transformData = function () {};
+    this._render = null;
```
b1lly: not sure that I understand the purpose of overriding these? can you elaborate? the data for instance should flow FROM this component, to the lower level components. 

Always try to think about data and configuration of components flowing from the top down of the application.

So not sure why you'd want to prevent people from being able to transformData at the top of the component hierarchy.

Also curious about the render override?

src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+    this.onMount = function () {
+      this.initResults();
+      this.initMap();
+    };
```
b1lly: style nit:

take this out of the constructor, and just create a method on the class

```
onMount() {

}
```

The same style/syntax for how you created initResults.

src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Locator param: The template name to use for rendering with handlebars
+     * @type {string}
+     */
+    this._templateName = 'results/locator';
+
+    /**
+     * Results param: the max amount of results to display
+     * @type {int}
```
b1lly: `number` is the proper type (fix everywhere)

src/ui/components/results/locatorcomponent.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * A helper method to mount the VerticalResults component
+   */
+  initResults () {
+    this.componentManager.create('VerticalResults', Object.assign({
+      parent: this,
+      container: '.js-yxt-locatorResults'
+    }, this._resultsParams));
+  }
```
b1lly, line 89: The other thing you can do is construct this through the markup syntax in the locator template. If that's something you preferred. I personally don't have a strong preference either way.

src/ui/components/results/locatorcomponent.js
aginsberg: I was trying to passthrough options from the Locator constructor to the appropriate components, but you're right, this doesn't make a whole lot of sense? I'll rework this to make it easier to configure the appropriate component

src/ui/components/results/locatorcomponent.js
benmcginnis: The main other thing for this @aginsberg is figuring out how you would add filters to it, like in citizens

aginsberg: > The main other thing for this @aginsberg is figuring out how you would add filters to it, like in citizens

oooo I hadn't considered that, shouldn't be too much to add an additional component though, I'll work on that

PR #124, opened 2019-08-01
Pass component config options to templates

Delerme:


src/ui/components/component.js
```diff
@@ -207,7 +207,8 @@ export default class Component {
   }
 
   setState (data) {
-    this._state.set(data);
+    const newState = Object.assign({}, this._config, data);
```
b1lly: this would actually merge the configuration as data, which could actually have some namespacing issues e.g. accidental collisions of names. It'd also be preferable to be explicit about what is data and what is configuration in the template.

If I were implementing the template, it'd be good to be able to do something like

```
 {{#if _config.showTitle}}
     {{title}}
 {{/if}}
```

In otherwords, probably something like:

```
Object.assign({}, { _config: this._config }, data);
```

Is safer and more explicit.

b1lly (changes requested):

src/ui/components/component.js
b1lly: @benmcginnis Thoughts? coming soon.

Basically, any configuration provided through the .addComponent('Type', { // config }) will be accessible in the templates under the `_config` object.

Delerme: I made it `config` since it's public. Done

b1lly: The suggestion for `_config` was so that it'd always avoid any type of "data" properties. Like if data ever came back from the server about a vertical config and used `config` as a key, then that'd be problematic.

Alternatively. we can namespace them both separately too? Not sure the impact of that but it'd be something like:

```
config.showTitle
data.title
```

Or, we can do something like: `config_.showTitle` if you want to be protected by not private.

`config` probably works, but just kind've exercising precaution so that we don't have to update all the templates if this were to ever have to change.

benmcginnis: I'm ok with this if this is what we want to do for now.  

I think we need to think a bit about the best way for the view model of a template to be constructed.  I'd prefer for the view model to be exactly what's exposed to the templates and have it's properties match one-to-one with the properties exposed in the template.

In my mind, I feel like you would never want to override server data with config, so perhaps what we should do is 
```
const model = Object.assign({}, data)
for (let key of Object.keys(this.config)) {
  if (key in this.config) {
     console.warn('key conflict', key);
   }
}
return Object.assign(model, config);
```

Delerme: made it _config to avoid conflicts for now. Most of this results logic is getting rewritten and like @benmcginnis said we'll have a real view model soon so let's just do this in the meantime.

b1lly (approved):

PR #138, opened 2019-08-07
Add number range and date range scalar filters

Delerme:


Delerme:
@benmcginnis When using a thirdparty date picker, you can replace the template, do any initialization that thing needs on mount, and use the `setMin` and `setMax` to integrate with the library (has to be in the yyyy-mm-dd format). Similar idea with number range, if you want to use some crazy knob control or something. Does that interface seem flexible enough?

src/ui/templates/controls/range.hbs
```diff
@@ -0,0 +1,13 @@
+<fieldset>
+  <legend>{{title}}</legend>
+
+  {{#if minLabel}}
+    <label for="js-yext-range-min-{{name}}">{{minLabel}}</label>
+  {{/if}}
+  <input id="js-yext-range-min-{{name}}" type="number" class="js-yext-range" placeholder="Min" max="{{max}}" data-key="min" value="{{minValue}}">
```
b1lly: style nit:

can we break these attributes onto new lines?

b1lly (approved):
LGTM

src/ui/templates/controls/range.hbs
Delerme: done

PR #141, opened 2019-08-09
Add missing config to scalar filters

Delerme:


b1lly (approved):
A lot of this boilerplate among filters makes me wonder if there should be an abstract class in the future iterations.

Delerme:
Yea I think an interface would be the best bet

PR #152, opened 2019-08-15
Support search limits and offset in core

Delerme:


src/core/core.js
```diff
@@ -78,11 +78,24 @@ export default class Core {
...
+   */
+  verticalSearch (verticalKey, query) {
     this.storage.set(StorageKeys.VERTICAL_RESULTS, VerticalResults.searchLoading());
     return this._searcher
-      .verticalQuery(queryString, verticalKey, filter)
+      .verticalQuery(verticalKey, query)
       .then(response => SearchDataTransformer.transformVertical(response))
+      .then(results => query.append
+        ? this.storage.getState(StorageKeys.VERTICAL_RESULTS).append(results)
+        : results)
```
b1lly, line 99: Can you expand this for readability?

Is the idea that all components that conduct searchers will have to implement their own query/throttling/error logic (e.g. Searchbar, Filters, etc)?

src/core/search/searchapi.js
```diff
@@ -1,62 +1,77 @@
...
+  /**
+   * Search in the context of a vertical
+   * @param {string} verticalKey vertical ID for the search
+   * @param {object} query The query details
+   * @param {string} query.input The input to search for
+   * @param {string} query.filter The filter to use in the search
+   * @param {number} query.limit The max number of results to include, max of 50
+   * @param {number} query.offset The results offset, for fetching more results of the same query
+   */
+  verticalQuery (verticalKey, { input, filter, limit, offset }) {
```
b1lly: do we want to name this verticalSearch? How come in one signature you have the expanded object and the other signature you have just the 'query' object. In terms of consistency and a pattern we should follow throughout our app, is there a preference?

b1lly (commented):
Nice refactor! This looks pretty solid, with the one exception of the vertical search appending situation. I'm not sure how I feel about it yet. Seems reasonable, though.

I would like to see the component implementations around this before approving.

src/core/core.js
Delerme: This seems readable, do you want `this.storage.getState` assigned to a variable?

All components already implement their own query and error logic, I'm not sure what you mean by throttling

src/core/search/searchapi.js
Delerme: Sure, I renamed it to verticalSearch. It's not expanded, it's still one argument, see how it's used in core.js. Destructuring it makes it easy to use the properties we care about directly, and also makes others see which properties verticalQuery cares about.

src/core/core.js
b1lly: I was suggesting not using a ternary for readability.

It's a style nit, but this seems more legible:

```
.then(results => {
  if (query.append) {
    return this.storage.getState(StorageKeys.VERTICAL_RESULTS).append(results);
  }
  return results;
})
```

src/core/search/searchapi.js
b1lly: Misuse of words. Yup, I agree it makes it easier to see the properties! That's what I like. I was suggesting being consistent everywhere with this pattern through our library.

Delerme: I don't think it's really a binary "do it everywhere or don't use it ever" thing, it really just depends on the function. In this case, the function needs to access every property so it's better to destructure. In core, it doesn't, so destructuring would be pointless, it's like a right tool for the job kind of thing

src/core/core.js
Delerme: that seems a good deal less readable to me, going to leave this for now

b1lly (approved):

Issue #153, opened 2019-08-19
Empty Map on Load


aginsberg:
Need to be able to display a map even when no results exist

b1lly:
Yeah, this should be easy to do. Feel free to send in a PR.

PR #156, opened 2019-08-19
adding option to show an empty map if there are no returned results, …

aginsberg:
…allow specifying default position for when there are no results (#153)


src/ui/components/map/providers/mapprovider.js
```diff
@@ -33,6 +33,18 @@ export default class MapProvider {
      */
     this._zoom = opts.zoom || 9;
 
+    /**
+     * The default coordinates to display if there are no results returned
+     * @type {Object}
+     */
+    this._defaultPosition = opts.defaultPosition || { lat: 37.0902, lng: -95.7129 };
```
Delerme, line 41: just double checking, is that default location meaningful and appropriate for all implementations? We may just want to error out is `showEmptyMap` is tru and `defaultPosition` is not provided

src/ui/components/map/providers/mapprovider.js
```diff
@@ -33,6 +33,18 @@ export default class MapProvider {
      */
     this._zoom = opts.zoom || 9;
 
+    /**
+     * The default coordinates to display if there are no results returned
```
Delerme, line 37: and `showEmptyMap` has to be true too right?

src/ui/components/map/providers/mapprovider.js
aginsberg: I thought it would make sense for the default position to be USA, since that's what the majority of our clients would want, but I'm open to changing it to be required

src/ui/components/map/providers/mapprovider.js
aginsberg: correct, I can update the comment so its clearer

src/ui/components/map/providers/mapprovider.js
Delerme: up to @b1lly

src/ui/components/map/providers/mapprovider.js
Delerme: sounds good

src/ui/components/map/providers/mapprovider.js
b1lly: I don't feel strongly. Seems fine.

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -71,21 +75,25 @@ export default class MapBoxMapProvider extends MapProvider {
       container: container,
       zoom: this._zoom,
       style: 'mapbox://styles/mapbox/streets-v9',
-      center: [mapData.mapCenter.longitude, mapData.mapCenter.latitude]
+      center: mapData && mapData.mapCenter && mapData.mapCenter.longitude && mapData.mapCenter.latitude
+        ? [mapData.mapCenter.longitude, mapData.mapCenter.latitude]
+        : { lng: this._defaultPosition.lng, lat: this._defaultPosition.lat }
```
b1lly: Style nit: 

I'd probably make a helper method for this.

Then its also testable in the future.

b1lly (approved):

src/ui/components/map/providers/mapboxmapprovider.js
aginsberg: Added!

PR #159, opened 2019-08-21
Collapse pins with the same lat/lng and support clicking pins

Delerme:


src/ui/components/map/providers/mapprovider.js
```diff
@@ -115,4 +121,34 @@ export default class MapProvider {
...
+          longitude: markers[0].longitude
+        };
+        collapsedMarkers.push(collapsedMarker);
+      } else {
+        collapsedMarkers.push(markers[0]);
+      }
+    }
+
+    return collapsedMarkers;
+  }
```
b1lly, line 153: Can you add a test for this method? I think the map provider might already have tests.

b1lly (commented):
Looks good. Just need to add a test for that one method if possible.

src/ui/components/map/providers/mapprovider.js
Delerme: Done.

src/ui/components/map/providers/mapprovider.js
```diff
@@ -115,4 +121,34 @@ export default class MapProvider {
...
+
+  /**
+   * Given a list of markers, combine markers with the same lat/lng into a single marker
+   * @param {object[]} markers The markers to collapse
+   */
+  _collapseMarkers (markers) {
+    const locationToItem = {};
+    markers.forEach(m => {
+      locationToItem[`${m.latitude}${m.longitude}`]
+        ? locationToItem[`${m.latitude}${m.longitude}`].push(m)
```
benmcginnis, line 133: I think this is probably not going to work perfectly because of the issue with the way the system does lat/lngs.  The system often does not produce the same lat/lng for the same address to all the digits supported by lat/lng.  It might be better to check if address.line1, address.region, address.city, and address.postalCode are all the same and then collapse using that instead, or potentially rounding to the 4th or 5th decimal place: https://en.wikipedia.org/wiki/Decimal_degrees

benmcginnis (approved):

src/ui/components/map/providers/mapprovider.js
Delerme: If it's not possible to ensure the same addresses produces the same lat/lng (which seems super weird that happens at all...) then I think it'd be better to have a collapse radius so all pins within that configurable radius are collapsed, but either way can do in another commit if it proves a problem for providence

PR #161, opened 2019-08-22
ui/qa: hook up question submit component + refactor

b1lly:
- creates a core model for the qa submission component, and decouples it from results storage dependency

- component/template handles a few various states for the component (errors, submit successful, and results)

- attempt a new pattern for default config component. with this pattern, if components need to reference config, its through `this._config.[prop]`. templates continue to reference through `_config.[prop]`. 

default values are provided via a separate object and not defined explicitly on the constructor.  this helps gets rid of a bunch of boilerplate code between the constructor assignment and setstate for passing configuration to templates. It also can unify how we handle configuration (between definition and validation).  

- because of the way the default config can work, added a validateConfig method that each component sub class can have for throwing errors relative to the component. in the future, this could be an extension of the base class if we like the pattern


src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -60,49 +91,103 @@ export default class QuestionSubmissionComponent extends Component {
...
-      emailLabel: this._emailLabel,
-      nameLabel: this._nameLabel,
-      questionLabel: this._questionLabel,
-      privacyPolicyLabel: this._privacyPolicyLabel,
-      buttonLabel: this._buttonLabel,
-      question: new SearchParams(window.location.search.substring(1)).get('query')
-    });
+  /**
+   * Validates the required fields (or rules) for the formData
+   * @param {HTMLElement} formData
```
b1lly: This actually is an {object} and not and HTMLElement. I'll fix this tomorrow.

src/core/core.js
```diff
@@ -90,33 +91,78 @@ export default class Core {
...
+      })
       .then(response => SearchDataTransformer.transformVertical(response))
       .then(results => query.append
         ? this.storage.getState(StorageKeys.VERTICAL_RESULTS).append(results)
         : results)
       .then(data => {
-        this.storage.set(StorageKeys.QUERY_ID, data[StorageKeys.QUERY_ID]);
-        this.storage.set(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
-        this.storage.set(StorageKeys.VERTICAL_RESULTS, data[StorageKeys.VERTICAL_RESULTS]);
+        this.storage.set(
```
Delerme: I find this a lot harder to read and follow, can we avoid multi-lining expressions if they are under 120 characters?

src/core/core.js
```diff
@@ -90,33 +91,78 @@ export default class Core {
...
         if (data[StorageKeys.DYNAMIC_FILTERS]) {
-          this.storage.set(StorageKeys.DYNAMIC_FILTERS, data[StorageKeys.DYNAMIC_FILTERS]);
+          this.storage.set(StorageKeys.DYNAMIC_FILTERS,
+            data[StorageKeys.DYNAMIC_FILTERS]);
         }
+
+        this.storage.set(
+          StorageKeys.QUESTION_SUBMISSION,
+          new QuestionSubmission({
+            questionText: query.input
```
benmcginnis: this will default the question text to the value of the user's query, correct?

src/core/storage/storagekeys.js
```diff
@@ -16,5 +16,6 @@ export default {
   QUERY: 'query',
   QUERY_ID: 'query-id',
   FACET_FILTER: 'facet-filter',
-  DYNAMIC_FILTERS: 'dynamic-filters'
+  DYNAMIC_FILTERS: 'dynamic-filters',
+  QUESTION_SUBMISSION: 'question-submission'
```
benmcginnis, line 20: you should be able to use a trailing comma here, to avoid two line diffs when you add a row in the future.

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -4,54 +4,85 @@ import Component from '../component';
...
-     * Optionally provided, otherwise defaults to `Name`
-     * @type {string}
-     */
-    this._nameLabel = opts.nameLabel || 'Name:';
+  /**
+   * The label to use for the name input
+   * Optionally provided, otherwise defaults to `Name`
+   * @type {string}
+   */
+  'nameLabel': '*Name:',
```
benmcginnis: I think you want to separate the * from the actual label.

src/ui/components/questions/questionsubmissioncomponent.js
benmcginnis: Does JSDoc respect the actual dom api models? If so, you might want to use {FormData}: https://developer.mozilla.org/en-US/docs/Web/API/FormData

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -0,0 +1,46 @@
+{{#if questionSubmitted}}
+  <div class="yxt-QuestionSubmission">Thanks for your submission!</div>
+{{else}}
+  <form class="yxt-QuestionSubmission">
+    {{#if errors}}
+      <div class="yxt-errors">
+        You're missing some required fields!
```
benmcginnis: I think you'd also want to add  the errors you made, right? Perhaps as a bulleted list?

benmcginnis (approved):

src/core/core.js
b1lly: Yes! Apologies. I made the adjustment already just didn't re-upload.

src/core/core.js
b1lly: Yeah, that was the original behavior Max wanted. Happy to change it though.

src/ui/templates/questions/questionsubmission.hbs
b1lly: Yeah, I left this simple because I wasn't exactly sure how we wanted to handle the errors. I kind've just plopped it in here to make sure the mechanism existed and worked.

For instance, there's a few ways:

1. Bulleted list
2. Each input appends an `yxt-error` class to highlight them red
3. Each input shows the error message above/below

Do we know of a preference?

src/core/core.js
b1lly: Do you know if this is something we can bake into the formatter? @Delerme

src/core/storage/storagekeys.js
b1lly: Yeah, I grew up with the trailing comma practice too, but the formatter we're using trims them unfortunately.

src/ui/templates/questions/questionsubmission.hbs
benmcginnis: I can take a crack at making it nice & accessible, but this is a good overview of the best practice: https://www.w3.org/WAI/tutorials/forms/notifications/

TL;DR:

```
<div class="yxt-errors" role="alert">
  <h3>You are missing {{errors.length}} required fields</h3>
  <ul>
      {{#each errors}}
          <li>{{error}}</li>
      {{/each}}
   </ul>
</div>
```

src/ui/components/questions/questionsubmissioncomponent.js
b1lly: Sure. Updated markup.

src/core/core.js
benmcginnis: just wanted to confirm as Rose asked me about making it a default

src/ui/templates/questions/questionsubmission.hbs
b1lly: added.

benmcginnis:
@b1lly, @creotutar pointed out to me that the analytics spec wants us to capture events whenever anyone focuses on the inputs in this form.  Do you want to add that as part of this PR or do another one?

.size-limit.js
```diff
@@ -1,7 +1,7 @@
 module.exports = [
   {
     path: 'dist/answers.min.js',
-    limit: '17 KB'
+    limit: '25 KB'
```
benmcginnis, line 4: 😢

b1lly: > @b1lly, @creotutar pointed out to me that the analytics spec wants us to capture events whenever anyone focuses on the inputs in this form. Do you want to add that as part of this PR or do another one?

I'm actually happy to implement now. Are we doing it on FOCUS or on click? I think there's a difference. What type of data are we actually capturing?

'User on input'?

Or is it 'User on email input'?

Can someone DM me a link on slack with the spec?

benmcginnis:
Focus should fire for both click and tabbing into the element, so I think
we want focus

src/core/core.js
Delerme: don't think so, there's no real hard and fast rule for multiline vs single, it's basically just whatever subjectively looks good, so you can't really represent that in a linter

src/core/core.js
```diff
@@ -2,6 +2,7 @@
 
 import SearchApi from './search/searchapi';
 import AutoCompleteApi from './search/autocompleteapi';
+import QuestionAnswerApi from './search/questionanswerapi';
```
b1lly, line 5: I plan on moving all of these into an `api` or `network` package.

src/core/core.js
b1lly: Ah poops. Doesn't the Java style thing do something like this? It's not always perfect but it tries to auto format and allows you to tweak it if its not ideal.

Delerme:
is this ready for review? take out of WIP status if so

b1lly:
> is this ready for review? take out of WIP status if so

Yup, should be in a good enough spot to review. I do want to add tests so wasn't sure if it should come out of WIP yet. There's a minor tweak also to how language should be provided for the component (waiting for product).

benmcginnis (approved):

b1lly:
This is done and ready to be reviewed. Sorry. I'll add tests separately.

benmcginnis (approved):
lgtm

src/core/core.js
```diff
@@ -76,6 +78,15 @@ export default class Core {
       answersKey: this._answersKey,
       locale: this._locale
     });
+
+    /**
+     * An abstraction
```
Delerme: ?

src/core/core.js
```diff
@@ -90,33 +101,64 @@ export default class Core {
    * @param {boolean} query.append If true, adds the results of this query to the end of the current results, defaults false
    */
   verticalSearch (verticalKey, query) {
-    this.storage.set(StorageKeys.VERTICAL_RESULTS, VerticalResults.searchLoading());
+    this.storage.set(
```
Delerme: single line

src/core/core.js
```diff
@@ -90,33 +101,64 @@ export default class Core {
...
+      })
       .then(response => SearchDataTransformer.transformVertical(response))
       .then(results => query.append
         ? this.storage.getState(StorageKeys.VERTICAL_RESULTS).append(results)
         : results)
       .then(data => {
         this.storage.set(StorageKeys.QUERY_ID, data[StorageKeys.QUERY_ID]);
         this.storage.set(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
-        this.storage.set(StorageKeys.VERTICAL_RESULTS, data[StorageKeys.VERTICAL_RESULTS]);
+        this.storage.set(
```
Delerme: single line

src/core/core.js
```diff
@@ -90,33 +101,64 @@ export default class Core {
...
+          StorageKeys.VERTICAL_RESULTS,
+          data[StorageKeys.VERTICAL_RESULTS]);
+
         if (data[StorageKeys.DYNAMIC_FILTERS]) {
-          this.storage.set(StorageKeys.DYNAMIC_FILTERS, data[StorageKeys.DYNAMIC_FILTERS]);
+          this.storage.set(
```
Delerme: single line

src/core/core.js
```diff
@@ -90,33 +101,64 @@ export default class Core {
...
+          new QuestionSubmission({
+            questionText: query.input
+          })
+        );
       });
   }
 
   search (queryString, urls) {
-    this.storage.set(StorageKeys.UNIVERSAL_RESULTS, UniversalResults.searchLoading());
+    this.storage.set(
```
Delerme: single line

src/core/core.js
```diff
@@ -167,6 +209,27 @@ export default class Core {
       });
   }
 
+  /**
+   * Submits a question to the server and updates the underlying questin model
```
Delerme: typo

src/core/models/questionsubmission.js
```diff
@@ -0,0 +1,53 @@
+/** @module QuestionSubmission */
+
+export default class QuestionSubmission {
```
Delerme, line 7: jsdoc

src/core/search/questionanswerapi.js
```diff
@@ -0,0 +1,47 @@
+/** @module QuestionAnswerApi */
+
+import ApiRequest from '../http/apirequest';
+import { AnswersBasicError, AnswersEndpointError } from '../errors/errors';
+
+/**
+ * QuestionAnswerApi exposes an interface for network related matters
+ * for all the autocomplete endpoints.
```
Delerme: ?

src/core/search/questionanswerapi.js
```diff
@@ -0,0 +1,47 @@
...
+     * @type {string}
+     * @private
+     */
+    if (!opts.apiKey) {
+      throw new AnswersBasicError('Api Key is required', 'QuestionAnswerApi');
+    }
+    this._apiKey = opts.apiKey;
+  }
+
+  submitQuestion (question) {
```
Delerme, line 35: jsdoc

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -3,55 +3,86 @@
...
+/**
+ * Configurable options for the component
+ * @type {Object}
+ */
+const DEFAULT_CONFIG = {
+  /**
+   * The entity identifier that the question is associated with.
+   * This is typically an organization object
+   * @type {number}
+   * @required
```
Delerme: I don't think this exists

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -60,49 +91,125 @@ export default class QuestionSubmissionComponent extends Component {
...
+      return;
+    }
+
+    this.bindAnalytics(formEl);
+    this.bindFormSubmit(formEl);
   }
 
-  initSubmit (formSelector) {
-    this._formEl = formSelector;
+  bindAnalytics (formEl) {
```
Delerme, line 138: jsdoc

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -60,49 +91,125 @@ export default class QuestionSubmissionComponent extends Component {
...
-    return false;
+  /**
+   * validationConfig contains a bunch of rules
+   * that are used to validate aginst configuration provided by the user
+   */
+  validateConfig () {
+    if (this._config.entityId === null ||
+        typeof this._config.entityId !== 'number') {
+      throw new AnswersComponentError(
+        '`entityId` is a required configuration option for Question Submission',
```
benmcginnis, line 113: Wait, this is strange, how do we know the entity Id for the question submission?

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -60,49 +91,125 @@ export default class QuestionSubmissionComponent extends Component {
...
+        });
     });
   }
 
-  static get type () {
-    return 'QASubmission';
+  /**
+   * Takes the form, and builds a object that represents the input names
+   * and text fields.
+   * @param {HTMLElement} formData
```
Delerme: formEl

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -60,49 +91,125 @@ export default class QuestionSubmissionComponent extends Component {
...
+
+    if (!formData.privacyPolicy || formData.privacyPolicy !== true) {
+      errors['privacyPolicy'] = 'Approving our privacy policy terms is required!';
+    }
+
+    if (!formData.questionText || typeof formData.questionText !== 'string') {
+      errors['questionText'] = 'Question text is required!';
+    }
+
+    return Object.keys(errors).length > 0 ? errors : false;
```
Delerme: does it make more sense to just return an empty object?

Delerme (changes requested):

src/ui/components/questions/questionsubmissioncomponent.js
b1lly: Entity ID in this case is the `organization` object which represents the brand. It's a single object that the account will have and will always be the one that you provide for the business/brand.

The point of this, is that it makes the downstream systems that depend on entities for things to work more easily as their primary relationship object.

src/ui/components/questions/questionsubmissioncomponent.js
b1lly: Probably not. I don't really feel strongly though. 

Otherwise you'd have to inspect it for errors.
I was doing this based off how `golang` handles errors for instance; the pattern there is to return an error, or `nil`.

let errors = checkErrors(stuff);
if (errors) {
  // do something
}

benmcginnis:
Great, thanks for the clarification.

src/ui/components/questions/questionsubmissioncomponent.js
b1lly: I'm an innovator.

src/ui/components/questions/questionsubmissioncomponent.js
Delerme: maybe null then? Just seems strange to have a return type that's either an object or a boolean. I get the pattern and although it's prevalent in Go, I don't think anyone in JS expects something like that

src/ui/components/questions/questionsubmissioncomponent.js
Delerme: I know you are,  but let's save the innovation for another day

Delerme (approved):

Issue #157, opened 2019-08-20
Default answers analytics


creotutar:
We would like to add default analytics for answers implementations according to the answers analytics doc

This would include eventOptions and eventType parameters for handlebars templates for Nav, Searchbar, Results, Result, Autocomplete, many other components

Ideally at the end of this, we will only have to supply analyticsOptions and transformData eventOptions for CTAs

PR #163, opened 2019-08-23
Added default analytics and eventtypes and eventoptions for implement…

creotutar:
…ation specific events (#157)


src/ui/templates/filters/filterbox.hbs
```diff
@@ -4,7 +4,12 @@
   {{/each}}
   {{#if showApplyButton}}
     <div>
-      <button type="button" class="js-yext-filterbox-apply">Apply</button>
+      <button type="button"
+              class="js-yext-filterbox-apply"
+              {{#if applyEventType}} data-eventtype="{{applyEventType}}"{{/if}}
+              {{#if applyEventPptions}} data-eventoptions='{{applyEventOptions}}'{{/if}}>
```
Delerme: typo

Delerme (approved):

b1lly (changes requested):
I'm not sure I understand what this is doing here, and i'm not sure this is the right "generic" solution.
Correct me if i'm wrong, but what you did here was bake variables in the template, from which you plan to use dataTransform to populate?

That is not the right approach. I'd expect each templates default data properties piped in through as core data.

benmcginnis:
Billy, when we discussed this it seemed like you weren't interested in
having us make that change.  In my mind, the templates shouldn't need to
know the event options at all, but this is what we could do in the short
term, and we have to have something for launch.

benmcginnis:
It's also what Christian & Julian discussed in their meeting for what we
were going to do for analytics in the short term.

src/ui/templates/filters/filterbox.hbs
```diff
@@ -4,7 +4,12 @@
   {{/each}}
   {{#if showApplyButton}}
     <div>
-      <button type="button" class="js-yext-filterbox-apply">Apply</button>
+      <button type="button"
+              class="js-yext-filterbox-apply"
+              {{#if applyEventType}} data-eventtype="{{applyEventType}}"{{/if}}
```
benmcginnis: I think this should be
```
{{if isFacet}}
  data-eventtype="FACETING"
{{else}}
  data-eventtype="FILTERING"
{{/if}}

File: src/ui/templates/filters/filterbox.hbs [hunk 3]
```diff
@@ -4,7 +4,12 @@
...
+              {{#if applyEventOptions}} data-eventoptions='{{applyEventOptions}}'{{/if}}>
```
benmcginnis: I think we should remove this, since there are no options in the spec.

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,6 +1,9 @@
 <nav class="yxt-Nav-container js-yxt-navContainer">
   {{#each tabs}}
-    <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}" href="{{url}}">
+    <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
+       href="{{url}}"
+       {{#if eventType}} data-eventtype="{{eventType}}"{{/if}}
```
benmcginnis: I think this should always be `data-eventtype="VERTICAL_SEARCHER_TAB"` based on the spec

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,6 +1,9 @@
...
+       {{#if eventOptions}} data-eventoptions='{{eventOptions}}'{{/if}}>
```
benmcginnis: I think we should remove this, since there are no options in the spec.

src/ui/templates/results/resultsitem.hbs
```diff
@@ -26,7 +26,12 @@
     <div class="yxt-Result-info">
       <h3 class="yxt-Result-title">
         {{#if link}}
-          <a class="yxt-Result-titleLink" href="{{link}}" data-eventtype="TITLE_CLICK">{{{title}}}</a>
+          <a class="yxt-Result-titleLink"
+             href="{{link}}"
+             data-eventtype="TITLE_CLICK"
+             {{#if titleEventOptions}} data-eventoptions='{{titleEventOptions}}'{{/if}}>
```
benmcginnis: @b1lly how should we get the vertical config id & searcher type here? If we work out how we want to pass that in we can remove the need for the implementor doing anything.  Otherwise they need to fill this out in transform data.  The entityId is the only thing we get on the model right now.

src/ui/templates/results/resultssectionheader.hbs
```diff
@@ -8,7 +8,10 @@
     </h2>
   </div>
   {{#if _config.verticalURL}}
-    <a class="yxt-Results-viewAllLink" href="{{_config.verticalURL}}">
+    <a class="yxt-Results-viewAllLink"
+       href="{{_config.verticalURL}}"
+       {{#if viewAllEventType}} data-eventtype="{{viewAllEventType}}"{{/if}}
```
benmcginnis: this should always be `data-eventtype="VERTICAL_SEARCHER_VEIW_ALL"` based on the spec.

@b1lly same question on how do you want to pass the vertical config id to this as that's the only eventoption?

src/ui/templates/results/resultssectionheader.hbs
```diff
@@ -22,7 +25,10 @@
     {{/each}}
     {{#if _config.showChangeFilters}}
       {{#if _config.verticalURL}}
-        <a class="yxt-Results-changeFilters" href="{{_config.verticalURL}}">
+        <a class="yxt-Results-changeFilters"
+           href="{{_config.verticalURL}}"
+           data-eventtype="FILTERING_WITHIN_SECTION"
+           {{#if filteringEventOptions}} data-eventoptions='{{filteringEventOptions}}'{{/if}}>
```
benmcginnis: same vertical config id question @b1lly

src/ui/templates/filters/filterbox.hbs
benmcginnis: I think we would want to determine the value of isFacet [here](https://github.com/yext/answers/blob/71fbeae7c59f83226229f216f3852ee02ae5d539/src/ui/components/filters/filterboxcomponent.js#L19), though I'm not sure exactly how we would

src/ui/templates/results/resultsitem.hbs
benmcginnis: I think we can get it via _config._parentOpts.verticalConfigId if we just make sure that's set in the config we pass in.

benmcginnis: same with searcher type if it's not available elsewhere

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -56,6 +56,16 @@ export class Tab {
...
 
+  /**
+   * helper to construct the eventOptions for analytics
+   * @returns {string}
+   */
+  get eventOptions () {
+    return JSON.stringify({
+      verticalConfigId: this.configId
+    });
+  }
```
b1lly, line 67: what's this being used for?

b1lly (approved):

src/ui/components/navigation/navigationcomponent.js
benmcginnis: spec says we need to send this for clicks on the nav tabs.

PR #165, opened 2019-08-26
Prevent setting the search loading state when loading additional results

Delerme:
To prevent flashing the loading state when making many back-to-back calls. The client will have to set its own loading state when making numerous requests and use the final promise to mark it as done


aginsberg (commented):
LGTM!

PR #167, opened 2019-08-26
templates: update NavigationComponent to add "more" dropdown

benmcginnis:
I'm opening this back up since it sounds like this is settled from Product's perspective and ready for trying to land in v0.8.0.  Original discussion in #107


benmcginnis:
Duplicated by #226

PR #181, opened 2019-08-28
templates: add search bar & autocomplete hover and focus states

benmcginnis:
Also adds some styling tweaks for the default styling.


b1lly (approved):
Typically we check in documentation generation separately from the actual commits.

PR #179, opened 2019-08-28
templates/icons: add icon component

benmcginnis:
Adds an icon component and a default set of SVG icons from design.  Allows
implementors to specify an icon by name or provide a custom icon via either URL
or by providing a string.

Updated the resultssectionheader.hbs and the searchbar.hbs to take advantage of
this new component.

PR #183, opened 2019-08-28
Refactor component mounting and lifecycles

Delerme:
Components render top-down recursively, so parents don't complete mounting until
all children finish. Additionally adjust lifecycle arguments, having lifecycles
provided via configuration not override the builtin hooks, while _Override
versions of each option are added to support overriding.


src/ui/components/component.js
```diff
@@ -376,19 +384,25 @@ export default class Component {
...
-    let childHTML = [];
-    for (let i = 0; i < childData.length; i++) {
-      let childComponent = this.addChild(childData[i], type, opts);
-      childHTML.push(childComponent.render());
-    }
-
-    DOM.append(domComponent, childHTML.join(''));
+    childData.reverse();
+    childData.forEach((data, index) => {
+      DOM.append(domComponent, `<div id=${this.name}-child-${index}></div>`);
```
Delerme: it's possible to remove this if we update the mount logic to _**not**_ replace the container's contents entirely, but instead to append. I think that's the right direction, but figured you might have some concerns so didn't make the change. If you're okay with that change let me know

tests/ui/components/component.js
```diff
@@ -95,100 +95,6 @@ describe('render component templates', () => {
   });
 });
 
-describe('render template components using markup syntax', () => {
```
Delerme, line 98: these tests are stale now, the template component step happens in mount, not in render. Unfortunately we haven't set up a good testing framework for mounting and testing components that way (maybe enzyme) so removing these tests entirely

src/ui/components/component.js
```diff
@@ -281,6 +300,7 @@ export default class Component {
 
   unMount () {
     DOM.empty(this._container);
+    this._children.forEach(c => c.remove());
```
Delerme, line 315: @benmcginnis This fixes the child problem

benmcginnis: do you not also need to clear the children array?  Otherwise we're just going to keep adding children

mhupman: unMount calls remove() in the children vs unMount()?  That seems surprising

src/ui/components/component.js
```diff
@@ -346,27 +366,16 @@ export default class Component {
...
   }
 
   _createSubcomponent (domComponent, data) {
-    let dataset = domComponent.dataset;
-    let type = dataset.component;
-    let prop = dataset.prop;
+    const dataset = domComponent.dataset;
+    const type = dataset.component;
+    const prop = dataset.prop;
     let opts = dataset.opts ? JSON.parse(dataset.opts) : {};
```
benmcginnis, line 390: it would be good to `let childData = data[prop] || {}` here.  I'm not 100% but I was working on #179 and it doesn't have childData so it was blowing up.

src/ui/components/component.js
mhupman: I could imagine a child implementing unMount but surprised it’s not being called

src/ui/components/component.js
benmcginnis: I think that's the right direction, IMHO

src/ui/components/component.js
Delerme: Hmm yea I guess child data is optional, will add

src/ui/components/component.js
Delerme: @benmcginnis yea I also clear it below (force pushed just before your comment)

@mhupman The call stack looks like
```
parent.mount
    parent.unmount // unmounts the previous, removes all children
        child.remove
        parent.children = []
    parent.createSubcomponents // populates children
    child.mount
        child.unmount // unmounts the previous, removes all children
    child.onMount
parent.onMount
```

So child unmounts do get called. Admittedly there's some waste where children get removed, repopulated, and then remove all of their children (but of course they don't have any) but I think that's better than components being aware they are subcomponents or not. Does that seem right?

src/ui/components/component.js
Delerme: yea thinking about it some more I think it's a change I'll push for. Will add

src/ui/components/component.js
benmcginnis: in your flow above @Delerme it seems like
```
parent.createSubcomponents
child.mount
    child.unmount // there's nothing to unmount, it was removed above & probably garbage collected
```

mhupman: I'm not sure I fully understand the call stack you drew out there.  Perhaps part of the issue is not understanding the difference between `remove()` and `unmount()`.  The former sounds destructive and the latter like it just deals with listeners, so its a bit strange to me that unmount ends up "destroying" the children.

mhupman: That said though, if this works and @benmcginnis is fine with it I'm more than happy to step aside, I'm not nearly as familiar with this codebase!

Delerme: Yea actually I do see your point, the above call trace is a little weird. Adjusted the code so now it's like
```
parent.unMount
    child.unMount
    child.onUnMount
    parent.removeChildren
    parent.onUnmount
parent.mount
    parent.render
    parent.createsubcomponents
    children.mount
        children.render
    children.onmount
parent.onmount
```

This way everything is unmounted before moving to mounted, and to your point children are unmounted (cleanup) then removed (deleted). Previously they were deleted outright without being able to cleanup. Seem better?

src/ui/components/component.js
Delerme: done

src/ui/components/component.js
benmcginnis: @mhupman I deployed #184 to yext answers staging and I believe the issue is addressed.  I've tried multiple faq queries on staging and they all seem to toggle correctly, please feel confirm if you can.

src/ui/components/component.js
Delerme: done.

src/ui/components/component.js
```diff
@@ -164,19 +164,37 @@ export default class Component {
...
      */
-    this.onCreate = config.onCreate || this.onCreate || function () {};
+    this.onCreate = config.onCreateOverride || this.onCreate || function () {};
 
     /**
      * The a local reference to the callback that will be invoked when a component is Mounted.
      * @type {function}
      */
-    this.onMount = config.onMount || this.onMount || function () { };
+    this.onMount = config.onMountOverride || this.onMount || function () {};
```
benmcginnis, line 174: is there any worry about `onMountOverride` not being able to call super?

Delerme: maybe a little. Added a `.bind(this)` so override functions can at least walk the prototype chain manually. Something like `Object.getPrototypeOf(Object.getPrototypeOf(this))` lol

src/ui/components/component.js
mhupman: LGTM, tested as well.

b1lly:
I think this is a nice quality of life improvement!

The only thing is, we need to be explicit and throw a fairly big warning about the interface change. For instance, I'm CERTAIN that a few of the consulting implementations will break, based on how their using the existing `onMount` configuration option as an `override` instead of an `extension`.

I think overall it's pretty good. I just want to take one more, more detailed pass and make sure there's nothing i'm overlooking. I think @benmcginnis and @mhupman noted most of the things. If you merge in the latest 0.8.1 stuff, there might be something that actually conflicts with the `unMount` comment that ben made.

Delerme:
Right, this is a breaking change so it's in 0.9.0, that'll definitely be well communicated before 0.9.0 makes it to master

b1lly:
This looks good to me. But it's super important that you tested all the use cases and components with this change before pushing it. Mainly, the universal results component including maps would be a good one.

b1lly (approved):

Issue #155, opened 2019-08-19
child components not garbage collected


aginsberg:
Conducting 2 searches causes the Accordion expand/collapse to stop working

benmcginnis:
I believe this is resolved in https://github.com/yext/answers/pull/79

benmcginnis:
It’s not solved by that, it seems that we’re not cleaning up old components

PR #184, opened 2019-08-28
address add child issue

benmcginnis:
add logging to diagnose the bug in #155


b1lly (approved):
@Delerme not sure if this impacts your lifecycle item.

PR #186, opened 2019-08-29
Templates/truncate details

benmcginnis:


src/core/models/resultfactory.js
```diff
@@ -62,4 +60,34 @@ export default class ResultFactory {
...
+      const word = words[i];
+      if (truncated.length + word.length + sep.length > max) {
+        truncated += trailing;
+        break;
+      }
+      truncated += i === 0 ? word : sep + word;
+    }
+
+    return truncated;
+  }
```
b1lly, line 94: This would be really good to have unit tests for.

b1lly (changes requested):
Otherwise LGTM.

Is it possible for you to add some tests for that helper?

b1lly (approved):
LGTM

PR #189, opened 2019-08-29
Add onStateChange init option

Delerme:
Adds an option to Answers init to listen to changes in the public state


Delerme:
This is a temporary hack until the global url state story gets done

b1lly (approved):

PR #192, opened 2019-08-30
search: disable spellcheck and autocorrect

benmcginnis:


src/ui/components/search/autocompletecomponent.js
```diff
@@ -148,7 +148,9 @@ export default class AutoCompleteComponent extends Component {
     }
 
     // Disable the native auto complete
-    DOM.attr(queryInput, 'autoComplete', 'off');
+    DOM.attr(queryInput, 'autocomplete', 'off');
+    DOM.attr(queryInput, 'autocorrect', 'off');
+    DOM.attr(queryInput, 'spellcheck', 'false');
```
b1lly: If you want to make the DOM.attr method support an object in addition to `prop` and `value` that would be a great facade.

Like for instance, if it worked both ways:

```
DOM.attr(queryInput, 'autocomplete', 'off')
```

AND

```
DOM.attr(queryInput, {
  'autocomplete': 'off',
  'autocorrect': 'off',
  'spellcheck': 'false'
});
```

That would be a nice enhancement.

b1lly (approved):
Minor suggestion to improve the interface but otherwise LGTM.

benmcginnis:
@b1lly I added the DOM helper you suggested, still LGTY?

b1lly (approved):

Issue #178, opened 2019-08-28
View All Button uses Previous Query


Mxs2019:
The View All button is using the wrong query when you run multiple searches. To recreate this behavior go to 

http://answers.yext.com.yextpages.net/index.html?query=leadership&tabOrder=.%2Findex.html%2Cleadership%2Csupport%2Cjobs%2Cpackages%2Cevents%2Cfaqs%2Cproducts%2Cpublishers%2Coffices%2Capp_directory_partners%2Ccase_studies%2Clinks

Run a search for `Leadership who graduated from cornell`

Then press view all

You'd expect to see leadership tab with query `Leadership who graduated from cornell`

Instead, you see this which is the original query. 
http://answers.yext.com.yextpages.net/leadership?query=leadership&tabOrder=.%2Findex.html%2Cleadership%2Csupport%2Cjobs%2Cpackages%2Cevents%2Cfaqs%2Cproducts%2Cpublishers%2Coffices%2Capp_directory_partners%2Ccase_studies%2Clinks

The bug seems to be that the JS library does not update it's representation of the latest query and only sends along the first query searched.

PR #193, opened 2019-08-30
Fix incorrect verticalurl reference

Delerme:


Delerme:
I _think_ this solves #178 but haven't seen the implementation so unclear. Definitely works locally with my setup now though

benmcginnis (approved):
Thanks!

PR #196, opened 2019-09-04
Add global state url param abstraction

Delerme:
Abstract away url search param manipulation behind a storage-like interface that syncs with regular storage. This way, components can explicitly point out which values they want to persist and restore across refreshes and history traversals, and the storage powering their state will automatically be updated


Delerme:
Putting this in draft for early feedback. Remaining work:

- Tests
- Removing manual param updating from filter components and navigation

src/ui/storage/persistentstorage.js
```diff
@@ -0,0 +1,115 @@
...
+     * The list of listeners to every storage update
+     * @type {function[]}
+     */
+    this._updateListeners = [];
+
+    /**
+     * The list of listeners to storage resets
+     * @type {function[]}
+     */
+    this._resetListeners = [];
```
Delerme: these and the above listeners can't be changed or removed, but I think that's fine for now. They are currently only used internally and components aren't really expected to listen on this.

Delerme:
Dev work is done. Skipping navigation component because I think that'll need a larger refactor, so it'll continue the url param hack for now.

Tests coming soon

Delerme:
Also heads up @benmcginnis the onStateChange function signature has changed, instead of just (params: string) it's (data: object, params: string) so the params string is now the second argument

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -170,11 +167,12 @@ export default class FilterSearchComponent extends Component {
...
 
-        window.history.pushState({}, '', '?' + params.toString());
-        this.core.storage.set(StorageKeys.PARAMS, params.toString());
-
         // save the filter to storage for the next search
-        this.core.setFilter(this.name, Filter.fromResponse(filter));
+        this.query = query;
+        this.filter = Filter.fromResponse(filter);
+        this.persistentStorage.set(`${StorageKeys.QUERY}.${this.name}`, this.query);
+        this.persistentStorage.set(`${StorageKeys.FILTER}.${this.name}`, this.filter);
```
b1lly: perhaps having a bulk interface would be ideal. Maybe following the setAll that we talked about.

src/ui/storage/persistentstorage.js
```diff
@@ -0,0 +1,119 @@
...
+    this._params.delete(key);
+    this._updateHistory();
+  }
+
+  _updateHistory () {
+    if (this._historyTimer) {
+      clearTimeout(this._historyTimer);
+    }
+
+    this._historyTimer = setTimeout(
```
b1lly, line 72: Can you add a comment for the purpose of this is "batching"? Also, if the bulk interface corrects the majority of these cases, maybe its best to leave this simpler for now. Unless of course you're noticing problems with the multiple components, then we can leave it.

This seems like a round about solution to the traditional `transactions` requirements of most storages. Probably fine for now the way it is, because it's not worth the engineering overkill, but just for longer-term thinking, it might be worth taking some inspiration from technologies that have already solved these problems well (e.g. java/go/mysql). I'm sure there some practices we can learn from there to apply to this interface (and how many threads can manage read/write)

src/ui/components/componentmanager.js
```diff
@@ -75,6 +84,11 @@ export default class ComponentManager {
     return this;
   }
 
+  setPersistentStorage (persistentStorage) {
+    this._persistentStorage = persistentStorage;
+    return this;
+  }
```
b1lly: Perhaps persistentStorage should be part of the core.

b1lly (commented):
Minor comments but this looks solid. Nice job!

src/ui/components/search/filtersearchcomponent.js
Delerme: punted as discussed

src/ui/storage/persistentstorage.js
Delerme: done. As discussed, will be necessary to avoid the url updating a bunch from different components, but I do think once we get indexing in it will likely become obsolete. I'm more comfortable with leaving a temporary solution in than no solution at all

src/ui/components/componentmanager.js
Delerme: done

b1lly (approved):
LGTM.

I'd test this thoroughly, and try to get one of the "fully featured" test apps (or demos) using the version of this JS to look for bugs, before pushing this, if possible.

PR #198, opened 2019-09-06
Add user options to geo filter

Delerme:


Delerme:
Backend is also working so tested with that

src/ui/templates/controls/geolocation.hbs
```diff
@@ -4,7 +4,9 @@
     <div class="yxt-GeoLocationFilter-query">
       <form class="yxt-GeoLocationFilter-form">
         <label class="yxt-GeoLocationFilter-label" for="query">{{labelText}}</label>
-        <input class="js-yxt-GeoLocationFilter-input yxt-GeoLocationFilter-input" type="text" name="query" value="{{#unless geoEnabled}}{{query}}{{/unless}}" placeholder="{{#if geoEnabled}}Current Location{{/if}}">
+        <input class="js-yxt-GeoLocationFilter-input yxt-GeoLocationFilter-input" type="text" name="query" value="{{#unless geoEnabled}}{{#unless geoLoading}}{{#unless geoError}}{{query}}{{/unless}}{{/unless}}{{/unless}}"
+          placeholder="{{#if geoEnabled}}{{enabledText}}{{else}}{{#if geoLoading}}{{loadingText}}{{else}}{{#if geoError}}{{errorText}}{{/if}}{{/if}}{{/if}}"
+        >
```
b1lly, line 9: Is all this logic something we can abstract into the view model as a single property value? Instead of having all this template logic?

For instance, can the component js do all this checking and just return a single value into the `setState` called {{geoLocationValue}} or something?

Delerme: yea just makes it harder for implementors to change. It's probably fine if they don't though. Done

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -119,6 +119,16 @@ export default class GeoLocationComponent extends Component {
...
+    let placeholder = '';
+    if (this._enabled) {
+      placeholder = this._config.enabledText;
+    }
+    if (data.geoLoading) {
+      placeholder = this._config.loadingText;
+    }
+    if (data.geoError) {
+      placeholder = this._config.errorText;
+    }
```
b1lly, line 131: We have a public 'dataTransform', might be something we want to bake internally in the future too. But otherwise LGTM here.

b1lly (approved):

PR #201, opened 2019-09-06
Switch autocomplete endpoint

Delerme:


b1lly (approved):
I hope this doesn't break anything consulting already has implemented. Are there any concerns? Seems like it should just work.

PR #202, opened 2019-09-06
Analytics/direct answer cta

benmcginnis:


src/ui/components/results/directanswercomponent.js
```diff
@@ -105,6 +105,24 @@ export default class DirectAnswerComponent extends Component {
     this.setState(newState);
   }
 
+  setState (data) {
+    return super.setState(Object.assign({}, data, {
+      eventOptions: this.eventOptions(data)
```
b1lly, line 110: I think in the future, the core component can provide event options as a default object into all components @Delerme what you think about that?

b1lly (approved):
@Delerme as a heads up, for our future iteration

src/ui/components/results/directanswercomponent.js
benmcginnis: I definitely thought about doing this, so I think it's probably a good idea.  I didn't pull the trigger because it seemed too large of a change at the time.

Issue #200, opened 2019-09-06
Add No Results screen


creotutar:
We would like to have a no results screen for the library when a user types in a query and the system returns no results (opposed to the current blank page).

The universal No Results page and the vertical No Results page should only differ in a link that will take you back to the universal search with the provided query.

(similar to the Yext Answers no results page: [here](https://answers.yext.com/app_directory_partners?query=skhdjasjhdgas&tabOrder=.%2Findex.html%2Capp_directory_partners%2Ccase_studies%2Cevents%2Cfaqs%2Cjobs%2Cleadership%2Coffices%2Cpackages%2Cproducts%2Cpublishers%2Csupport))

PR #203, opened 2019-09-08
Added NoResults screen for both universal and vertical searchers

creotutar:


src/ui/components/results/resultscomponent.js
```diff
@@ -75,6 +75,8 @@ export default class ResultsComponent extends Component {
       render: config.renderItem,
       template: config.itemTemplate
     });
+
+    this._universalUrl = config.universalUrl;
```
creotutar, line 83: I have this as a config option for Results for now. The universal url is used in the no results template to provide a link back to the universal searcher from vertical searcher no results. There are a few options I see rn

(1) Have this universalUrl as a config option in the UniversalResults (on the client to maintain that the url provided in Navigation and the url provided in UniversalResults is consistent)
(2) Search through the navigation tabs for "All" instead (brittle, can have label change easily). Could also add a isUniversal option to the All tab in navigation to make this search easier

src/ui/templates/results/universalresults.hbs
```diff
@@ -1,6 +1,12 @@
-<div class="universal-result-list">
-  <div class="universal-result-sections"
-      data-component="VerticalResults"
-      data-prop="sections">
-  </div>
-</div>
\ No newline at end of file
+{{#if isSearchComplete}}
```
creotutar, line 1: I added isSearchComplete checks because if we have one of NoResults or Results as a default, then it will always quickly flash the incorrect screen if we are not showing the default.

src/ui/components/results/resultscomponent.js
```diff
@@ -90,14 +92,18 @@ export default class ResultsComponent extends Component {
...
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
     return super.setState(Object.assign({ results: [] }, data, {
       isPreSearch: searchState === SearchStates.PRE_SEARCH,
       isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
       isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE,
       includeMap: this._config.includeMap,
       mapConfig: this._config.mapConfig,
-      eventOptions: this.eventOptions()
+      eventOptions: this.eventOptions(),
+      universalUrl: this._universalUrl + window.location.search,
```
benmcginnis: You should check if it's not set first

benmcginnis: Because it might not be provided

src/ui/templates/results/results.hbs
```diff
@@ -1,13 +1,17 @@
...
-      <div class="yxt-Results-map"
-           data-component="Map"
-           data-prop="map"
-           data-opts='{ "isStatic": true }'>
-      </div>
-    {{/if}}
-    {{> results/resultitems}}
-  </section>
+{{#if isSearchComplete}}
+  {{#if showNoResults}}
```
benmcginnis, line 2: why are we doing it this way instead of `{{#if results}}`? Does that not work?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -36,11 +36,14 @@ export default class UniversalResultsComponent extends Component {
...
 
   setState (data, val) {
+    const sections = data.sections || [];
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
     return super.setState(Object.assign({ sections: [] }, data, {
       isPreSearch: searchState === SearchStates.PRE_SEARCH,
       isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
-      isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE
+      isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE,
+      showNoResults: sections.length === 0 && searchState === SearchStates.SEARCH_COMPLETE,
```
benmcginnis: based on how this looks in the handlebars, this logic seems redundant as showNoResults is below the isSearchComplete check

src/ui/templates/results/results.hbs
creotutar: are you referring to the showNoResults variable?
that's b/c universal results checks based off of data.sections and vertical results checks based off of data.results

src/ui/components/results/resultscomponent.js
Delerme: I think this is fine for now, I do like the idea of 2), where a tab has a "home" or "universal" property, but I think we should save that for the larger results refactor

src/ui/templates/results/universalresults.hbs
Delerme: Nice, that was the intended use of that, good to know it worked

Delerme (commented):
lgtm, to @b1lly

Delerme (approved):

benmcginnis (approved):

creotutar:
closes #200

PR #205, opened 2019-09-09
Version 0.8.1

### LINKED ISSUE #97: IFRAME support: JSLib should expose hooks for application state
The JS library should expose some sort of interface hooks for being able to retrieve the query when its updated, or the tab order when its updated.

The reason being, consulting would need to grab this type of state and push it out of the iframe so that they can manage their URL params properly.

### LINKED ISSUE #133: facets should get serialized to URL params
Facets should work like filters, when you select or deselect a facet, it should get persisted to the URL so that users can copy/paste. This would enable the back button to also work properly.

THE ONE open product question (which needs an answer before we push state), is if we're pushing state for every search, the browser history would effectively be navigating through each option instead of going to the previous page.

It's worth figuring out the behavior before we implement.

### LINKED ISSUE #155: child components not garbage collected

### LINKED ISSUE #178: View All Button uses Previous Query

Delerme:
# Version 0.8.1

# Additions
- onStateChange init option (closes #97)
- GeoLocation static filter component
- AccordionResult template
- Icon Component
- Default Analytics

# Changes
- Direct Answer Styling Tweaks
- Truncated results text
- Vertical search uses vertical autocomplete

# Fixes
- Incorrect View All url (fixes #178)
- Clear child array when unmounting (fixes #155)
- URL state saving (fixes #133)


b1lly (approved):
savage

PR #208, opened 2019-09-09
Do not rewrite falsy values in moduledata

Delerme:


benmcginnis:
interesting, not sure I’m clear but why does not copying the data solve the problem?

Delerme:
storage powers component state, and this overwrites any falsy values like empty string, boolean false, and empty array with empty object, which then gets propagated to components

src/core/storage/moduledata.js
```diff
@@ -24,20 +24,18 @@ export default class ModuleData extends EventEmitter {
    * @param {*} data the data to replace the current data
    */
   set (data) {
-    const newData = data || {};
```
b1lly, line 27: 2 questions:

1. Is being able to set null or undefined data going to cause trouble?
2. In terms of "immutability", is this something we want to "clone" and assign?

b1lly (changes requested):

src/core/storage/moduledata.js
Delerme: 1) Null and undefined values throw errors in global storage. ModuleData should be a simple container that gives you whatever you put in, the users should be the ones handling falsy values. I actually found 2 more places this happens
2) yea definitely, but unfortunately going from mutable to immutable can break existing code, and this is meant to be a bugfix, not a breaking change, so I think we should change it in a future commit

b1lly (approved):

PR #211, opened 2019-09-09
More default styling

creotutar:


src/ui/sass/modules/_Result.scss
```diff
@@ -167,6 +185,12 @@
     min-width: 60px;
     text-align: center;
 
+    @media (max-width: 768px)
+    {
```
benmcginnis, line 189: Looks like it’s 768 here vs 767 elsewhere. Perhaps we should add a variable?

benmcginnis (approved):

PR #212, opened 2019-09-10
Extend external interface

Delerme:
Allow external users to subclass the base component class, register custom
components, remove components by name, and access the core api without having to
create a component


Delerme:
@benmcginnis Here's what extending in old (IE-friendly) JS looks like

```
function Thing (config) {
  ANSWERS.Component.call(this, config);
}

Thing.prototype = Object.create(ANSWERS.Component.prototype);
Thing.prototype.constructor = Thing;
Object.defineProperty(Thing, 'type', { get: () => 'Thing' });
Thing.areDuplicateNamesAllowed = () => false;
Thing.defaultTemplateName = () => 'default';
```

b1lly (commented):
Can we please update documentation for these use cases?

Delerme:
Done, added to the base README

README.md
```diff
@@ -175,6 +177,74 @@ ANSWERS.addComponent('SearchBar', {
...
+function MyCustomComponent (config) {
+  ANSWERS.Component.call(this, config);
+
+  this.myProperty = config.myProperty;
+}
+
+MyCustomComponent.prototype = Object.create(ANSWERS.Component.prototype);
+MyCustomComponent.prototype.constructor = MyCustomComponent;
+MyCustomComponent.defaultTemplateName = function () { return 'default' };
+MyCustomComponent.areDuplicateNamesAllowed = function () { return false };
```
b1lly, line 216: Is this any different than:
```
MyCustomComponent.prototype = {
  constructor: MyCustomComponent,
  defaultTemplateName: function() {}
  areDuplicateNamesAllowed: function() {}
}
```

I'd also consider not showing an ES5 example, since its not something the library supports in general without polyfills. But we can add it later?

b1lly (approved):
LGTM

README.md
Delerme: I'm a little wary of recommending people completely override the prototype object. I'll leave it in for now but yea maybe when we revisit our documentation we'll remove it

PR #218, opened 2019-09-12
Add support for field formatters

Delerme:


src/core/models/resultfactory.js
```diff
@@ -33,14 +43,17 @@ export default class ResultFactory {
...
+  static fromGeneric (data, formattedData, index) {
     return new Result({
       raw: data,
-      title: data.name,
-      details: this.truncate(data.description),
-      link: data.website,
+      formatted: formattedData,
+      title: formattedData.name || data.name,
+      details: formattedData.description || this.truncate(data.description),
+      link: formattedData.website || data.website,
```
benmcginnis: probably don't need this formattedData key, I can't imagine actually reformatting a website url in the real world, if someone needed to do that then I think they should do so in transformData rather than us building it in by default

src/core/search/searchdatatransformer.js
```diff
@@ -13,21 +13,21 @@ import DynamicFilters from '../models/dynamicfilters';
...
   }
 
-  static transformVertical (data) {
+  static transformVertical (data, formatters) {
     return {
       [StorageKeys.QUERY_ID]: data.response.queryId,
-      [StorageKeys.NAVIGATION]: new Navigation(), // Veritcal doesn't respond with ordering, so use empty nav.
-      [StorageKeys.VERTICAL_RESULTS]: VerticalResults.from(data.response),
+      [StorageKeys.NAVIGATION]: new Navigation(), // Vertical doesn't respond with ordering, so use empty nav.
+      [StorageKeys.VERTICAL_RESULTS]: VerticalResults.from(data.response, formatters),
```
benmcginnis, line 30: so in this case, does it know the verticalConfigId? because previously it didn't

benmcginnis: so the formatters would not be able to do any vertical based logic and would have to have a default case to work properly

benmcginnis (approved):
Looks GREAT to me

src/core/models/resultfactory.js
Delerme: Yea true, done.

benmcginnis:
@Delerme can we do this on an 0.8.x version? It would be nice to get testing this _before_ 0.9.0 is out.

src/core/search/searchdatatransformer.js
Delerme: yea formatters take in three params, (val, profile, verticalId) where verticalId can be null. So your formatter can use it or ignore it, i.e.
```js
function nameFormatter (name, profile, verticalId) {
  if (verticalId === 'leadership') {
    // show first name only for leadership
    return name.split(' ')[0];
  } else {
    // uppercase all other entity names
    return name.toUpperCase();
  }
}
```

Delerme:
Yea I guess it doesn't depend on anything in 0.9.0

Moved it to 0.8.3

PR #223, opened 2019-09-13
Autofocus with autocomplete and invert autocomplete highlighting

Delerme:
Give this a try before approving if you can, my dev environment is a little wonky


tests/core/models/autocompletedata.js
```diff
@@ -60,15 +60,15 @@ describe('constructing from a response', () => {
         shortValue: 'Jesse Sharps',
         filter: {},
         matchedSubstrings: [ { offset: 8, length: 2 } ],
-        highlightedValue: 'Jesse Sh<strong>ar</strong>ps'
+        highlightedValue: '<strong>Jesse Sh</strong>ar<strong>ps</strong>'
```
benmcginnis, line 63: for these tests, which value is the input search string?

Delerme: It's applying the matchedSubstrings to the `value` strings above, but the matched substrings are kind of arbitrary and don't really reflect a particular query. The server takes the query and figures out the substrings, while the library doesn't care about the original query and just uses the matched substrings, so these tests similarly just make up some substrings with no query in context. 

It's a little confusing, especially with the multi substring test below, but main thing is just look at the matchedSubstrings for what should be (or now that it's inverted, should not be) highlighted

benmcginnis: great ok, I think we also need to change the default scss [here](https://github.com/yext/answers/blob/ed5c0c992abf496a7f8a015539fc49361dc14fb1/src/ui/sass/modules/_AutoComplete.scss#L12) to `$font-weight-normal` and [here](https://github.com/yext/answers/blob/ed5c0c992abf496a7f8a015539fc49361dc14fb1/src/ui/sass/modules/_AutoComplete.scss#L55) to `$font-weight-semibold`

Delerme: Done

benmcginnis (approved):

PR #225, opened 2019-09-15
null check data

aginsberg:


benmcginnis (approved):

Delerme:
_might_ need a similar fix in resultsfactory

PR #226, opened 2019-09-16
Add "More" dropdown functionality to navbar

Delerme:
Posted demo link in slack, let me know if you need the link again


Delerme:
Left the styling pretty simple so you can apply some of your better stylings @benmcginnis which we can merge back in a separate commit

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -97,6 +98,24 @@ export default class NavigationComponent extends Component {
...
+     * The label to show on the dropdown menu button
+     * @type {string}
+     */
+    this.dropdownLabel = config.dropdownLabel || 'More';
+
+    /**
+     * The optional icon to show on the dropdown menu button
+     * @type {string}
+     */
+    this.dropdownIcon = config.dropdownIcon || null;
```
benmcginnis: should probably default to "kabob"

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -132,6 +158,90 @@ export default class NavigationComponent extends Component {
...
+  onMount () {
+    if (!this.static) {
+      this.refitNav();
+      DOM.on(DOM.query(this._container, '.yxt-Nav-more-button'), 'click', this.toggleMoreDropdown.bind(this));
+    }
+  }
+
+  refitNav () {
+    const container = DOM.query(this._container, '.yxt-Nav-container');
+    const moreButton = DOM.query(this._container, '.yxt-Nav-more-button');
```
benmcginnis: for performance reasons, I recommend not making any dom updates inside of a resize callback, but rather doing them inside of requestAnimationFrame.  Not sure why this technique isn't up on MDN anymore but here's the wayback machine link: https://web.archive.org/web/20190116150203/https://developer.mozilla.org/en-US/docs/Web/Events/resize#requestAnimationFrame

benmcginnis (commented):
Do you want to just grab the stylings from #167  ?

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,10 +1,22 @@
...
-    <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
-       href="{{url}}"
-       data-eventtype="VERTICAL_SEARCHER_TAB"
-       data-eventoptions='{{eventOptions}}'>
-      {{label}}
-    </a>
-  {{/each}}
+<nav class="yxt-Nav-container js-yxt-navContainer {{#if static}}yxt-Nav-static{{/if}}">
+  <ul class="yxt-Nav-expanded-items">
+    {{#each tabs}}
```
benmcginnis, line 3: should we do this in setState / updateState? moving the tabs between more & not more.  Not sure if that would blink or not but if you use requestAnimation frame above it probably would work

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,10 +1,22 @@
...
+          data-eventoptions='{{eventOptions}}'>
+          {{label}}
+        </a>
+      </li>
+    {{/each}}
+  </ul>
+  <ul class="yxt-Nav-collapsed-items yxt-Nav-hidden"></ul>
+  <button class="yxt-Nav-more-button {{#if static}}yxt-Nav-hidden{{/if}}">
+    {{#if dropdownIcon}}
+      <img src="{{dropdownIcon}}" class="yxt-Nav-more-icon" alt="More Navigation Links">
```
benmcginnis: this should probably use the icon component, so it would be:
```
<div class="yxt-Nav-moreIconWrapper">
      <div class="yxt-Nav-moreIcon"
      data-component="IconComponent"
      data-opts='{"iconName": "{{#if dropdownIcon}}{{dropdownIcon}}{{/if}}"}'
      data-prop="icon"></div>
    </div>
```

src/ui/components/navigation/navigationcomponent.js
Delerme: as in all implementations should default to kabob? Currently by default it shows "more"

src/ui/components/navigation/navigationcomponent.js
Delerme: Added a 100ms debounce

src/ui/templates/navigation/navigation.hbs
Delerme: Yea I'd like to but that means untangling the tab reordering stuff. I think I'll do that in a separate commit

src/ui/templates/navigation/navigation.hbs
Delerme: done

Delerme (commented):
Added most of the styles from the other PR too

src/ui/components/navigation/navigationcomponent.js
benmcginnis: the `kabob` icon name is the name of the three vertical dots icon that you see next to "more" [here](https://answers.yext.com/), the idea is that you show the text _and_ the icon.

src/ui/sass/modules/_Nav.scss
```diff
@@ -6,20 +6,96 @@
...
+  }
+
+  &-expanded-items
+  {
+    display: inline-block;
+    list-style-type: none;
+    white-space: nowrap;
+  }
+
+  &-collapsed-items
```
benmcginnis: do you mind camel casing module descendant selectors? e.g. `.yxt-Nav-collapsedItems` an descendant selector being a child of a module. https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#ComponentName-descendentName

benmcginnis:
Also same request that this go into an 8.x version, please.  I'd prefer not to open up the 0.9.0 can of worms on the existing implementations.

Delerme:
Yea I put this one on 0.9.0 since it's component changes, and affects the implementation. I guess if we need it for current implementations I can backport it. Is the idea that we stay on 0.8.x until onward?

benmcginnis:
I'd like to get all of the "core" functionality that we've built into the implementations into an 8.x version so that we can just lock the implementations on that until we're ready to test the 9.0 changes and ensure we don't have any regressions.

Delerme:
Gotcha, what's left on that list after nav?

benmcginnis:
Nothing, though calls to action would be nice, I hadn't opened a PR because it's still wip, but I can push it up if you're interested in taking a look.

Delerme:
Those are components too right? I'd prefer to put it on 0.9.0

Alright I'll backport this then

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,10 +1,27 @@
...
-  {{#each tabs}}
-    <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
-       href="{{url}}"
-       data-eventtype="VERTICAL_SEARCHER_TAB"
-       data-eventoptions='{{eventOptions}}'>
-      {{label}}
-    </a>
-  {{/each}}
+<nav class="yxt-Nav-container js-yxt-navContainer {{#if static}}yxt-Nav-static{{/if}}">
+  <ul class="yxt-Nav-expanded-items">
```
benmcginnis: I don't think we actually want these two side by side `<ul>` elements, and it would be better to use this style from #167 https://github.com/yext/answers/pull/167/files#diff-88c35f3580ae030fbd06d720c842fe1bR9 . this will ensure the user can tab through with their keyboard properly

benmcginnis: the only additional part that we would need for all the styling to work is changing the class on the nav-item when it gets moved to the dropdown, as done here: https://github.com/yext/answers/pull/167/files#diff-011a08758bbc7f4e509ccdcdd4cb335fR338

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,10 +1,27 @@
...
+          href="{{url}}"
+          data-eventtype="VERTICAL_SEARCHER_TAB"
+          data-eventoptions='{{eventOptions}}'>
+          {{label}}
+        </a>
+      </li>
+    {{/each}}
+  </ul>
+  <ul class="yxt-Nav-collapsed-items yxt-Nav-hidden"></ul>
+  <button class="yxt-Nav-more-button {{#if static}}yxt-Nav-hidden{{/if}}">
```
benmcginnis: this should also have aria-controls and aria-expanded attributes, where the aria-controls points to the ID of the container of the hidden nav items and the aria-expanded attribute is "false" when it's collapsed and "true" when it's open.

src/ui/components/navigation/navigationcomponent.js
Delerme: Ah we want both, gotcha, set.

src/ui/sass/modules/_Nav.scss
Delerme: Ended up removing this but will keep in mind

src/ui/templates/navigation/navigation.hbs
Delerme: Copied over all the template and styling stuff from your PR

src/ui/templates/navigation/navigation.hbs
Delerme: Added aria stuff

Delerme (commented):
Backported to 0.8.5. Made some tiny template changes so let me know if you see anything weird on your end

benmcginnis:
@Delerme I think you need to rebase this and drop the 9.0 commits

Delerme:
oof yea rebasing back and forth caught some stragglers, fixed

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,10 +1,31 @@
...
+      >
+        {{label}}
+      </a>
+    {{/each}}
+  </div>
+  <button id="yxt-Nav-moreButton" class="yxt-Nav-item yxt-Nav-more yxt-Nav-item--more js-yxt-navMore"
+    aria-haspopup="listbox"
+    aria-labelledby="yxt-Nav-moreButton"
+  >
+    <div class="yxt-Nav-moreIconWrapper">
```
benmcginnis: this is probably duplicative in reality, of the span below it. Also, I divs are invalid inside buttons which I realized after doing some audits.

src/ui/sass/modules/_Nav.scss
```diff
@@ -52,9 +78,86 @@
...
+  &-moreTextWrapper
+  {
+    display: flex;
+    align-items: center;
+  }
+
+  &-moreIconWrapper
+  {
+    margin-top: 1px;
+    margin-right: 7px;
```
benmcginnis: these styles can live in the moreIcon selector, I think

src/core/models/autocompletedata.js
```diff
@@ -78,6 +78,10 @@ export class AutoCompleteResult {
     let highlightedValue = '';
     let nextStart = 0;
 
+    if (invertedSubstrings.length === 0) {
+      return val;
+    }
```
b1lly: is this related to nav?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -87,8 +87,7 @@ export default class FilterBoxComponent extends Component {
   setState (data) {
     super.setState(Object.assign(data, {
       filterConfigs: this._filterConfigs,
-      showApplyButton: !this._searchOnChange,
-      eventType: this._isDynamic ? 'FACETING' : 'FILTERING'
+      showApplyButton: !this._searchOnChange
```
b1lly: ^

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -4,6 +4,13 @@ import Component from '../component';
 import { AnswersComponentError } from '../../../core/errors/errors';
 import StorageKeys from '../../../core/storage/storagekeys';
 import SearchParams from '../../dom/searchparams';
+import DOM from '../../dom/dom';
+
+/**
+ * The debounce duration for resize events
```
b1lly, line 10: Can you describe the purpose of this in the comment?

b1lly (changes requested):
Here's how I think i'd approach this problem:

1. Create a `moreTabs` property on the data model that contains the tabs to be rendered under "more"
2. Template renders using the `tabs` and `moreTabs` data

To handle resizing:
1. When resize happens, compute the total number of "tabs" that can fit from the DOM.
2. Using the total number of "tabs" that can fit, re-generate `tabs` and `moretabs`.
3. Set State with new data to trigger re-render.

That would eliminate all the DOM manipulation we're doing for 'refitNav' and hopefully reduce some of that complexity here.

Thoughts?

src/ui/templates/navigation/navigation.hbs
Delerme: deleted

src/ui/sass/modules/_Nav.scss
Delerme: done

src/core/models/autocompletedata.js
Delerme: no you were seeing old commit because this pr became stale, updated

src/ui/components/filters/filterboxcomponent.js
Delerme: no you were seeing old commit because this pr became stale, updated

src/ui/components/navigation/navigationcomponent.js
Delerme: resize events can trigger many times a second (as the user drags the window), this makes it so we only really do all refitting when the browser size has settled (the user has stopped dragging)

Delerme:
As mentioned on one of Ben's comments, I wanted to make it based on state, but that would require tangling with all of the taborder logic which is currently tightly coupled to setState. Since this is needed for existing implementations, I figured this was factored well enough to be moved to stateful when we start rewriting all the components

b1lly:
> As mentioned on one of Ben's comments, I wanted to make it based on state, but that would require tangling with all of the taborder logic which is currently tightly coupled to setState. Since this is needed for existing implementations, I figured this was factored well enough to be moved to stateful when we start rewriting all the components

@Delerme  How long would it take to refactor this to be based on state? Also, @benmcginnis what is the expected timeline/need for this component being in the library?

If it's urgent, i'm ok merging this in now to unblock them and make a high priority item to fix it. I'd prefer this stateful issue to be resolved before this story is called complete.

Delerme:
Better part of a week, and I'd have port it _back_ to 0.9.0 for the component changes (which is fine, but it just sucks that I wasted a bunch of time porting to 8.x because we needed it soon, and will then have to basically rewrite the component from scratch in 0.9.0). I know we don't plan on bringing in preact anytime soon, but I think we should save rewriting this component until then, so we don't lose x days to redoing it now, just to again redo it later in preact. I think there are more pressing components like autocomplete which has super flaky keyboard and mouse support that we should rewrite before onward for compliance. In this nav case it seems like a refactor would be nice to have, but not a pressing concern

benmcginnis:
I'd like to have this done so we can pin the 10 implementations onto a
reliable version of the library.

benmcginnis:
And eliminate prototype branch.

Delerme:
@b1lly

b1lly:
Ok. merge it in! 

Lets make a separate item to fix the stateful issue. I want to try to be more proactive with tech debt instead of burying ourselves further.

benmcginnis:
@b1lly & @Delerme there's one bug with this that I had mentioned before but doesn't look like it got fixed.  On mobile the "more" dropdown appears centered on the page rather than aligned with the "more" button".  I also think the more button should probably be aligned to the right side.

Screen Shot 2019-09-25 at 1 33 01 PM

benmcginnis:
Oh, also, setting the "static" key causes side-scrolling
Screen Shot 2019-09-25 at 1 42 10 PM

benmcginnis:
And the more button is still in the DOM if you set "static", so it the nav doesn't take up the full width if you set "overflow: hidden"

Delerme:
Made the changes here: https://github.com/yext/answers/pull/247

PR #227, opened 2019-09-16
Search for non string queries

Delerme:


src/ui/components/search/searchcomponent.js
```diff
@@ -131,7 +131,7 @@ export default class SearchComponent extends Component {
   }
 
   onCreate () {
-    if (this.query && this.query.length > 0) {
+    if (this.query) {
```
benmcginnis, line 134: well that's interesting, man

benmcginnis: what about `00000` i feel like that's probably a valid zip somewhere

Delerme: `00000` worked for me, I think it gets coerced into a string, but admittedly `0` doesn't

Delerme: ended up just `String()`ing everything

b1lly (approved):

PR #229, opened 2019-09-17
Use string values for all persistent keys and autofocus behavior

Delerme:


benmcginnis (approved):
LGTM, only approving given the urgency @b1lly.

PR #230, opened 2019-09-17
Don't cache fuzzy versions

Delerme:


benmcginnis (commented):
I think we probably want `no-cache` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Delerme:
That's what it's set to, see line 29

b1lly (approved):

PR #231, opened 2019-09-17
Update analytics event

Delerme:


src/ui/templates/navigation/navigation.hbs
```diff
@@ -2,7 +2,7 @@
   {{#each tabs}}
     <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
        href="{{url}}"
-       data-eventtype="VERTICAL_SEARCHER_TAB"
+       data-eventtype="VERTICAL_TAB_NAVIGATION"
```
benmcginnis: I think you should probably do: 
`data-eventtype="{{#if isFirst}}ALL_TAB_NAVIGATION{{else}}VERTICAL_TAB_NAVIGATION{{/if}}"`
in order to make the "ALL" tab work, since "ALL" is always first right now.

Delerme: alright, let's not accidentally make "ALL" second or something between now and getting the config stuff out

PR #233, opened 2019-09-17
Don't submit search on load for searchbars that redirect

Delerme:


src/ui/components/search/searchcomponent.js
```diff
@@ -138,7 +138,7 @@ export default class SearchComponent extends Component {
   }
 
   onCreate () {
-    if (this.query) {
+    if (this.query && !this.redirectUrl) {
       this.core.setQuery(this.query);
```
b1lly, line 142: do we want to set the query, but not search?

b1lly (changes requested):

src/ui/components/search/searchcomponent.js
Delerme: I was a little worried of listeners on query doing unexpected stuff but it's probably fine, done

b1lly (approved):

PR #238, opened 2019-09-18
date wrapper styling tweaks, collapse accordians by default (#114)

aginsberg:


src/ui/templates/results/resultsaccordion.hbs
```diff
@@ -9,11 +9,11 @@
         {{#each results}}
           <li id="accordion-{{this.id}}"
               class="yxt-AccordionResult js-yxt-AccordionResult
-                    {{#if this.collapsed}} is-collapsed{{/if}}
+                    is-collapsed
```
aginsberg: @benmcginnis Is this okay for having them be collapsed by default? AFAIK, if want to pass in the option by default, the change would need to happen in the answers-experience-builder

benmcginnis: please set is-collapsed to true in the component js, rather than here.

benmcginnis (approved):
lgtm

PR #239, opened 2019-09-18
wip

benmcginnis:


benmcginnis:
@Delerme / @b1lly I'm putting up what I have so far on this for feedback on the architecture.  Also, HitchHikers want to automatically create CTAs by default from a new custom field type that they're going to add to all implementations.  The problem right now is that I can't realistically implement that on the model because the results don't know enough to create their analytics eventOptions (they need the vertical config id & searcher type).  If, as we discussed on slack, we start returning that information from the API and giving it to the results factory, this would be _a lot_ cleaner.

benmcginnis:
Closing as this is going to be taken over by the OOBX initiative

PR #240, opened 2019-09-18
Add IE bundle

Delerme:


Delerme:
IE still has one issue with this line
`	    var e = new Event(event, Object.assign({`
which prevents autocomplete from working, but I was able to search stuff and see results

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -69,7 +129,7 @@ function watchCSS (cb) {
 }
 
 exports.default = parallel(
-  series(bundle, minifyJS),
+  series(bundle, legacyBundle, minifyJS, minifyLegacy),
```
b1lly: minor nit, you can probably move legacyBundle and minifyLegacy to a separate series right?

b1lly (changes requested):
seems solid. what you mean by there is still an IE bug?

@benmcginnis are you aware of that, or know the proper resolution there?

conf/gulp-tasks/library.gulpfile.js
Delerme: done

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -40,13 +42,85 @@ function bundle () {
...
+      format: 'iife',
+      name: NAMESPACE,
+      sourcemap: true
+    },
+    plugins: [
+      resolve(),
+      insert.prepend(`
+        import '@babel/polyfill/noConflict';
+        import 'whatwg-fetch';
+        if (!Element.prototype.matches) { // Element.matches polyfill from MDN
```
benmcginnis, line 58: you might want to create a polyfills.js file to hold all of these and then do a prepend `import polyfills` that way it will be a little nicer to update

benmcginnis (approved):

conf/gulp-tasks/library.gulpfile.js
Delerme: agreed, punting til I have better internet and don't have to wait 3 seconds between key presses

PR #242, opened 2019-09-19
adding config option to specifiy whether or not to collapse pins (#241)

aginsberg:
Needed for cases where clients would like an individual pin per entity, with the corresponding label. On Citizens, we are seeing multiple pins with the same label https://answers.citizensbank.com/locations?tabOrder=all%2Clocations%2Cfaqs%2Clearning%2Cother_links%2Cproducts&query=pennsylvania&DynamicFilters.filterbox.filter0=%5B%5D&DynamicFilters.filterbox.filter1=%5B%22ATM+Voice+Assisted%22%2C%22Supermarket+Branch%22%5D (filter on 'ATM voice assisted' and 'supermarket branch'). There is a pin with label 2 in the bottom right and in the top left


benmcginnis (approved):
LGTM, can you update the commit message to explain the bug to @b1lly & @Delerme?

PR #245, opened 2019-09-23
Checking source instead of specific field when ResultFactory transforms response

bwangYext:
TEST=auto,manual
npm run test


Delerme:
Looks good to me. Out of curiosity, is `source` set for knowledge graph results? Doesn't have to happen right now, but wondering if we also want to treat knowledge graph results separately.

To @b1lly

src/core/models/resultfactory.js
```diff
@@ -29,11 +30,16 @@ export default class ResultFactory {
         });
       }
 
-      if (data.htmlSnippet || data.htmlTitle) {
+      if (source === 'GOOGLE_CSE' || source === 'CUSTOM_SEARCHER') {
```
b1lly, line 33: This is probably OK for now, but curious if the "custom searcher" fall in the "default" case instead of the google case.

b1lly (approved):
LGTM

PR #246, opened 2019-09-25
Remove static map support in results, fix map styling

Delerme:
This prevents stretching out the static maps. Users should be specifying the right (or at least the max) map dimensions in the config so the provider can correctly make the static image


b1lly (approved):

b1lly:
Seems fine, but just wanna confirm that it doesn't break anything that @benmcginnis is depending on for his implementations.

src/ui/sass/modules/_Results.scss
```diff
@@ -134,11 +134,7 @@
   &-map
   {
     display: flex;
+    justify-content: center;
```
benmcginnis: Are you sure this works for the dynamic map in the universal results?  The designs want them to be full width, so can you double check your change against the designs? If it comes out of the box looking like what UX wants, I'm fine.  You can check with this repo: https://github.com/yext/answers-oobx-test and do a search for "locations in london"

Delerme: static maps can't be full width unless the user passes the right width in the config, since the image is generated based on that and would be stretched

You can full width dynamic maps though, so I made that the default (100% width and height). The user can still override dynamic map size in the config

benmcginnis:
So, just to be clear, generally images need to have their width set to 100%
of their parent so that they resize correctly as their parent container
changes size.  It's a basic responsive image technique.  That can cause
pixelation on the image if it's not large enough.  That said, Answers
should probably produce an image large enough to work with the
default designs, rather than having the implementer have to specify a size
that makes the image work with the default designs.

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -49,9 +49,12 @@ export default class GoogleMapProvider extends MapProvider {
       this._pinConfig
     );
 
+    const width = this._width || 600;
```
benmcginnis: these should be large enough so that they provide a retina-capable image for the default designs.

Delerme: This might be worth a bigger discussion with product, but both mapbox and google have some pretty strict limits on static map sizes, with google requiring a premium (not pay-as-you-go) plan and mapbox requiring you to contact sales after 100 big images. I would definitely like for us to default to nice big images but I'm a little worried of setting these as defaults

Delerme: I feel like a good solution might be default to dynamic maps (already the case in this PR) and then have no defaults for static maps, you must specify the size, scale, and zoom. Thoughts?

benmcginnis:
Can we just remove the static maps capability then?  Since it doesn't
really work with the designs and it's not WCAG compliant at all, we might
as well just drop teh functionality.

Delerme:
static maps are dead, long live dynamic maps

benmcginnis (approved):

PR #247, opened 2019-09-25
Update nav styling

Delerme:
Move "more" button and dropdown to right side, horizontally
scroll navbar if static


src/ui/sass/modules/_Nav.scss
```diff
@@ -21,7 +21,7 @@
   &-container
   {
     display: flex;
-    justify-content: flex-start;
+    justify-content: space-between;
```
benmcginnis: this is going to have a weird effect on sites with few nav items.  I recommend either setting a `margin-left: auto` or a `justify-self: flex-end` on the more button.

src/ui/templates/navigation/navigation.hbs
```diff
@@ -12,18 +12,20 @@
...
-    data-opts='{"iconName": "{{dropdownIcon}}"}'
-    data-prop="icon"></span>
-    {{dropdownLabel}}
-  </button>
-  <div class="yxt-Nav-modal js-yxt-navModal {{#if expanded}}is-active{{/if}}"
-    role="listbox"
-    aria-labelledby="yxt-Nav-moreButton"
-  ></div>
+  {{#unless static}}
+    <button id="yxt-Nav-moreButton" class="yxt-Nav-item yxt-Nav-more yxt-Nav-item--more js-yxt-navMore"
```
benmcginnis: I think you also need a common wrapper of the more button and the dropdown in order to properly position the dropdown relative to the button rather than relative to the whole bar.

src/ui/sass/modules/_Nav.scss
Delerme: true, done

src/ui/templates/navigation/navigation.hbs
Delerme: yea I figured it'd be equivalent since the button was aligned right, but probably better if it actually tracks the button. Done.

benmcginnis (approved):

benmcginnis:
Julian I think we could cut an 8.7 release with this if y'all are up for it.

PR #248, opened 2019-09-25
Add configVersion param

Delerme:
Currently optional to support v1 configs, I'm assuming we'll make it required in 1.0.0


src/core/search/questionanswerapi.js
```diff
@@ -37,6 +44,7 @@ export default class QuestionAnswerApi {
       endpoint: '/v2/accounts/me/questions',
       apiKey: this._apiKey,
       params: {
+        'version': this._configVersion,
```
b1lly: Does this work? This endpoint is owned by the reviews team (slanden probably can help). 

I don't think this would ever be used by them.

b1lly (approved):
Only question I have is about the QA endpoint. As long as it doesn't break anything LGTM.

Yeah, is there concerns with making it required now?

src/core/search/questionanswerapi.js
Delerme: removed

Delerme:
Making it required now would make 0.9.0 incompatible with v1 configs, meaning users would have to upgrade configs before using 0.9.0. I don't think there's a good reason to make that a blocker right this minute. We can take our time rolling out config v2 and make it required next version

b1lly:
SGTM. LGTM.

PR #250, opened 2019-09-25
Create HighlightedValue model and extract highlight logic from autocomplete

bwangYext:
TEST=auto,manual
npm run test


src/core/models/highlightedvalue.js
```diff
@@ -0,0 +1,103 @@
+/** @module HighlightedValue */
+
+/**
+ * Model representing a highlighted value
+ *
```
Delerme: extra line

src/core/models/highlightedvalue.js
```diff
@@ -0,0 +1,103 @@
...
+    this.shortValue = data.shortValue || this.value;
+    this.matchedSubstrings = data.matchedSubstrings || [];
+  }
+
+  /**
+   * get highlighted value string
+   * @returns {string}
+   */
+  get () {
+    return this._getValue(false);
```
Delerme: hmm to avoid having a function with two different behaviors depending on a flag, we can probably make the flow a little more functional.

- Both functions need to validate and clean the input (lines 31-47) so maybe pull that into a validate() function
- once validated, get() can call _buildHighlightedValue directly
- getInverted() takes the output of validate() and inverts it, so you can rename _getHighlightSubstrings to invertSubstrings and remove the shouldInvert flag

So get() would look like
```
validate();
return _buildHighlightedValue(...)
```

and getInverted would look like
```
validate()
const invertedSubstrings = _invertSubstrings(...)
return _buildHighlightedValue(...)
```

then you can get rid of the invert flags and the _getValue function entirely

Delerme (commented):
nice. Just to confirm, this works for the "did you mean" response too right?

src/core/models/highlightedvalue.js
bwangYext: Done.

src/core/models/highlightedvalue.js
bwangYext: Done. Since default matchedSubstrings is [], we do not need to check if it is undefined. Also the other methods can handle empty array. So have a new method called _sortMatchedSubstrings().

bwangYext:
Yes, HighlightedValue in autocomplete and spell check responses are the same.

Delerme (approved):

PR #253, opened 2019-09-26
Support did you mean in ui

bwangYext:


src/core/models/spellcheck.js
```diff
@@ -0,0 +1,51 @@
...
+     * @type {string}
+     */
+    this.query = data.query || null;
+
+    /**
+     * The corrected query
+     * @type {string}
+     */
+    this.correctedQuery = data.correctedQuery || null;
+    /**
```
Delerme, line 23: new line above

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,59 @@
+/** @module SpellCheckComponent */
+
+import Component from '../component';
+import SearchParams from '../../dom/searchparams';
+import StorageKeys from '../../../core/storage/storagekeys';
+
+/**
+ * SpellCheckComponent will support displaying suggestion, autocorrect and combined(maybe in the future)
+ * provided from speelling correction.
```
Delerme: typo

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,59 @@
...
+
+  static get type () {
+    return 'SpellCheck';
+  }
+
+  static defaultTemplateName () {
+    return 'search/spellcheck';
+  }
+
+  init (opts) {
```
Delerme: Is this needed? The superclass one should get called automatically

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,59 @@
...
+    super.init(opts);
+    return this;
+  }
+
+  setState (data, val) {
+    return super.setState(Object.assign(data, {
```
Delerme: Object.assign modifies the first argument, so to avoid doing that pop in `{}` as the first argument so it creates a new object

Delerme (commented):
Nice, can you also update the README in root

src/core/models/spellcheck.js
```diff
@@ -0,0 +1,51 @@
...
+     * The original query
+     * @type {string}
+     */
+    this.query = data.query || null;
+
+    /**
+     * The corrected query
+     * @type {string}
+     */
+    this.correctedQuery = data.correctedQuery || null;
```
b1lly, line 21: is this the suggestedQuery?

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,59 @@
...
+
+  _buildRedirectQueryUrl (query) {
+    if (query === undefined) {
+      return '';
+    }
+    let params = new SearchParams(window.location.search.substring(1));
+    params.set('query', query);
+    params.set('spellCheck', false);
+    return window.location.href.split('?')[0] + '?' + params.toString();
+  }
```
b1lly, line 50: is it possible to generate this from the core and then be part of the core spellcheck model?

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,59 @@
...
+
+  _getHelpText (type) {
+    switch (type) {
+      case 'SUGGEST':
+        return 'Did you mean:';
+      default:
+        return '';
+    }
+  }
```
b1lly, line 59: same question, can this be part of the core model? or is it influenced by component configuration?

b1lly (changes requested):
Nice start dude :)

src/core/models/spellcheck.js
bwangYext: Done

src/ui/components/search/spellcheckcomponent.js
bwangYext: Done

src/ui/components/search/spellcheckcomponent.js
bwangYext: Removed.

src/ui/components/search/spellcheckcomponent.js
bwangYext: Not needed after moving to model.

src/core/models/spellcheck.js
bwangYext: Yes, the response field name is called corrected query.

src/ui/components/search/spellcheckcomponent.js
bwangYext: Done.

src/ui/components/search/spellcheckcomponent.js
bwangYext: Done.

src/core/models/spellcheck.js
```diff
@@ -48,4 +68,24 @@ export default class SpellCheck {
...
+
+  _buildRedirectQueryUrl (query, type) {
+    if (query === null) {
+      return '';
+    }
+    let params = new SearchParams(window.location.search.substring(1));
+    params.set('query', query.value);
+    params.set('skipSpellCheck', true);
+    params.set('queryTrigger', type.toLowerCase());
+    return window.location.href.split('?')[0] + '?' + params.toString();
```
b1lly: So one thing the core should never have is a dependency on the DOM.

For instance, we shouldn't be accessing "window" directly here. Is it possible to get the existing URL from the  `storage`, and provide it to the model as input that it can use to build the new URL?

b1lly (changes requested):

src/core/models/spellcheck.js
Delerme: As you said, the core shouldn't really have a dependency on the DOM. Is there a strong reason to go through the trouble of passing around the URL through core? It seems to make more sense just keeping the logic in the component

bwangYext: Passed params to the model. Do we need to replace !, '. (, ) as SearchParams.encode() does?

b1lly: The base URL, or the url params string? I think in an ideal world there would be a single source of truth that powers all the data for the application, which includes data for the URL and URL params. 

In that world, the `storage` would be the thing that contains all that information-- which would eliminate the need for it to be "passed around". 

I think it's time to revisit the storage as a whole. I'll setup a meeting to get a brain storming sessions started.

Delerme: @b1lly in the short term is it fine to keep the url logic in the component? It seems very specific to that component and if we later revisit storage it won't be too difficult to move over

b1lly: I'm open to moving it, but what makes you feel that the data for the view model should be generated in the component instead of being part of the storage? 

For me, keeping the core view models and their data part of the core feels correct. Part of the purpose of the core is to encapsulate the transformation logic for models, and to provide the core models that are needed to power the UI.

I also think it's better for testability and robustness... And if someone were to ever build their own UI components, they would already have access to the proper models that power the view.

I agree that it hasn't always been a consistent practice across our components, but I think that's mostly when configuration from the UI has to impact the data in the component.

Delerme: Yea it sounds like we have two different ideas for storage

I think storage should be used exclusively for global app knowledge. Kind of like how redux is used in other applications. The data there can power components, but not all components and state should be powered by storage. I.e. it's just used for shared knowledge and components explicitly use it when needed, but they should not automatically be powered by storage. Therefore, if the knowledge (in this case the url) is not needed elsewhere it should not be in storage.

I think your idea of storage is a single source of truth for app state that powers all the components. I.e. everything that components need should go into storage, which automatically triggers rerenders on the components. Therefore, since this url is needed for rendering, it should be in storage.

Definitely think it's worth stepping back and investigating our architecture for the future, but I'd argue it should not block this PR and the url logic should just stay in the component for now.

Delerme: As discussed, let's leave in the component for this PR while we revisit the higher level storage architecture

@bwangYext go ahead and keep the url creation logic in the component like you first had

bwangYext: @Delerme @b1lly Moved back to component.

src/core/core.js
```diff
@@ -125,7 +125,7 @@ export default class Core {
         ...query,
         isDynamicFiltersEnabled: this._isDynamicFiltersEnabled
       })
-      .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters))
+      .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, this.persistentStorage.getAll()))
```
Delerme: can these persistenststorage calls be removed now?

src/core/search/searchdatatransformer.js
```diff
@@ -13,22 +14,24 @@ import DynamicFilters from '../models/dynamicfilters';
  * component library and core storage understand.
  */
 export default class SearchDataTransformer {
-  static transform (data, urls = {}, formatters) {
+  static transform (data, urls = {}, formatters, params) {
```
Delerme: params is not needed now right?

bwangYext: cleaned up.

Delerme (approved):

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,60 @@
...
+  _buildRedirectQueryUrl (query, type) {
+    if (query === undefined) {
+      return '';
+    }
+    let params = new SearchParams(window.location.search.substring(1));
+    params.set('query', query.value);
+    params.set('skipSpellCheck', true);
+    params.set('queryTrigger', type.toLowerCase());
+    return '?' + params.toString();
+  }
```
b1lly, line 50: I don't think it matters too much for this function since SearchParams is technically a DOM API dependency, but just for context, one good practice that is worth following is to always "provide" the input that is being operated on.

For instance, instead of "looking up the URL string" from the DOM as part of the function, you provide it to the function.

e.g.
```
_buildRedirectQueryUrl (urlString, query, type) { ... }
```

Then to use it:
```
_buildRedirectQueryUrl(window.location.search.substring(1), ...);
```
That enables you to write unit tests on your helper method without the DOM dependency.

Again, since SearchParams is technically a DOM dependency, I wouldn't worry too much here. But just for general context.

b1lly (approved):
LGTM

PR #260, opened 2019-10-01
Zendesk vertical

vramesh1:
Adding support for the Zendesk Vertical, and minor changes.
1) _answers/src/core/models/resultfactory.js_
- map Zendesk API server results to JS lib Result objects
2) _answers/conf/gulp-tasks/library.gulpfile.js_
- change to dev build config to also build legacy bundle on changes
3) _README.md_
- minor fix to documentation


src/core/models/resultfactory.js
```diff
@@ -40,6 +40,11 @@ export default class ResultFactory {
         continue;
       }
 
+      if (source === 'ZENDESK') {
```
benmcginnis, line 43: perhaps this would be better restructured as a switch statement on the source

benmcginnis (approved):

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -133,7 +133,7 @@ function compileCSS () {
 function watchJS (cb) {
   return watch(['./src/**/*.js'], {
     ignored: './dist/'
-  }, series(bundle));
```
Delerme: what's the use case for this? Ideally we'd keep watch builds as fast as possible

vramesh1: if we're working off the generated answers.js file locally, then we'd need this change to enable hot compilation and regeneration of this file from changes to JS files.
we're doing this for answers-modern.js currently.
is that the preferred generated source file instead of answers.js?
isn't answers.js the preferred generated source file that works across all browsers, that we should use or at least cover during development?

Delerme: the additional code is just polyfills, so it takes longer to build and is a larger bundle and doesn't really contribute to testing anything locally. It's similar to us not minifying the bundle here. Minifying is slow and doesn't affect logic, so there's no need to do that in a watch loop

vramesh1: ok, so answers.js contains polyfills for IE, and answers-modern.js contains everything you need for other browsers?
and as long as we're not using IE locally, we should always use answers-modern.js instead of answers.js?

Delerme: right, `answers-modern` is the library. `answers` has additional code to interface with older browsers (so unless you need to test that additional code, there is no reason to build this in a loop). Minifying is the same logic but takes up less space, so `answers*.min` is the same thing as the non-minified logically

vramesh1: ok, undoing this change

src/core/models/resultfactory.js
vramesh1: sure, will defer to the other reviewers on whether to restructure this.
don't have a strong opinion on that.
would prefer to make that change separately.

Delerme (approved):

PR #263, opened 2019-10-02
Add location biasing

Delerme:
When geolocation is enabled, bias all searches. When near me intent is detected
and promptLocation is true on the searchbar, prompt for geolocation


src/ui/components/filters/geolocationcomponent.js
```diff
@@ -213,13 +213,21 @@ export default class GeoLocationComponent extends Component {
...
+    const filter = this._buildFilter(position);
     this.core.persistentStorage.set(`${StorageKeys.QUERY}.${this.name}`, query);
     this.core.persistentStorage.set(`${StorageKeys.FILTER}.${this.name}`, filter);
     this.core.setFilter(this.name, filter);
+    const coords = {
+      lat: position.coords.latitude,
+      lng: position.coords.longitude,
+      radius: position.coords.accuracy
+    };
+    this.core.globalStorage.set(StorageKeys.GEOLOCATION, coords);
```
b1lly: why not just inline the object?

```
this.core.globalStorage.set(StorageKeys.GEOLOCATION, {
  // ...
})
```

src/ui/components/search/searchcomponent.js
```diff
@@ -141,6 +150,23 @@ export default class SearchComponent extends Component {
     if (this.query && !this.redirectUrl) {
       this.core.setQuery(this.query);
     }
+    if (this._promptForLocation) {
```
b1lly, line 153: style nit:

to reduce excessive indentation it's sometimes better to short circuit functions instead of wrapping them in control.

```
if (this._promptForLocation === false) {
  return this;
}

this.core.globalStorage.on('update', ...)
```

src/ui/components/search/searchcomponent.js
```diff
@@ -141,6 +150,23 @@ export default class SearchComponent extends Component {
...
+      this.core.globalStorage.on('update', StorageKeys.INTENTS, i => {
```
b1lly: ```i = intent``` ? can we use descriptive variable names for readability purposes?

src/ui/components/search/searchcomponent.js
```diff
@@ -141,6 +150,23 @@ export default class SearchComponent extends Component {
...
+        if (i.nearMe && !this.core.globalStorage.getState(StorageKeys.GEOLOCATION)) {
```
b1lly: can we be more explicit about our expressions? I found in large applications there's a lot of benefit to following this practice. I think it'd be super beneficial to us long-term if we were to adapt this practice and use it more consistently. 

by using `!variable`, we're vaguely describing what we want because it can mean many things (e.g. undefined, null, false, empty string, etc) and can cause confusion. And because of the misunderstanding, it can also lead to wonky side effects in the future.

basically what i'm suggesting is to use stricter type comparisons for our expressions.

src/ui/components/search/searchcomponent.js
```diff
@@ -141,6 +150,23 @@ export default class SearchComponent extends Component {
...
+                lat: position.coords.latitude,
+                lng: position.coords.longitude,
+                radius: position.coords.accuracy
+              };
+              this.core.globalStorage.set(StorageKeys.GEOLOCATION, coords);
+              this.search(this.query);
+            }
+          );
+        }
+      });
```
b1lly: Just to clarify... a response from the server WILL NOT prompt the user for permission to use HTML5 geolocation, correct? It will just use it if it's already enabled?

b1lly (changes requested):

src/ui/components/filters/geolocationcomponent.js
Delerme: done

src/ui/components/search/searchcomponent.js
Delerme: yea I usually try to do this, only weird thing here is we may have additional onCreate logic in the future, so it's better to be explicit and say "do this action if this flag is present" as opposed to "exit this onCreate function if this flag isn't present). The indentation is ugly though, so I pulled it out into a method, let me know what you think

src/ui/components/search/searchcomponent.js
Delerme: done

src/ui/components/search/searchcomponent.js
Delerme: yea typically `!variable` is to ensure we properly handle all the falsey values without spelling them out one by one, i.e we don't want to check for `undefined` and accidentally forget `null` or empty string. In this case though, I think we can guarantee storage returns null or the value, so I added a `!== null` check. Since `nearMe` is a boolean, we can continue to use `!`

src/ui/components/search/searchcomponent.js
Delerme: It will prompt as requested by Max

src/ui/components/search/searchcomponent.js
b1lly: Yup, SGTM. The other common practice in Javascript is to build out a suite of comparative helpers to enable us to do these type of things easier.

E.g. isString, isEmpty, isNull, isFalsy, etc. That makes being more explicit a little less verbose and a little more prosaic to in some of the "permutation" cases.

src/ui/components/search/searchcomponent.js
b1lly: OK, that might not be WCAG compliant, but if Max requested it then I guess it's OK for now. I'm pretty sure we're going to have to revisit this mechanism in the future.

src/ui/components/search/searchcomponent.js
b1lly: LGTM

src/ui/components/search/searchcomponent.js
Delerme: Max actually said there weren't WCAG concerns, check the slack thread where I sent the demo

b1lly (approved):

PR #267, opened 2019-10-04
Add clearbutton option to search component

Delerme:


src/ui/templates/search/search.hbs
```diff
@@ -7,6 +7,17 @@
       <label class="yxt-SearchBar-label" for="query">{{labelText}}</label>
       {{!// TODO (bmcginnis): migrate js class names}}
       <input class="js-yext-query yxt-SearchBar-input" type="text" name="query" value="{{query}}">
+      <button
+        type="button"
+        class="js-yxt-SearchBar-clear yxt-SearchBar-clear yxt-SearchBar-hidden"
+        data-eventtype="SEARCH_CLEAR_BUTTON"
+      >
+        <div class="yxt-SearchBar-buttonImage"
```
benmcginnis: I don't actually think you need the div within the button.  I think you can just set the data-component on the button and it will just become the container of the icon

benmcginnis: if you want to have it as a child, it should be a span.  buttons can't contain block level elements

Delerme: true, done

benmcginnis:
It seems a little janky right now with the set state implementation, it seems like it should appear if I start typing anything and if the value becomes empty it should disappear. You may want to just do that with an on change listener

src/ui/components/search/searchcomponent.js
```diff
@@ -164,11 +178,41 @@ export default class SearchComponent extends Component {
...
+    }
+
     if (this.autoFocus === true && this.query.length === 0 && this.autocompleteOnLoad) {
       DOM.query(this._container, this._inputEl).focus();
     }
   }
 
+  initClearButton () {
+    const button = DOM.query(this._container, '.js-yxt-SearchBar-clear');
+    this._showClearButton = this._showClearButton || this.query;
```
benmcginnis, line 192: I think maybe you should make a getter for showClearButton rather than setting a property, i think that's what's causing the odd behavior

benmcginnis: Also if this is called at all, you probably don't need to check the value of this.showClearButton, just this.query.length

src/ui/components/search/searchcomponent.js
```diff
@@ -164,11 +178,41 @@ export default class SearchComponent extends Component {
...
+    button.classList.toggle('yxt-SearchBar-hidden', !this._showClearButton);
+
+    DOM.on(button, 'click', () => {
+      this.query = '';
+      this._showClearButton = false;
+      button.classList.add('yxt-SearchBar-hidden');
+      this.setState({});
+      this.core.persistentStorage.set(StorageKeys.QUERY, this.query);
+      this.core.setQuery(this.query);
```
benmcginnis, line 201: I don't think this is actually resetting the classes to hide the button, at least it's not in my testing.

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -97,6 +97,11 @@
     }
   }
 
+  &-hidden
```
benmcginnis: this is technically a state, so it would be `&--hidden` rather than a child selector which is what `&-hidden` represents

Delerme: done

benmcginnis (approved):

PR #268, opened 2019-10-04
Version 0.8.8

### LINKED ISSUE #259: CUSTOM_SEARCHER using wrong model
https://github.com/yext/answers/blob/dist/src/core/models/resultfactory.js#L33

Custom searcher results are being passed into 'fromGoogleCustomSearchEngine', which then tries to access fields that don't exist, i.e. 'htmlTitle'.

### LINKED ISSUE #264: View All using incorrect event name
I believe that the event type for View All links should be VERTICAL_VIEW_ALL, unless something will be changing soon. That is what it is on Citizens and it seems to be firing properly there. Sites using this template return an error when trying to fire the event https://github.com/yext/answers/blob/master/src/ui/templates/results/resultssectionheader.hbs#L18, i.e. Yext answers

Delerme:
# Version 0.8.8

# Additions
- Zendesk vertical support
- "Near Me" search intent and location biasing
- Search clear button

# Changes
- Empty queries ignored in universal search

# Fixes
- Custom vertical parsing (fixes #259)
- Update's "view all" analytics event (fixes #264)


benmcginnis (approved):
LGTM

PR #269, opened 2019-10-04
Actually passing skipSpellCheck and queryTrigger params with Search r…

bwangYext:
TEST: npm run test, manual


src/core/storage/globalstorage.js
```diff
@@ -66,6 +66,15 @@ export default class GlobalStorage {
     return data;
   }
 
+  /**
+   * Remove the data in storage with the given key to the provided data,
+   * @param {string} key the storage key to delete
+   */
+  delete (key) {
+    // Note: Do we need to clean up listeners here?
```
Delerme, line 74: I think it's handled

The ModuleData gets deleted and garbage collected, and since that extends EventEmitter (which tracks the listeners) everything should be cleaned up

Delerme (approved):

PR #275, opened 2019-10-07
Add sample Answers app

data-enabler:
Screen Shot 2019-10-08 at 7 36 48 PM
Screen Shot 2019-10-08 at 7 37 23 PM
Screen Shot 2019-10-08 at 7 37 34 PM


sample-app/public/app.js
```diff
@@ -0,0 +1,93 @@
...
+const config = JSON.parse(document.getElementById('config').textContent);
+
+window.initAnswers = async function initAnswers () {
+  const {
+    apiKey,
+    answersKey,
+    verticalKey,
+    mapProvider,
+    mapApiKey
+  } = config;
```
b1lly, line 12: Style nit:

Lets try to be consistent with our declarations
```
const config = {
   ..
}
```

sample-app/public/app.js
```diff
@@ -0,0 +1,93 @@
...
+  const verticalConfig = config.verticals[verticalKey];
+  const mapConfig = {
+    includeMap: true,
+    mapConfig: {
+      mapProvider,
+      apiKey: mapApiKey
+    }
+  };
```
b1lly, line 20: new line

sample-app/public/app.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-env es6 */
```
b1lly: I don't think we use eslint?

sample-app/public/style.css
```diff
@@ -0,0 +1,32 @@
...
+  flex-grow: 1;
+  justify-content: space-around;
+}
+
+.yxt-QuestionSubmission::before {
+  content: "Can't find your answer?";
+  display: block;
+  font-size: 200%;
+  font-weight: bolder;
+}
```
b1lly: I'd prefer if we kept the default styling, and baked this type of stuff into the sdk based on the UX mocks.

It's not a big deal for now.

sample-app/index.js
```diff
@@ -0,0 +1,89 @@
...
+    markup: `
+    <main>
+      <div class="search-container"></div>
+      <div class="direct-answer-container"></div>
+      <div class="universal-results-container"></div>
+      <div class="question-submission-container"></div>
+    </main>`
+  });
+  writeFile('index.html', universalHtml)
+    .catch(console.error);
```
b1lly: Minor nit: can we just push both operations into helpers?

And/or move the templates to static files, instead of inlined? It's fine for them to be ES6 templates in the short-term, but I can imagine cases more long-term where keeping the templating language consistent (e.g. handlebars) would be beneficial.

```
writeFile('index.html', generateUniversalHtml(config))
  .catch(console.error);

writeFile('vertical.html', generateVerticalHtml(config))
  .catch(console.error);
```

b1lly (changes requested):
SAVAGE!

sample-app/public/app.js
data-enabler: Not sure what you mean here. Are you just suggesting that I don't use object destructuring?

sample-app/public/app.js
data-enabler: oops

sample-app/public/app.js
data-enabler: `semistandard` is just a set of configuration options for ESLint.
However, I think the issue I was dealing with here was that my IDE is using the default ESLint config because this repo doesn't have its own eslintrc.

Will remove.

sample-app/public/style.css
data-enabler: I'll trim this down the the minimum required to actually look like a webpage.

sample-app/index.js
data-enabler: Not a problem; I was on the fence about making these helpers actually.

sample-app/index.js
```diff
@@ -1,9 +1,20 @@
 const fs = require('fs');
 const path = require('path');
 const util = require('util');
-
 const writeFile = util.promisify(fs.writeFile);
 
+const Handlebars = require('handlebars');
+const metadataTemplate = Handlebars.compile(
+  fs.readFileSync('./metadata.handlebars', 'utf8')
```
b1lly: for consistency, can we just name all the handlebars templates *.hbs?

b1lly (changes requested):
Minor templates file name changes. Otherwise LGTM.

sample-app/index.js
data-enabler: done

b1lly (approved):

PR #277, opened 2019-10-09
Send url to analytics for blue links

Delerme:


src/ui/components/results/resultsitemcomponent.js
```diff
@@ -59,7 +68,7 @@ export default class ResultsItemComponent extends Component {
    */
   setState (data) {
     return super.setState(Object.assign(data, {
-      eventOptions: this.eventOptions(data.id)
+      eventOptions: this.eventOptions(data.id, data.link)
```
b1lly, line 71: should it just take in data?

Delerme: I think it's good to keep it explicit for now. If this gets any bigger I'll swap it

b1lly (approved):

PR #280, opened 2019-10-10
Update result factory to handle Algolia response

bwangYext:
Discussed with product, we are passing objectID for now since it is the only general field on Algolia response.

TEST=auto
npm run test


src/core/models/resultfactory.js
```diff
@@ -45,6 +45,11 @@ export default class ResultFactory {
         continue;
       }
 
+      if (source === 'ALGOLIA') {
+        results.push(ResultFactory.fromAlgoliaSearchEngine(data));
+        continue;
+      }
+
```
Delerme: I think we've hit the point where we should probably refactor these 4 `if`s into a switch statement with a default case. Do you mind taking care of that in this commit?

bwangYext: Done.

src/core/models/resultfactory.js
```diff
@@ -115,6 +115,21 @@ export default class ResultFactory {
     });
   }
 
+  /**
+   * Converts an API result object into a result view model.
+   * Maps view model fields based on the API data for a Algolia Search Engine object.
+   * Deatiails field is set to objectID since response has only one general field objectID.
```
Delerme: typo

Delerme (commented):
nice thanks

src/core/models/resultfactory.js
bwangYext: Done.

Delerme (approved):

b1lly (commented):
Are there any tests we need to add to the ResultFactory?

bwangYext:
> Are there any tests we need to add to the ResultFactory?

Currently there are only tests for truncate().

PR #282, opened 2019-10-14
Add 'showAllOnLoad' option to vertical results

Delerme:


src/ui/components/results/resultscomponent.js
```diff
@@ -83,6 +90,17 @@ export default class ResultsComponent extends Component {
...
 
+  onCreate () {
+    const query = this.core.globalStorage.getState(StorageKeys.QUERY);
+    if (this._showAllOnLoad && (query === null || query === '')) {
+      if (!this._verticalKey) {
+        throw new AnswersComponentError('showOnLoad requires a verticalKey');
+      }
+      this.core.globalStorage.set(StorageKeys.QUERY_TRIGGER, 'onLoad');
+      this.core.verticalSearch(this._verticalKey, { input: '' });
+    }
```
b1lly, line 108: I have a few questions:

1. Should this live in the SearchBar instead of in the results?
2. If not the SearchBar, it's starting to feel like there is "global settings" and behaviors we want to happen that are not "bound" to a component per say (e.g. Navigation Tabs, Searching on Load), etc.
3. Should we support a configurable `defaultQuery`, thats value is defaulted to ""? I can imagine a world where someone might want to explicitly show certain "subset" of results on the default page load.

b1lly (changes requested):
This looks good, but what about being able to provide the `defaultQuery` ?

src/ui/components/results/resultscomponent.js
Delerme: 1. Max & Rose suggested the results component, it's in a conversation on the JIRA story
2. They considered that too, but went with results
3. Max said empty query, but makes sense to me, will add just in case

PR #283, opened 2019-10-14
Highlighting

vramesh1:
List of changes :

1) Result Factory separation for Knowledge Manager
- previously, Knowledge Manager and Custom searchers fell into a Generic Result Factory function
- creating a separate Result Factory function to create Results for Knowledge Manager backends
- refactor to move KM Result creation and Field Formatting logic into this function
2) Highlighting
- KM backends return highlighting instructions when available to highlight entity field values
- apply this highlighting to substrings within entity field values
- adding a property to the Result class to contain highlighted fields
- in addition to the property that contains the raw (without highlighting) entity profile data 
3) Highlighting available to Field Formatters
- pass in the highlighted field value to field formatters functions
- developers can choose to use either the raw entity field value or the highlighted version
4) Field Formatter interface
- modified the interface for Field Formatters to take in an object containing named formatter params
- instead of a growingly long list of parameters
5) Scope of Highlighter member function
- to consolidate logic to manipulate strings for highlighting
- re-using the HighlightedValue class' buildHighlightedValue function
- this was previously a private member function
- making it public for use by ResultFactory as well
6) Automated Tests
- added test suite for highlighting entity profile data
- changed field formatting test suite to use the modified field formatter interface


src/core/models/highlightedvalue.js
```diff
@@ -25,7 +25,31 @@ export default class HighlightedValue {
   getInverted () {
     this._sortMatchedSubstrings();
     const invertedSubstrings = this._getInvertedSubstrings(this.matchedSubstrings, this.value.length);
-    return this._buildHighlightedValue(this.value, invertedSubstrings);
+    return this.buildHighlightedValue(this.value, invertedSubstrings);
+  }
+
+  buildHighlightedValue (val, highlightedSubstrings) {
```
b1lly, line 77: Can you add a comment on top of the function to describe what the structure of these values look like?

You can use JSDoc to help communicate that, and/or also give a example object.

src/core/models/resultfactory.js
```diff
@@ -45,23 +35,133 @@ export default class ResultFactory {
...
+  /**
+   * Applies field formatters to Knowledge Manager Entity Profile Data
+   *
+   * @param data Entity Profile Data
+   * @param formatters Developer specified Field Formatters
+   * @param verticalId Identifier for Vertical
+   * @param highlightedFields KM specified highlighting instructions to highlight certain Fields
+   * @returns {Object} Subset of Entity Profile Data Fields with field formatters applied
+   */
+  static getFormattedData (data, formatters, verticalId, highlightedFields) {
```
b1lly: Style nit: can we name `data` = `entityProfile` or `entityProfileData`?

src/core/models/resultfactory.js
```diff
@@ -45,23 +35,133 @@ export default class ResultFactory {
...
+    const formattedData = {};
+
+    if (Object.keys(formatters).length > 0) {
```
b1lly: style nit: to avoid all the indentations can we just short circuit?

```
if (Object.keys(formatters.length) === 0) {
  return formattedData;
}

Object.entries(data).forEach([....]) => {
  // This can if (formatters[fieldName] === undefined) { ... }
  // Or, you can explicitly ensure that it's a function (e.g. below)...
  // It just depends on how safe you want to be and how you want to handle exceptions
  if (typeof formatters[fieldName] !== 'function') {
    // This skips the current iteration
    return;
  }

  let highlightedFieldObject = ResultFactory.getHighlightedData(profileData, highlightedField);
  ...
}

return formattedData;
```

b1lly (commented):
Nice dude! Minor style nits.

Delerme:
this'll need a rebase on a current version of v0.9.0 before review

vramesh1:
@b1lly @Delerme , changes made per comments and ready for review when you get a chance. thanks.

src/core/models/highlightedvalue.js
```diff
@@ -25,7 +25,77 @@ export default class HighlightedValue {
...
+   *
+   *  example object :
+   *  {
+   *    name: 'ATM',
+   *    featuredMessage: {
+   *      description: 'Save time & bank on your terms at over 1,800 ATMs'
+   *    }
+   *  }
+   *
+   * @param highlightedSubstrings highlighting specifiers to apply to input object
```
b1lly: JSDoc is like this:

```
@param name {type} description
```

e.g.

```
@param highlightedSubstrings {object} highlighting specifies to apply to input object
@returns {string} The highlighted value
```

At some point, we use a type checking compiler, but in the short term, at least it gives the developers a little more context into the structure of the object.

Your example is amazing btw.

src/core/models/resultfactory.js
```diff
@@ -43,25 +33,140 @@ export default class ResultFactory {
...
+    // if no field formatters specified, nothing to format
+    if (Object.keys(formatters).length === 0) {
+      return {};
+    }
+
+    const formattedData = {};
+
+    Object.entries(entityProfileData).forEach(([fieldName, fieldVal]) => {
+      // check if a field formatter function exists for the current entity profile field
+      if (formatters[fieldName] === undefined || typeof formatters[fieldName] !== 'function') {
```
b1lly: This is almost like swallowing an exception (the typechecking case). Just wanted to make sure you were explicit about what behavior you wanted here.

For instance, if someone did:
```
fieldFormatters: {
  'name': "NERD"
};
```
In your case, it would just ignore it (which is fine). The alternative is to explicitly throw an AnswersError that tells them they did something bad.

I don't feel strongly either way, was just making sure the behavior was intended.

src/core/models/resultfactory.js
```diff
@@ -43,25 +33,140 @@ export default class ResultFactory {
...
+        return;
+      }
+
+      let highlightedFieldObject = ResultFactory.getHighlightedData(entityProfileData, highlightedFields);
```
b1lly: Style nit: If it's an annoying refactor, ignore this comment, but typically I tend to avoid `getName` methods unless it's an explicit read only operation. For instance, in this case, i'd probably name it "computeHighlightData" since the operation is actually doing work.

Reasoning: A lot of people fall into the trap of overloading the `get[Name]` practice, which can cause others to use the methods improperly.

For instance, i've seen a lot of cases where people implement a `getMethod` which turns out to be an `expensive` operation, and then other developers end up using it in a tight loop without realizing the operation was actually doing something heavy.

Because of that, I try to be as explicit as I can with my naming... to help others avoid making mistakes.

I know you didn't name the original, but wanted to give some context.

src/core/models/resultfactory.js
```diff
@@ -43,25 +33,140 @@ export default class ResultFactory {
...
+        highlightedFieldVal = highlightedFieldObject[fieldName];
+      }
+
+      let fieldFormatterInputParams = {
+        entityProfileData: entityProfileData,
+        entityFieldValue: fieldVal,
+        highlightedEntityFieldValue: highlightedFieldVal,
+        verticalId: verticalId,
+        isDirectAnswer: false
+      };
```
b1lly: Nit: You could probably just inline this into the consumer, instead of storing a reference.

```
formattedData[fieldName] = formatters[fieldName]({
  ...
});
```

src/core/models/resultfactory.js
```diff
@@ -43,25 +33,140 @@ export default class ResultFactory {
...
+
+  /**
+   * Applies highlighting to substrings within Knowledge Manager Entity Field Values
+   * according to highlighting specifiers returned from the Knowledge Manager Search Backend
+   *
+   * @param entityProfileData Entity Profile Data
+   * @param highlightedFields KM specified highlighting instructions to highlight certain Fields
+   * @returns {Object} Subset of Entity Profile Data Fields with highlighting applied
+   */
+  static getHighlightedData (entityProfileData, highlightedFields) {
```
b1lly: computeHighlightedData ?

b1lly (changes requested):
You're a savage! Mostly style nits.

Sorry I didn't catch these in the first iteration, but this looks pretty solid.

src/core/models/highlightedvalue.js
vramesh1: Sure. Adding Type.
Looks like it's actually 

> @param type name

instead of the other way around.
Not sure if that matters anyway for now.

src/core/models/resultfactory.js
vramesh1: sure, adding exception handling for a couple of cases across formatting and highlighting functions

src/core/models/highlightedvalue.js
b1lly: yeah, woops!

src/core/models/resultfactory.js
```diff
@@ -43,25 +34,144 @@ export default class ResultFactory {
...
    * @returns {Result}
    */
-  static fromGeneric (data, formattedData, index) {
+  static fromKnowledgeManager (data, formattedData, highlightedFields, index) {
     return new Result({
       raw: data,
       formatted: formattedData,
+      highlighted: ResultFactory.computeHighlightedData(data, highlightedFields),
       title: formattedData.name || data.name,
       details: formattedData.description || this.truncate(data.description),
```
b1lly: Woops, don't we want to use the highlighted data here by default, if it exists?

src/core/models/resultfactory.js
```diff
@@ -43,25 +34,144 @@ export default class ResultFactory {
         case 'ALGOLIA':
           results.push(ResultFactory.fromAlgoliaSearchEngine(data));
           break;
+        case 'KNOWLEDGE_MANAGER':
+          const highlightedFields = resultsData[i].highlightedFields || {};
+          const formattedData = ResultFactory.computeFormattedData(
+            data, formatters, verticalId, highlightedFields);
```
b1lly: what made you decide to process the formatting here, instead of within the factory? but you process the highlighted data within the factory?

Also, I think you're computing the highlighted data twice.

Perhaps you should move both computes into the factory, but do the highlighting first, and then pipe the results into the formatter.

src/core/models/resultfactory.js
```diff
@@ -43,25 +34,144 @@ export default class ResultFactory {
...
+    const formattedData = {};
+
+    Object.entries(entityProfileData).forEach(([fieldName, fieldVal]) => {
+      // check if a field formatter exists for the current entity profile field
+      if (formatters[fieldName] === undefined) {
+        return;
+      }
+      // verify the field formatter provided is a formatter function as expected
+      if (typeof formatters[fieldName] !== 'function') {
+        throw new AnswersCoreError('Field Formatter is not of expected type', 'ResultFactory');
```
b1lly: Maybe better: "Field formatter is not of expected type `function`"

Just a little more descriptive.

src/core/models/resultfactory.js
```diff
@@ -43,25 +34,144 @@ export default class ResultFactory {
...
+      }
+
+      let highlightedField = highlightedFields[highlightedFieldName];
+
+      // check for nested fields
+      if (typeof highlightedField === 'object' &&
+          Object.keys(highlightedField).length > 0 &&
+          highlightedField['matchedSubstrings'] === undefined) {
+        // recurse to children fields
+        highlightedData[highlightedFieldName] = ResultFactory.computeHighlightedData(
```
b1lly, line 128: See comment above.

b1lly (changes requested):

Delerme (commented):
Looks good, will let billy approve after the minor changes

b1lly:
I think we need to rebase this before we can merge it back in.

Should we rebase this onto the v0.9.2 branch or should this be part of 1.0.0.
@Delerme thoughts?

Delerme:
Up to you all, I haven't heard anything about 1.0.0 so am not sure what the plan is there

vramesh1:
Created a branch for v1.0.0 and merged v0.9.2 changes into it.
Updated the base branch of this PR to v1.0.0.

vramesh1:
created a separate PR to merge these changes into new base branch v1.0.0.
https://github.com/yext/answers/pull/325

vramesh1:
closing in favour of 
https://github.com/yext/answers/pull/325

PR #284, opened 2019-10-14
Temporary fix for passing immutable data to transfromData() when render

bwangYext:
As Julian suggested, the right solution is to go all the way to stat.js and make that immutable from start. But that will require a lot more changes. We probably need to bear with this fix for now.

TEST=auto,manual
npm run test


src/ui/components/component.js
```diff
@@ -372,7 +372,15 @@ export default class Component {
...
-    data = this.transformData(data);
+    // Temporary fix for passing immutable data to transformData().
+    // Excluding fields to avoid circular reference.
+    data = this.transformData(JSON.parse(JSON.stringify(data, (key, value) => {
+      let excludedFields = ['componentManager', 'core', 'renderer', 'analyticsReporter', 'parent'];
+      if (excludedFields.includes(key)) {
+        return undefined;
+      }
+      return value;
+    })));
```
b1lly: It might be time to separate "userConfig" and "systemConfig" into two separate objects that are managed separately in the Component. I don't think the "local state" of the component should ever need access to system config. 

Per our discussion offline, can we tweak the code to see if that approach isn't too difficult?

If it causes problems, we can wait until a larger refactor. But basically, the change would be to edit `ComponentManager` and the `Component` class, and separate the `systemConfiguration` (core, renderer, componentmanager, analytics, etc), from the `userConfig` (container, tabs, etc).

b1lly (changes requested):

src/ui/components/component.js
bwangYext: Extracting system config from config when systemConfig is not provided. Also passing parent container only instead of the whole parent object. 

Tried some components and it seems that things are not breaking but wondering if there is a more comprehensive way to test.

src/ui/components/component.js
```diff
@@ -16,14 +16,25 @@ import { AnswersComponentError } from '../../core/errors/errors';
...
+    if (Object.keys(systemConfig).length === 0) {
+      systemConfig = {
+        core: config.core,
+        componentManager: config.componentManager,
+        renderer: config.renderer,
+        analyticsReporter: config.analyticsReporter
+      };
+
+      this.removeSystemConfig(config);
+    }
```
b1lly: I'm glad this works! Can we do the refactor for all the components as well? You can do it as a separate commit here, but as part of the same PR.

I think we can also deprecate the type argument, and the type check on line 22.

b1lly (changes requested):

src/ui/components/component.js
b1lly: @bwangYext, have you tried @Delerme component tester app? Can you point him to it Julian?

bwangYext: Split into config and system config and changed components constructor. Testing with simulacrum.

b1lly (approved):
For the comment of this commit, can you ensure it details the reason for this change? I think it evolved a bit.

@benmcginnis as a heads up -- This might be a breaking change for some of the implementations you have, so I just wanted to make you aware.

This should fix the dataTransform issue you were having for direct answer, where the data transform was running on a mutated object instead of the original (after the first time)

src/ui/components/component.js
```diff
@@ -372,7 +365,8 @@ export default class Component {
    */
   render (data = this._state.get()) {
     this.beforeRender();
-    data = this.transformData(data);
+    // Temporary fix for passing immutable data to transformData().
+    data = this.transformData(JSON.parse(JSON.stringify(data)));
```
benmcginnis, line 369: @b1lly would this be the only spot you're worried about a breaking change, or was there something else?

b1lly: Sort of.

I don't think this would be a breaking change at all UNLESS the implementation of your dataTransform hook is dependent on two different models: the first iteration (e.g. original model), and the second iteration (e.g. the post transformed one).

This was the bug that was causing your data-transforms to wonk-out. That should no longer be the case. This change ensures that every time the `dataTransform` gets data, it is always the same.

PR #285, opened 2019-10-14
Pretty print answers errors, add debug option

Delerme:
When debug is true, ugly print errors


src/answers-umd.js
```diff
@@ -129,6 +129,20 @@ class Answers {
...
+        console.error(e.error.toString());
+        if (config.debug) {
+          console.log({ ...e.error });
+        }
+      }
+    };
+
+    // Print detailed error messages
+    window.addEventListener('error', printError);
+    window.addEventListener('unhandledrejection', printError);
```
b1lly: Can we move this to a helper? Or better yet, into the ErrorReporter implementation?

If the ErrorReporter is mainly to be a networking an abstraction, i'd suggest creating a separate ErrorReporting implementation that encapsulates this DOM integration and uses the ErrorReportingApi for sending data.

That way lines 132-153 are all encapsulated in the implementation itself.

b1lly (changes requested):

src/answers-umd.js
Delerme: done

src/answers-umd.js
```diff
@@ -129,11 +129,16 @@ class Answers {
...
+    // Report errors to console & server
+    this._errorReporter = new ErrorReporter({
+      apiKey: config.apiKey,
+      experienceKey: config.experienceKey,
+      experienceVersion: config.experienceVersion,
+      printVerbose: config.debug,
+      sendToServer: !config.suppressErrorReports
+    });
+    window.addEventListener('error', e => this._errorReporter.report(e.error));
+    window.addEventListener('unhandledrejection', e => this._errorReporter.report(e.error));
```
b1lly: I'd consider these DOM dependencies to be part of the "ErrorReporting" implementation. Thoughts?

b1lly (changes requested):
Mmm... that looks so much better!

src/answers-umd.js
Delerme: done

b1lly (approved):

PR #292, opened 2019-10-15
Load Google maps api js over https

Delerme:


src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -37,7 +37,7 @@ export default class GoogleMapProvider extends MapProvider {
         this._onLoaded();
       },
       async: true,
-      src: `http://maps.googleapis.com/maps/api/js?${this.generateCredentials()}`
+      src: `https://maps.googleapis.com/maps/api/js?${this.generateCredentials()}`
```
b1lly, line 40: If you remove the protocol completely, would that solve the problem?

e.g.

```
src: `//maps.googleapis.com/maps....`
```

b1lly (changes requested):
If you remove the protocol completely, would that solve the problem

src/ui/components/map/providers/googlemapprovider.js
Delerme: we really should be pointing to https, I don't think we want to continue using insecure resources (this doesn't prevent it from working in http contexts)

b1lly (approved):

PR #303, opened 2019-10-16
Fix nav in IE11

Delerme:


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -195,8 +195,13 @@ export default class NavigationComponent extends Component {
       : container.offsetWidth - moreButton.offsetWidth;
     let numBreakpoints = this._navBreakpoints.length;
 
-    if (mainLinks.offsetWidth > navWidth) {
-      this._navBreakpoints.push(mainLinks.offsetWidth);
+    let mainLinksWidth = 0;
+    for (let el of mainLinks.children) {
+      mainLinksWidth += el.offsetWidth;
+    }
```
b1lly, line 203: Do you want to add an IE11 comment to explain why this logic is here. That way some future developer doesn't  accidentally delete purposeful logic if they were to e.g. refactor?

b1lly (approved):

src/ui/components/navigation/navigationcomponent.js
Delerme: done

PR #308, opened 2019-10-17
Prevent duplicate searches with multiple searchbars

Delerme:
Inform components if they are "twins" i.e. there exists another another
component of the same type, so they may properly split responsibilities


b1lly (approved):
LGTM

PR #317, opened 2019-10-22
Mock services

data-enabler:
Was a bit confused by the exports in `core/index.js`; Should `answers-umd` only use exports from this file?


data-enabler:
Also, names aren't the best because I chose not to rename the existing files to avoid merge conflict nightmares

src/answers-umd.js
```diff
@@ -71,21 +78,58 @@ class Answers {
...
+        experienceKey: config.experienceKey,
+        experienceVersion: config.experienceVersion,
+        locale: locale
+      });
+
+    const questionAnswerService = config.mock
+      ? new MockQuestionAnswerService()
+      : new QuestionAnswerApi({
+        apiKey: config.apiKey
+      });
```
b1lly: This works, alternatively

const services = config.mock ? this.getMockServices() : this.getServices();


```
// example getServices()
// note, this can probably just be a literal unless we wanted to require certain services to be constructed
return Services({
  searchService: new SearchApi(),
  errorReporter: new ErrorReporter(),
  analyticsReporter: new AnalyticsReporter()
  ...
})

// example getMockServices()
return Services({
  searchService: new MockSearchService()
  errorReporter: new ConsoleErrorReporter(),
  analyticsReporter: NoopAnalyticsReporter()
  ...
})

```

And then we'd just reference the services object throughout the code instead of having the mock configuratino check sprinkled everywhere:

```
// usage
services.searchService;
```

src/answers-umd.js
```diff
@@ -71,21 +78,58 @@ class Answers {
...
     const globalStorage = new GlobalStorage();
     const persistentStorage = new PersistentStorage({
       updateListener: config.onStateChange,
       resetListener: data => globalStorage.setAll(data)
     });
     globalStorage.setAll(persistentStorage.getAll());
 
+    const searchService = config.mock
+      ? new MockSearchService()
+      : new SearchApi({
```
b1lly: I think you mentioned this in the comments, but how come we didn't update this to `SearchService` from the SearchAPI?

b1lly (changes requested):
Savage.

Two minor comments.

Otherwise, code organization discussion:

I was thinking we should organize the packages like so:

core/services/*[interfaces]
core/infra/*[implementations]

Services would be named e.g. SearchService

Implementation would be named: HttpSearchService, MockSearchService

Or something like that? Is there anything we can learn from the way we organize our java code that might be applicable here?

src/answers-umd.js
data-enabler: Well `SearchService` is the name of the interface, so it would get renamed to `ApiSearchService`. Yeah, like I mentioned before, I avoided renaming the classes in this commit because it made merge more likely and annoying to deal with and also messed with Git history for the classes where the interface takes the name that the implementation was using.

I figure the names can be renamed in a separate commit and it'll be less disruptive.

src/answers-umd.js
data-enabler: Good idea. Only problem is that AnalyticsReporter depends on Core, which depends on the rest of the services, so I've left it out for now.

Moved the service interfaces to `services/`, but held of on moving the other files for the same merge-conflict and git-history reasons I gave below.

src/answers-umd.js
```diff
@@ -83,21 +120,35 @@ class Answers {
...
       globalStorage: globalStorage,
       persistentStorage: persistentStorage,
       experienceKey: config.experienceKey,
       fieldFormatters: config.fieldFormatters,
       experienceVersion: config.experienceVersion,
-      locale: config.locale
+      locale: config.locale,
+      searchService: this._services.searchService,
+      autoCompleteService: this._services.autoCompleteService,
+      questionAnswerService: this._services.questionAnswerService
```
b1lly, line 151: is there an disadvantage to providing the core the 'services' object instead of explicitly listing out each of them?

b1lly (approved):

src/answers-umd.js
data-enabler: Not against doing that and was already considering it, but it would involve moving the `Services` type into its own file, and I don't think it's quite ready for that since there's things missing from it (AnalyticsReporterService).

PR #320, opened 2019-10-22
Add pagination component

Delerme:


Delerme:
Message me on slack if you want a demo link

src/ui/components/results/paginationcomponent.js
```diff
@@ -0,0 +1,117 @@
...
+    const nextPageButton = DOM.query(this._container, '.js-yxt-Pagination-next');
+    const firstPageButton = DOM.query(this._container, '.js-yxt-Pagination-first');
+    const lastPageButton = DOM.query(this._container, '.js-yxt-Pagination-last');
+    const maxPage = Math.trunc((results.resultsCount - 1) / limit);
+
+    DOM.on(previousPageButton, 'click', () => this.updatePage(offset - limit));
+    DOM.on(nextPageButton, 'click', () => this.updatePage(offset + limit));
+    DOM.on(firstPageButton, 'click', () => this.updatePage(0));
+    DOM.on(lastPageButton, 'click', () => this.updatePage(maxPage * limit));
+  }
```
b1lly, line 97: Is there a disadvantage to just using a list of links? The urls for the pagination can just be precomputed, and the page would reload and use it as part of the initial search.  I think that would eliminate a lot of the state that is being managed in the component as well.

src/core/models/searchconfig.js
```diff
@@ -0,0 +1,20 @@
+import { AnswersConfigError } from '../errors/errors';
+
+/** @module SearchConfig */
+
+export default class SearchConfig {
```
b1lly, line 5: I LIKE THIS!

src/core/models/searchconfig.js
```diff
@@ -0,0 +1,20 @@
...
+  constructor (config = {}) {
+    // The max results per search.
+    // Also defines the number of results per page, if pagination is enabled
+    this.limit = config.limit || 20;
```
b1lly, line 9: What do you think about this config supporting vertical key too?

src/ui/components/results/paginationcomponent.js
```diff
@@ -0,0 +1,117 @@
...
+export default class PaginationComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+
+    /**
+     * The vertical key to use for searches
+     * @type {string}
+     * @private
+     */
+    this._verticalKey = config.verticalKey;
```
b1lly: Could this be part of the search config?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -204,6 +204,8 @@ export default class FilterSearchComponent extends Component {
     const searchQuery = this.core.globalStorage.getState(StorageKeys.QUERY) || '';
     const facetFilter = this.core.globalStorage.getAll(StorageKeys.FACET_FILTER)[0];
 
+    this.core.persistentStorage.delete(StorageKeys.SEARCH_OFFSET);
+    this.core.globalStorage.delete(StorageKeys.SEARCH_OFFSET);
```
b1lly, line 208: This seems like an unfortunate, but necessary situation with the existing way we do things.

Do you have any ideas in regards to how this would work in e.g. the model where "search controller" didn't live inside the component?

b1lly (changes requested):

src/ui/components/results/paginationcomponent.js
Delerme: Ideally we wouldn't trigger a full refresh for each page. Our sample apps might be relatively tiny and have only the answers logic to init on load, but our customers' pages might have all sorts of logic that happen on every load, some of which may be super slow. We don't want to be the bad neighbor on our clients' pages that constantly reloads the entire page and creates a bad user experience

src/core/models/searchconfig.js
Delerme: done

src/ui/components/results/paginationcomponent.js
Delerme: Done

src/ui/components/search/filtersearchcomponent.js
Delerme: Yea the controller would deal with this stuff

src/ui/components/results/paginationcomponent.js
Delerme: As discussed, talked with product and they decided we should just submit a new search instead of refreshing the whole page

b1lly: This is fine as long as it works well for the IFRAME integrations. Has it been tested? Is there something we can do with the test app that would enable us to test the IFRAME integration cases better @data-enabler ?

b1lly (dismissed):
As long as its been tested with the IFRAME integrations it seems fine. I'm still not too warm about the single page experience for pagination, it doesn't seem to be a thing that a lot of the other search competitors do -- and for the added complexity it might not be warranted.

BUT, if that's what product wants *shrug*.

LGTM

Delerme:
Fixed merge conflicts

src/ui/components/results/paginationcomponent.js
data-enabler: Hmm, I guess it wouldn't be too hard to generate iframe'd versions for each page.

data-enabler (approved):

PR #322, opened 2019-10-23
Fix searchbar overflow

Delerme:
Gets rid of that weird growth under the searchbar that overflows its container


b1lly (approved):
I trust you :P

PR #325, opened 2019-10-23
Highlighting Entity Profile Field Values

vramesh1:
List of changes :

1) Result Factory separation for Knowledge Manager
- previously, Knowledge Manager and Custom searchers fell into a Generic Result Factory function
- creating a separate Result Factory function to create Results for Knowledge Manager backends
- refactor to move KM Result creation and Field Formatting logic into this function
2) Highlighting
- KM backends return highlighting instructions when available to highlight entity field values
- apply this highlighting to substrings within entity field values
- adding a property to the Result class to contain highlighted fields
- in addition to the property that contains the raw (without highlighting) entity profile data 
3) Highlighting available to Field Formatters
- pass in the highlighted field value to field formatters functions
- developers can choose to use either the raw entity field value or the highlighted version
4) Field Formatter interface
- modified the interface for Field Formatters to take in an object containing named formatter params
- instead of a growingly long list of parameters
5) Scope of Highlighter member function
- to consolidate logic to manipulate strings for highlighting
- re-using the HighlightedValue class' buildHighlightedValue function
- this was previously a private member function
- making it public for use by ResultFactory as well
6) Automated Tests
- added test suite for highlighting entity profile data
- changed field formatting test suite to use the modified field formatter interface


vramesh1:
potentially replace
https://github.com/yext/answers/pull/283

b1lly (approved):

PR #329, opened 2019-10-24
Add external interface for analytics events

Delerme:


Delerme:
this allows you to do something like `ANSWERS.AnalyticsReporter.report(new ANSWERS.AnalyticsEvent('CUSTOM'));`

b1lly (dismissed):

b1lly:
Can we update the documentation too?

Delerme:
Added

README.md
```diff
@@ -735,7 +735,13 @@ ANSWERS.addComponent('SpellCheck', {
...
 # Analytics
 
-Answers will track some basic interaction analytics automatically, such as search bar impressions and Call-To-Action clicks. You may add additional, custom analytic events to templates using certain data attributes, explained below.
+Answers will track some basic interaction analytics automatically, such as search bar impressions and Call-To-Action clicks. You may add additional, custom analytic events to templates using certain data attributes, explained below. You may also send analytics from external code with the below interface.
+
+```js
+  const event = new ANSWERS.AnalyticsEvent('CUSTOM');
+  event.addOptions({ myData: 'data' });
+  ANSWERS.AnalyticsReporter.report(event)
+```
```
benmcginnis, line 744: would it be possible for us to expose all the default event types too? So we don't have to worry about string typos, e.g. `new ANSWERS.AnalyticsEvent.ctaClick` ?

benmcginnis (approved):

README.md
Delerme: The library sadly doesn't know the default types, hence us often sending types that the server doesn't support

benmcginnis: I mean, sort of.  We have the spec, so I feel like it makes a lot of sense to formalize them as constants in the library and export them so implementers don't need to worry about getting the string exactly right.  It also seems like it will make it a lot easier to identify disconnects between the analytics API and the library.

Delerme: Yea ideally the api and library are powered off the same source of truth, otherwise the constants become a false sense of security. I think it's definitely something we should tackle, but just making a bunch of constants seems incomplete to me

PR #341, opened 2019-10-29
codelab: add component

apokellypse:


src/ui/components/search/timelinecomponent.js
```diff
@@ -0,0 +1,51 @@
...
+    this.timelineEvents = [];
+    this.core.globalStorage.on('update', StorageKeys.UNIVERSAL_RESULTS, r => {
+      console.log(r);
+      this.timelineEvents.push({
+        state: r.searchState,
+        timestamp: r.timestamp
+      });
+      this.setState();
+    });
+    this.moduleId = StorageKeys.TIMELINE;
```
Delerme, line 23: moduleId means the component state is automatically updated when this storage key changes, but doesn't look like anything is published to it, so the key and moduleId here are unecessary

src/core/models/universalresults.js
```diff
@@ -14,6 +14,7 @@ export default class UniversalResults {
      * @type {string}
      */
     this.searchState = data.searchState || SearchStates.SEARCH_COMPLETE;
+    this.timestamp = data.timestamp || new Date().toLocaleString;
```
Delerme, line 17: this seems to be more of a display concern, at least for the use case provided, so you would probably avoid making changes to core and instead record the timestamp in your UNIVERSAL_RESULTS update callback in the timeline component, recording the timestamp in your events

src/ui/components/search/timelinecomponent.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * TimelineComponent displays the timeline of searches and allows the user to clear it
+ * @extends Component
+ */
+export default class TimelineComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+    this.timelineEvents = [];
+    this.core.globalStorage.on('update', StorageKeys.UNIVERSAL_RESULTS, r => {
+      console.log(r);
```
Delerme, line 16: oof

Delerme (commented):
nice, looks good in general

Delerme:
Good stuff, closing for completion

PR #342, opened 2019-10-30
Facets

Delerme:
Message me on a slack for a demo link


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -14,79 +14,141 @@ const SUPPORTED_CONTROLS = [
...
+        config.previousOptions = [];
+      }
+    }
+    let selectedOptions = config.previousOptions || [];
+    this.options = this.options.map(o => ({
+      ...o,
+      selected: selectedOptions.length
+        ? selectedOptions.includes(o.label)
+        : o.selected
+    }));
```
b1lly, line 124: Nit: Might be able to push this logic to a helper.

this.options = this.computeOptions(config.previousOptions)

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -14,79 +14,141 @@ const SUPPORTED_CONTROLS = [
...
+    super(config, systemConfig);
+
+    let previousOptions = this.core.globalStorage.getState(this.name);
+    this.config = new FilterOptionsConfig({
+      previousOptions,
+      ...config
+    });
+
+    this.expanded = true;
+    this.allShown = false;
```
b1lly, line 179: jsdoc?

b1lly (changes requested):
Nice dude.

Can you clean up the ReadMe to support all the new configurations. And just strengthen the documentation there a little bit with some of the use cases?

src/ui/components/const.js
```diff
@@ -52,7 +52,7 @@ export const COMPONENT_MANAGER = new ComponentManager()
   .register(FilterOptionsComponent)
   .register(RangeFilterComponent)
   .register(DateRangeFilterComponent)
-  .register(DynamicFiltersComponent)
+  .register(FacetsComponent)
```
benmcginnis, line 56: So, just to confirm, this is going to be a breaking change because the component name is different now?

src/ui/sass/_fonts.scss
```diff
@@ -19,3 +19,4 @@ $line-height-lg: 1.5;
 $line-height-xlg: 5/3;
 $line-height-xxlg: 1.7;
 
+$font-family: system-ui,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif !default;
```
benmcginnis, line 22: seems like this is a conflict with #345

src/ui/sass/modules/_FilterBox.scss
```diff
@@ -0,0 +1,63 @@
...
+    @include Text(
+      $size: $font-size-md-lg,
+      $line-height: $line-height-lg,
+      $weight: $font-weight-semibold);
+
+    text-transform: uppercase;
+    margin-left: 13px;
+  }
+
+  &-filter:not(.yxt-FilterBox-filter--first) {
```
benmcginnis: is this the equivalent of doing `&-filter + &-filter`?

src/ui/sass/modules/_FilterBox.scss
```diff
@@ -0,0 +1,63 @@
...
+      $size: $font-size-md,
+      $weight: $font-weight-semibold,
+      $color: $color-brand-primary);
+
+    @include TextButton();
+
+    text-decoration: underline;
+
+    &:not(:disabled):hover {
+      text-decoration: none;
```
benmcginnis, line 58: would this make sense in the `TextButton` mixin?

src/ui/sass/modules/_FilterOptions.scss
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  &-showToggle {
+    @include Text(
+      $color: $color-brand-primary,
+      $style: italic);
+    @include TextButton($padding: 0);
+
+    svg {
```
benmcginnis, line 85: is this just meant to affect the icon component? if so I don't think you actually need to set this styling on the svg element.  If you set the font-size of the icon it should shrink down appropriately if it's square.  additionally i recommend adding a selector to the icon's parent and setting the left margin there.  I also recommend setting the margin with an `ex` unit as it will scale with the font-size without an implementer having to change it explicitly.

src/ui/sass/modules/_FilterOptions.scss
```diff
@@ -0,0 +1,145 @@
...
+  &-collapseUp, &-collapseDown {
+    vertical-align: middle;
+  }
+
+  &-collapseUp {
+    padding-top: 2px;
+  }
+
+  &-collapseUp svg {
+    transform: rotate(270deg);
```
benmcginnis, line 134: do you need the svg as part of this selector or does rotating the parent work?

src/ui/templates/filters/filterbox.hbs
```diff
@@ -1,13 +1,28 @@
-<div>
+<div class="yxt-FilterBox-container">
+  {{#if title}}
+    <div class="yxt-FilterBox-titleContainer">
```
benmcginnis: you should use an h level element for a title, such as an h2 or h3 depending on the overall document outline

src/ui/components/const.js
Delerme: Yea you can more of less consider this deleting the old facets and creating a new component. So any usages of the old facets are now broken

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/components/filters/filteroptionscomponent.js
Delerme: done

src/ui/sass/_fonts.scss
Delerme: yea kelly stole that from this review, it should merge fine

src/ui/sass/modules/_FilterBox.scss
Delerme: as a flag? This button works the opposite from normal text buttons (shows underline default, then hides on hover, whereas textbutton does the opposite)

src/ui/sass/modules/_FilterOptions.scss
Delerme: deleted

src/ui/sass/modules/_FilterOptions.scss
Delerme: has to be svg

src/ui/templates/filters/filterbox.hbs
Delerme: yea I'm worried we don't really know the document outline since these components are used on thirdparty sites, so we should remain h-agnostic, thoughts?

benmcginnis: Yeah, I think probably what we should do is add an optional "staring h level" or some such to the init config and have it cascade down from there.  The issue with not having them is that it's an accessibility problem.

Delerme: Yea that's an annoying situation. This should probably be discussed at a high level with product and ux and fix the headings across the library all at once, do you think we should wait for that before merging this?

src/ui/sass/modules/_FilterBox.scss
Delerme: yup, changed

src/ui/sass/modules/_FilterBox.scss
benmcginnis: Sorry I didn't realize it was special it seemed like it was part of the style

src/ui/templates/filters/filterbox.hbs
benmcginnis: I think we can maybe just pick a sensible h level here so we don't lose this line?

benmcginnis: 2 probably makes the most sense in isolation, as currently the searchbar sets an h1

benmcginnis:
Since this is a breaking change, should it be against the 0.10.0 version?

src/ui/sass/modules/_FilterOptions.scss
benmcginnis: Interesting and unexpected

src/ui/templates/filters/filterbox.hbs
Delerme: set to 2

Delerme:
Moved to 10

b1lly (approved):

data-enabler:
> Since this is a breaking change, should it be against the 0.10.0 version?

Speaking of which, this broke the sample app.

PR #343, opened 2019-10-31
Add LocationBias component

bwangYext:
Backend changes for this are still in CR but should be out soon.

TEST=auto,manual
Tested it locally with backend change.


src/ui/components/search/locationbiascomponent.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @extends Component
+ */
+export default class LocationBiasComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+
+    this.moduleId = StorageKeys.LOCATION_BIAS;
+
+    this._updateLocationEl = config.updateLocationEl || '.js-locationBias-update-location';
```
b1lly, line 35: can you add JS doc to each of these?

Also, can you update the documentation to include updateLocationEl?

src/ui/components/search/locationbiascomponent.js
```diff
@@ -0,0 +1,70 @@
...
+    DOM.on(this._updateLocationEl, 'click', (e) => {
+      if ('geolocation' in navigator) {
+        navigator.geolocation.getCurrentPosition((position) => {
+          this.core.globalStorage.set(StorageKeys.GEOLOCATION, {
+            lat: position.coords.latitude,
+            lng: position.coords.longitude,
+            radius: position.coords.accuracy
+          });
+        });
+      }
```
b1lly, line 75: @Delerme is this navigator logic shared somewhere else in code?

What do you think about creating an GeoLocation abstraction class that abstracts away the implementation details and is shared amongst the system?

src/ui/components/component.js
```diff
@@ -230,6 +230,7 @@ export default class Component {
         this.unMount();
         this.mount();
       } catch (e) {
+        console.log(e);
```
b1lly, line 233: woops

src/core/core.js
```diff
@@ -186,6 +187,7 @@ export default class Core {
         this.globalStorage.set(StorageKeys.SPELL_CHECK, data[StorageKeys.SPELL_CHECK]);
         this.globalStorage.delete('skipSpellCheck');
         this.globalStorage.delete('queryTrigger');
+        this.globalStorage.set(StorageKeys.LOCATION_BIAS, data[StorageKeys.LOCATION_BIAS]);
```
b1lly: style nit:

can we keep the sets together? or is there a reason it needs to happen after the deletes?

src/ui/components/search/searchcomponent.js
```diff
@@ -146,6 +146,9 @@ export default class SearchComponent extends Component {
     this._promptForLocation = config.promptForLocation === undefined
       ? true
       : Boolean(config.promptForLocation);
+    this.core.globalStorage.on('update', StorageKeys.GEOLOCATION, e => {
+      this.search(this.query);
+    });
```
b1lly: Is this a product requirement? Where whenever we remove geolocation we should be automatically be triggering searches?

b1lly (changes requested):
Savage! I'd like to be CC'd on the backend change please.

src/ui/sass/modules/_LocationBias.scss
```diff
@@ -0,0 +1,65 @@
...
+  &-bullet
+  {
+    height:  $locationBias-bullet-size;
+    width:  $locationBias-bullet-size;
+    display: inline-block;
+    border-radius: 100%;
+    background: $locationBias-bullet-color-default;
+  }
+
+  &-bullet-unknown-loc
```
benmcginnis: Similar to my comment [here](https://github.com/yext/answers/pull/350#discussion_r342734680), I'd really appreciate it if we could use single level class names following the [SUIT CSS naming convention](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md).

src/ui/components/search/locationbiascomponent.js
bwangYext: Done

src/ui/components/search/locationbiascomponent.js
bwangYext: Let me know if you want to do this now.

src/ui/components/component.js
bwangYext: Removed.

src/core/core.js
bwangYext: Done.

src/ui/components/search/searchcomponent.js
bwangYext: Calling search in component.

src/ui/sass/modules/_LocationBias.scss
bwangYext: Renamed.

src/ui/sass/modules/_LocationBias.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  &-locSource
+  {
+    @media (max-width: $breakpoint-mobile-max)
+    {
+      display: block;
+    }
+  }
+
+  a
```
benmcginnis: you should have a class selector on this element, rather than using the `a` selector

src/ui/sass/modules/_LocationBias.scss
benmcginnis: Thank you!!

src/ui/sass/modules/_LocationBias.scss
bwangYext: Done

bwangYext:
Poke

src/ui/components/search/locationbiascomponent.js
b1lly: I would have preferred to do it now, but because of my absence of this review -- seems like this item might be expected to be done already.

I think it's ok to push the refactor for this area to the next version, and merge this as is.

b1lly (dismissed):

b1lly (approved):

PR #345, opened 2019-10-31
directanswers: style string text return option

apokellypse:


apokellypse:
Forgot to Select the correct version branch. Closing

vramesh1:
the latest changes on v0.9.3 will need to be pulled into your branch.
could you rebase your branch on top of v0.9.3?

Delerme (dismissed):
You can use the 'update branch' button below to merge in v0.9.3 changes, since we use the 'squash & merge' option.

Also for visual changes send a screenshot or simulacrum demo via slack

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -59,6 +73,10 @@
 
   &-content
   {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 60px;
```
creotutar: nit: do we want this to relative to `$base-spacing-sm`

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -74,6 +92,7 @@
   {
     display: flex;
     justify-content: flex-end;
+    height: 24px;
```
creotutar: nit: do we want this to relative to `$base-spacing-sm`

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -84,6 +103,7 @@
 
   &-footerText
   {
+    margin-right: 16px;
```
creotutar: nit: do we want this to relative to `$base-spacing`

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -162,4 +183,10 @@
     display: inline-flex;
     align-items: center;
   }
+
+  &-viewMoreIcon
+  {
+    $line-height: $direct-answer-view-details-line-height;
+    margin-left: 4px;
```
creotutar, line 190: nit: do we want this to relative to `$base-spacing`

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -15,20 +15,21 @@
...
-  $direct-answer-footer-font-size: 13.5px !default;
+  $direct-answer-footer-font-size: $font-size-md !default;
   $direct-answer-footer-line-height: $line-height-md !default;
   $direct-answer-footer-font-weight: $font-weight-normal !default;
-  $direct-answer-footer-color: $color-text-primary !default;
+  $direct-answer-footer-color: $color-text-secondary !default;
 
   &-title
   {
+    height: 26px;
```
creotutar: we usually avoid giving explicit heights to things that arent images b/c some divs can expand to two lines of text. is this the case here? maybe we want to use padding to achieve this?

src/ui/sass/modules/_DirectAnswer.scss
creotutar: same height comment here, are we guaranteed the content will never exceed 60px in the case of wrapping text / number of lines for a direct answer?

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: the footer height seems independent of horizontal spacing, so I don't think so in this case. will add as base variable $module-footer-height so it is more visible.

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: good point, this won't handle wrapping. removing for now

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: Yes, we do. Will fix

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: don't think so; we want icons to be close to their action text regardless of what may eventually override base-spacing

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: we can definitely use padding or margin to achieve this!

creotutar (dismissed):
🎺 🎺 🎺 LGTM

src/ui/sass/_fonts.scss
```diff
@@ -19,3 +19,4 @@ $line-height-lg: 1.5;
 $line-height-xlg: 5/3;
 $line-height-xxlg: 1.7;
 
+$font-family: system-ui,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif !default;
```
benmcginnis, line 22: ❤️

src/ui/sass/modules/_DirectAnswer.scss
benmcginnis: I would recommend using the `ex` measurement as that is close to a single space of the font in question.  Usually `0.8ex` is pretty close to two spaces for any given font

benmcginnis (approved):

src/ui/sass/_fonts.scss
apokellypse: props to Julian for giving me the font families and their orders!

src/ui/sass/modules/_DirectAnswer.scss
apokellypse: what do you think about using `em`? I looked into it and saw that `ex` is relative to the height of the font, while `em` is relative to the width

PR #347, opened 2019-11-02
Allow specifying default initial search queries for vertical searchers

data-enabler:


src/ui/components/search/searchcomponent.js
```diff
@@ -167,7 +167,7 @@ export default class SearchComponent extends Component {
   }
 
   onCreate () {
-    if (this.query && !this.redirectUrl) {
+    if (this.query != null && !this.redirectUrl) {
```
b1lly, line 170: @Delerme Can you confirm this doesn't add any wonky side effects? I can't remember the values this was supposed to avoid.

b1lly (dismissed):
Otherwise LGTM

src/ui/components/search/searchcomponent.js
Delerme: Empty queries, but probably fine

data-enabler: If an empty query is provided via url the query in storage still ends up being null, so I don't think this should change existing behavior around that. Noticed an issue while doing some extra testing, which should be fixed now.

vramesh1 (approved):

PR #349, opened 2019-11-04
cookies: session tracking opt in feature

apokellypse:
I set it to true by default. I think we want some way for the client to customize it further, but it didn't occur to me how we wanted to do this, so decided to put up a patch for now.


apokellypse:
Sorry, I investigated and realized I misunderstood the prompt. Closing this for now.

apokellypse:
After some discussion, It turns out that this approach is on the right track, and the client should be able to set the Session opt in, similar to how conversion tracking is done. Reopening.

data-enabler:
Looks like my commits are still being included in this? You can always just rebase locally and `push -f` if you need to.

apokellypse:
Got rid of the commit sandwich! Ended up using cherry-picking after interactive rebase didn't handle my merge commit and initial commit well.

src/answers-umd.js
```diff
@@ -140,6 +140,7 @@ class Answers {
     });
     globalStorage.setAll(persistentStorage.getAll());
     globalStorage.set(StorageKeys.SEARCH_CONFIG, config.search);
+    globalStorage.set(StorageKeys.SESSIONS_OPT_IN, config.sessionTrackingEnabled);
```
data-enabler: If we're adding a config option, we should add it to the table in the readme.

src/core/services/searchservice.js
```diff
@@ -18,6 +18,7 @@ export default class SearchService {
    * @param {boolean} query.isDynamicFiltersEnabled If true, asks the server to return dynamic filters
    * @param {string} query.skipSpellCheck The boolean as string to indicate if it should skip spell checking
    * @param {string} query.queryTrigger The source that triggers query such as suggest
+   * @param {string} query.sessionTrackingEnabled The boolean as string to indicate if session tracking is on
```
data-enabler: Seems weird to require users of the interface to deal with stringifying the boolean, and we already have a boolean param in `isDynamicFiltersEnabled`.

apokellypse: good point, will make this a boolean

src/answers-umd.js
apokellypse: will do

src/answers-umd.js
```diff
@@ -140,6 +140,8 @@ class Answers {
     });
     globalStorage.setAll(persistentStorage.getAll());
     globalStorage.set(StorageKeys.SEARCH_CONFIG, config.search);
+    let sessionTrackingEnabled = config.sessionTrackingEnabled || true;
```
data-enabler: Doesn't this always evaluate to true?

data-enabler (changes requested):

src/answers-umd.js
apokellypse: Yes, will fix

data-enabler (approved):

PR #350, opened 2019-11-04
Q&A component styling changes

vramesh1:
Styling changes to the Q&A Component.

Screenshot with desktop layout : 

QA_component_desktop

Screenshot with mobile layout :
QA_component_mobile


src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -0,0 +1,178 @@
...
+    {
+      display: flex;
+      align-items: center;
+    }
+
+    &-right
+    {
+      display: flex;
+      align-items: center;
+    }
```
Delerme: can combine under a single rule
`&-left, &-right`

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -0,0 +1,178 @@
...
+      line-height: $question-submission-title-bar-text-line-height;
+      font-weight: $font-weight-semibold;
+      font-size: $font-size-md-lg;
+      text-transform: uppercase;
+    }
+
+    &-description
+    {
+      text-align: right;
+      float: left;
```
Delerme: avoid if possible, switch to flex

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -0,0 +1,178 @@
...
+
+          textarea
+          {
+            width:100%;
+          }
+
+          input
+          {
+            width:100%;
+          }
```
Delerme: can combine
`textarea, input {`

src/ui/sass/modules/_QuestionSubmission.scss
vramesh1: removing the text-align and float property.
not needed.
replacing with a justify-content property on the parent to take care of this.

Delerme (commented):
looks good, have you sent to product/ux to QA? I usually find it easier to let them look during PR instead of having to revisit later. I can help you set up a demo to send them in the qa app if you want

vramesh1:
sounds good, will set up a demo for Product/UX

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -0,0 +1,166 @@
...
+      margin-bottom: $base-spacing / 2;
+      margin-left: $base-spacing;
+      margin-right: $base-spacing;
+
+      @media (max-width: $breakpoint-mobile-max)
+      {
+        flex-direction: column;
+      }
+
+      &-field
```
benmcginnis: I know we don't have an agreed upon naming convention, but I'd really really appreciate it if we could use this one:
https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

so you'd do `.yxt-QuestionSubmission-row` and `.yxt-QuestionSubmission-field` rather than `.yxt-QuestionSubmission-form-row` and `yxt-QuestionSubmission-form-row-field`.  The benefit being that the class structure isn't strictly tied to the DOM hierarchy

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -1,59 +1,126 @@
...
+                      placeholder="{{_config.questionInputPlaceholder}}"
+                      required>{{questionText}}</textarea>
+          </div>
+        </div>
+      </div>
+
+      <div class="yxt-QuestionSubmission-form-row">
+        <div class="yxt-QuestionSubmission-form-row-field">
+          <div class="yxt-QuestionSubmission-form-row-field-label">
+            <label for="name">
```
benmcginnis: for doesn't work unless you use an id on the input.  given that the library could allow more than one of these forms on the page you probably want to wrap the input with the label to avoid having to use an id

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -1,59 +1,126 @@
...
-            {{#each errors}}
-              <li>{{this}}</li>
-            {{/each}}
-           </ul>
-        </div>
+
+<section class="yxt-QuestionSubmission">
+
+  <div class="yxt-QuestionSubmission-titleBar">
+    <div class="yxt-QuestionSubmission-titleBar-left">
```
benmcginnis: same thing about the double `-` nesting

vramesh1:
sent a demo link to Product.
will address feedback from Ben in the meantime while the demo is being tested.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -0,0 +1,169 @@
...
+    font-weight: bold;
+    margin-bottom: $base-spacing / 4;
+  }
+
+  &-formInput
+  {
+    display: flex;
+    width:100%;
+
+    ::placeholder
```
benmcginnis, line 121: Do y'all have autoprefixer setup in gulp? If not you may need to include some vendor prefixes for this: https://stackoverflow.com/questions/22199047/placeholder-css-not-being-applied-in-ie-11

vramesh1: not aware of having that set up at the moment.
will add the vendor prefixes.
thanks.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -1,59 +1,121 @@
...
+              name="name"
+              placeholder="{{_config.requiredInputPlaceholder}}"
+              value="{{name}}"
+              required>
+          </label>
+        </div>
+
+        <div class="yxt-QuestionSubmission-formField">
+          <label>
+            <div class="yxt-QuestionSubmission-formLabel">
```
benmcginnis: I think this should be a `span`: https://stackoverflow.com/questions/18609554/is-div-inside-label-block-correct . for context, HTML validation is a wcag compliance thing: https://www.w3.org/TR/WCAG21/#parsing

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -167,18 +173,27 @@
...
-      height: 40px;
-      box-sizing: border-box;
-      width: 170px;
-      color: $question-submission-cta-foreground-color;
-      border: none;
-      padding-top: $base-spacing / 2;
-      padding-bottom: $base-spacing / 2;
-      padding-left: $base-spacing;
-      padding-right: $base-spacing;
+      button
```
benmcginnis: do you need the element selector?

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -108,9 +108,9 @@
             </div>
           </div>
 
-          <button type="submit"
-            class="yxt-QuestionSubmission-submitButton"
-            data-eventtype="QUESTION_SUBMIT">{{_config.buttonLabel}}</button>
+          <div class="yxt-QuestionSubmission-submitButton">
+            <button type="submit" data-eventtype="QUESTION_SUBMIT">{{_config.buttonLabel}}</button>
```
benmcginnis: I don't think you need to have a div wrapper if you're just using it for layout

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -158,6 +159,11 @@
   &-ackCheckboxInput
   {
     margin-right: $base-spacing / 2;
+
+    input:focus
```
benmcginnis: can you simply add a class to the actual input rather than using the element selector?

src/ui/templates/questions/questionsubmission.hbs
vramesh1: added the div wrapper and element selector to override the following in _base.scss :
`
*:focus
{
  outline: none;
}`
should we perhaps remove this from base.css instead and leave it to the components?
or modify the focus behavior in base so that it is consistent in all components?
need to add focus behavior for checkboxes and buttons for WCAG guidelines.

vramesh1: probably best to add type specific focus stylings at the base stylesheet level.
input[type='checkbox']:focus
so that other components get this standardized behavior.

benmcginnis: the base level is probably best, yeah.  the outline none is there because people were really unhappy with the browser default outline styling on buttons, links, and the searchbar input, all of which have custom focus states

vramesh1: gotcha, ok.
will do something like this in base then :
```scss
*:focus
{
  outline: none;
}

button:focus
{
  outline: ... ;
}

input[type='checkbox']:focus
{
  outline: ... ;
}

File: src/ui/sass/modules/_QuestionSubmission.scss (hunk 11 above)
Comment (vramesh1) [reply src/ui/sass/modules/_QuestionSubmission.scss]: will do, related to the prev comment

File: src/ui/templates/questions/questionsubmission.hbs (hunk 10 above)
Comment (benmcginnis) [reply src/ui/templates/questions/questionsubmission.hbs]: yeah that would work, another alternative would be:
``` scss
a[class],
button[class]
{
  &:focus: {
    outline: none;
  }
}
// separately in SearchBar.scss
.yxt-SearchBar-input:focus { outline: none; }
```

this would ensure all buttons and links with their own classes applied would not have the standard outline and that only the searchbar would have it's outline removed thus ensuring that any links that are inline in the content and all inputs on any other form would have the outline

vramesh1: going with the former for now since it seems simpler and more maintainable, but can change that if you have a strong opinion on this.
the SearchBar component has some outline related styling already, so will definitely test alongside that component as well.

src/ui/sass/modules/_QuestionSubmission.scss
vramesh1: done. resolving.

src/ui/templates/questions/questionsubmission.hbs
vramesh1: made the change. resolving.

src/ui/sass/modules/_QuestionSubmission.scss
vramesh1: made the change, resolving.

src/ui/templates/questions/questionsubmission.hbs
vramesh1: done

vramesh1:
my bad. it looks like the auto-prefixer is hooked up already.
verified in the generated dist/answers.css file that vendor prefixes are being added.
will remove the vendor prefixes from my source sass file.

src/ui/sass/modules/_QuestionSubmission.scss
vramesh1: discussed and resolved in a later comment.
resolving this one.

benmcginnis (approved):

PR #356, opened 2019-11-06
Version 0.9.3

vramesh1:
# Version 0.9.3

## Changes

* Conversion Tracking Opt-In
  * Major change in behavior
  * Conversion Tracking was previously enabled by default
  * With this version, Conversion Tracking will no longer be enabled by default
  * To enable Conversion Tracking, the following line of code will need to be added to the Application
  * `ANSWERS.setConversionsOptIn(true)`
  
* New Pagination Component
* Vertical Search
  * Allow specifying a default initial search query for vertical searchers
  * Introduce boolean to toggle allowing an empty search
* Sample App enhancements


data-enabler:
> Allow specification of default initial search queries on empty vertical searches

This kinda makes it sound like this something that will happen for every empty search. I would just say "Allow specifying a default initial search query for vertical searchers"

data-enabler (approved):

PR #357, opened 2019-11-08
Generator

b1lly:
There's definitely a few more iterations to organizing this a little better.

We need to:

1. Organize the generator code into its own package and files
2. Have the SDK be a dependency thats pulled in to the app and speed up the compilation of this step
3. Define the TOP LEVEL configuration object for the siteConfig more concretely instead of just depending on an object literal. For instance, we should have a concrete model defined in the library of all the configuration options we allow.


sample-app/README.md
```diff
@@ -22,4 +22,7 @@ or
 ```
 npm start [config_file]
 ```
-Open `index.html` in your browser of choice
+
+Once done, all the assets for the generated site will live in the `generated` directory by default.
+
+You can `cd` into that directory and open `index.html` in your browser of choice
```
data-enabler: Probably simpler to just say that they should open `generated/index.html`

sample-app/config/config.json
```diff
@@ -7,4 +8,4 @@
     },
     "jobs": {}
   }
-}
+}
```
data-enabler: Missing newline

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
-const fs = require('fs');
+// TODO(billy) Move to ES6 imports when Node supports ES6 modules
```
data-enabler: Node does support ES6 modules, so I'd just remove the latter part of this comment.

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
+const fs = require('fs-extra');
 const path = require('path');
 const util = require('util');
 const writeFile = util.promisify(fs.writeFile);
```
data-enabler, line 4: Side note: apparently as of Node 10 there's promise-based versions of the fs methods
https://nodejs.org/api/fs.html#fs_fs_promises_api

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
-  writeUniversalHtml();
+  // Optionally create our `output directory`
+  if (!fs.existsSync(config.outputDir)) {
+    fs.mkdirSync(config.outputDir);
+  }
+
+  // Copy our assets into the `output directory`
+  fs.copy(config.assetsDir, `${config.outputDir}/${config.assetsDir}`, function (err) {
+    if (err) {
+      return console.error(err);
```
data-enabler: Promise APIs are nice for simple error handling like this, since you can say:
```js
fs.copy(config.assetsDir, `${config.outputDir}/${config.assetsDir}`)
  .catch(console.error)
  .then(() => console.log('Copied assets to generated site!'));
```

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
-  templateUrl
-});
+
+class Configuration {
+  constructor (credentialsFile, configurationFile) {
+    /**
+     * The directory of the output from the generation
+     * @type {string}
+     */
+    this.outputDir = './generated';
```
data-enabler, line 13: It occurs to me that specifying paths like this makes the script not compatible with Windows. We should start using the `path` api at some point.

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
-const credentials = require('./credentials.json');
-const configFile = process.argv[2]
-  ? path.join(process.cwd(), process.argv[2])
-  : './sample-config.json';
-const siteConfig = require(configFile);
-const config = Object.assign({}, credentials, siteConfig, {
-  templateUrl
-});
+
+class Configuration {
```
data-enabler, line 7: If we're going to bother writing JSDoc, I think this could just be a `typedef` + a `DEFAULTS` object.

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
+     */
+    this.sdkBundleDir = '../dist';
+
+    /**
+     * Configuration used for generation of each page and vertical
+     * @type {object}
+     */
+    const configFile = process.argv[2] ? path.join(process.cwd(), process.argv[2]) : configurationFile;
+    this.siteConfig = Object.assign({},
+      require('./config/credentials.json'),
```
data-enabler, line 39: If we're cleaning stuff up, it probably makes sense to just parse the file normally, since this is kind of a hack.

sample-app/index.js
```diff
@@ -1,63 +1,127 @@
...
+      require(configFile));
+  }
+}
+
+// TODO(billy) This should just be pulled in as a dependency to the application
+class SDKBundle {
+  constructor (config) {
+    this.bundleUrl = `${config.assetsDir}/answers-modern.js`;
+    this.templateUrl = `${config.assetsDir}/answerstemplates.compiled.min.js`;
```
data-enabler, line 48: This isn't currently being used, is it? Before, this was being added to the config so that `app.js` could pass it to `ANSWERS.init`.

data-enabler (changes requested):

sample-app/index.js
b1lly: Isn't it only supported under an experimental flag?

vramesh1: https://timonweb.com/posts/how-to-enable-es6-imports-in-nodejs/

sample-app/config/config.json
data-enabler: Actually, this file shouldn't have even been in the repo at all; I think I just accidentally committed it at some point.

sample-app/index.js
data-enabler: @b1lly Yeah, but it would just be a matter of changing the buildHtml script to use the flag.
@vramesh1 That article is out of date, at the end it links to a more recent version that uses the `esm` module (which is what I personally use to handle ES6 imports.

Point is, there isn't really anything stopping us from using ES6 imports right now, so I don't see a reason to phrase it like we're waiting on outside stuff. (Also, if we were to use TypeScript we'd get ES6 module compatibility for free 😉)

sample-app/index.js
b1lly: You mean with fetch? or with the fs module?

sample-app/index.js
b1lly: This is a temporary object. I think it actually should be pulled in as a dependency into the library via a package manager.

sample-app/index.js
b1lly: Agreed. Also, we'll have to revisit some of these values and variables because some are "src" of data and some are "dest" of the directory to put it in.

To keep things simple it just uses the same dest (e.g. assets) as it does from source.

sample-app/index.js
data-enabler: With `fs`, rather than using `require`

sample-app/README.md
```diff
@@ -25,4 +25,4 @@ npm start [config_file]
 
 Once done, all the assets for the generated site will live in the `generated` directory by default.
 
-You can `cd` into that directory and open `index.html` in your browser of choice
+You can `cd` into that directory and open `generated/index.html` in your browser of choice
```
data-enabler, line 28: They don't need to `cd` if they're using the `generated/index.html` path.

sample-app/index.js
data-enabler: To be clear, I'm talking about `templateUrl` specifically. The sample app generator currently sets the `templateUrl` property passed to `ANSWERS.init` to point at the locally compiled templates, but it looks like we're no longer doing that.

sample-app/index.js
```diff
@@ -1,63 +1,120 @@
...
+  // Copy our assets into the `output directory`
+  fs.copy(config.assetsDir, `${config.outputDir}/${config.assetsDir}`)
+    .catch(console.error)
+    .then(() => console.log('Copied assets to generated site!'));
 
-  for (const verticalKey of Object.keys(config.verticals)) {
+  // Copy our sdk bundle into the `output directory`
+  fs.copy(config.sdkBundleDir, `${config.outputDir}/${config.assetsDir}`)
+    .catch(console.error)
+    .then(() => console.log('Copied SDK bundles to generated site!'))
```
data-enabler, line 85: Missing semicolon

data-enabler (changes requested):

vramesh1:
Closing for now since I'm not sure if this will work will continue.
Pls re-open if needed.

PR #359, opened 2019-11-11
mapproviders: change default zoom from 9 to 14 and consolidate logic …

apokellypse:
…between different providers


src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -58,7 +58,7 @@ export default class MapBoxMapProvider extends MapProvider {
       center: this.getCenterMarker(mapData)
     });
 
-    if (mapData) {
+    if (mapData && mapData.mapMarkers.length) {
```
vramesh1, line 61: should we check for the existence of the mapMarkers property too?

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -89,12 +93,18 @@ export default class GoogleMapProvider extends MapProvider {
           bounds.extend(marker.position);
         }
 
-        this.map.fitBounds(bounds);
-      } else {
-        this.map.setCenter(new google.maps.LatLng(this._defaultPosition.lat, this._defaultPosition.lng));
+        if (googleMapMarkerConfigs.length >= 2) {
```
vramesh1, line 96: is this to ensure we display a map only if we have more than one location/pin?

apokellypse: this is to ensure that we are zoomed out to include all location pins if this result contains more than one location config. with one location, we use the default zoom and center in on it.

src/ui/components/map/providers/mapboxmapprovider.js
apokellypse: due to the logic in section.js we can safely assume that if mapData exists it will have a mapMarkers property

vramesh1 (approved):

PR #363, opened 2019-11-15
questionsubmission: add show/hide trigger and interaction

apokellypse:
Please add ?w=1 param to view this without the whitespace changes.

More context here:
https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing


apokellypse:
Here's a link to see the diffs without whitespace changes:

https://github.com/yext/answers/pull/363/files?utf8=%E2%9C%93&diff=unified&w=1

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,98 +18,108 @@
       <div class="yxt-QuestionSubmission-titleDescription">
         {{_config.sectionDescription}}
       </div>
+      <div class="{{#if questionExpanded}}yxt-QuestionSubmission-expandWrapper{{else}}yxt-QuestionSubmission-collapseWrapper{{/if}} js-content-visibility-toggle"
```
data-enabler: I think if we're actually doing BEM this should be yxt-QuestionSubmission__toggle--expanded, though a quick grep shows that there's no `__` in the codebase. But we should at least make these suffixes `--expanded` and `--collapsed`.

data-enabler: This isn't keyboard accessible; you should wrap it in a `button`. I think there should be an example in `resultsaccordion.hbs`

data-enabler (changes requested):

src/ui/templates/questions/questionsubmission.hbs
apokellypse: fixed in latest commit

apokellypse: fixed in latest commit

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -164,6 +164,12 @@ export default class QuestionSubmissionComponent extends Component {
   }
 
   onMount () {
+    let triggerEl = DOM.query(this._container, '.js-content-visibility-toggle');
+    if (triggerEl === null) {
+      return;
```
data-enabler, line 180: It would be nice if  not having a toggle element didn't prevent the component from working. I think that would make it possible to use the new code with an old template without things breaking.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,98 +18,109 @@
...
+          <div class="yxt-QuestionSubmission-acknowledgementBar">
+            <div class="yxt-QuestionSubmission-acknowledgement">
+              <input type="checkbox"
+                    name="privacyPolicy"
+                    class="yxt-QuestionSubmission-ackCheckboxInput js-question-field"
+                    value="true"
+                    required>
+              <div class="yxt-QuestionSubmission-privacyPolicyText">
+                {{_config.privacyPolicyLabel}}
+                <a class="yxt-QuestionSubmission-privacyPolicyUrl" href="{{_config.privacyPolicyUrl}}">{{_config.privacyPolicyUrlLabel}}</a>
```
data-enabler, line 114: This link should open in a separate tab/window.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -33,29 +65,53 @@
...
+    border: none;
+
+    svg {
+      height: 1.25em;
+      width: 1.25em;
+    }
   }
 
-  &-title
+  &--expanded
```
data-enabler, line 82: `yxt-QuestionSubmission--expanded` doesn't really make sense as the selector when it's being applied specifically to the toggle element. Let's make this `yxt-QuestionSubmission__toggle--expanded`.

@b1lly Are we actually using BEM notation for the sdk, or did we cook up our own thing?

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -33,29 +65,53 @@
...
   {
-    $question-submission-title-bar-text-color: $color-text-primary !default;
-    $question-submission-title-bar-text-line-height: $line-height-lg !default;
+    transform: rotate(-90deg);
+    cursor: pointer;
```
data-enabler, line 85: I think this shouldn't be necessary anymore since it's inside a button

data-enabler (changes requested):

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,98 +18,109 @@
       <div class="yxt-QuestionSubmission-titleDescription">
         {{_config.sectionDescription}}
       </div>
+      <button class="{{#if questionExpanded}}yxt-QuestionSubmission--expanded{{else}}yxt-QuestionSubmission--collapsed{{/if}} js-content-visibility-toggle"
```
data-enabler, line 21: Another thing: screen readers can't make sense of a toggle button like this unless you use the `aria-expanded` and `aria-controls` attributes. You can see an example in `resultsaccordion.hbs`. In the ANS-325 description they suggested that we should reuse the implementation between that component and this one, but I'm not sure how easy that would actually be with our current component architecture.

It's also probably worthwhile to make the whole header the toggle rather than the icon.

apokellypse:
will apply these comments to the new branch; right now separating these out into separate reviews

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: fixed in patch 2 of new pull request

src/ui/templates/questions/questionsubmission.hbs
apokellypse: fixed in patch 2 of new pull request

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: it's still necessary, unfortunately; otherwise it's a default pointer

src/ui/templates/questions/questionsubmission.hbs
apokellypse: fixed in patch 2 of new pull request

apokellypse: header is not toggle-able yet so I'll wait to resolve

src/ui/sass/modules/_QuestionSubmission.scss
vramesh1: @data-enabler seems like we've been doing our own thing for now.
to avoid mixing notations within the same stylesheet, should we keep consistent with the current methodology for the time being, while we document/communicate team-wide the methodology to implement in new components going forward?
so in this case, keeping with current methodology, it would probably be something closer to `yxt-QuestionSubmission-toggle-expanded` or `yxt-QuestionSubmission-expanded`?

apokellypse: let's continue this conversation in https://github.com/yext/answers/pull/389, am closing this review because it is no longer needed after it was split out into a functional PR and a styling PR

PR #380, opened 2019-11-19
Do not show update location link after denying location permission

bwangYext:
Using Permissions API. Talked to Rose about browswer compatibility and she said that it was fine for browsers which do not support this showing the link when refreshes.

TEST=auto,manual


Delerme (approved):
going forward, we probably want to move all logic in onMount to functions that onmount optionally calls to avoid weird scenarios like someone wanting to add more onMount behavior below this but line 53 would return early. Anyway not important here, just something to think about

PR #386, opened 2019-11-20
qacomponent: functional changes for the following:

apokellypse:
- add show/hide trigger and interaction
- add link to privacy policy
- new error handling


src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -251,17 +281,26 @@ export default class QuestionSubmissionComponent extends Component {
    * @returns {Object|boolean} errors object if any errors found
    */
   validateRequired (formData) {
+    var chromiumEmailRegex = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
data-enabler: Browsers have built-in validation for `input type="email"`; we should be able to use this by calling [`checkValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/checkValidity), even though we're using `novalidate` on the form.

We should be able to do this for the other required fields as well, which is nice since it means we don't have to duplicate validation requirements in the html and js (having it in the html is desirable for screen reader accessibility).

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -58,7 +53,7 @@
               <span class="yxt-QuestionSubmission-formLabel">
                 {{_config.questionLabel}}
               </span>
-              <textarea class="yxt-QuestionSubmission-formInput js-question-field js-question-text"
+              <textarea class="yxt-QuestionSubmission-formInput js-question-field js-question-text{{#if errors.questionTextRequired}} required-error{{/if}}"
```
data-enabler: If you're just planning on using `required-error` for styling, we could probably use `yxt-QuestionSubmission-formInput:invalid` instead

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -85,31 +81,39 @@
...
-                  class="yxt-QuestionSubmission-ackCheckboxInput js-question-field"
-                  value="true"
-                  required>
-            <div class="yxt-QuestionSubmission-privacyPolicyText">
-              {{_config.privacyPolicyLabel}}
-            </div>
+                id="yxt-QuestionSubmission-ackCheckboxInput"
+                name="privacyPolicy"
+                class="yxt-QuestionSubmission-ackCheckboxInput js-question-field"
+                value="true"
```
data-enabler: Since we're using the `checked` property, I think we can get rid of this.
Also, if I'm reading this correctly, seems like the original check for this never worked lol (since `.value` would always return `'true'`).

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,36 +18,31 @@
...
+        data-prop="icon"
+      ></button>
     </div>
   </div>
-  {{#if questionSubmitted}}
-    <div class="yxt-QuestionSubmission-form">
-      <div class="yxt-QuestionSubmission-formRow">
-        {{_config.questionSubmissionConfirmationText}}
+
+  {{#if questionExpanded}}
```
data-enabler, line 31: Torn on whether it makes more sense to just hide this with CSS when it's collapsed. We're not using a virtual dom, so there isn't really a performance benefit.

Looking at ResultsAccordion, it looks like we're hiding collapsed content with CSS there.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,18 +18,20 @@
       <div class="yxt-QuestionSubmission-titleDescription">
         {{_config.sectionDescription}}
       </div>
-      <button class="{{#if questionExpanded}}yxt-QuestionSubmission--expanded{{else}}yxt-QuestionSubmission--collapsed{{/if}} js-content-visibility-toggle"
+      <button class="{{#if questionExpanded}}yxt-QuestionSubmission__toggle--expanded{{else}}yxt-QuestionSubmission__toggle--collapsed{{/if}} js-content-visibility-toggle"
         data-component="IconComponent"
         data-opts='{ "iconName": "chevron" }'
         data-prop="icon"
+        aria-controls="yxt-QuestionSubmission-form"
```
data-enabler: Looks like you didn't add this class to the form itself.

Also, we unfortunately need to account for the fact that users could put multiple of these components on the page, so any ids we use need to be unique per component. I think `name` is supposed to be unique, so we can probably just append that. Looks like that's what the range and date components do.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,18 +18,20 @@
       <div class="yxt-QuestionSubmission-titleDescription">
         {{_config.sectionDescription}}
       </div>
-      <button class="{{#if questionExpanded}}yxt-QuestionSubmission--expanded{{else}}yxt-QuestionSubmission--collapsed{{/if}} js-content-visibility-toggle"
+      <button class="{{#if questionExpanded}}yxt-QuestionSubmission__toggle--expanded{{else}}yxt-QuestionSubmission__toggle--collapsed{{/if}} js-content-visibility-toggle"
```
data-enabler: Looked around at some css files, and it seems like our current style is basically BEM, but we use a hyphen to separate block and element instead of double underscore. Whether we should use BEM proper is outside of the scope for this, so let's just be consistent with our other stuff.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -85,31 +83,39 @@
...
-            <div class="yxt-QuestionSubmission-privacyPolicyText">
-              {{_config.privacyPolicyLabel}}
-            </div>
+                id="yxt-QuestionSubmission-ackCheckboxInput"
+                name="privacyPolicy"
+                class="yxt-QuestionSubmission-ackCheckboxInput js-question-field"
+                value="true"
+                required>
+            <span>
+              <label for="yxt-QuestionSubmission-ackCheckboxInput"
```
data-enabler: If you put the checkbox inside the label element, you don't have to worry about creating an id for it.

data-enabler (changes requested):
Make sure the tests continue to pass. You can run them locally with `npm run test`.

src/ui/templates/questions/questionsubmission.hbs
apokellypse: yeah lol the original check for this never worked

src/ui/templates/questions/questionsubmission.hbs
apokellypse: I wanted to separate them out for ease of styling. Based on the design, the checkbox effectively has its own column, so no text should appear under it with word wrapping. Any suggestions?

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: won't this mean we can't validate based on the formData object alone, but now need to depend on using references to the dom object to call checkValidity?

I'm ok with that, but some part of me really liked validating using what the form submitted, and this will make it so that I need to validation logic on the submit handler, separating validation in two stages

apokellypse: I guess validateRequired is fine as a separate step actually, because this is validating format, which is different

src/ui/templates/questions/questionsubmission.hbs
apokellypse: the red styling provided by required-error should only show up after the form submits, but :invalid will select it in real-time, which doesn't make it the perfect replacement at this time

src/ui/components/questions/questionsubmissioncomponent.js
data-enabler: The way I see it, there isn't any validation logic here that can't be checked using `checkValidity`, so our validation code becomes "if `checkValidity` returns false for anything, then show the appropriate error".

apokellypse: sgtm

src/ui/templates/questions/questionsubmission.hbs
data-enabler: Ah, fair enough. In that case, I think we probably want to prefix the classes so that it doesn't interfere with any of the client's CSS? Not sure where we stand on that, since there are other places that run afoul of this.

src/ui/templates/questions/questionsubmission.hbs
data-enabler: You can probably either
1. Make the label `display: flex` and put the text inside a div
1. Use `margin-left` and negative `textIndent`

src/ui/templates/questions/questionsubmission.hbs
apokellypse: prefixing sgtm

src/ui/templates/questions/questionsubmission.hbs
apokellypse: since there doesn't seem to be a big difference between the two methods, I'm in favor of keeping it this way because it's less html to render

src/ui/templates/questions/questionsubmission.hbs
apokellypse: done

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -275,32 +275,30 @@ export default class QuestionSubmissionComponent extends Component {
...
-    if (!formData.privacyPolicy || formData.privacyPolicy !== true) {
-      errors['privacyPolicy'] = '* You must agree to the privacy policy to submit feedback.';
-    }
-
-    if (!formData.questionText || formData.questionText.trim().length === 0) {
-      errors['questionTextRequired'] = true;
+    const fields = DOM.queryAll(formEl, '.js-question-field');
+    for (let i = 0; i < fields.length; i++) {
+      if (!fields[i].checkValidity()) {
+        if (fields[i].name === 'email') {
```
data-enabler: `switch (fields[i].name) {`

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -31,13 +31,13 @@
   {{#if questionExpanded}}
 
     {{#if questionSubmitted}}
-      <div id="yxt-QuestionSubmission-form" class="yxt-QuestionSubmission-form">
+      <div class="yxt-QuestionSubmission-form">
```
data-enabler: I think we want the id to be on both elements, which is fine since only one is rendered at a time. Also, you didn't update the `aria-controls` attribute.

data-enabler (changes requested):

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: tru

src/ui/templates/questions/questionsubmission.hbs
apokellypse: gotcha, will do

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -284,18 +284,23 @@ export default class QuestionSubmissionComponent extends Component {
...
-        } else if (fields[i].name === 'privacyPolicy') {
-          errors['privacyPolicyErrorText'] = '* You must agree to the privacy policy to submit feedback.';
-          errors['privacyPolicyError'] = true;
-        } else if (fields[i].name === 'questionText') {
-          errors['questionTextError'] = true;
+        switch (fields[i].name) {
+          case 'email':
+            errors['emailError'] = true;
+            if (!fields[i].validity.valueMissing) {
+              errors['emailErrorText'] = '* Please enter a valid email address';
```
data-enabler: According to Max, all user-facing text needs to be configurable. Seems like the error messages were overlooked in the spec.

data-enabler (changes requested):

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: ahh yes

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -85,31 +83,37 @@
...
               </span>
-              <input class="yxt-QuestionSubmission-formInput js-question-field"
+              <input class="yxt-QuestionSubmission-formInput js-question-field{{#if errors.emailError}} yxt-required-error{{/if}}"
                   name="email"
+                  type="email"
                   placeholder="{{_config.requiredInputPlaceholder}}"
                   value="{{email}}"
                   required>
             </label>
+            <div class='yxt-QuestionSubmission-inputError'>{{#if errors.emailErrorText}}{{errors.emailErrorText}}{{/if}}</div>
```
data-enabler: I think this should be `strong` instead of `div`.
https://www.w3.org/TR/WCAG20-TECHS/G138.html

Though this makes me wonder if we need to have better cues for errors for missing required fields.

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -101,7 +101,31 @@ const DEFAULT_CONFIG = {
...
+   * The default privacy policy url
+   * @type {string}
+   */
+  'privacyPolicyUrl': '',
+
+  /**
+   * The default privacy policy error text, shown when the user does not agree
+   * @type {string}
+   */
+  'privacyPolicyErrorText': '* You must agree to the privacy policy to submit feedback.',
```
data-enabler, line 122: Don't forget to add these to the readme.

data-enabler (changes requested):

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -104,9 +104,9 @@
                     class="yxt-QuestionSubmission-ackCheckboxInput js-question-field{{#if errors.privacyPolicyError}} yxt-required-error{{/if}}"
                     required>
                   {{_config.privacyPolicyText}}
+                  <a class="yxt-QuestionSubmission-privacyPolicyUrl" href="{{_config.privacyPolicyUrl}}" target="_blank">{{_config.privacyPolicyUrlLabel}}</a>
```
data-enabler, line 107: If you're going with the approach we talked about the text and link will need to be wrapped in a span or div, unless you were planning on doing that alongside the styling changes.

Also, looks like the indentation's a bit weird here.

data-enabler (approved):

PR #389, opened 2019-11-25
qa-component: styling

apokellypse:
styling changes for:
    
    - icon sizing to 18px by 18px
    - font sizes in form
    - using new mixins
    - focus state for input elements
    - error states
    - link to privacy policy


src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -1,28 +1,65 @@
...
+  $question-submission-inputError-font-weight: $font-weight-normal !default;
+  $question-submission-inputError-color: $color-error !default;
+
+  $question-submission-border: $border-default !default;
 
   $question-submission-error-text-color: rgb(148, 0, 0);
 
+  $question-submission-title-bar-background: $color-background-highlight !default;
+
+  $acknowledgement-bar-background: $color-background-highlight !default;
```
data-enabler: It's kinda weird for this to be the only variable without the `question-submission-` prefix. You could replace `background` with `bg` to stop the variable from being too long.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -1,28 +1,65 @@
...
+  $question-submission-inputError-font-size: $font-size-sm !default;
+  $question-submission-inputError-line-height: $line-height-md-sm !default;
+  $question-submission-inputError-font-weight: $font-weight-normal !default;
+  $question-submission-inputError-color: $color-error !default;
+
+  $question-submission-border: $border-default !default;
 
   $question-submission-error-text-color: rgb(148, 0, 0);
 
+  $question-submission-title-bar-background: $color-background-highlight !default;
```
data-enabler: Thinking about the variable organization, it probably makes sense to just alphabetize them? I was expecting this to be near the rest of the title variables.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -33,29 +70,53 @@
     justify-content: space-between;
   }
 
-  &-titleIconWrapper
+  &-titleIconWrapper, &-toggle--expanded, &-toggle--collapsed
```
data-enabler: Missed this in the previous review, but the idea with BEM is that we would put both `yxt-QuestionSubmission-toggle` and `yxt-QuestionSubmission-toggle--expanded` on the element, so styling common to both states would go on the former and we wouldn't need to write selectors like this.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -75,11 +136,14 @@
...
     display: flex;
     flex-direction: row;
-    margin-top: $base-spacing / 2;
-    margin-bottom: $base-spacing / 2;
+    margin-bottom: $base-spacing;
     margin-left: $base-spacing;
     margin-right: $base-spacing;
 
+    &:first-child {
+      margin-top: $base-spacing;
```
data-enabler: In my experience, `:first-child` can cause issues if, say, you added a hidden input at the beginning of the form. I'd suggest sibling selectors instead:
```scss
&-formRow {
  margin: $base-spacing;

  & + & {
    margin-top: 0;
  }
}
```
(Also personal preference on my part to use the special-case to remove margins rather than adding them.)

If the container weren't `flex`, we could just take advantage of margin collapsing.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -75,11 +136,14 @@
   {
     display: flex;
     flex-direction: row;
-    margin-top: $base-spacing / 2;
-    margin-bottom: $base-spacing / 2;
+    margin-bottom: $base-spacing;
```
data-enabler: Unrelated: there's a `margin-block-end` a few lines up that we should change, since that property isn't supported in IE.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -155,12 +240,39 @@
...
+    margin: 0;
     margin-right: $base-spacing / 2;
+    margin-top: 0.15rem;
   }
 
-  &-privacyPolicyText
+  &-privacyPolicy
   {
-    margin-right: $base-spacing;
+    text-indent: -1.75rem;
```
data-enabler: I'd rather go with the `flex`-based approach we discussed, since the `text-indent` method is kinda hacky.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -155,12 +240,39 @@
...
+    margin-left: 1.75rem;
+    margin-right: 1rem;
+  }
+
+  &-privacyPolicyUrl
```
data-enabler, line 283: There's some `Link` mixins we should probably use here.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -1,28 +1,65 @@
...
+  $question-submission-input-color: $color-text-primary !default;
+
+  $question-submission-inputError-font-size: $font-size-sm !default;
+  $question-submission-inputError-line-height: $line-height-md-sm !default;
+  $question-submission-inputError-font-weight: $font-weight-normal !default;
+  $question-submission-inputError-color: $color-error !default;
+
+  $question-submission-border: $border-default !default;
 
   $question-submission-error-text-color: rgb(148, 0, 0);
```
data-enabler: Is this unused?

data-enabler (changes requested):

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: I think that if things are in the same level of importance, alphabetizing them is probably good

apokellypse: *or within the same level of context

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: oh, I see. I'll make that change and add in the template file

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: it's used. will rename to make it more obvious

data-enabler: More concerned about the color literal. It's identical to `$color-error`, right?

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: the margins seem to be collapsing with flex :o

apokellypse: jk they are not collapsing

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: It is; I'll make it equal to $color-error

apokellypse: In the most recent commit, it's already been fixed with the color literal removed.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -240,28 +238,20 @@
...
+    margin-top: 0.2rem;
   }
 
   &-privacyPolicy
   {
-    text-indent: -1.75rem;
-    margin-left: 1.75rem;
-    margin-right: 1rem;
+    display: flex;
+    flex-flow: row nowrap;
```
data-enabler, line 279: We should use `align-items: flex-start` since we want the checkbox to be at the top of the container. The default behaves like `stretch`, which happens to align checkboxes to the top in Chrome, but not in Firefox.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -97,16 +97,23 @@
...
-                    id="yxt-QuestionSubmission-ackCheckboxInput"
-                    name="privacyPolicy"
-                    class="yxt-QuestionSubmission-ackCheckboxInput js-question-field{{#if errors.privacyPolicyError}} yxt-required-error{{/if}}"
-                    required>
-                  {{_config.privacyPolicyText}}
-                  <a class="yxt-QuestionSubmission-privacyPolicyUrl" href="{{_config.privacyPolicyUrl}}" target="_blank">{{_config.privacyPolicyUrlLabel}}</a>
-              </label>
-              <strong class='yxt-QuestionSubmission-inputError'>{{#if errors.privacyPolicyErrorText}}{{errors.privacyPolicyErrorText}}{{/if}}</strong>
+              <input type="checkbox"
+                  id="yxt-QuestionSubmission-ackCheckboxInput"
```
data-enabler: `id`s still need to be unique.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -97,16 +97,23 @@
...
+                  name="privacyPolicy"
+                  class="yxt-QuestionSubmission-ackCheckboxInput js-question-field{{#if errors.privacyPolicyError}} yxt-required-error{{/if}}"
+                  required
+              >
+              <span>
+                <label class="yxt-QuestionSubmission-privacyPolicyText">
```
data-enabler: Looks like you're missing the `for`?

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -97,16 +97,23 @@
...
+                  <span>
```
data-enabler: Is this span doing anything? it seems unnecessary.

data-enabler (changes requested):

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: ah, thanks for this useful tidbit

src/ui/templates/questions/questionsubmission.hbs
apokellypse: done

src/ui/templates/questions/questionsubmission.hbs
apokellypse: yeah it was missing; put it back in

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -33,52 +69,78 @@
...
+    border: none;
+
+    svg {
+      height: 1.25em;
+      width: 1.25em;
+    }
   }
 
-  &-title
+  &-toggle--expanded
```
data-enabler, line 99: @vramesh1 We're already deviating from BEM to match the existing stylesheets. The CSS here doesn't the `__{element}` syntax, which we aren't using anywhere yet, but does use the `--{modifier}` syntax, which is already present in some files.

src/ui/templates/questions/questionsubmission.hbs
apokellypse: not necessary so removed it

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -71,7 +71,7 @@ const DEFAULT_CONFIG = {
    * The description to display in the title bar
    * @type {string}
    */
-  'sectionDescription': 'Can’t find what you are looking for? Provide your feedback below.',
+  'teaser': 'Can’t find what you are looking for? Provide your feedback below.',
```
data-enabler: This reminds me that we need to make sure the readme includes all the configuration options, though maybe not in this PR.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -38,60 +40,63 @@
       </div>
     {{else}}
       <form id="yxt-QuestionSubmission-form-{{name}}" class="yxt-QuestionSubmission-form" novalidate>
+        <input style="display: none" type="text" name="decoyusernameforchromeautofill">
+        <input style="display: none" type="email" name="decoyemailforchromeautofill">
```
data-enabler: Do we specifically not want this form to autofillable? Can't really see a good reason for disabling.

Also, if that is what we want, is `autocomplete="off"` not an option?

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -18,12 +22,10 @@
       <div class="yxt-QuestionSubmission-titleDescription">
         {{_config.teaser}}
       </div>
-      <button class="yxt-QuestionSubmission-toggle{{#if questionExpanded}}--expanded{{else}}--collapsed{{/if}} yxt-QuestionSubmission-toggle js-content-visibility-toggle"
+      <button class="yxt-QuestionSubmission-toggle{{#if questionExpanded}}--expanded{{else}}--collapsed{{/if}} yxt-QuestionSubmission-toggle"
```
data-enabler: If this no longer performs actions, it shouldn't be a button.

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -1,7 +1,11 @@
 
 <section class="yxt-QuestionSubmission">
 
-  <div class="yxt-QuestionSubmission-titleBar">
+  <div
+    class="yxt-QuestionSubmission-titleBar js-content-visibility-toggle"
```
data-enabler, line 5: Just adding a click handler to a `div` doesn't account for keyboard users, who won't be able to navigate to and activate the element. I also don't know if the aria attributes here are enough for screen readers to present the element as a control.

https://www.w3.org/TR/wai-aria-practices/examples/button/button.html

Often it's easier to just make the element a button and override the styling as necessary, since you get all the behavior for free.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -273,7 +272,7 @@
 
     color: white;
     width: 100px;
-    height: 40px;
+    height: auto;
```
data-enabler: Could we just remove the line instead? I think `auto` should be the default.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -152,11 +154,20 @@
...
     {
       margin-bottom: $base-spacing;
+      &:only-child
+      {
+        margin-bottom: 0;
+      }
+
+      & + &
+      {
+        margin-bottom: 0;
```
data-enabler: I don't think this works correctly when there's 3+ fields in the row.

I think what we probably want is:
```scss
      & + &
      {
        margin-top: $base-spacing;
      }
```

data-enabler (changes requested):

README.md
```diff
@@ -800,20 +800,27 @@ when a search query is run.
...
+  privacyPolicyErrorText: '* You must agree to the privacy policy to submit feedback.',
+  emailFormatErrorText: '* Please enter a valid email address.'
+  requiredInputPlaceholder: '(required)',
+  questionSubmissionConfirmationText: 'Thank you for your question!',
+  buttonLabel: 'Submit',
+  entityId: 123,                                  // Required.
+  language: 'en',
+  expanded: true
 })
 ```
```
rosiegrant: ANSWERS.addComponent('QASubmission', {
  container: '.question-submission-container',
  formSelector: '.js-form',                       // Defaults to native form node within container
  nameLabel: 'Name', //Label for name input
  emailLabel: 'Email', //Label for email input
  questionLabel: 'Question', //Label for question input
  sectionTitle: 'Ask a question', //Title displayed for the form
  teaser: 'Can\'t find what you’re looking for? Ask a question below.', //Teaser displayed for the form, next to the title
  description: 'Enter your question and contact information, and we\'ll get back to you with a response shortly.' // Description for the form. 
  privacyPolicyText: 'By submitting my email address, I consent to being contacted via email at the address provided.', //Text before the privacy policy link
  privacyPolicyUrlLabel: 'Learn more here.', //Label for the privacy policy url 
  privacyPolicyUrl: 'https://mybiz.com/policy',   // Required. Defaults to ''.
  privacyPolicyErrorText: '* You must agree to the privacy policy to submit feedback.', //error message displayed when the privacy policy is not selected
  emailFormatErrorText: '* Please enter a valid email address.'  //error message displayed when an invalid email is not submitted.
  requiredInputPlaceholder: '(required)', //placeholder displayed in all required fields
  questionSubmissionConfirmationText: 'Thank you for your question!', //Confirmation displayed once a question is submitted
  buttonLabel: 'Submit', //Label displayed on the button to submit a question 
  entityId: 123,                                  // Required. Entity ID of the organization entity in the Knowledge Graph. 
  language: 'en', 
  expanded: true //whether or not the form is expanded by default when a user arrives on the page. 
})

src/ui/templates/questions/questionsubmission.hbs
apokellypse: @rosiegrant we want autofill to be off because it's a bad experience right? I personally don't like the Chrome autofill, but I don't know the official reason.

This solution was proposed in stackoverflow to work, but it doesn't seem to work all the time. Will look into the autocomplete solution.

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: it's already in the pr, rip

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: yeah & done

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: oh i like this much better.

README.md
apokellypse: done!

src/ui/templates/questions/questionsubmission.hbs
apokellypse: thanks for the link! I went with using the <button> element

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -61,6 +61,11 @@
     background-color: $question-submission-title-bar-background;
     border-bottom: $question-submission-border;
     cursor: pointer;
+
+    &:focus
+    {
+      box-shadow: inset 0 0 0 1px $color-text-primary;
```
data-enabler: `outline` is another alternative to border, and is what's typically used by browsers for focus styling.

apokellypse: whoaaaa

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -179,5 +312,11 @@
     letter-spacing: 0.5px;
     font-size: 14px;
     font-weight: $font-weight-semibold;
+    overflow-wrap: break-word;
+
+    &:focus
+    {
+      padding: 7px 13px;
```
data-enabler, line 318: This is kind of unfortunate, but the fix would have to happen in the `Button` mixin so it's not worth addressing in this commit.

src/ui/sass/modules/_QuestionSubmission.scss
```diff
@@ -155,21 +265,44 @@
...
+    margin-top: $base-spacing / 4;
+    display: block;
+  }
+
   &-submitButton
   {
     @include Button;
 
     color: white;
     width: 100px;
```
data-enabler: Is there a reason for the fixed width? Anything longer than the default text is going to wrap in a weird way. Seems like we could just remove this, but if not then `min-width` would be more appropriate.

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: Understood, will remove this from my commit

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -37,80 +40,90 @@
         </div>
       </div>
     {{else}}
-      <form id="yxt-QuestionSubmission-form-{{name}}" class="yxt-QuestionSubmission-form" novalidate>
+      <form id="yxt-QuestionSubmission-form-{{name}}" class="yxt-QuestionSubmission-form" novalidate autocomplete="false">
```
data-enabler: I don't think it's worth trying to use hacks like `"new-password"` to get around browser autofill, since browsers are always free to ignore html attributes and do whatever they want. I'd just put `autocomplete="off"` on the form and call it a day. Not sure if `autocomplete="false"` is supposed to do something special, but it's not a standard autocomplete value.

data-enabler (changes requested):

src/ui/sass/modules/_QuestionSubmission.scss
data-enabler: No, leave this as-is. My point was that we either have to adjust the padding here or change how the border works in the `Button` mixin, and it's not worth doing the latter right now.

I'm just leaving this comment for anyone in the future who wonders why things were done this way.

apokellypse: ahh gotcha

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: based on some previous conversations it seems like we want to allow for button text that take up multiple lines. I'll check with Rose

src/ui/templates/questions/questionsubmission.hbs
apokellypse: the `autocomplete="false"` was another hack that some SO users reported working. I'll remove that and the other hacks and go with `autocomplete="off"` as recommended.

src/ui/sass/modules/_QuestionSubmission.scss
apokellypse: offline discussion conclusion: flex will take care of evenly distributing words across lines, so removing fixed-width will get us what we want

src/ui/templates/questions/questionsubmission.hbs
rosiegrant: correct we wanted it to be off b/c it's a bad experience

data-enabler (approved):

PR #393, opened 2019-12-02
Add enzyme & enzyme adapter for component testing

Delerme:
Add enzyme to dev dependencies along with jest setup and a custom enzyme adapter
for the Answers component framework. To support this, adjust a few Component
requirements and move the component registry to a static object so a mock
component manager can use the same set of components


Delerme:
will update the now obsolete component tests but wanted to get this up for CR in the meantime

src/ui/components/registry.js
```diff
@@ -36,43 +34,54 @@ import QuestionSubmissionComponent from './questions/questionsubmissioncomponent
...
+
+  // Helper Components
+  IconComponent
+];
+
+/**
+ * The component registry is a map that contains
+ * all available component classes used for creation or extension.
+ * Each component class has a unique type, which is used as the key for the registry
+ * @type {Object.<string, function>}
```
data-enabler: Even more accurate would be `{Object.<string, Component>}`

src/ui/components/registry.js
```diff
@@ -36,43 +34,54 @@ import QuestionSubmissionComponent from './questions/questionsubmissioncomponent
...
-  .register(PeopleResultsItemComponent)
-
-  .register(MapComponent)
-
-// Questions Components
-  .register(QuestionSubmissionComponent)
-
-// Helper Components
-  .register(IconComponent);
+export const COMPONENT_CLASS_LIST = [
```
data-enabler: Does this actually need to be exported?

tests/setup/mockcomponentmanager.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * It is not necessary to create components to test via this
+ * manager, but it is recommended to provide a mock manager to
+ * each unit test component if manually created.
+ */
+export default class MockComponentManager {
+  constructor (mockCore) {
+    this._core = mockCore;
+
+    this.renderer = new Renderers.Handlebars();
```
data-enabler, line 19: Any significance to `_core` having the leading underscore but `renderer` not? I assume we don't need to reference this from outside the class.

tests/setup/mockcomponentmanager.js
```diff
@@ -0,0 +1,48 @@
+import { Renderers } from '../../src/ui';
+import templates from '../../dist/answerstemplates.compiled.min';
+import { COMPONENT_REGISTRY } from '../../src/ui/components/registry';
+
+/** @module MockComponentManager */
```
data-enabler, line 5: Been meaning to ask: does `@module` really do anything if there aren't exported functions/variables outside the class?

tests/setup/enzymeadapter.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+}
+
+function toComponentRstNode (answersComponent) {
+  const container = answersComponent._container;
+  const children = container.hasChildNodes() ? container.childNodes : [];
+  const childRstNodes = [];
+  for (let child of children) {
+    childRstNodes.push(toDomRstNode(child));
```
data-enabler: `children.map(toDomRstNode)`?

tests/setup/enzymeadapter.js
```diff
@@ -0,0 +1,160 @@
...
+      return domElement.textContent;
+    }
+    return null;
+  }
+
+  const props = {};
+  const attr = domElement.attributes;
+  for (let i = 0; i < attr.length; i++) {
+    if (attr[i].name === 'class') {
+      props['className'] = attr[i].value;
```
data-enabler, line 189: Does enzyme actually require this? Doesn't seem like the sort of thing we'd want to copy from React otherwise.

tests/setup/enzymeadapter.js
```diff
@@ -0,0 +1,160 @@
...
+
+function toDomRstNode (domElement) {
+  const children = domElement.hasChildNodes() ? domElement.childNodes : [];
+  const childRstNodes = [];
+  for (let child of children) {
+    childRstNodes.push(toDomRstNode(child));
+  }
+
+  if (domElement instanceof Text) {
+    const text = Text.wholeText;
```
data-enabler: Should this be `domElement.wholeText`?

data-enabler (changes requested):
Seems like it might be worth sprinkling some more JSDoc around the new files.

src/ui/components/registry.js
Delerme: guess not, removed

src/ui/components/registry.js
Delerme: true, done

tests/setup/enzymeadapter.js
Delerme: done

tests/setup/enzymeadapter.js
Delerme: yup, fixed

tests/setup/enzymeadapter.js
Delerme: yea it's how `hasClass` functions. I figured it's fine to support that, and `getDOMNode().classList` also works as expected so seems like no downside really

tests/setup/mockcomponentmanager.js
Delerme: I forget exactly, but it was only useful for generating jsdoc. I figure if we remove the generated jsdoc we'll probably clean these declarations then

tests/setup/mockcomponentmanager.js
Delerme: was mostly copy pasta, moved to `.core` since tests may want to access each to mess with them

src/ui/components/component.js
```diff
@@ -87,10 +87,12 @@ export default class Component {
...
-        throw new Error('Missing `container` option for component configuration. Must be of type `string`.');
+      if (typeof config.container === 'string') {
+        this._container = DOM.query(config.container) || null;
+        if (this._container === null) {
+          throw new Error('Cannot find container DOM node: ' + config.container);
+        }
       }
-      this._container = DOM.query(config.container) || null;
     } else {
       this._container = DOM.query(this._parentContainer, config.container);
```
b1lly, line 97: What happens if this is null?

src/ui/components/componentmanager.js
```diff
@@ -54,15 +43,11 @@ export default class ComponentManager {
     this._analyticsReporter = null;
   }
 
-  static setInstance (instance) {
+  static getInstance () {
```
b1lly, line 46: This originally was meant to be a singleton, what was the advantage of making a helper that returned a new instance every time? Should we just deprecate this helper all together if we don't want it to be a singleton anymore?

src/ui/components/registry.js
```diff
@@ -36,43 +34,54 @@ import QuestionSubmissionComponent from './questions/questionsubmissioncomponent
...
+  EventResultsItemComponent,
+  PeopleResultsItemComponent,
+  MapComponent,
+
+  // Questions Components
+  QuestionSubmissionComponent,
+
+  // Helper Components
+  IconComponent
+];
```
b1lly, line 76: In the future, I think all "components" should be exported from the package.

And this Manager could just do something like:

psuedo:

```
import * as COMPONENTS;

const manager = new ComponentManager();
for (component in COMPONENTS) {
  manager.register(component);
}

```

Obviously the above wont work right out of the box, but the idea being:

1. All component `classes` are exported from the package
2. The manager just automatically registers everything that was exported

b1lly (commented):
Just some minor nuggets:

@data-enabler for transparency

src/ui/components/componentmanager.js
Delerme: This is still a singleton, you can't really make a private constructor in JS, but the getInstance method is now more inline with how singletons work in other languages. We were actually wondering why this was a singleton in the first place, if you're okay with this no longer being a singleton we can make that change next commit

src/ui/components/registry.js
Delerme: sounds good, can you make a story for that?

src/ui/components/component.js
Delerme: It'll fail during the render instead of during the constructor. I think the goal going forward would be to split rendering and mounting logic from the component. Component instances should be able to be created without knowing things like what container they'll eventually make it to.

tests/setup/mockcomponentmanager.js
```diff
@@ -14,15 +14,16 @@ import { COMPONENT_REGISTRY } from '../../src/ui/components/registry';
...
   constructor (mockCore) {
-    this._core = mockCore;
+    this.core = mockCore;
 
     this.renderer = new Renderers.Handlebars();
     this.renderer.init(templates);
   }
 
-  create (componentType, config) {
+  create (componentType, config, analyticsReporter) {
```
data-enabler: Pretty sure this breaks a bunch of stuff. At the very least, the `create` call in `answers-umd` needs to be updated.

Alternatively, we could use the existing `setAnalyticsReporter` method on `ComponentManager`. But we shouldn't need to have both.

tests/ui/components/component.js
```diff
@@ -1,185 +1,150 @@
...
-    let renderEl = DOM.create(component.render(data));
-    let testEl = DOM.create(expected);
-    expect(renderEl.isEqualNode(testEl)).toBeTruthy();
+    const domOn = jest.spyOn(DOM, 'on');
 
-    component.setState(data);
-    component.mount();
+    const wrapper = mount(component);
+    const div = wrapper.find('#test').getDOMNode();
+    expect(div.getAttribute('data-eventtype')).toBe('test_event');
```
data-enabler: This seems like it's just testing that the template rendered, and isn't really relevant to testing the analytics functionality.

tests/ui/components/component.js
```diff
@@ -1,185 +1,150 @@
...
-
-    component.setState(data);
-    component.mount();
-    component.mount();
-    expect(domOn).toHaveBeenCalledTimes(2);
+    const wrapper = mount(component);
+    const div = wrapper.find('#test').getDOMNode();
+    expect(div.getAttribute('data-eventtype')).toBe('test_event');
+    expect(div.getAttribute('data-eventoptions')).toBe('{"name":"Vig"}');
+    expect(div.getAttribute('data-is-analytics-attached')).toBe('true');
```
data-enabler: This feels like it's testing the implementation.

data-enabler (changes requested):

tests/setup/mockcomponentmanager.js
Delerme: yea I should keep them to be the same interface, updated the mock

tests/ui/components/component.js
Delerme: true, removed

tests/ui/components/component.js
Delerme: yup, removed

data-enabler (approved):

Issue #306, opened 2019-10-17
Back button requires two clicks after a spell correction


Mxs2019:
**Steps to Reproduce**
1. Search for something that has spellcheck (http://answers.yext.com.yextpages.net/?query=howard+lemran&beta=true)
2. Click the correction
3. Press back

It will not take you back to the previous query. 

**Expected Behavior**
If you click a correction and then click back it should take you back to the pre-corrected query.

bwangYext:
This fix has been pushed to v0.10.1.

vramesh1:
closing

PR #396, opened 2019-12-04
Allow persistentstorage optionally replacing state instead of pushing new one when updates history

bwangYext:
This fixes issue #306.

TEST=auto,manual


Delerme (approved):
hmm ideally components wouldn't have to worry about this sort of thing, but think this is fine for now. The search rearchitecture will likely fix this in a better way

PR #403, opened 2019-12-10
Update navigation component for oobx

bwangYext:
https://yexttest.atlassian.net/browse/ANS-323

TEST=auto,manual


src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,33 +1,33 @@
...
-  <div class="yxt-Nav-expanded">
+<nav class="yxt-Nav-container js-yxt-navContainer {{#unless showCollapse}} yxt-Nav-static{{/unless}}">
+  <div class="yxt-Nav-expanded" role="tablist" aria-label="{{ariaLabel}}">
     {{#each tabs}}
       <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
         href="{{url}}"
         data-eventtype="{{#if @first}}ALL_TAB_NAVIGATION{{else}}VERTICAL_TAB_NAVIGATION{{/if}}"
         data-eventoptions='{"verticalConfigId": "{{configId}}"}'
-        role="option"
+        role="tab"
```
data-enabler: I don't think we should be using the tab role here. These may visually look like tabs, but they're just links to separate pages.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role
In particular, there's nothing that makes sense to put in the `aria-controls` attribute.

I think just having links inside the nav element should be enough, though we should probably be using a `<ul>`.
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/navigation.html

bwangYext: Using `<ul>` for both main links and collapsed links. Removed role and aria-selected.

bwangYext:
Ready for review

README.md
```diff
@@ -70,6 +70,7 @@ Below is a list of configuration options that can be used during initialization.
 | experienceVersion | string or number | The Answers Experience version to use for api requests | optional |
 | debug | boolean | Prints full Answers error objects when set to `true` | optional |
 | sessionTrackingEnabled | boolean | default: `true`. If true, the search session is tracked. If false, there is no tracking. | optional |
+| navigation | object | Provide navigation configuartion including tab configurations | optional |
```
data-enabler: Typo in the first "configuration"

README.md
```diff
@@ -758,17 +759,10 @@ ANSWERS.addComponent('', {
...
-```
+Tab configurations should be provided in initial configuration.
 
 ```js
-ANSWERS.addComponent('Navigation', {
-  container: '.navigation-container',
-  dropdownLabel: 'More',     // The label to display on the dropdown menu button
-  dropdownIcon: null,        // If provided, show this icon on the dropdown button instead
-  static: false,             // If true, shows all tabs in a horizontally scrolling container
+navigation: {
```
data-enabler: The pattern we've been using is to have init configuration objects like this in a section below Configuration Options. Here's how it was done for the `search` property: https://github.com/yext/answers/pull/379/files

README.md
```diff
@@ -780,13 +774,23 @@ ANSWERS.addComponent('Navigation', {
...
-      label: 'Employees'     // The label used for the navigation element
-      url: 'employees.html'  // The link for the navigation element
     }
   ]
+}
+```
+
+
+```html
+<nav class="navigation-container"></nav>
```
data-enabler: Since we use a `nav` element in the NavigationComponent template, using one for the container results in two `nav` elements. We should just change this to be a `div`; that's consistent with current Answers implementations.

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -93,6 +96,19 @@ describe('navigation component configuration', () => {
...
   });
 
+  const mockedCore = (tabsConfig) => {
+    const navigationConfig = new NavigationConfig({
+      tabs: tabsConfig
+    });
+    return {
+      globalStorage: {
+        getState: () => {
+          return navigationConfig;
```
data-enabler, line 108: You should check that the string passed in getState is the navigation config key. This code is making an assumption about how the component uses global storage, which is fine, but if that changes then this code should throw an error.

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,33 +1,30 @@
...
+      </li>
     {{/each}}
-  </div>
-  {{#unless static}}
+  </ul>
+  {{#if showCollapse}}
     <div class="yxt-Nav-moreContainer">
       <button id="yxt-Nav-moreButton" class="yxt-Nav-item yxt-Nav-more yxt-Nav-item--more js-yxt-navMore"
-        aria-haspopup="listbox"
         aria-labelledby="yxt-Nav-moreButton"
```
data-enabler: Is there a reason this element is labeled by itself?

README.md
```diff
@@ -758,17 +759,10 @@ ANSWERS.addComponent('', {
 The Navigation Component adds a dynamic experience to your pages navigation experience.
 When using multiple vertical searches in a universal search, the navigation ordering will be automatically updated based on the search results. By default, tabs that do not fit in the container will go inside a dropdown menu.
 
-
-```html
-<nav class="navigation-container"></nav>
-```
+Tab configurations should be provided in initial configuration.
```
data-enabler, line 785: Make sure you update the sample app, so it doesn't break

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -216,6 +267,7 @@ export default class NavigationComponent extends Component {
 
       if (collapsedLinks.children.length === 0) {
         moreButton.classList.add('yxt-Nav-item--more');
+        this.setAriaHidden('true');
```
data-enabler: If we use `visibility: hidden` instead of `opacity: 0` on `yxt-Nav-item--more`, then we don't need this. It also takes care of an issue the nav bar has currently where you can still tab to the More button even when it's hidden

README.md
```diff
@@ -780,13 +774,23 @@ ANSWERS.addComponent('Navigation', {
...
+<nav class="navigation-container"></nav>
+```
+
+```js
+ANSWERS.addComponent('Navigation', {
+  container: '.navigation-container',
+  mobileOverflowBehavior: 'COLLAPSE'     // optional, options which are COLLAPSE, INNERSCROLL to control if navigation shows scroll bar or dropdown for mobile
+  ariaLabel: 'Search Page Navigation'    // optional, the aria-label to set on the navigation
+  overflowLabel: 'More',    // optional, the label to display on the dropdown menu button when overflow
+  overflowIcon: null        // optional, show this icon on the dropdown button instead when overflow
```
data-enabler: "when overflow" -> "when it overflows"
On the previous line as well.

README.md
```diff
@@ -780,13 +774,23 @@ ANSWERS.addComponent('Navigation', {
...
+
+
+```html
+<nav class="navigation-container"></nav>
+```
+
+```js
+ANSWERS.addComponent('Navigation', {
+  container: '.navigation-container',
+  mobileOverflowBehavior: 'COLLAPSE'     // optional, options which are COLLAPSE, INNERSCROLL to control if navigation shows scroll bar or dropdown for mobile
```
data-enabler: "// optional, controls if navigation shows a scroll bar or dropdown for mobile. Options are COLLAPSE and INNERSCROLL"

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -281,16 +333,28 @@ export default class NavigationComponent extends Component {
...
   getUrlParams () {
     return new SearchParams(window.location.search.substring(1));
   }
 
+  shouldCollapse () {
+    const container = DOM.query(this._container, '.yxt-Nav-container') || this._container;
+    const navWidth = container.offsetWidth;
+    switch (this._mobileOverflowBehavior) {
+      case MOBILE_OVERFLOW_BEHAVIOR_OPTION.COLLAPSE:
+        return true;
```
data-enabler, line 348: Seems wasteful to bother bother querying the DOM if we're just going to return true in this case.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -155,25 +183,47 @@ export default class NavigationComponent extends Component {
...
+        clearTimeout(this._checkMobileOverflowBehaviorTimer);
+      }
+
+      this._checkMobileOverflowBehaviorTimer = setTimeout(this.setState.bind(this), RESIZE_DEBOUNCE);
+    });
   }
 
   onMount () {
-    if (!this.static) {
+    this.unbindOverflowHandlers();
```
data-enabler: This should go in `onUnMount`, which should also mean that you don't need to call `bindOverflowHandlers` in `onCreate`.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -155,25 +183,47 @@ export default class NavigationComponent extends Component {
...
+      this._debounceTimer = setTimeout(this.refitNav.bind(this), RESIZE_DEBOUNCE);
+    });
+    DOM.on(window, 'click', this.checkOutsideClick.bind(this));
+  }
+
+  unbindOverflowHandlers () {
+    if (this._debounceTimer) {
+      clearTimeout(this._debounceTimer);
+    }
+    DOM.off(window, 'click', this.checkOutsideClick.bind(this));
```
data-enabler: In order for `removeEventListener` to work you have to pass the same function that was used in `addEventListener`. `.bind` creates a new function every time you call it.

Either call `this.checkOutsideClick = this.checkOutsideClick.bind(this);` in the constructor, or just declare it as an arrow function.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -155,25 +183,47 @@ export default class NavigationComponent extends Component {
...
-        }
-
-        this._debounceTimer = setTimeout(this.refitNav.bind(this), RESIZE_DEBOUNCE);
-      });
-      DOM.on(window, 'click', this.checkOutsideClick.bind(this));
+    if (this.shouldCollapse()) {
+      this.bindOverflowHandlers();
     }
+
+    DOM.on(window, 'resize', () => {
```
data-enabler: We never remove either of the resize handlers; this causes errors if you ever remove the navigation component from the page.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -155,25 +183,47 @@ export default class NavigationComponent extends Component {
...
+      if (this._checkMobileOverflowBehaviorTimer) {
+        clearTimeout(this._checkMobileOverflowBehaviorTimer);
+      }
+
+      this._checkMobileOverflowBehaviorTimer = setTimeout(this.setState.bind(this), RESIZE_DEBOUNCE);
```
data-enabler: Re-rendering and re-mounting the component every time the window changes size isn't great. I guess it's probably not worth the time to clean it up if this works as-is, but I would at least leave a TODO about it.

Currently, the `_debounceTimer` timeout never even activates, since this runs first.

data-enabler (changes requested):
Honestly, there's a lot of things that this component isn't really doing correctly for keyboard navigation and screen reader support, but It's probably not worth the time to address them all, since the new team might end up rewriting things anyway.

If you're curious, W3C has a guide on how to implement components like this:
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

README.md
bwangYext: Done.

README.md
bwangYext: Done

README.md
bwangYext: Done.

tests/ui/components/navigation/navigationcomponent.js
bwangYext: Done.

src/ui/templates/navigation/navigation.hbs
bwangYext: Removed.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Done.

README.md
bwangYext: Done.

README.md
bwangYext: Done.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Done.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Moved to onUnMount.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Done.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Added calls to remove both handlers.

src/ui/components/navigation/navigationcomponent.js
bwangYext: Added to do. Could you explain why _debounceTimer timeout never activates? It seems that the it is refitting nav when resize.

README.md
bwangYext: Done.

src/ui/components/navigation/navigationcomponent.js
data-enabler: The code registers two `resize` event handlers, which get called in the order they were registered. That means that this handler gets called before the one in `bindOverflowHandlers`; Javascript is single-threaded, so the handlers get called sequentially. Both handlers call `setTimeout` with the same delay; those timeouts will also happen sequentially. In this timeout we call `setState`, which causes the component to be unmounted, then re-mounted. In `onMount` we call `unbindOverflowHandlers`, which clears the `_debounceTimer` before it gets a chance to fire.

The reason the resizing works correctly is because we also call `refitNav` in `onMount`.

README.md
data-enabler: Looks like this wasn't changed.

src/ui/templates/navigation/navigation.hbs
```diff
@@ -1,33 +1,27 @@
...
+        data-opts='{"iconName": "{{overflowIcon}}"}'
         data-prop="icon"></span>
-        {{dropdownLabel}}
+        {{overflowLabel}}
       </button>
-      <div class="yxt-Nav-modal js-yxt-navModal {{#if expanded}}is-active{{/if}}"
-        role="listbox"
-        aria-labelledby="yxt-Nav-moreButton"
-      ></div>
+      <ul class="yxt-Nav-modal js-yxt-navModal {{#if expanded}}is-active{{/if}}"></ul>
```
data-enabler: It's fine to leave the `aria-labelledby` on this element.

data-enabler (changes requested):

src/ui/templates/navigation/navigation.hbs
bwangYext: Done

README.md
bwangYext: Done

src/ui/components/navigation/navigationcomponent.js
bwangYext: Thanks! Removed `_debouncerTimer`.

data-enabler (approved):

Issue #395, opened 2019-12-04
SearchBar Component: Autocomplete Remains Expanded After Search


rosiegrant:
Sometimes, when you enter a query, the autocomplete remains expanded 

See second 18 in this clip:
https://drive.google.com/file/d/18S3DGdUYPuSoSjevt_sGjQZ0ESB1axit/view

Browser: Chrome 78
Dignity is on v0.9.6 in staging

rosiegrant:
I've verified this works so am closing, but anyone can feel free to reopen

PR #409, opened 2019-12-11
Only show autocomplte dropdown when it is focused on query input

bwangYext:
Fix for https://github.com/yext/answers/issues/395. The cause was that autocomplte respone came back later than query response.


src/ui/components/search/autocompletecomponent.js
```diff
@@ -137,6 +142,10 @@ export default class AutoCompleteComponent extends Component {
     }));
   }
 
+  isQueryInputFocused () {
+    return document.activeElement.className.includes(this._inputEl.substring(1));
```
vramesh1, line 146: is the purpose of `substring(1)` to exclude a leading underscore in the input element name?
will there always be an underscore, or should we exclude it only if it exists?

bwangYext: It is used to remove the period in front of input element selector. But we are not validating the selector given by clients.

vramesh1: oh I see. ok.

vramesh1: probably would be clearer with a comment, but sounds good

vramesh1 (approved):

PR #412, opened 2019-12-12
send analytics events for form

apokellypse:


src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -168,32 +183,50 @@ export default class QuestionSubmissionComponent extends Component {
...
-   * bindFormSubmit handles submitting the question to the server.
+   * bindFormSubmit handles submitting the question to the server,
+   * and submits an event to serverside reporting
    * @param {HTMLElement} formEl
    */
   bindFormSubmit (formEl) {
     DOM.on(formEl, 'submit', (e) => {
       e.preventDefault();
+
+      // we report that we are attempting to submit form data
```
data-enabler: Putting a comment before a block of code is usually a good sign that it should be pulled out into a helper function.

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -119,6 +119,21 @@ export default class QuestionSubmissionComponent extends Component {
      */
     this.moduleId = StorageKeys.QUESTION_SUBMISSION;
 
+    /**
+     * verticalKey is used for analytics
+     * @type {string}
+     * @private
+     */
+    this._verticalKey = config.verticalKey;
```
data-enabler: What I was trying to say before is that now that we have `verticalKey` available in the global config `ANSWERS.init()`, we don't need to set it for each individual component.

[I've suggested that in the future the global config should be passed to components](https://docs.google.com/document/d/1-iS_Z1ur3_2MMw8mIj6GCVp5m_PGmzLPSpv7mYxufX4/edit#heading=h.pifayunu5mve), but in the meantime you can retrieve it from global storage [like we do in PaginationComponent](https://github.com/yext/answers/blob/master/src/ui/components/results/paginationcomponent.js#L18)

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -119,6 +119,21 @@ export default class QuestionSubmissionComponent extends Component {
...
+
+    /**
+     * isUniversal is used for analytics and passed to children and is set to
+     * true if this component is added by the UniversalResultsComponent
+     * @type {boolean}
+     * @private
+     */
+    this._isUniversal = config.isUniversal || false;
```
data-enabler: Isn't this just `!verticalKey`?

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -168,32 +183,50 @@ export default class QuestionSubmissionComponent extends Component {
...
     if (this.analyticsReporter === null) {
       return;
     }
 
-    const questionTextEl = DOM.query(formEl, '.js-question-text');
-    DOM.on(questionTextEl, 'focus', () => {
-      this.analyticsReporter.report(new AnalyticsEvent('QUESTION_FOCUS'));
-    });
+    const questionFields = DOM.queryAll(formEl, '.js-question-field');
+    for (let i = 0; i < questionFields.length; i++) {
```
data-enabler: For loops are so last-year; you could use `Array.prototype.foreach` instead and not have to deal with indices. I'd suggest `for ... of`, but I think we're trying to avoid that in the SDK because it transpiles to a lot of code and bloats the IE bundle.

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -168,32 +183,50 @@ export default class QuestionSubmissionComponent extends Component {
...
+  bindFormFocus (formEl) {
     if (this.analyticsReporter === null) {
       return;
     }
 
-    const questionTextEl = DOM.query(formEl, '.js-question-text');
-    DOM.on(questionTextEl, 'focus', () => {
-      this.analyticsReporter.report(new AnalyticsEvent('QUESTION_FOCUS'));
-    });
+    const questionFields = DOM.queryAll(formEl, '.js-question-field');
```
data-enabler: From the discussion, it sounded like Rose only wants events to get sent when the question input itself gets focused, not name, email, etc.

I'd suggest the `'[name="questionText"]'` selector.

data-enabler (changes requested):

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: Thanks for the note, I didn't know about that caveat with for...of, so that's a cool fact.

Turns out after limiting the focus event to the question itself iteration won't be necessary, so I won't need to use foreach or for after all.

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: there is an existing selector js-question-text that was already hooked up with an event that I can also use

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: Could there be a situation where the client messes up and defines verticalKey but the component was added by UniversalResults, so config.isUniversal is true? Or, pending the new behavior with verticalKey available in the global config, is that scenario not possible?

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: good point, will do

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: ahhh what I thought was that it was going to happen, not that I could use it now. Thanks for the links.

src/ui/components/questions/questionsubmissioncomponent.js
apokellypse: turns out QA component isn't a child of UniversalResults, so it doesn't have a meaningful isUniversal value. will use !verticalKey instead

data-enabler (approved):

PR #414, opened 2019-12-13
change default text values in qa component and update readme. now sup…

apokellypse:
…ports expanded state as well.


README.md
```diff
@@ -800,20 +800,34 @@ when a search query is run.
 ```
 
 ```js
+// Unless noted, fields are optional and show default values
```
data-enabler, line 803: `container` is required as well

README.md
```diff
@@ -800,20 +800,34 @@ when a search query is run.
...
+                                                  // Error message displayed when the privacy policy is not selected
+  emailFormatErrorText: '* Please enter a valid email address.'
+                                                  // Error message displayed when an invalid email is not submitted
+  requiredInputPlaceholder: '(required)',         // Placeholder displayed in all required fields
+  questionSubmissionConfirmationText: 'Thank you for your question!',
+                                                  // Confirmation displayed once a question is submitted
+  buttonLabel: 'Submit',                          // Label displayed on the button to submit a question
+  entityId: 123,                                  // Required. Set this to the Entity ID of the organization entity in the Knowledge Graph
+
+  language: 'en',
```
data-enabler: Currently checking with Rose if we even want to keep this config options, but we can leave it be in the meantime.

data-enabler (changes requested):

README.md
data-enabler: Heard back from her and it doesn't sound like there's a good reason for it. So I guess we're free to use the global config locale, although I am now remembering that that isn't an easy thing to do... 😩

README.md
apokellypse: thanks, I will add a comment

README.md
apokellypse: In questionsubmissioncomponent we already have
`
'questionLanguage': this._config.language,`

which doesn't seem that hard; am I missing something

data-enabler: Pretty sure that refers to the language passed in with the component config, not the language set on the global config. What we've been doing for now is to store parts of the global config in global storage (see PaginationComponent).

data-enabler (changes requested):

README.md
apokellypse: Ah I see, thanks! Most recent commit should have this corrected.

sample-app/public/app.js
```diff
@@ -37,6 +38,7 @@ window.initAnswers = async function initAnswers () {
     businessId,
     search: searchConfig,
     sessionTrackingEnabled: sessionTrackingEnabled,
+    locale: locale,
```
data-enabler: You should set a locale in `sample-config.json`. Seems like the code as-is will pass `locale: undefined` to the SDK, and the SDK currently doesn't validate locales.

Also, you can just to `locale,` instead of `locale: locale,`, to be consistent with the other lines.

data-enabler (changes requested):

sample-app/public/app.js
apokellypse: sgtm!

data-enabler (approved):

PR #424, opened 2019-12-18
Add targeted reset at the start of the Answers CSS.

tmeyer2115:
This CR adds a targeted CSS reset at the beginning of the Answers CSS. This
will clear certain user agent stylings for elements with a "yxt-..." class
name. Note that this will not impact and customer stylings on the page.

J=SPR-1491
TEST=auto, manual

Reran all Jest tests. Manually verified that the problematic user agent
stylings I saw previously were no more.


benmcginnis (approved):
LGTM

Delerme (approved):

PR #426, opened 2019-12-18
increment version to v0.11.0

vramesh1:


data-enabler:
What's the goal here? This seems like a weird spot to cut a release, considering that the only changes from 0.10.0 are the component testing stuff, which is irrelevant to library users, and the QASubmission stuff, half of which is still in review (#389, #414).

Also, we need to merge in bug fixes from 0.10.x, right?

vramesh1:
Discussed with Max.
We'll hold off on cutting the release until the outstanding Q&A and Nav Component PRs are merged in with outstanding changes requested made.
Will re-request review when ready.

data-enabler (approved):

PR #431, opened 2020-01-07
SearchBar: Assorted high priority QA fixes.

tmeyer2115:
This CR addresses a number of the high priority QA issues with the SearchBar.
Specifically, it fixes the following:

- Default styling of promptHeader is different than default styling of
  actual autocomplete items.
- Default styling of the bar's title. Previously, with the default styling,
  the title did not appear.
- Adds a shadow and outline for the list of autocomplete results on mobile
  (when properly focused).
- Adds an implementation of delete to SearchParams. This is necessary for
  IE11 compliance (when URLSearchParams is not available). I was seeing
  'delete undefined' errors in IE11.

J=ANS-322
TEST=auto, manual

Manually tested the above fixes using a local dev site. Ran all Jest tests.
Used BrowserStack to test IE11 compatibility of SearchParams.


src/ui/dom/searchparams.js
```diff
@@ -89,6 +89,14 @@ export default class SearchParams {
     return query in this._params;
   }
 
+  /**
+   * delete removes the given query param and its associated value from the params object
+   * @param {string} name the query param key
+   */
+  delete (name) {
+    delete this._params[String(name)];
```
apokellypse, line 97: Is casting to a String needed?

src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -34,32 +35,47 @@
...
+        $weight: $autocomplete-text-font-weight,
+        $color: $autocomplete-text-color
+      );
+
+      &.yxt-selected
+      {
+        background-color: $autocomplete-selected-background-color;
+      }
+
+      strong
```
apokellypse, line 70: I have mixed feelings about styling outside of normalizing styling that targets html tags vs classes, but I am unsure of our conventions.

benmcginnis: this should probably be in the reset nested under a ‘yxt-‘ attribute selector

benmcginnis (approved):

src/ui/sass/modules/_AutoComplete.scss
creotutar: we are generally a fan of using classes over html tags in these situations

data-enabler: I think it's reasonable here, since at least it's prefixed with a Yext class.

The question here is if we really want to generally target `strong` elements, or if this styling is specifically for something like query highlighting. If the server were returning the markup I'd say the former, but since it's generated client-side I think it's the latter and this should ideally have a specific class.

src/ui/dom/searchparams.js
tmeyer2115: I'm not sure if the cast is strictly necessary. Casting to a String is being performed in the other methods of this class (get, set, etc.) so I was trying to be consistent.

src/ui/sass/modules/_AutoComplete.scss
tmeyer2115: We were using this html tag previously, I just refactored the styling in here slightly (to make a distinction between the styling of the yxt-AutoComplete-option--item and yxt-AutoComplete-option--promptHeader classes). Would it be okay to do what Ben suggested as part of another PR (investigate if this styling can be included as part of the CSS reset).

src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -34,32 +35,47 @@
...
-    );
-
-    cursor: pointer;
     padding-left: $base-spacing;
     padding-right: $base-spacing;
     padding-top: $base-spacing / 4;
     padding-bottom: $base-spacing / 4;
 
-    &.yxt-selected
+    &--promptHeader
```
data-enabler, line 43: So I'm really confused about what promptHeader is supposed to be. The documentation says

> optioanl [sic], the query text to show as the first item for auto complete

which along with the fact that it's placed within a `yxt-AutoComplete-results` list implies that it's some sort of hard-coded autocomplete option, but the name (and the fact that clicking it doesn't do anything) implies that it's just a decorative header. I'm guessing it's the latter, and the documentation and markup are just wrong? Maybe @b1lly can clarify.

If that's the case, the selector for this should definitely use a different element than `-option`; so this should probably be `yxt-AutoComplete-promptHeader`.

data-enabler (commented):
I have a couple of concerns, but fixing them would require changing the AutoComplete template, which I think is out of scope for this.

Also any particular reason this is going on 0.12.0 instead of 0.11.x if this is a qa fix? I personally don't really care, but I wouldn't be surprised if product asks you to put this into 0.11 as well.

PR #432, opened 2020-01-07
SearchBar & AutoComplete: Assorted medium priority QA fixes.

tmeyer2115:
This CR includes medium priority QA fixes for the SearchBar and AutoComplete
components. These fixes include:

- Updating the documentation about the SearchBar's configuration options.
- Preventing an autocomplete request from firing twice when initially clicking
  the search input on page load. To do this, I removed the 'click' handler
  the AutoComplete component added to the <input> element. I believe this
  handler was unnecessary. A 'click' event triggers a 'focus' event. This
  'focus' event gets resolved first. There is already a 'focus' handler
  specified in AutoComplete. It launches the same autocomplete API call that
  the 'click' handler does. So, the additional 'click' handler seemed
  unnecessary.I need to talk with Rose about the correct behavior for an
  AutoComplete component with autoFocus.
- Updating styling of the clear button.
- Updating default label text of the SearchComponent.

J=ANS-322
TEST=auto, manual

Manually tested the above fixes using a local dev site. Reran all Jest
tests.


src/ui/components/search/autocompletecomponent.js
```diff
@@ -201,12 +203,6 @@ export default class AutoCompleteComponent extends Component {
       this.handleSubmitResult(e.keyCode, queryInput.value, e);
     });
 
-    if (this._autoFocus) {
-      DOM.once(queryInput, 'click', () => {
-        this.autoComplete(queryInput.value);
-      });
-    }
-
```
data-enabler: `SearchComponent` passes `this.autoFocus && !this.autocompleteOnLoad` as the value for AutoComplete's `autoFocus`, so I feel like there's some bad naming going on here. I think the reasoning here was that if we autofocus the search bar without automatically opening the dropdown, we want to trigger an autocomplete request if the user clicks the searchbar even though it was already focused. 

If we're fine with losing that functionality this seems fine, otherwise the right fix is probably to register this handler, but remove it when the searchbar is `blur`-ed.

tmeyer2115: Reverted this change as part of a new commit in this branch

data-enabler (approved):

PR #436, opened 2020-01-08
AutoComplete: Add spacing before first autocomplete option.

tmeyer2115:
This CR makes a minor UI change to the AutoComplete component. Specifically,
10px of spacing is added before the first autocomplete option (be it an actual
option or the promptHeader). This spacing was removed in a previous CR, but
after playing around some more, UX decided the spacing should be put back.

J=ANS-322
TEST=manual

Manually verified that the correct amount of spacing appeared.


src/ui/sass/modules/_AutoComplete.scss
```diff
@@ -20,6 +20,7 @@
     @media (min-width: $breakpoint-mobile-min) {
       margin-left: $base-spacing;
       margin-right: $base-spacing;
+      margin-bottom: 10px;
```
benmcginnis: maybe do ⅔ base-spacing? That way if the core spacing is adjusted the proportions remain the same.

tmeyer2115: Good call! Updated this in a second commit.

benmcginnis (approved):

PR #437, opened 2020-01-08
Autocomplete: Fix logic to determine if query input is focused.

tmeyer2115:
This CR fixes a bug in the logic of the isQueryInputFocused method. This method
assumed that document.activeElement was never null. However, if nothing in the
DOM has focus, activeElement can be either <body> or null. In IE11, I was
seeing that activeElement was null in this case. This lead to us attempting
to access the className property of something that was null. A null
activeElement should mean that isQueryInputFocused is false. This CR adds that
check.

J=ANS-322
TEST=auto, manual

Manually verified that the SearchBar and AutoComplete components worked as
expected after this change. Also made sure I was not seeing the invalid access
errors any more in IE11.


bwangYext (approved):

src/ui/components/search/autocompletecomponent.js
```diff
@@ -150,7 +150,8 @@ export default class AutoCompleteComponent extends Component {
   }
 
   isQueryInputFocused () {
-    return document.activeElement.className.includes(this._inputEl.substring(1));
+    return document.activeElement &&
```
data-enabler, line 153: Be aware that this changes the method signature from returning `boolean` to returning `boolean | null`. This could cause subtle bugs down the line.

PR #438, opened 2020-01-08
SearchBar: Add clearText config option and improve WCAG for clear.

tmeyer2115:
This CR adds a new configuration option for the SearchBar component, clearText.
This is in support of improving the accessibility of the SearchBar component.
Specifically, the clear button has some challenges when navigated to using
Voiceover. In Voiceover, when a user focuses on the clear button, it just
says 'Button' and gives no indication as to what the clear button does. Now,
a description of the clear button can be given with the clearText configuration
option. The default for clearText is 'Clear'. This follows the same pattern
we have where submitText is used to provide context to Voiceover about the
behavior of the search button.

J=ANS-322
TEST=manual

Navigated a local dev site manually using Voiceover. Voiceover gave the
expected context when I focused on the clear button. I made sure that things
worked as expected with Voiceover when the clear button was hidden too.


benmcginnis (approved):

rosiegrant:
LGTM! 

(we discussed updating it to `clearSrOnlyText`, and `submitText` to `submitSrOnlyText`, but from a loose poll of HHs, "srOnly" isn't clear, so we can leave the label as-is)

PR #444, opened 2020-01-13
SearchBar: Handle duplicate searches with near me intent.

tmeyer2115:
With this CR, before a query is issued, we compute the intent of the query and,
if appropriate, attempt to get location information. Namely, if the intent of
the query is NEAR_ME, we will attempt to get location information before
issuing the actual query. This information, if the user consents, will be
added to the query when it is issued. This is in contrast to the current
behavior of the system. Currently, the SearchBar issues the query without
first checking it's intent. Once that query is performed, we attempt to get
location info from the user (if the query had a NEAR_ME intent). If we get this
information, we perform a second query with the location info. This CR
simplifies the process by not issuing the query until we've had a chance to get
the location information. Note that this behavior of checking intents before
issuing the query can be turned off by disabling promptForLocation in the
SearchBar's config.

Part of the requirements for this item were to remove all references to the
PermissionsAPI from the SearchBar component.

J=ANS-404
TEST=auto, manual

Fixed the Jest tests broken by this change. Added a new unit test for the
AutocompleteData class. Tested the following scenarios manually:

- Location info was initially unset in Chrome. Picked an autocomplete option
  that had NEAR_ME intent. I allowed the site to access my location. In the
  Network tab, I saw exactly one query was issued and it included my location
  info. I manually entered a second query with NEAR_ME intent. I was not
  prompted again to allow access to my location info. Again, exactly one query
  was issued and it had my location info.
- Location info was initially unset in Chrome. Entered a query with NEAR_ME
  intent. Denied access to my location. Ensured only one query was issued and
  it did not include location information. Entered another query with NEAR_ME
  intent to ensure I wasn't again prompted for access to my location.
- Location info was initially unset in Chrome. I performed a search by editing
  the query param in the site's URL. The query had NEAR_ME intent. I was
  prompted for access to my location. I granted the access and verified only
  one query was issued and that it had my location info. Navigated to a
  vertical tab and ensured I wasn't prompted for that info again. Made sure
  one query was executed on that page and it had my location info.
- Disabled promptForLocation. Performed a query with NEAR_ME intent. Ensured
  I was not prompted for my location info and only one query was executed.


apokellypse:
Lgtm; is this working in other Browsers besides Chrome?

tmeyer2115:
@apokellypse yep! this was tested in Safari as well.

src/ui/components/search/searchcomponent.js
```diff
@@ -410,6 +405,33 @@ export default class SearchComponent extends Component {
     }
   }
 
+  /**
+   * A helper method that computes the intents of the provided query. If the query was entered
+   * manually into the search bar, its intents will have been stored already in globalStorage.
```
data-enabler: Maybe I'm misunderstanding, isn't it when the query was input via autocomplete that we have the intent data already?

src/ui/components/search/searchcomponent.js
```diff
@@ -410,6 +405,33 @@ export default class SearchComponent extends Component {
...
+      return autocompleteRequest.then(data => data.inputIntents);
+    } else {
+      // There are two alternatives to consider here. The user could have selected the query
+      // as an autocomplete option or manually input it themselves. If the former, use the intents
+      // of the corresponding autocomplete option. If the latter, use the inputIntents of the
+      // autocompleteData.
+      const results = autocompleteData.sections.flatMap(section => section.results);
+      const matchingResult = results.find(result => result.value === query);
+      const queryIntents = matchingResult ? matchingResult.intents : autocompleteData.inputIntents;
+      return new Promise((resolve, reject) => resolve(queryIntents));
```
data-enabler: `Promise.resolve(queryIntents)`

src/ui/components/search/searchcomponent.js
```diff
@@ -410,6 +405,33 @@ export default class SearchComponent extends Component {
...
+   * @returns {Promise} A promise containing the intents of the query.
+   */
+  fetchQueryIntents (query) {
+    const autocompleteData =
+      this.core.globalStorage.getState(`${StorageKeys.AUTOCOMPLETE}.${this.name}.autocomplete`);
+    if (!autocompleteData) {
+      const autocompleteRequest = this._verticalKey
+        ? this.core.queryVertical(query, this._verticalKey, this._barKey)
+        : this.core.queryUniversal(query);
+      return autocompleteRequest.then(data => data.inputIntents);
```
data-enabler, line 434: This seems pretty wasteful? We're going from only repeating queries when we have new geolocation data to just always repeating the query.

Ideally we would only repeat the query if the server returns a NEAR_ME intent that wasn't known when the query was originally send, and the user provides location data. This is pretty much what the original behavior was, even before the permission-checking optimization; I think the proper approach here would be to detect a difference in intents between the request and response and defer displaying the results until we resend with location data, if necessary. This would require hooking into the query flow in a way we haven't had to before, but I think it might be as simple as tossing an extra callback in the `then` chain.

src/core/core.js
```diff
@@ -197,15 +197,36 @@ export default class Core {
...
+
+  /**
+   * Given an input, query for a list of similar results in the provided vertical
+   *
+   * @param {string} input the string to autocomplete
+   * @param {string} verticalKey the vertical key for the experience
+   * @param {string} barKey the bar key for the experience
+   * @returns {Promise<AutoCompleteData>} the autocomplete results
+   */
+  queryVertical (input, verticalKey, barKey) {
```
data-enabler: Do we really need these in addition to `autoCompleteUniversal` `autoCompleteVertical`? Those also add the autocomplete data to global storage, which seems desirable.

src/ui/components/search/searchcomponent.js
data-enabler: Oops, due to the naming I didn't notice this was calling autocomplete and not performing an actual query. Carry on.

src/core/core.js
tmeyer2115: The queryUniversal and queryVertical methods are currently only called when we don't already have autocomplete data in storage. That corresponds to the case where a user performs a query by manually editing the URL's query param or navigating to a vertical tab. I was concerned about potential side effects if I added the autcomplete data to the global store in these cases. Conceptually, it seemed that we should store autocomplete data in the store only if it was generated by manually typing in the search bar. If we don't anticipate any issues always storing the autocomplete data in the store, I can definitely just get rid of these methods.

tmeyer2115: Read through the code a bit more and I think I was worried for nothing. The fetchQueryIntents method I create is the only consumer of the state's AUTOCOMPLETE info. So, there shouldn't be any weird side effects if I just use autoCompleteUniversal and autoCompleteVertical.

tmeyer2115: Updated this in my most recent commit.

src/ui/components/search/searchcomponent.js
tmeyer2115: Done.

src/ui/components/search/searchcomponent.js
tmeyer2115: Updated docs.

data-enabler (approved):

PR #446, opened 2020-01-14
Filters: remove storeOnChange config option

oshi97:
This commit removes the storeOnChange config option for the DateRangeFilter
and RangeFilter. The setting, which was false by default  didn't make much
sense within the context of either filter - we would always want to update
the filter in global storage when it is changed. The only current usage
of these filters currently has this setting toggled true, and this setting
is also currently undocumented for these components.

This commit does not fix the issue with filters not being loaded into
global storage from persistent storage on answers init.

TEST=manual
Tested local dev site with date range filter and range filter,
and checked that the request was filtering as expected.


tmeyer2115 (dismissed):

data-enabler:
Did some more reading, and it looks like storeOnChange is meant to be used for filters inside a filter box component, that way the filter box can control when the filters get applied. I don't think this was intended to be set by users; user config being used to pass around internal values is an unfortunate limitation of the current component architecture, that I was hoping to address with this redesign: https://docs.google.com/document/d/1-iS_Z1ur3_2MMw8mIj6GCVp5m_PGmzLPSpv7mYxufX4/edit#heading=h.pifayunu5mve

Seems like the right thing to do here is probably to just have it default to true? FilterBoxComponent already sets it to false when creating filter components.

oshi97:
Ok that makes sense to me, updated!

tmeyer2115 (approved):

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -51,7 +51,7 @@ export default class DateRangeFilterComponent extends Component {
      * @type {boolean}
      * @private
      */
-    this._storeOnChange = config.storeOnChange || false;
+    this._storeOnChange = config.storeOnChange || true;
```
data-enabler, line 54: This always evaluates to `true`.
What you actually want is `config.storeOnChange === undefined ? true : config.storeOnChange`.

PR #448, opened 2020-01-14
SpellCheck: styling changes

oshi97:
This commit changes the spell check styling to be closer to the mock.

TEST=manual
Tested that styling looked like mock


src/ui/sass/modules/_SpellCheck.scss
```diff
@@ -6,9 +6,18 @@
   $spellCheck-text-font-weight: $font-weight-normal !default;
   $spellCheck-text-color: $color-text-primary !default;
 
+  padding-top: 32px;
```
benmcginnis: probably want this to be based on a variable, I think it would be `$base-spacing*2`;

src/ui/sass/modules/_SpellCheck.scss
```diff
@@ -6,9 +6,18 @@
...
   padding-bottom: $base-spacing;
   display: block;
 
+  .yxt-SpellCheck-helpText {
+    font-size: $font-size-lg;
+  }
+
+  .yxt-SpellCheck-container {
+    height: 20px;
```
benmcginnis: same thing here though i'm not sure what the base-ten name is

src/ui/sass/modules/_SpellCheck.scss
```diff
@@ -17,6 +26,7 @@
 
   strong
   {
+    font-style: italic;
```
benmcginnis, line 35: interesting, it's unusual to see a strong element styled in italics

benmcginnis (dismissed):

src/ui/sass/modules/_SpellCheck.scss
oshi97: Yeah the spec asked for the auto-corrected part of the search text to be bold + italics and before it was just bold but I'm not sure if this is the best way?

benmcginnis: The strong element comes back from the API, yeah? If so then this is definitely the way to do it.

src/ui/sass/modules/_SpellCheck.scss
oshi97: ah ok I couldn't find anything either so I added one

benmcginnis (dismissed):

tmeyer2115 (approved):

PR #450, opened 2020-01-15
Answers SDK: Include 1st party cookie when sending user events to Ana…

tmeyer2115:
…lytics.

With this CR, a 1st party cookie ID can be included when sending user events to
the realtime analytics ingester. This behavior is guarded by enabling
conversion tracking, which is an opt-in feature. If opted out, user events will
be sent as before this change, with no 1st party cookie information. Note that
to enable conversion tracking, a user must call the ANSWERS.setConversionOptIn
method and include the ytag library via a <script/> tag in their HTML. This
library was provided by Analytics and is responsible for the cookie generation.

Note that if the user calls setConversionOptIn to enable conversion tracking,
but does not include the <script/> tag, an AnswersAnalyticsError will be thrown
when they attempt to report a user event.

J=SPR-1601
TEST=auto, manual

Updated and added Jest tests. Tested the following scenarios manually:

- Enabled conversion tracking with setConversionOptIn and added the <script/>
  tag. Hit the clear search button multiple times. Each time, I obeserved the
  request payload sent to Analytics for the CLEAR_SEARCH event. The payload
  included the same 1st party cookie ID each time (even after I refreshed the
  page).
- I did not have conversion tracking enabled and issued a CLEAR_SEARCH event.
  The payload did not include any 1st party cookie info.
- I called setConversionOptIn to enable conversion tracking but did not include
  the ytag library via <script/> tag. I saw the AnswersAnalyticsError whenever
  I attempted to send a CLEAR_SEARCH event.


tripproberts (commented):
Awesome, thanks for handling. Will let Billy approve but LGTM

PR #458, opened 2020-01-16
Implement SortOptionsComponent

oshi97:
This commit implements the sort options component for adding a
filter component that lets users filter vertical results
based on configuration.

TEST=manual
Ran test site with component locally


src/ui/components/component.js
```diff
@@ -214,6 +214,7 @@ export default class Component {
       this.setState(opts.data || opts.state || {});
       this.onCreate();
       this.userOnCreate();
+      // console.log('init', this);
```
tmeyer2115: Remove.

src/ui/components/component.js
```diff
@@ -227,6 +228,7 @@ export default class Component {
         this.userOnUpdate();
         this.unMount();
         this.mount();
+        // console.log('uddate', this);
```
tmeyer2115: Remove.

src/ui/sass/modules/_SortOptions.scss
```diff
@@ -0,0 +1,11 @@
+/** @define SortOptions */
+.yxt-SortOptions {
```
tmeyer2115, line 3: Was this all the styling necessary to match the mocks?

src/core/core.js
```diff
@@ -119,7 +119,8 @@ export default class Core {
         isDynamicFiltersEnabled: this._isDynamicFiltersEnabled,
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
-        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN)
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
+        sortBys: JSON.stringify(this.globalStorage.getState(StorageKeys.SORT_BY) || undefined)
```
tmeyer2115: Is the || undefined necessary here? Won't getState(StorageKeys.SORT_BY) be undefined if sortBys are not present in the state?

src/core/core.js
```diff
@@ -267,6 +268,14 @@ export default class Core {
       });
   }
 
+  /**
+   * Stores the given sortBys into storage, to be used for the next search
+   * @param {Array<Object>} sortBys
+   */
+  setSortBys (sortBys) {
+    this.globalStorage.set(StorageKeys.SORT_BY, sortBys);
```
tmeyer2115: Do we also want the sortBy info to wind up in persistent storage? Does it do this already?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+import Component from '../component';
+import { AnswersBasicError } from '../../../core/errors/errors';
+import DOM from '../../dom/dom';
+import StorageKeys from '../../../core/storage/storagekeys';
+import Filter from '../../../core/models/filter';
+
+/**
+ * Renders configuration options for sorting Vertical Results.
+ */
+export default class SortOptionsComponent extends Component {
```
tmeyer2115, line 14: We'll want to add Jest tests for this component. We should try and have these tests exercise as much of the component's functionality as possible. Take a look at the tests that already exist in tests/ui/components for some examples.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+}
+
+function throwConfigError (message) {
+  throw new AnswersBasicError(message, 'SortOptions');
+}
+
+function assignDefaults (config) {
+  const updatedConfig = {};
+
+  // Optional, The label used for the “default” sort (aka sort the order provided by the config).
```
tmeyer2115: "(aka the sort specified by the experience config)"?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+  updatedConfig.defaultSortLabel = config.defaultSortLabel || 'Best Match';
+
+  // Array of search options, where an option has type, label, and if is type FIELD also a label and direction
+  updatedConfig.options = config.options || throwConfigError('Options are required in config');
+  const OPTION_TYPES = ['RELEVANCE', 'ENTITY_DISTANCE', 'RELEVANCE_RANDOM'];
+  for (let i = 0; i < updatedConfig.options.length; i++) {
```
tmeyer2115: Minor nit, but let's use updatedConfig.options.forEach

src/core/core.js
oshi97: it's sorta weird but getState returns null when the key isn't found, and JSON.stringify(null) returns 'null' rather than null

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    const option = updatedConfig.options[i];
```
tmeyer2115: Will this loop also modify the config.options? I know arrays in JS are passed by copy of a reference. I'm not sure if that also applies to assignments (like the one on line 156). We'd definitely want to avoid mutating the input config.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    option.type = option.type || throwConfigError('option.type is required for option: ' + option);
+    const isField = OPTION_TYPES.indexOf(option.type) === -1;
```
tmeyer2115: Why don't we have FIELD as part of OPTION_TYPES? Seems like it would make this cleaner.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    if (isField && !option.field) {
+      throwConfigError('option.field is required for option: ' + option);
+    } else if (isField && !option.direction) {
+      throwConfigError('option.direction is required for option: ' + option + ', valid values are "ASC" or "DESC"');
+    }
+    option.label = option.label || (isField ? 'FIELD' : option.type);
```
tmeyer2115: I think the default logic here is if label is not present: 
- Display name of the field (i.e. "First Name") if option corresponds to a field. Do we have access to this info? 
- Distance if the type is ENTITY_DISTANCE
- Relevance if the type is RELEVANCE
Not sure what the display name should be for RELEVANCE_RANDOM. May be a question for Rose

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    // Flag for whether all options are shown or if they are hidden behind a show more link
+    this.allShown = false;
+
+    // Component has default option checked on init
+    this.currentlySelectedOption = -1;
+  }
+
+  setState () {
+    let options = this.options;
+    if (this.showMore && !this.allShown) {
```
tmeyer2115: I think things might be clearer if this conditional was just if (this.hideExcessOptions) { blah }. We could replace allShown with hideExcessOptions. The logic for initializing hideExcessOptions would be: 
this.showMore && # of options > limit.
Then we could toggle hideExcessOptions like you have on line 96.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+      label: this.label,
+      allShown: this.allShown,
+      resetLabel: this.resetLabel,
+      optionSelector: this._config.optionSelector
+    }));
+    this.updateSelectedOption();
+  }
+
+  updateSortBys (sortBys) {
+    if (this._searchOnChange) {
```
tmeyer2115: How does the searchOnChange behavior work elsewhere? Do we need to have the this.core.setSortBys call before we call this._search()?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  onMount () {
+    // Handle radio button selections
+    DOM.on(
+      DOM.query(this._container, '#yxt-SortOptions-fieldSet'),
+      'change',
+      evt => {
+        const index = parseInt(evt.target.value);
+        if (index !== -1) {
```
tmeyer2115: What does the value of -1 correspond to? Nothing selected?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  updateSortBys (sortBys) {
+    if (this._searchOnChange) {
+      this._search();
+    }
+    this.core.setSortBys(sortBys);
+  }
+
+  updateSelectedOption () {
```
tmeyer2115: Is it not possible to move this logic to the HBS template? I think there would be a couple ways to do this. One would be to give the options an isSelected property. Another way would be to pass currentlySelectedOption to the template. In the template, you'd conditionally set 'checked'.

src/ui/templates/controls/sortoptions.hbs
```diff
@@ -0,0 +1,43 @@
...
+
+  {{#each options}}
+  <div>
+    <input type="radio" name="yxt-SortOptions-radio" id="yxt-SortOptions-option_{{@index}}" value="{{@index}}" class="{{../optionSelector}}">
+    <label for="yxt-SortOptions-option_{{@index}}">{{this.label}}</label>
+  </div>
+  {{/each}}
+
+  {{#if showMore}}
+    <button class="yxt-SortOptions-showToggle">
```
tmeyer2115: showMoreToggle?

src/ui/sass/modules/_SortOptions.scss
oshi97: nope but luckily looks like filteroptions has identical styling

src/core/core.js
oshi97: yes

oshi97: *Yes we want it no it doesn't do it already

src/ui/components/filters/sortoptionscomponent.js
oshi97: I had a lot of fun adding tests, not quite sure why might need to see some kind of therapist

src/ui/components/filters/sortoptionscomponent.js
oshi97: english hard

src/ui/components/filters/sortoptionscomponent.js
oshi97: was a little worried about "Modifying local variable from inside lambda" but I guess that's just a java issue

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yes it will mutate the input config

src/ui/components/filters/sortoptionscomponent.js
oshi97: Don't think we can get the name of the field without either adding a hardcoded map or making some call, will ask rose about RELEVANCE_RANDOM

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yes sounds good to me

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yeah default value

src/ui/components/filters/sortoptionscomponent.js
oshi97: The only other filter this is implemented on is the geolocationfilter, which runs a search every time after it updates persistent/global storage. We need to call this.core.setSortBys (which just updates global storage) so the search has the updated filter value.

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yeah, it's probably a good thing this came up because it forced me to clean up the default logic to not use -1

src/ui/templates/controls/sortoptions.hbs
oshi97: Think this is fine as it is since it's more for whether or not the option is enabled in config, maybe shouldShowMore?

src/ui/components/filters/sortoptionscomponent.js
tmeyer2115: According to the spec, if it's a field, the default should be the display name. Talk with Rose and see how she wants to proceed, given that we don't have access to that info.

src/core/core.js
```diff
@@ -267,6 +268,19 @@ export default class Core {
       });
   }
 
+  /**
+   * Stores the given sortBy into storage, to be used for the next search
+   * @param {string} namespace the namespace to use for the storage key
+   * @param {Object} sortBy
+   */
+  setSortBy (namespace, sortBy) {
```
tmeyer2115: Is the namespace what is used to link this to a particular SearchBar?

src/core/core.js
```diff
@@ -267,6 +268,19 @@ export default class Core {
...
+    if (sortBy === null) {
+      this.globalStorage.delete(`${StorageKeys.SORT_BY}.${namespace}`);
```
tmeyer2115: Is this behavior we rely on where we set sortBy to null to delete? The other setters in core don't implement this kind of deleting behavior. How do we clear values for those things (such as filter)?

src/core/core.js
```diff
@@ -119,7 +119,8 @@ export default class Core {
         isDynamicFiltersEnabled: this._isDynamicFiltersEnabled,
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
-        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN)
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
+        sortBys: JSON.stringify(this.globalStorage.getAll(StorageKeys.SORT_BY))
```
tmeyer2115: Do we need to scope this using a namespace?

src/core/search/searchapi.js
```diff
@@ -59,7 +59,7 @@ export default class SearchApi {
   }
 
   /** @inheritdoc */
-  verticalSearch (verticalKey, { input, filter, facetFilter, limit, offset, id, geolocation, isDynamicFiltersEnabled, skipSpellCheck, queryTrigger, sessionTrackingEnabled }) {
+  verticalSearch (verticalKey, { input, filter, facetFilter, limit, offset, id, geolocation, isDynamicFiltersEnabled, skipSpellCheck, queryTrigger, sessionTrackingEnabled, sortBys }) {
```
tmeyer2115, line 69: Can we have multiple sortBys here or only the one?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -157,6 +157,13 @@ export default class FilterBoxComponent extends Component {
      */
     this._filters = [];
 
+    /**
```
tmeyer2115: Lets make the integration with Filterbox a separate PR. I think it would be good to break the commits up into: 
- Create a stand-alone SortOptions.
- Integrate SortOptions into Filterbox.
Whenever it makes sense, I think it's good to break things into multiple, smaller commits.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+    this.showReset = this._config.showReset;
+    this.showMore = this._config.showMore && (this.options.length > this._config.showMoreLimit);
+    this.showMoreLabel = this._config.showMoreLabel;
+    this.showLessLabel = this._config.showLessLabel;
+    this.label = this._config.label;
+    this.resetLabel = this._config.resetLabel;
+    this.optionSelector = this._config.optionSelector;
+
+    // Flag for whether all options are shown or if they are hidden behind a show more link
+    this.hideExcessOptions = this.showMore;
```
tmeyer2115: Would the default logic be:
this.hideExcessOptions = this.showMore && # of options > showMoreLimit?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+
+    // Component has default option checked on init
+    const optionIndex = parseInt(this.core.globalStorage.getState(`${this.name}`)) || 0;
```
tmeyer2115: Does there need to be a StorageKeys.SORT_BY here?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+    this.currentlySelectedOption = optionIndex;
```
tmeyer2115: Minor nit, but let's call this selectedOptionIndex or some such thing. Calling it currentlySelectedOption makes me think it's an option object.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+    this.options[optionIndex].isSelected = true;
+
+    // Update selected option in component and persistentStorage
+    const option = this.options[optionIndex];
+    this.currentlySelectedOption = optionIndex;
+    this.core.persistentStorage.set(`${this.name}`, optionIndex);
+
+    // Update sortBys in global storage
+    if (optionIndex === 0) {
+      this.core.setSortBy(this.name, null);
```
tmeyer2115: Do we need to explicitly handle the optionIndex of 0 case? Will it cause problems on the backend if we send the same sortBy as the default sort?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+        direction: option.direction
+      });
+    }
+
+    // If search on change, search
+    if (this._searchOnChange) {
+      this._search();
+    }
+
+    // Run additional on change function, either from user config or FilterBox
```
tmeyer2115: We'll need to add a section to the README about this new component. In that section, we should make clear that the onChange handler is run after the search (if searchOnChange is true).

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,253 @@
...
+    const totalFilter = allFilters.length > 1
+      ? Filter.and(...allFilters)
+      : allFilters[0];
+
+    const query = this.core.globalStorage.getState(StorageKeys.QUERY);
+    const facetFilter = this.core.globalStorage.getAll(StorageKeys.FACET_FILTER)[0];
+    this.core.verticalSearch(this._verticalKey, {
+      input: query,
+      filter: JSON.stringify(totalFilter),
+      facetFilter: JSON.stringify(facetFilter)
```
tmeyer2115, line 123: Do we need to pass the sortBy here too?

src/core/core.js
oshi97: I think they just call delete directly, maybe a deleteSortBy method would be better?

src/core/search/searchapi.js
oshi97: I just wanted to match the request param name, but I do need to change the code to restrict it to just option in the array at a time

src/ui/components/filters/filterboxcomponent.js
oshi97: Sure

src/ui/components/filters/sortoptionscomponent.js
oshi97: This was just to initialize the component from persistent storage, but it seemed like the current pattern in the code was to only use persistent storage to seed global store and then getState off of global storage

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yeah if I send the backend an option with just label or even as an empty { } it returns an error, I could move the massaging to the setSortBy method though?

src/ui/components/filters/sortoptionscomponent.js
oshi97: I wanted to put the getting of the sortBys from global storage inside of verticalSearch instead of outside of it since that doesn't seem like something most components should be worried about

src/ui/components/filters/sortoptionscomponent.js
oshi97: decided to make label required

src/ui/components/filters/filterboxcomponent.js
oshi97: sure

src/ui/components/filters/sortoptionscomponent.js
oshi97: I think if the component can display all options without going over the limit showMore should be false, and no button should be rendered at all

src/ui/templates/controls/sortoptions.hbs
oshi97: Sorry this statement isn't completely correct, right now it's for whether some kind of show more/show less button needs to be rendered at all

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+import StorageKeys from '../../../core/storage/storagekeys';
+import Filter from '../../../core/models/filter';
+
+/**
+ * Renders configuration options for sorting Vertical Results.
+ */
+export default class SortOptionsComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+    this._config = assignDefaults(config);
```
b1lly: can you not do

super(assignDefaults(config), systemConfig));?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    // config needed in template
+    this.defaultSortLabel = this._config.defaultSortLabel;
+    this.options = this._config.options;
+    this.showReset = this._config.showReset;
+    this.showMore = this._config.showMore;
+    this.showMoreLabel = this._config.showMoreLabel;
+    this.showLessLabel = this._config.showLessLabel;
+    this.label = this._config.label;
+    this.resetLabel = this._config.resetLabel;
+    this.optionSelector = this._config.optionSelector;
```
b1lly: Is there a reason these are needed? since they are already part of the config?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,201 @@
...
+      defaultSortLabel: this.defaultSortLabel,
+      options,
+      showReset: this.showReset,
+      showMore: this.showMore,
+      showMoreLabel: this.showMoreLabel,
+      showLessLabel: this.showLessLabel,
+      label: this.label,
+      allShown: this.allShown,
+      resetLabel: this.resetLabel,
+      optionSelector: this._config.optionSelector
```
b1lly: you shouldn't need ANY of this, because the template should already have access to the config object.

In the template, you'd do something like `{{_config.optionSelector}}`

b1lly (changes requested):

src/ui/components/filters/sortoptionscomponent.js
oshi97: after a little thinking I think ideally we would store this in persistent storage as this.name -> array containing necessary args from persistent store, have an additional constructor argument called something like ...persistentState, and then in during ComponentManager.create get the components persistentState array and pass that in

src/core/core.js
oshi97: after talking with rose we decided on not worrying about mutltiple sort options for now

src/ui/components/filters/sortoptionscomponent.js
oshi97: I wanted to be explicit about which fields were used by the template if somebody else came along and tried to understand the code

src/ui/components/filters/sortoptionscomponent.js
oshi97: I didn't want to run any code before calling the inherited constructor but yeah I think I could do that

src/ui/components/filters/sortoptionscomponent.js
oshi97: I wanted to be explicit about which fields were being used by the template

README.md
```diff
@@ -918,6 +919,67 @@ ANSWERS.addComponent('LocationBias', {
 })
 ```
 
+## Sort Options Component
+
+The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search.
+There may only one sort options component per page.
```
tmeyer2115: Currently, there may be only one of these components per page.

src/core/core.js
```diff
@@ -267,6 +268,28 @@ export default class Core {
...
+   */
+  setSortBys (...sortByOptions) {
+    const sortBys = sortByOptions.map(option => {
+      return {
+        type: option.type,
+        field: option.field,
+        direction: option.direction
+      };
+    });
+    this.globalStorage.set(`${StorageKeys.SORT_BYS}`, JSON.stringify(sortBys));
```
tmeyer2115: Don't think you need a template string here.

src/core/core.js
```diff
@@ -267,6 +268,28 @@ export default class Core {
...
+  }
+
+  /**
+   * Clears the sortBys key in global storage.
+   */
+  clearSortBys () {
+    this.globalStorage.delete(`${StorageKeys.SORT_BYS}`);
```
tmeyer2115: Same.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -290,7 +292,7 @@ export default class FilterBoxComponent extends Component {
       ? Filter.and(...allFilters)
       : allFilters[0];
 
-    const query = this.core.globalStorage.getState(StorageKeys.QUERY);
+    const query = this.core.globalStorage.getState(StorageKeys.QUERY) || '';
```
tmeyer2115: Are these changes needed for the standalone SortOptions component?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -50,8 +50,7 @@ class FilterOptionsConfig {
      * If true, show a button to reset the current filter selection
      * @type {boolean}
      */
-    this.showReset = config.showReset || false;
-    this.showReset = this.showReset & this.options.length > 0;
+    this.showReset = config.showReset && this.options.length > 0;
```
tmeyer2115: Same q about being needed for the standalone component

src/ui/components/filters/sortoptionscomponent.js
tmeyer2115: Yeah, lets go with Billy's suggestion. I think that's good practice, we can view assignDefaults as creating a "sane" config. We'd want to pass that "sane" config to the super too.

src/ui/components/filters/sortoptionscomponent.js
tmeyer2115: Having them explicit like this breaks the pattern we follow in our other components (where we access them in the template via _config.). It adds a lot of unnecessary info to our state. Architecturally, I believe things specified in the config should not be duplicated in the state's data. This isn't how all of the components work, but is how the newer ones (such as QuestionSubmission) do. I think this is a pattern we want to follow from now on.

src/ui/components/filters/sortoptionscomponent.js
b1lly: Part of the design is that all configuration options can be used by templates. Regardless of what you're doing here, they're all still available within the template via the config object.

What you're doing HERE is merging configuration options with template data instead of keeping the concerns separate.

By using the mechanism that exists for configuration and templates, you're saving yourself a ton of boilerplate in the class, and helping to keep the concerns separate. I believe it's EASIER to read a template and see anything that is prefixed with `{{_config.foo}}` and understand that its configuration, versus reading `{{foo}}`.

It's also clearer when reading through the class, to understand what properties are configuration, because you'd access them through `this._config.foo`

Take a look at this example:
https://github.com/yext/answers/blob/master/src/ui/components/questions/questionsubmissioncomponent.js

If you follow the pattern there, you can create two objects if you'd like:

```
const DEFAULT_CONFIG = {};

const TEMPLATE_CONFIG = {}
```

That would give the people reading the component the understanding of which ones are for which, even though they'd still all be available everywhere... but at least your being organized and explicit about their purpose in the definition.

b1lly (dismissed):

src/ui/components/filters/sortoptionscomponent.js
tmeyer2115: Ack.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,217 @@
...
+  static get type () {
+    return 'SortOptions';
+  }
+
+  static defaultTemplateName () {
+    return 'controls/sortoptions';
+  }
+}
+
+function throwConfigError (message) {
```
tmeyer2115: Minor nit, but lets move all these functions into the class.

tmeyer2115 (changes requested):

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yeah I'm dumb changing it is way cleaner

src/ui/components/filters/sortoptionscomponent.js
oshi97: yeah it was much cleaner to use _config, thanks :)

src/ui/components/filters/filterboxcomponent.js
oshi97: no, I can put the bugfix in a separate pr

src/ui/components/filters/sortoptionscomponent.js
oshi97: I thought I was doing this._config = assignDefaults(this._config) but I'll still change it to what Billy suggested

src/ui/components/filters/sortoptionscomponent.js
oshi97: honestly this probably shouldn't have a method

oshi97: don't think I can move assignDefaults into the class since I have to call this before super if I do that?

oshi97: 'this' the javascript keyword sorry if that was unclear

README.md
```diff
@@ -28,6 +28,7 @@ Outline:
    - [QASubmission Component](#qa-submission-component)
```
tmeyer2115, line 28: The PR is attempting to merge the SortOptions stuff directly into master. We want to merge this into v0.12.1. I'm not sure if Git lets you edit the target of a PR. May need to make a new one.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -203,7 +203,9 @@ export default class FilterBoxComponent extends Component {
           showReset: this.config.resetFilter,
           resetLabel: this.config.resetFilterLabel,
           showExpand: this.config.expand,
-          onChange: (filter) => {
+          showMore: this.showMore,
```
tmeyer2115: Do we need to make any changes to the filterbox in this PR? I think it would be better to hold off on the integration with FilterBox and add that as a separate PR.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * Renders configuration options for sorting Vertical Results.
+ * TODO: how to deal with multiple instances of this component (and filters in general),
+ * ideally "identical" filters/sorts would be synced up.
+ */
+export default class SortOptionsComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(assignDefaults(config), systemConfig);
+    this.options = this._config.options;
+    console.log(this);
```
tmeyer2115: Remove.

src/ui/sass/modules/_SortOptions.scss
```diff
@@ -0,0 +1,106 @@
...
+
+.yxt-SortOptions {
+  &-container {
+    @include Text($line-height: $line-height-md);
+  }
+
+  &-fieldSet {
+    border: 0;
+    padding: 0;
+    margin: 12px 0;
```
tmeyer2115: We should be trying to use variables as much as possible instead of hardcoding px. Sometimes making something a variable doesn't make a whole lot of sense, but we should always be looking for opportunities to use them. Here, for instance we use 12px in a few different places. Might be good then to create a common size variable. That variable should be related to sizing in _layout.scss.

tmeyer2115: One example of not really needing a var would be 0px.

src/ui/sass/modules/_SortOptions.scss
```diff
@@ -0,0 +1,106 @@
...
+  &-apply {
+    @include Text(
+      $size: $font-size-md,
+      $weight: $font-weight-semibold,
+      $color: white);
+
+    @include Button();
+
+    width: 90px;
+    height: 40px;
```
tmeyer2115: Should we be hard coding width and height here? Does it change on mobile?

src/ui/templates/controls/sortoptions.hbs
```diff
@@ -0,0 +1,57 @@
+<fieldset class='yxt-SortOptions-fieldSet'>
+  <legend class='yxt-SortOptions-legend'>
+    <div class='yxt-SortOptions-legendLabel'>
```
tmeyer2115, line 3: Do we need this inner div or can we just have <legend class=...>{{_config.label}}</legend>?

src/ui/templates/controls/sortoptions.hbs
```diff
@@ -0,0 +1,57 @@
...
+      {{_config.label}}
+    </div>
+  </legend>
+
+  {{#if _config.showReset}}
```
tmeyer2115: Rose wants to only show reset if showReset is true and the default value isn't selected. In the component, then, maybe the showReset we pass in isn't the _config.showReset, but a showReset living in state that's computed according to the way Rose specified.

tmeyer2115 (changes requested):

src/ui/components/filters/filterboxcomponent.js
oshi97: oh this was more of a bugfix but yeah I'll move everything off

README.md
oshi97: updated

src/ui/sass/modules/_SortOptions.scss
oshi97: ok got it

src/ui/sass/modules/_SortOptions.scss
oshi97: I just copied the css from filterbox + filteroptions but probably shouldn't be hard coded

src/ui/templates/controls/sortoptions.hbs
oshi97: there's some styling I think I want to be specific to it

tmeyer2115 (approved):

PR #459, opened 2020-01-17
Answers SDK: Include 1st party cookie when sending user events to Ana…

tmeyer2115:
…lytics.

With this CR, a 1st party cookie ID can be included when sending user events to
the realtime analytics ingester. This behavior is guarded by enabling
conversion tracking, which is an opt-in feature. If opted out, user events will
be sent as before this change, with no 1st party cookie information. Note that
to enable conversion tracking, a user must call the ANSWERS.setConversionOptIn
method and include the ytag library via a <script/> tag in their HTML. This
library was provided by Analytics and is responsible for the cookie generation.

Note that if the user calls setConversionOptIn to enable conversion tracking,
but does not include the <script/> tag, an AnswersAnalyticsError will be thrown
when they attempt to report a user event.

J=SPR-1601
TEST=auto, manual

Updated and added Jest tests. Tested the following scenarios manually:
- Enabled conversion tracking with setConversionOptIn and added the <script/>
  tag. Hit the clear search button multiple times. Each time, I obeserved the
  request payload sent to Analytics for the CLEAR_SEARCH event. The payload
  included the same 1st party cookie ID each time (even after I refreshed the
  page).
- I did not have conversion tracking enabled and issued a CLEAR_SEARCH event.
  The payload did not include any 1st party cookie info.
- I called setConversionOptIn to enable conversion tracking but did not include
  the ytag library via <script/> tag. I saw the AnswersAnalyticsError whenever
  I attempted to send a CLEAR_SEARCH event.


tmeyer2115:
I'm adding an item to the Spruce backlog to update how we use ytag here once the compiled library can be directly copied into the SDK (instead of via script tag).

b1lly:
LGTM. 

As a note, this introduces an external dependency that can cause breakages to the historical versions of the library IF that dependency changes the interface or behavior of the interface.

The external dependency will NOW, based on this change, be tightly coupled to all implementations of Answers from this version forward -- so they need to be ESPECIALLY careful when making changes.

b1lly (approved):

b1lly:
The proper way to handle this is 1. the external library is compiled into the source, or 2. it should be injected through a plugin by the implementer

PR #460, opened 2020-01-17
SearchBar: Give the CSS reset for the input and buttons lower specifi…

tmeyer2115:
…city.

The high specificity of the CSS reset for the SearchBar's input and buttons
was interfering with Consulting's custom styling. This CR lowers the
specificity. The default styling of the input and buttons is still maintained,
but we will no longer clobber Consulting's custom styles.

TEST=manual

Manually verified that the default styling for the SearchBar was preserved.


src/ui/sass/modules/_SearchBar.scss
```diff
@@ -23,15 +23,6 @@
   height: 51px;
   margin-bottom: 0;
 
-  // Targeted CSS reset that prevents any user agent styles from interfering
-  // with styling of SearchBar's buttons and input. Note that this resets CSS styling only for
-  // elements with 'yxt' in the class name. It will not impact any customer styling.
-  button[class*="yxt-"], input[class*="yxt-"]
-  {
-    appearance: none;
```
benmcginnis, line 31: i think we may want to keep the appearance reset here and just move the boarder.  Appearance is a browser styling not a custom one.  I can definitely see why the border would be an issue

benmcginnis: @creotutar thoughts?

tmeyer2115: I can add that back in if you think it would help. Removing it didn't seem to have an impact on the SearchBar's default styling (in Chrome and Safari at least).

benmcginnis: i think it will on mobile safari

tmeyer2115: Added a second commit to this PR which keeps the appearance reset the way it was.

creotutar (approved):
👍 👍 👍 LGTM

Issue #419, opened 2019-12-16
Pass sessionTrackingEnabled Param to Autocomplete Endpoint


rosiegrant:
**Background:** In the SDK, in the `ANSWERS.init`, an implementer can set `sessionTrackingEnabled: false`

However, on our /autocomplete calls, we do not appear to be passing the `sessionTrackingEnabled` param, which means the server is still attempting to set a cookie when it's been set to `false` in the SDK. 

**Desired Behavior**: The `sessionTrackingEnabled` param should be appended to **all** requests, including autocomplete.  

No one is using this param yet, so this is not hugely urgent, but we should still fix ASAP.

rosiegrant:
@data-enabler you had some thoughts on how to implement this, feel free to add them here

data-enabler:
Ideally this should be done at a level where it's not possible to accidentally omit the param from Answers API requests

tmeyer2115:
Resolved by commit #463 in v0.12.1

rosiegrant:
just noting that I tested this on /dev/v0.12.1 and correctly, I do not see a sessionId being set when sessionTrackingEnabled: false 

![image](https://user-images.githubusercontent.com/22160788/73017695-3d68a380-3dee-11ea-8e0e-9caaf8b71255.png)


And when sessionTrackingEnabled: true, confirming i do see the sessionID

<img width="1552" alt="Screen Shot 2020-01-23 at 2 42 10 PM" src="https://user-images.githubusercontent.com/22160788/73017933-97696900-3dee-11ea-9b56-cb40ab4c3715.png">

PR #463, opened 2020-01-21
Pass sessionTrackingEnabled with all non-Analytics API requests

tmeyer2115:
This CR ensures that the sessionTrackingEnabled param is passed along with all non-Analytics API requests issued by the SDK. The ApiRequest class now expects sessionTrackingEnabled as part of the configuration object passed to its constructor. If this value is not present in the config, an error will be thrown. This offers a bit of future proofing by ensuring that anyone adding new
ApiRequests must pass in a sessionTrackingEnabled param. This should resolve issue #419 .

J=ANS-436
TEST=auto, manual

Updated Jest tests. Verified that I saw the sessionTrackingEnabled param with the following requests:
    - Vertical Autocomplete
    - Universal Autocomplete
    - Vertical Search
    - Universal Search
    - Error reporting
    - Q&A submission


data-enabler:
> I removed the setSessionsOptIn method from answers-umd. Users can already set the sessionTrackingEnabled variable in their Answers config and adding the ability to dynamically change this value didn't seem especially valuable.

Was this discussed with Product? At any rate, we shouldn't be changing the public api in a bugfix release.

tmeyer2115:
@data-enabler Added the setSessionsOptIn method back. Passed globalStorage to the various services so that they can pass an up-to-date sessionTrackingEnabled value when constructing ApiRequests.

src/answers-umd.js
```diff
@@ -123,81 +123,79 @@ class Answers {
   }
 
   init (config) {
-    config = Object.assign({}, DEFAULTS, config);
-    if (typeof config.apiKey !== 'string') {
+    const parsedConfig = this.parseConfig(config);
+    if (typeof parsedConfig.apiKey !== 'string') {
```
apokellypse: would recommend the error checking of the config be factored out into a validateConfig method, similar to how you did parseConfig, which was nice.

apokellypse (dismissed):

data-enabler (commented):
Also, prefixing the commit message with "Answers SDK" is pretty redundant.

src/core/http/apirequest.js
```diff
@@ -24,6 +25,16 @@ export default class ApiRequest {
...
 
+    /**
+     * A boolean indicating if sessionTracking is enabled
+     * @type {boolean}
+     * @private
+     */
+    if (typeof opts.sessionTrackingEnabled !== 'boolean') {
+      throw new AnswersBasicError('Must indicate if session tracking is enabled', 'ApiRequest');
+    }
+    this._sessionTrackingEnabled = opts.sessionTrackingEnabled;
```
data-enabler: When I was talking about passing globalStorage through, I meant passing it all the way through to ApiRequest. It's not a very clean solution either way, but at least we won't have boilerplate all over the place.

Also, we should leave a TODO somewhere in this class to make it clear that this isn't really a desired state, and we want to move to something else like a service interface with dependency injection.

src/answers-umd.js
```diff
@@ -226,6 +224,21 @@ class Answers {
...
 
+  /**
+   * Parses the config provided by the user. In the parsed config, any options not supplied by the
+   * user are given default values.
+   * @param {Object} config The user supplied config.
+   */
+  parseConfig (config) {
+    const parsedConfig = Object.assign({}, DEFAULTS, config);
+    let sessionTrackingEnabled = true;
+    if (typeof config.sessionTrackingEnabled === 'boolean') {
```
data-enabler, line 229: This is fine for now, but In the long term we should probably be validating the config with JSON Schema or something, so we'll be able to use Object.assign to handle the defaults for this as well.

data-enabler (changes requested):

src/answers-umd.js
tmeyer2115: Done.

src/core/http/apirequest.js
tmeyer2115: Done.

src/core/http/apirequest.js
```diff
@@ -62,6 +53,18 @@ export default class ApiRequest {
...
      */
     this._params = opts.params || {};
+
+    // TOOD (tmeyer): Create an interface for fetching sitesTrackingEnabled. Inject an
+    // implementation of that interface here.
+    /**
+     * The global storage of the Answers experience
+     * @type {GlobalStorage}
+     * @private
+     */
```
data-enabler, line 65: JSDoc needs to go immediately before the declaration. I would also get rid of the superfluous description.

src/core/http/apirequest.js
```diff
@@ -62,6 +53,18 @@ export default class ApiRequest {
      * @private
      */
     this._params = opts.params || {};
+
+    // TOOD (tmeyer): Create an interface for fetching sitesTrackingEnabled. Inject an
+    // implementation of that interface here.
```
data-enabler: Not really what I meant. An `ApiService` implementation should get passed as a dependency to `ErrorReporter`, `AutoCompleteApi`, etc. It's fine for the `ApiService` implementation to have a reference to `GlobalStorage`.

data-enabler (changes requested):

data-enabler (approved):

PR #464, opened 2020-01-23
Add support for running the SDK against Sandbox.

tmeyer2115:
This CR enables the SDK to be run against Sandbox or Production. The
requirement for running against Sandbox came from the Hitchhikers. A notion of
environment was added to the SDK config. Ultimately, the environment is just
used to determine which version of the various endpoints to hit. By default,
the environment is assumed to be Production. To indicate that the SDK should
be run against Sandbox, a hitchhiker will need to preface their apiKey with
'sandbox-'. This functionality was a product request.

This commit further highlights the need for an ApiService, an implementation of
which can be injected into the current consumers of ApiRequest. An item has
been added to the Spruce backlog for this.

J=SPR-1619
TEST=auto, manual

Updated Jest tests. Made sure I could hit the following endpoints successfully
when running against Production or Sandbox:
     - Vertical Autocomplete
     - Vertical Search
     - Universal Autocomplete
     - Universal Search
     - Question submission
     - Analytics reporting (with and without conversion tracking enabled)
     - Error reporting


oshi97 (approved):
lgtm I originally thought we would do the environment at the time of request but this is probably better

data-enabler (approved):

Issue #456, opened 2020-01-16
Reset Css for SearchBar / GeoLocationFilter High Specificity 


creotutar:
On v0.12.0

Trying to style the GeoLocationFilter search bar input with the following class selector `yxt-GeoLocationFilter-input` but the styling is overridden by the reset css `.yxt-GeoLocationFilter input[class*=yxt-]` (e.g. when trying to add a border to a search bar). This is a change that would affect all css for sites upgrading from pre v0.12.0 to v0.12.0 if they are overridden in the reset css.

We can match specificity of the reset css with `.yxt-GeoLocationFilter input[class*=yxt-]` but that's not clear or super consumer friendly. Could we have the reset css have specificity lower than the single element classes like `yxt-GeoLocationFilter-input`? Or do we suggest another way of overriding the reset css?

tmeyer2115:
This should be resolved by #460

creotutar:
hey @tmeyer2115 this is also happening for yxt-Results-filters

is it possible to have styling changes be reset generally in a less specific manner?

<img width="273" alt="Screen Shot 2020-01-21 at 2 34 00 PM" src="https://user-images.githubusercontent.com/32468132/72836670-175eca00-3c5b-11ea-82da-9fd52e9e7f95.png">

tmeyer2115:
@creotutar Yep, should be possible. I can add those resets as part of a single element class's styling. Seems like we shouldn't be trying to do any global, high specificity resets like that as it causes problems for custom styling.

tmeyer2115:
Fixed in v0.12.1.

PR #465, opened 2020-01-23
Remove the global SCSS reset for Answers.

tmeyer2115:
This CR removes _reset.scss which contained global CSS resets for Answers. The specificity of these resets was too high and was clobbering the custom styling from Consulting. Initially, these resets were added during the default styling of the SearchBar and AutoComplete components. The resets needed for the default styling of the AutoComplete were moved to the SCSS file for that component and given lower specificity. The remaining resets contained in _reset.scss were no longer needed and so the file could be safely deleted. The net result of this change is that the default styling for the AutoComplete component is preserved and any custom styling applied by consulting won't be clobbered. This resolves #456 .

TEST=manual

Verified that the default styling of AutoComplete was preserved.


src/ui/sass/_reset.scss
```diff
@@ -1,15 +0,0 @@
-// This file performs a targeted CSS reset that prevents any user agent styles from interfering
-// with Answers styling. Note that this resets CSS styling only for elements with 'yxt' in the
-// class name. It will not reset any customer styling.
-
-[class*="yxt-"], [class*="yxt-"]::before, [class*="yxt-"]::after
-{
-  box-sizing: border-box;
```
creotutar, line 7: hmm, i do like having `box-sizing: border-box` but i do not know how we would accomplish having this for our elements (as it isn't inherited)

creotutar (approved):
🥁 🥁 🥁 LGTM

PR #475, opened 2020-01-30
Tests: Add tests for search data transformer current behavior

alexisgrow:


tests/core/search/searchdatatransformer.js
```diff
@@ -0,0 +1,58 @@
...
+import Navigation from '../../../src/core/models/navigation';
+import VerticalResults from '../../../src/core/models/verticalresults';
+import SpellCheck from '../../../src/core/models/spellcheck';
+import DynamicFilters from '../../../src/core/models/dynamicfilters';
+import SearchIntents from '../../../src/core/models/searchintents';
+import LocationBias from '../../../src/core/models/locationbias';
+import ResponseNoResults from '../../fixtures/responseNoResults.json';
+import ResponseWithResults from '../../fixtures/responseWithResults.json';
+
+jest.mock('../../../src/core/http/httprequester');
```
tmeyer2115: Do we need to mock the httprequester here? The transformer's not issuing network calls, is it?

tests/core/search/searchdatatransformer.js
```diff
@@ -0,0 +1,58 @@
...
+  beforeEach(() => {
+    responseNoResults = ResponseNoResults;
+    responseWithResults = ResponseWithResults;
+  });
+
+  it('vertical search response with data without formatters', () => {
+    const data = responseWithResults;
+    const formatters = [];
+    const result = SearchDataTransformer.transformVertical(data, formatters);
+    expect.assertions(1);
```
tmeyer2115: expect.assertions is usually only needed when testing async code: https://jestjs.io/docs/en/expect#expectassertionsnumber
I don't think we need it here.

tests/core/search/searchdatatransformer.js
```diff
@@ -0,0 +1,58 @@
...
+        [StorageKeys.VERTICAL_RESULTS]: VerticalResults.from(data.response, formatters),
+        [StorageKeys.DYNAMIC_FILTERS]: DynamicFilters.from(data.response),
+        [StorageKeys.INTENTS]: SearchIntents.from(data.response.searchIntents),
+        [StorageKeys.SPELL_CHECK]: SpellCheck.from(data.response.spellCheck),
+        [StorageKeys.LOCATION_BIAS]: LocationBias.from(data.response.locationBias)
+      }
+    );
+  });
+
+  it('vertical search response no data without formatters', () => {
```
tmeyer2115: I think just the first test will be sufficient for SearchDataTransformer. We shouldn't need to worry about coverage for the no results case here. That will be covered in the test for VerticalResults.from. In this test, we'll assume that all the various .from methods work as expected with whatever input they're given.

tmeyer2115: I think we can just have two tests here: it('transforms vertical response correctly') (with exactly what you have above) and it('transforms universal response correctly') (which we can add later on).

tests/core/search/searchdatatransformer.js
alexisgrow: You're right, we don't need it! removing

tests/core/search/searchdatatransformer.js
alexisgrow: Ah true oof, I think I was following an example that made a request, makes sense. Removed!

tests/core/search/searchdatatransformer.js
alexisgrow: Oh sweet makes sense, updated

tmeyer2115 (approved):

PR #477, opened 2020-01-30
Standard card + Vertical Results

oshi97:
Implement StandardCard and usage of card templates in VerticalResults

TEST=manual, auto
Tested with local site, created bare minimum unit tests


src/answers-umd.js
```diff
@@ -92,6 +93,11 @@ class Answers {
      */
     this._onReady = function () {};
 
+    /**
+     * @type {Object}
+     */
+    this.cardTemplates = cardTemplates;
```
tmeyer2115, line 99: Is this variable needed by answers-umd?

src/ui/components/cards/ctas.js
```diff
@@ -0,0 +1,68 @@
...
+  newWindow: false,
+
+  /**
+   * The list of eventOptions needed for the event to fire.
+   * @type {Object}
+   */
+  eventOptions: null
+};
+
+export default function calculateCTAs (configCTAs, data) {
```
tmeyer2115, line 39: Add some documentation to this method to explain it's role, the inputs, output, etc.

src/ui/components/cards/ctas.js
```diff
@@ -0,0 +1,68 @@
...
+  if (!configCTAs) {
+    return null;
+  }
+  const callsToAction = configCTAs.map(cta => {
+    if (typeof cta === 'function') {
```
tmeyer2115, line 44: What would a cta of type function look like? I don't see that included in the spec (just CTAs defined by a field name or an object containing mappings).

src/ui/components/cards/ctas.js
```diff
@@ -0,0 +1,68 @@
...
+      return Object.assign({}, CTA_DEFAULTS, cta(data));
+    } else if (typeof cta === 'object') {
+      const calculatedCTA = {};
+      Object.entries(cta).forEach(([key, value]) => {
+        if (typeof value === 'function') {
+          try {
+            calculatedCTA[key] = value(data);
```
tmeyer2115, line 51: What is data here? The profile data?

src/ui/components/cards/ctas.js
```diff
@@ -0,0 +1,68 @@
...
+          } catch (err) {
+            console.debug('StandardCard - error parsing cta: ', value);
```
tmeyer2115, line 53: I don't think we want these console debug statements. It feels like we should return one of our Answers JS errors here instead of a console.error.

src/ui/components/cards/ctas.js
```diff
@@ -0,0 +1,68 @@
+const CTA_DEFAULTS = {
```
tmeyer2115, line 1: What would you think of making CTAs an actual component class? In my head, I thought CTAs would be their own component. We could create a CTABuilder interface that exposes a method mapping profile data to a list of CTAs. Right now, we'd have one implementation of that interface corresponding to the callsToAction property specified here: https://docs.google.com/document/d/1_iisd_3DoHfyLcn-1K9l8bZt8fq2pf6nyInehyndEqE/edit#heading=h.todouj689k0h. This would be a bit like the method you have below.

src/ui/components/cards/consts.js
```diff
@@ -0,0 +1,16 @@
+// Uncomment others when implemented
+export const cardTemplates = {
```
tmeyer2115, line 2: Could we not just add StandardCard directly to the registry? The parent component instantiating would now how to parse its config to determine the correct component type for the cards. It would then just call addChild. Let me know if that doesn't make sense.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+  showMoreLimit: null,
+
+  // Shows to the left of the result
+  image: null,
+
+  // subtitle
+  subtitle: null,
+
+  // Template mappings
+  templateMappings: () => {}
```
tmeyer2115, line 28: What are the templateMappings? I don't see those referenced in the spec with the other Standard card template options.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+/**
+ * Card components expect to receive a data config option, containing data regarding entity result
+ * each card is assigned to, including all field data in data._raw.
+ */
+export default class StandardCardComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(Object.assign({}, DEFAULT_CONFIG, config), systemConfig);
+    const data = this._config.data || {};
```
tmeyer2115, line 38: Is this profile data? If so, let's rename the variable. The names data and raw are very general.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+    this.details = this._config.details || data.details || raw.description || '';
+
+    // If populated, link out the title field.
+    this.url = this._config.url || data.link || raw.website;
+
+    /**
+     * A local reference to the analytics reporter, used to report events for this component
+     * @type {AnalyticsReporter}
+     */
+    this.analyticsReporter = systemConfig.analyticsReporter || null;
```
tmeyer2115, line 56: All instances of component should already have a this.analyticsReporter.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+
+    this.showToggle = !!(this._config.showMoreLimit && (this.details.length > this._config.showMoreLimit));
```
tmeyer2115, line 58: There are two '!' marks. Does that mean there shouldn't be any?

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+    this.hideExcessDetails = this.showToggle;
+  }
+
+  setState (data) {
+    let details = this.details;
+    if (this.hideExcessDetails) {
+      details = details.substring(0, this._config.showMoreLimit);
+      details = details.concat('...');
```
tmeyer2115, line 66: Use a template literal for this

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,132 @@
...
+    if (this.showToggle) {
+      const el = DOM.query(this._container, '.yxt-StandardCard-showToggle');
+      DOM.on(el, 'click', () => {
+        this.hideExcessDetails = !this.hideExcessDetails;
+        this.setState();
+      });
+    }
+  }
+
+  bindCTA () {
```
tmeyer2115, line 88: One nice thing about creating a component class for CTAs is that this logic would live there.

src/ui/components/registry.js
```diff
@@ -61,15 +56,13 @@ const COMPONENT_CLASS_LIST = [
   // Results Components
   DirectAnswerComponent,
   UniversalResultsComponent,
-  ResultsComponent,
+  VerticalResultsComponent,
```
tmeyer2115, line 59: Is this the new version of ResultsComponent? Is it basically the same except for accepting card templates? Should removing ResultsComponent be part of this CR?

tmeyer2115 (changes requested):

src/answers-umd.js
oshi97: This is here so that the cardTemplates can be accessed in an enum-like way inside ANSWERS.init. I originally intended to do a static getter, but I think the way we made ANSWERS a singleton made it so static methods cannot be called. would something like

`const ANSWERS = new Answers();`
`ANSWERS.cardTemplates = cardTemplates;`
`export default ANSWERS;`

be better?

oshi97:
shelving this pull request for a cleaner workflow

PR #478, opened 2020-01-30
patch filter options storeOnChange

oshi97:
was not setting default to true


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -44,7 +44,7 @@ class FilterOptionsConfig {
      * If true, stores the filter to storage on each change
      * @type {boolean}
      */
-    this.storeOnChange = config.storeOnChange || false;
+    this._storeOnChange = config.storeOnChange === undefined ? true : config.storeOnChange;
```
tmeyer2115: The convention in JS is that things prefixed with an '_' are private by convention (although still functionally public, haha): https://stackoverflow.com/questions/4484424/underscore-prefix-for-property-and-method-names-in-javascript
Because this this.config.storeOnChange is used by the FilterOptionsComponent, it should be considered public.

oshi97: Oh yeah I changed this in the second commit already after reading the PR again

tmeyer2115 (approved):

PR #479, opened 2020-01-30
Add label field to filters, move filter fetching into core

oshi97:
This commit adds a label metadata to Filters (and by extension Facets), to be used for showAppliedFilters in VerticalResults. In the future if we were to add additional metadata it might be good to have that in a single field, and then ignore just that field when stringifying the filter/facet.


oshi97:
Was not aware applied filters were returned by the backend

PR #480, opened 2020-01-31
Remove deprecated item templates

oshi97:


src/ui/components/registry.js
```diff
@@ -20,15 +20,10 @@ import SortOptionsComponent from './filters/sortoptionscomponent';
 
 import DirectAnswerComponent from './results/directanswercomponent';
 import ResultsComponent from './results/resultscomponent';
-import AccordionResultsComponent from './results/accordionresultscomponent.js';
```
tmeyer2115: I don't think we want to get rid of the AccordionResultsComponent (at least not just yet). Billy said the other three were deprecated and that the ResultsComponent should only use the generic ResultsItemComponent. It doesn't seem like these templates are ever used (or should be used), so removing them doesn't pose a problem. 

It seems like AccordionResultsComponent is only used by UniversalResults. Billy didn't mention if that was deprecated. Looking at the code, it seems like someone could easily configure UniversalResults to use this component. So, removing it could negatively impact them. Additionally, in this feature branch, we won't be updating UniversalResults (that will be a later item) to use sections/card templates. If we were, I think it would be fine to remove this now, since they could replicate the same functionality by choosing the appropriate card template.

src/ui/components/results/resultscomponent.js
```diff
@@ -35,37 +25,6 @@ export default class ResultsComponent extends Component {
     this._isUniversal = config.isUniversal || false;
 
     this.moduleId = StorageKeys.VERTICAL_RESULTS;
-    this._itemConfig = {
```
tmeyer2115, line 38: Were we allegedly supporting the ability to supply a custom render function for the items? Was this actually supported in the code? If so, we should check with Billy about removing the renderItem functionality as well. He just said that the other built-in templates were deprecated, he didn't say if renderItem was deprecated too.

src/ui/components/results/resultscomponent.js
```diff
@@ -121,64 +80,12 @@ export default class ResultsComponent extends Component {
    * @returns {string}
    * @override
    */
-  static defaultTemplateName (config) {
```
tmeyer2115: I think you need to leave config as an arg since this overrides a method from the Component class.

src/ui/components/results/resultscomponent.js
```diff
@@ -121,64 +80,12 @@ export default class ResultsComponent extends Component {
...
-    this._itemConfig[clazz.type].template = template;
-  }
-
-  setItemRender (type, render) {
-    let clazz = this.getItemComponent(type);
-    this._itemConfig[clazz.type].render = render;
-  }
-
-  getItemComponent (type) {
-    let comp = ResultsItemComponent;
```
tmeyer2115, line 157: In terms of testing this change, did you render VerticalResults and just see it use ResultsItemComponent?

oshi97: Yes

src/ui/components/results/resultscomponent.js
oshi97: Added back

src/ui/components/results/resultscomponent.js
oshi97: Done

src/ui/components/results/resultscomponent.js
oshi97: On second thought, shouldn't this kind of functionality be the card's responsibility rather than vertical results'?

tmeyer2115 (commented):
Looks like these changes may be causing a test to fail. I think it would also be good to add to the README showing how to use renderItem and itemTemplate.

tmeyer2115 (approved):

PR #483, opened 2020-02-03
feature/vertical-cards: rename results -> verticalresults

oshi97:
TEST=manual
Test that vertical results and universal results still load


src/ui/sass/modules/_VerticalResults.scss
```diff
@@ -1,5 +1,5 @@
-/** @define Results */
-.yxt-Results
+/** @define VerticalResults */
+.yxt-VerticalResults
```
tmeyer2115: Here and in verticalresults.hbs, I don't think it would be the end of the world if we left the class names as is. The more confusing things are the file names. By not changing the class names, we'll also be helping consulting a bit. They won't have to update their custom CSS styling.

tmeyer2115 (changes requested):
We should update the README so that any references to ResultsComponent are also renamed.

src/ui/sass/modules/_VerticalResults.scss
oshi97: I think consulting will still have to update their css since the html is (probably) going to have a different structure in the next few commits. In this case would it still make their life easier to keep the top level class name the same? If not I would still prefer to change it even though it's a really minor thing

tmeyer2115 (approved):

PR #485, opened 2020-02-03
refactor ResultItemComponent to Cards

oshi97:
following commits will change implementation


src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,75 @@
...
+
+import Component from '../component';
+import { cardTemplates, templateToType } from './consts';
+
+export default class StandardCardComponent extends Component {
+  constructor (opts = {}, systemConfig = {}) {
+    super(opts, systemConfig);
+
+    /**
+     * verticalConfigId used for analytics and passed to children
```
tmeyer2115: Remove "and passed to children". This component has no children, right?

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,75 @@
...
+     * @type {string}
+     * @private
+     */
+    this._verticalConfigId = opts.verticalConfigId;
+
+    /**
+     * isUniversal is used for analytics and passed to children and is set to
+     * true if this component is added by the UniversalResultsComponent
```
tmeyer2115, line 19: Same thing about referring to children in the docs.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,75 @@
...
+     * @type {boolean}
+     * @private
+     */
+    this._isUniversal = opts.isUniversal || false;
+  }
+
+  static get type () {
+    return templateToType[this.defaultTemplateName()];
```
tmeyer2115: Could we flip this logic around? We could hardcode the string 'StandardCard' here and refer to the static method in the consts.js map.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * The template to render
+   * @returns {string}
+   * @override
+   */
+  static defaultTemplateName (config) {
+    return cardTemplates.Standard;
```
tmeyer2115, line 36: Same thing about hard coding the string here and using that in the consts.js map.

src/ui/components/cards/standardcardcomponent.js
oshi97: Changed to cardTypes and cardTemplates I think doing two references like I was doing is a little silly

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,47 @@
+/** @module StandardCardComponent */
```
tmeyer2115: Update.

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,47 @@
...
+    // we can't just pass all config because it will overwrite things like
+    // parentContainer and name
+    const childConfig = {
+      verticalConfigId: this._config._verticalConfigId,
+      isUniversal: this._config._isUniversal
+    };
+    return super.setState(Object.assign({}, data, {
+      childConfig: JSON.stringify(childConfig),
+      cardType: cardTypes.Standard,
+      data
```
tmeyer2115: Do we need data here? Is it not enough to have data as the 2nd arg to Object.assign?

src/ui/components/registry.js
```diff
@@ -59,7 +60,8 @@ const COMPONENT_CLASS_LIST = [
   UniversalResultsComponent,
   VerticalResultsComponent,
   PaginationComponent,
-  ResultsItemComponent,
+  StandardCardComponent,
```
tmeyer2115: Minor organizational nit, but lets create a new section in this list for the various card components:
// Card Templates
CardComponent,
StandardCardComponent,
AccordionCardComponet,
...

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,85 @@
...
+      Object.assign(this, templateMappings(data));
+    } catch (err) {
+      console.error(err);
+    }
+  }
+}
+
+export default class CardComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
```
tmeyer2115: I think we want to follow the pattern set by QuestionSubmissionComponent. We build what we want the config object to be, then pass it into super(config, systemConfig). This automatically sets this._config for us and ensures consistency between the config provided to super(...) and the config we use in this class. If we call super first, then make changes to config and manually set this._config, I worry it could have some weird side effects. The super class has a different config than this class.

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,85 @@
...
+     * TODO(oshi) also something a Component might want in general
+     * @type {string|undefined}
+     */
+    this.itemTemplate = config.itemTemplate;
+
+    /**
+     * Template mappings is a function specified in the config
+     * that returns config based on the data passed into card
+     */
+    const data = config.data;
```
tmeyer2115: Could we update this doc? Not sure what data is exactly. Don't see it listed in the spec. Are the attributes in template mapping what we see on the card?

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,85 @@
...
+    const templateMappings = config.templateMappings || (() => {});
+    try {
+      Object.assign(this, templateMappings(data));
+    } catch (err) {
+      console.error(err);
+    }
+  }
+}
+
+export default class CardComponent extends Component {
```
tmeyer2115, line 32: How are templateMappings used in this class?

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,85 @@
...
+   */
+  static defaultTemplateName (config) {
+    return 'cards/card';
+  }
+
+  static areDuplicateNamesAllowed () {
+    return true;
+  }
+
+  addChild (data, type, opts) {
```
tmeyer2115: It seems like we could do things in such a way that we don't have to override addChild. We can just manually build opts (should just be two attributes). I share your aversion to overriding addChild, so I think we should avoid it here, when we can.

tmeyer2115 (changes requested):

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Template mappings is a function specified in the config
+     * that returns config based on the data passed into card
+     */
+    const data = config.data;
+    const templateMappings = config.templateMappings || (() => {});
+    try {
+      Object.assign(this, templateMappings(data));
+    } catch (err) {
+      console.error(err);
```
b1lly: Whats the point of try->catch->console log?

If you care about the error, instead of console logging it, you should throw a new AnswersError here.... probably AnswersComponentError to be specific, or AnswersUiError.

e.g.

```
try {
   ...
} catch (e) {
  throw new AnswersComponentError('CardComponent: Invalid template mappings provided', e);
}
```

OR, you can provide specific details 


https://github.com/yext/answers/blob/master/src/core/errors/errors.js

By throwing an `AnswersError` it'll integrate more effectively with the error reporter.

b1lly (changes requested):

src/ui/components/cards/cardcomponent.js
oshi97: I did it this way because I don't think the card component should ever have any children besides it's one child card, and putting the opts as stringified into data-opts puts all into the html which is pretty hideous by its own respect. I'll change it but we definitely need to refactor how we pass config to children.

src/ui/components/cards/cardcomponent.js
oshi97: I can move them to the next commit

src/ui/components/cards/cardcomponent.js
oshi97: How I understood it was config.data is passed in from data-prop in vertical results. In this case we expect it to be profile data for the result this card cares about. This gets passed into templateMappings to construct the config for the card.

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,78 @@
...
+     * Custom render method
+     * @type {function|undefined}
+     */
+    this.renderItem = config.renderItem;
+
+    /**
+     * Custom template string
+     * @type {string|undefined}
+     */
+    this.itemTemplate = config.itemTemplate;
```
tmeyer2115: If you do: 
this.render = config.renderItem
this.template = config.template
You can save yourself the call to setRender and setTemplate below.

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    if (typeof this._config.itemTemplate === 'string') {
+      this.setTemplate(this._config.itemTemplate);
+    }
+  }
+
+  setState (data) {
+    return super.setState(Object.assign({}, data, {
+      cardType: this._config.cardType,
+      cardOpts: JSON.stringify(new CardConfig(this._config))
```
tmeyer2115: This is a minor thing, but let's be explicit here about what cardOpts actually contains. It just needs verticalConfigId and isUniversal, right? I think we should just JSON.stringify and Object containing those. Then, it will be clear to someone reading the code what we're passing down. We shouldn't pass it a config with more than it needs.

tmeyer2115 (approved):

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -111,21 +116,17 @@ export default class VerticalResultsComponent extends Component {
...
-    }));
-
-    if (typeof this._config.renderItem === 'function') {
-      comp.setRender(this._config.renderItem);
-    }
-    if (typeof this._config.itemTemplate === 'string') {
-      comp.setTemplate(this._config.itemTemplate);
-    }
-    return comp;
+    return super.addChild(data, type, opts);
```
b1lly, line 130: `opts` is an old name. Can we rename it to `config` everywhere so that its consistent? That way we avoid confusion, and we always know what we're actually dealing with.

src/ui/components/cards/cardcomponent.js
```diff
@@ -0,0 +1,62 @@
...
+  constructor (config = {}, systemConfig = {}) {
+    super(new CardConfig(config), systemConfig);
+  }
+
+  setState (data) {
+    const { verticalConfigId, isUniversal } = this._config;
+    return super.setState(Object.assign({}, data, {
+      cardType: this._config.cardType,
+      cardOpts: JSON.stringify({ verticalConfigId, isUniversal }),
+      cardData: data
```
b1lly, line 42: I'd probably do most of this through configuration and component-markup-syntax.

Something like this:

```
this._config = Object.assign({}, this._config, {
   cardType: ...,
   childConfig: { 
      verticalConfigId,
      isUniversal
   }
});
```

And then the template becomes:

```
<div data-component="{{_config.cardType}}"
  data-prop="cardData"
  data-opts="{{_config.childConfig}}">
</div> 
```
Also note my other comment. I'd like to rename `data-opts`, to `data-config` and standardize the naming to keep it consistent. `opts` was the old name. 

As you work through areas of the code with the legacy name, might be worth updating it! That way you don't have to update it everywhere at once.

Separately, for child components and configuration, i'd probably what i'd probably think about is a better architectural pattern that makes configuration flow more easily to child components without having to do things like this. Or define a pattern/practice for how to do it thats consistent.

PR #486, opened 2020-02-04
Security template updates

alexisgrow:


oshi97 (commented):
Looks good to me! Only thing I'm not sure about is if whether it's necessary to add the SafeString method but I'm not really an expert on code cleanliness

src/ui/components/icons/iconcomponent.js
```diff
@@ -80,7 +54,8 @@ export default class IconComponent extends Component {
    */
   setState (data) {
     return super.setState(Object.assign(data, {
-      image: this.image,
+      iconUrl: this.iconUrl,
+      iconName: this.iconName ? this.iconName : 'default',
```
tmeyer2115: Given the default on line 19, I think we can skip the defaulting behavior here.

tmeyer2115 (commented):
Looks great! Are there any stray references to customIcon in the README? If so, we should remove those.

src/ui/components/icons/iconcomponent.js
alexisgrow: Oh true, updating

tmeyer2115 (approved):

PR #487, opened 2020-02-05
StandardCard Styling

oshi97:
This card updates the styling for the standard card to use
the new spec instead of the original ResultsItemComponent
styling. It adds a ShowMoreComponent to separate some logic.

TEST=manual

I don't know if this ShowMoreComponent actually helps that much, it seemed nice in theory so I put it in the CR just so people could take a look at what it would be like.


src/ui/components/cards/cardcomponent.js
```diff
@@ -35,12 +40,12 @@ export default class CardComponent extends Component {
   }
 
   setState (data) {
-    const { verticalConfigId, isUniversal } = this._config;
-    return super.setState(Object.assign({}, data, {
+    return super.setState({
+      ...data,
```
tmeyer2115, line 38: Seems kind of strange that we are essentially including data twice here. We have all it's attributes here and cardData: data below. Can we consolidate?

src/ui/components/cards/cardcomponent.js
```diff
@@ -14,18 +14,23 @@ class CardConfig {
...
     /**
-     * isUniversal is set to true if this component is added by the UniversalResultsComponent
-     * @type {boolean}
-     * @private
+     * config.data comes from the data-prop attribute passed in
+     * from the parent component.
+     * @type {Object}
      */
-    this.isUniversal = config.isUniversal || false;
+    this.data = config.data || {};
```
tmeyer2115: I'm a bit confused. Is this data the actual result data used to populate the card? If this is passed in via the data-prop attribute, why is it available in the config? I though only stuff in data-opts was available via the config.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -3,73 +3,92 @@
...
 
-export default class StandardCardComponent extends Component {
-  constructor (config = {}, systemConfig = {}) {
-    super(config, systemConfig);
+class StandardCardConfig {
+  constructor (config = {}) {
+    Object.assign(this, config);
+
+    const data = config.data || {};
+    const raw = data._raw || {};
```
tmeyer2115: What is the difference between data and raw? Is that something we get from LiveAPI?

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -3,73 +3,92 @@
...
+    this.showMoreLimit = config.showMoreLimit;
 
     /**
-     * verticalConfigId used for analytics
+     * Image url to display
      * @type {string}
-     * @private
      */
-    this.verticalConfigId = config.verticalConfigId;
+    this.image = config.image;
```
tmeyer2115: It feels a bit weird calling all these things configuration options. They feel more like the state and the templateMappings are the card's configuration option.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -3,73 +3,92 @@
...
+/**
+ * Card components expect to receive a data config option, containing data regarding entity result
+ * each card is assigned to, including all field data in data._raw.
+ */
+export default class StandardCardComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(new StandardCardConfig(config), systemConfig);
+  }
+
+  setState (data) {
```
tmeyer2115, line 99: If we have templateMappings in the config for StandardCard, could we have Card pass the result data (item) as data-prop to StandardCard? We could then make a method here mapping the item data to title, subtitle, details, etc.

src/ui/components/helper/showmorecomponent.js
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Whether a 'show more' toggle button needs to be rendered at all
+     */
+    const detailsOverLimit = this.details.length > this.showMoreLimit;
+    this.showToggle = this.showMoreLimit && detailsOverLimit;
+  }
+}
+
+/**
+ * A simple component that displays text and has a show more/less button
```
tmeyer2115: Would this be shared across multiple card templates?

src/ui/components/cards/standardcardcomponent.js
oshi97: I think we do some processing on our side to produce `data`, and raw is the raw data from the api

src/ui/components/cards/cardcomponent.js
oshi97: data-prop places the referenced data inside a 'data' attribute in the config

src/ui/components/cards/cardcomponent.js
oshi97: I think it's difficult because the first ...data is expected by setState and the second is purely to pass data into the child, and we also don't want to include cardType and cardOpts in the child's data

src/ui/components/cards/standardcardcomponent.js
oshi97: The way I did it was to have CardComponent process templateMappings into these config options

src/ui/components/cards/standardcardcomponent.js
oshi97: That would require instantiating the child component in javascript, which I think is fine but just wanted to let you know since neither way is perfect

src/ui/components/helper/showmorecomponent.js
oshi97: I'm just going to remove this it's not as helpful as I hoped

src/ui/components/cards/cardcomponent.js
tmeyer2115: Great, that makes so much sense 🙄. Could we call the variable something more descriptive here though? "Data" doesn't really indicate what this is/does.

src/ui/components/cards/cardcomponent.js
tmeyer2115: Got it. I need to get used to this pattern, but man can it be confusing.

src/ui/components/cards/cardcomponent.js
```diff
@@ -35,12 +34,18 @@ export default class CardComponent extends Component {
...
-      cardOpts: JSON.stringify({ verticalConfigId, isUniversal }),
-      cardData: data
-    }));
+      result: this._config.data
+    });
+  }
+
+  addChild (data, type, opts) {
+    return super.addChild(data, type, {
+      templateMappings: this._config.templateMappings,
```
tmeyer2115, line 46: Is it basically the case that we always have to override addChild if the child's config would include some kind of function that the parent needs to pass down?

src/ui/templates/cards/standard.hbs
```diff
@@ -1,87 +1,31 @@
...
-  <div class="yxt-Result-wrapper">
-    {{#if _config._parentOpts.showOrdinal}}
-      <div class="yxt-Result-ordinalWrapper">
-        <div class="yxt-Result-ordinal">
-          {{ordinal}}
-        </div>
-      </div>
+<div class='yxt-StandardCard'>
+  {{#if _config.image}}
+    <div style="background-image: url('{{_config.image}}');" class='yxt-StandardCard-image'></div>
```
tmeyer2115: What was the reasoning again for this not being an  tag?

src/ui/sass/modules/_Results.scss
```diff
@@ -133,4 +133,9 @@
     height: 300px;
```
tmeyer2115, line 133: Are these stylings shared by both results for a vertical query and a universal query? Is that the reasoning for the rename?

src/ui/components/cards/standardcardcomponent.js
tmeyer2115: Eh, I guess the "Details" is really the only thing that's stateful, so this prolly makes sense.

src/ui/components/cards/standardcardcomponent.js
tmeyer2115: Again, let's make these names a bit more descriptive. I would even be fine with rawResult and formattedResult or something like that.

alexisgrow: Can these also have block comments above them explaining what they are and their type?

src/ui/templates/cards/standard.hbs
```diff
@@ -1,87 +1,31 @@
...
+  {{/if}}
+  <div class='yxt-StandardCard-textWrapper'>
+    {{#if _config.title}}
+      <a href="{{url}}" class='yxt-StandardCard-title' target="{{#if _config.newWindow}}_blank{{else}}_self{{/if}}">
```
alexisgrow: Does it make sense to null check the {{url}} here?

alexisgrow (commented):
overall LGTM, agree with Tom's comments and left a few

src/ui/templates/cards/standard.hbs
oshi97: being bad at css, learned about object-fit today after some experimentation

src/ui/components/cards/cardcomponent.js
oshi97: Currently yes, if we don't change how we embed components

src/ui/components/cards/cardcomponent.js
oshi97: Yeah I hope we can refactor how we embed components soon otherwise it'll just get more annoying later

src/ui/templates/cards/standard.hbs
oshi97: updated!

src/ui/sass/modules/_Results.scss
oshi97: Yes, also yxt-Results was basically only styling the `resultssectionheader` and there was practically no styling before for the actual list of results, so I thought it made more sense to keep it as _Results.scss to be more consistent

src/ui/components/cards/standardcardcomponent.js
oshi97: Added!

src/core/models/verticalresults.js
```diff
@@ -7,6 +7,7 @@ export default class VerticalResults {
   constructor (data = {}) {
     Object.assign(this, { searchState: SearchStates.SEARCH_COMPLETE }, data);
     Object.freeze(this);
+    console.log(this);
```
tmeyer2115: Remove console.log

README.md
```diff
@@ -475,8 +475,6 @@ You define all the options at the top level object.
 ```js
 ANSWERS.addComponent('VerticalResults', {
   container: '.vertical-results-container',
-  // The max number of search results to return, defaults to 20
-  limit: 25,
```
tmeyer2115, line 479: Why did we remove limit?

src/ui/components/cards/cardcomponent.js
```diff
@@ -14,18 +14,17 @@ class CardConfig {
     this.cardType = config.cardType || cardTypes.Standard;
 
     /**
-     * isUniversal is set to true if this component is added by the UniversalResultsComponent
-     * @type {boolean}
-     * @private
+     * config.data comes from the data-prop attribute passed in
+     * from the parent component.
+     * @type {Result}
```
tmeyer2115, line 19: Is Result an actual type? If not, keep it as Object

src/ui/components/cards/cardcomponent.js
```diff
@@ -35,12 +34,18 @@ export default class CardComponent extends Component {
...
+    return super.setState({
+      ...data,
+      cardType: this._config.cardType
+    });
+  }
+
+  addChild (data, type, opts) {
+    return super.addChild(data, type, {
+      templateMappings: this._config.templateMappings,
+      result: this._config.result,
```
tmeyer2115: Would it be more correct to include result as part of the data passed to super.addChild?

src/ui/templates/cards/card.hbs
```diff
@@ -1,4 +1,2 @@
-<div data-component="{{cardType}}"
-  data-prop="cardData"
-  data-opts="{{cardOpts}}">
-</div>
\ No newline at end of file
+<div data-component="{{cardType}}">
```
tmeyer2115: I think it made the code a bit easier to understand when we were passing cardData as a data-prop here. Is that something we could still do (although it's now called something else)? I think we're having addChild to more work than strictly necessary by having it handle the actual result data too and not just template mappings.

src/ui/sass/_mixins.scss
```diff
@@ -100,3 +100,13 @@
     $hover-decoration: none
   );
 }
+
+@mixin Image(
```
tmeyer2115: Do we need this mixin anymore?

src/ui/components/cards/cardcomponent.js
oshi97: It's src/core/models/result.js but I didn't want to import it just for the jsdoc

README.md
oshi97: I missed it before but the spec says to remove it in favor of having the SearchConfig handle it

src/ui/components/cards/cardcomponent.js
oshi97: updated!

src/ui/sass/_mixins.scss
oshi97: nope

src/ui/templates/cards/standard.hbs
```diff
@@ -1,87 +1,36 @@
...
+  {{#if _config.image}}
+    <div class='yxt-StandardCard-img'>
+      <img src='{{_config.image}}'>
+    </div>
+  {{/if}}
+  <div class='yxt-StandardCard-textWrapper'>
+    {{#if _config.title}}
+      <a class='yxt-StandardCard-title'
+        {{#if url}} href="{{url}}" {{/if}}
+        target="{{#if _config.newWindow}}_blank{{else}}_self{{/if}}"
```
alexisgrow: when target="_blank", could you  add `rel="noopener noreferrer nofollow"` to the <a> tag? Security reasons https://developers.google.com/web/tools/lighthouse/audits/noopener

alexisgrow (approved):
🔥 LGTM

src/ui/templates/cards/standard.hbs
oshi97: updated!

PR #488, opened 2020-02-05
Spellcheck, Autocomplete: add escaping for highlighted value

alexisgrow:


tests/core/models/autocompletedata.js
```diff
@@ -42,92 +42,6 @@ describe('constructing from a response', () => {
     expect(actualResults).toEqual(expectedResults);
   });
 
-  it('properly creates highlightedValue from matchedSubstrings', () => {
```
alexisgrow, line 45: this case is covered in the highlighted value test cases (tests/core/models/highlightedvalue.js)

tests/core/models/autocompletedata.js
```diff
@@ -42,92 +42,6 @@ describe('constructing from a response', () => {
...
-      }
-    ];
-
-    const data = AutoCompleteData.from(response);
-    const actualResults = data.sections[0].results;
-    expect(actualResults).toHaveLength(2);
-    expect(actualResults).toEqual(expectedResults);
-  });
-
-  it('properly highlights multiple matchedSubstrings', () => {
```
alexisgrow, line 93: same comment, covered in the highlighted value test cases (tests/core/models/highlightedvalue.js)

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -134,5 +143,27 @@ export default class HandlebarsRenderer extends Renderer {
...
+
+      let highlightedVal = new HighlightedValue({
+        value: escapedInput,
+        matchedSubstrings: name.matchedSubstrings
+      }).getInverted();
+
+      return self.SafeString(highlightedVal);
+    });
+
+    self.registerHelper('highlightedValue', function (name, value, options) {
```
alexisgrow: i don't love how these two helpers are nearly identical. concerns -> seems like duplicated stuff + is annoying to see the difference between them

src/ui/templates/search/autocomplete.hbs
```diff
@@ -22,7 +22,7 @@
             data-eventtype="AUTO_COMPLETE_SELECTION"
             data-eventoptions='{"suggestedSearchText": "{{value}}"}'
         >
-            {{{highlightedValue}}}
+            {{highlightedValue highlightedValue true}}
```
tmeyer2115: Minor thing, but could we rename the first arg to be something besides highlightedValue? It's a bit confusing to see the same name twice. And, with this change, highlightedValue isn't yet highlighted. Maybe valueToHighlight or something?

src/core/models/autocompletedata.js
```diff
@@ -31,7 +29,7 @@ export default class AutoCompleteData {
 export class AutoCompleteResult {
   constructor (data = {}) {
     this.filter = data.filter || {};
-    this.highlightedValue = new HighlightedValue(data).getInverted();
+    this.highlightedValue = data;
```
tmeyer2115: I think it might make things a bit cleaner if we didn't have highlightedValue as a property of AutoCompleteResult at all. It's a domain model and highlightedValue is purely display concern. It's also a bit weird because we essentially have everything contained in data in the AutoCompleteResult twice (at different levels of nesting). In the AutoCompleteComponent, we can just add a valueToHighlight to the state that is an object containing all the stuff you need when calling the Handlebars helper.

tmeyer2115 (commented):
Much better! Don't know why this wasn't done in the first place 😛

src/ui/templates/search/autocomplete.hbs
alexisgrow: Yes, updating it to highlightValue so it's got a verb in it! Updated the data name to `this` due to changes from second comment

src/core/models/autocompletedata.js
alexisgrow: oooh amazing i like that, updated ✨

tmeyer2115 (approved):

PR #489, opened 2020-02-10
Implement CTAs for Cards

oshi97:
This commit implements CTAs by creating CTAComponent as a subcomponent
of StandardCardComponent.

TEST=manual


src/ui/components/helper/ctacomponent.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+export default class CTAComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(new CTAConfig(config), systemConfig);
+  }
+
+  onMount () {
+    const el = DOM.query(this._container, `.yxt-CTA`);
```
alexisgrow, line 55: do we / should we have a convention of using a `.js-` prefix for class selectors that we use in JS? It's a way of distinguishing which classes are used for CSS vs JS. (example here: https://gomakethings.com/javascript-selectors-in-html/)

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +106,64 @@
...
+        @include CTA(
+          $label-margin: 0 0 0 $standard-card-base-spacing/2,
+          $margin: 0,
+        );
+      }
+
+      &-item:only-child .yxt-CTA {
+        flex-direction: row;
+        @include CTA(
+          $label-margin: 0 0 0 0,
```
alexisgrow, line 162: nit: i think this can be simplified as one 0

oshi97:
I think actually this can be merged on top of feature/vertical-cards instead of into it since it's pretty separate

PR #491, opened 2020-02-10
CTAs for cards

oshi97:
This commit implements CTAs by creating CTAComponent as a subcomponent
of StandardCardComponent.

TEST=manual


src/ui/components/cards/cardcomponent.js
```diff
@@ -25,6 +25,32 @@ class CardConfig {
...
+
+    const callsToAction = config.callsToAction || [];
+    /**
+     * Array of call to action configuration objects
+     * @type {Array<Object>}
+     */
+    this.callsToAction = callsToAction.map(cta => this.handleCTA(cta));
+  }
+
+  handleCTA (cta) {
```
tmeyer2115: I think all the logic for mapping a result to a CTA should live inside the CTAComponent itself. This way, CTAComponent is just kind of a black box that the CardComponent can call into.

src/ui/components/cards/cardcomponent.js
```diff
@@ -25,6 +25,32 @@ class CardConfig {
...
+    if (typeof cta === 'function') {
+      return cta(this.result);
```
tmeyer2115: Confirm with Rose, but I think the 3 cases we want to handle for CTA are:

1. item => list of CTAs
2. There is a list of CTAs, where each CTA is an object whose values are determined by a function on item.
3. FieldNames.

Are we leaving option 3 for a later PR? If so, that's fine. I think, then, handleCTA needs to return a list of CTAs, since one mapping can potentially yield multiple CTAs.

src/ui/sass/modules/helper/_CTA.scss
```diff
@@ -0,0 +1,3 @@
+.yxt-CTA {
+  @include CTA();
```
tmeyer2115: I feel like we should get rid of the Mixin and move all the CSS into this class. I feel like the Mixin for CTAs has way too many arguments. Additionally, cards are the only consumers of CTAs, so the Mixin would not be widely used. Can you just override the default styling for the CTA in _StandardCard.scss in your various .yxt-CTA selectors?

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +106,64 @@
...
+
+      .yxt-CTA {
+        @include CTA(
+          $svg-height: $standard-card-svg-height,
+          $svg-width: $standard-card-svg-width,
+          $word-break: break-word,
+        );
+      }
+
+      &:only-child .yxt-CTA {
```
tmeyer2115: So the mocks call for lots of customization of the CTAs depending on if it's mobile or there's only one item? Kinda annoying.

tmeyer2115 (changes requested):

src/ui/components/cards/cardcomponent.js
oshi97: updated to use CTACollectionComponent

src/ui/components/cards/cardcomponent.js
oshi97: Updated

src/ui/sass/modules/helper/_CTA.scss
oshi97: Yes I can, I was hoping the mixing would make life easier for people but I guess it did the opposite

src/ui/sass/modules/_StandardCard.scss
oshi97: Yeah, I think it's important to customize for mobile but not sure about the only child thing

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,69 @@
...
+        const _cta = this.handleCTA(cta);
+        return {
+          optsString: JSON.stringify(_cta),
+          shouldRender: _cta.url && _cta.label
+        };
+      });
+    }
+  }
+
+  handleCTA (cta) {
```
tmeyer2115: I think the work of translating the callsToAction mappings in the config to actual ctas should happen in the component, not the config class. That feels like too much responsibility for the config class. I don't think we always need a ComponentConfig class either, especially if the config is relatively straightforward.

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,69 @@
...
+     */
+    if (typeof this.callsToAction === 'function') {
+      this.callsToAction = this.callsToAction(this.result);
+    }
+    if (Array.isArray(this.callsToAction)) {
+      this.callsToAction = this.callsToAction.map(cta => {
+        const _cta = this.handleCTA(cta);
+        return {
+          optsString: JSON.stringify(_cta),
+          shouldRender: _cta.url && _cta.label
```
tmeyer2115: What's this shouldRender?

src/ui/components/ctas/ctacollectioncomponent.js
tmeyer2115: For clarity, I think we should do some renaming. We could call this method resolveCTAMapping or something to give a better indication for what it does. I think it should have params ctaMapping and result. We could add some docs too.

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+    }
+  }
+
+  handleCTA (cta) {
+    if (typeof cta === 'function') {
+      return cta(this.result);
+    } else if (typeof cta === 'object') {
+      const calculatedCTA = { ...cta };
+      for (let [ctaOption, optionValue] of Object.entries(cta)) {
```
tmeyer2115: [ctaAttribute, attributeMapping]?

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,69 @@
...
+        };
+      });
+    }
+  }
+
+  handleCTA (cta) {
+    if (typeof cta === 'function') {
+      return cta(this.result);
+    } else if (typeof cta === 'object') {
+      const calculatedCTA = { ...cta };
```
tmeyer2115: Wouldn't we want to initialize calculatedCTA as empty?

src/ui/components/ctas/ctacomponent.js
```diff
@@ -0,0 +1,74 @@
...
+     * Whether the click should open in a new window
+     * @type {boolean}
+     */
+    this.newWindow = config.newWindow;
+
+    /**
+     * The eventOptions needed for the event to fire, passed as a string from templateMappings.
+     * @type {Object}
+     */
+    this.eventOptions = config.eventOptions ? JSON.parse(config.eventOptions) : {};
```
tmeyer2115, line 45: Why do we need to JSON.parse here? Were we JSONifying the eventOptions at some point higher in the chain?

src/ui/sass/answers.scss
```diff
@@ -8,6 +8,9 @@
 // General styling
 @import "base";
 
+// Helper component Styling
+@import "modules/helper/CTA";
```
tmeyer2115, line 12: I don't think we need to create a helper dir for this. I think it should just live in modules.

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +108,64 @@
...
+      margin-top: $base-spacing-sm;
+      align-items: center;
+
+      .yxt-CTA {
+        word-break: break-word;
+        @media (max-width: $breakpoint-mobile-max) {
+          flex-direction: row;
+          margin: 0;
+          &-iconLabel {
+            margin: 0 0 0 $standard-card-base-spacing/2;
```
alexisgrow: Instead of the four numbers here, margin-left can be more readable

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: Rose didn't want to render the cta if there was no label and url

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: no because we want to keep any static values

src/ui/components/ctas/ctacomponent.js
oshi97: The spec has eventOptions as a json string not sure which is better

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +108,64 @@
...
+          height: $standard-card-svg-height;
+          width: $standard-card-svg-width;
+        }
+      }
+
+      &:only-child .yxt-CTA {
+        $flex-direction: column;
+        margin: 0;
+        text-align: center;
+        word-break: break-word;
```
alexisgrow: looks like 
```
        margin: 0;
        text-align: center;
        word-break: break-word;
``` 
can be moved up to .yxt-CTA bc it's shared

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +108,64 @@
...
+        @media (max-width: $breakpoint-mobile-max) {
+          flex-direction: row;
```
alexisgrow: i think this is also shared with the regular .yxt-CTA one

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -105,6 +108,64 @@
...
+        }
+
+        .Icon svg {
+          height: $standard-card-svg-height-solo;
+          width: $standard-card-svg-width-solo;
+        }
+
+        &-iconLabel {
+          margin: $standard-card-base-spacing-sm/2 0 0 0;
```
alexisgrow: margin-top?

src/ui/sass/modules/helper/_CTA.scss
```diff
@@ -0,0 +1,31 @@
+.yxt-CTA {
+  display: flex;
+  flex-direction: row;
```
alexisgrow: this is default for display:flex, nit, you can omit!

src/ui/sass/modules/helper/_CTA.scss
```diff
@@ -0,0 +1,31 @@
...
+  text-decoration: none;
+  word-wrap: break-word;
+  margin: 0 0 0 $base-spacing;
```
alexisgrow: margin-left :octocat:

src/ui/sass/modules/helper/_CTA.scss
```diff
@@ -0,0 +1,31 @@
...
+    text-decoration: underline;
+  }
+
+  .Icon svg {
+    height: $base-spacing * 1.125;
+    width: $base-spacing * 1.125;
+  }
+
+  &-iconLabel {
+    margin: 0 0 0 $base-spacing-sm/2;
```
alexisgrow: same 👾

src/ui/sass/modules/helper/_CTA.scss
```diff
@@ -0,0 +1,31 @@
...
+  flex-direction: row;
+  text-decoration: none;
+  word-wrap: break-word;
+  margin: 0 0 0 $base-spacing;
+  text-align: left;
+  word-break: normal;
+  color: $color-brand-primary;
+  align-items: center;
+
+  &:hover, &:focus {
```
alexisgrow: :active as well?

alexisgrow (commented):
LG, left a few CSS comments

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: updated!

src/ui/sass/modules/_StandardCard.scss
oshi97: tried to update all the css things lmk what can still be improved!

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: updated to use the json hbs helper!

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Handles resolving ctas from a cta mapping which is either
+   * 1. a function that returns a cta's config
+   * 2. an object that has a per-attribute mapping of either a
+   *    a) static value
+   *    b) function that takes in resut data and returns the given attributes value
+   * 3. an api field name that keys into the result data which contains the cta config as a json string
+   * @param {Array<Function|Object|string>} ctaMapping
```
tmeyer2115: Doesn't this just take in a single Function?

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Either a function that spits out an array of CTA config objects or an array of CTA config objects
+     * or api fieldnames
+     * @type {Function|Array<Object|string>}
+     */
+    let callsToAction = this._config.callsToAction || [];
+
+    if (typeof callsToAction === 'function') {
+      callsToAction = callsToAction(this._config.result) || [];
```
tmeyer2115: Can this case be handled by resolveCTAMapping? According to the doc, it supports a function that returns a CTA's config. Is this handled separately because the function could return a list of CTA objects? If so, we could modify resolveCTAMapping to return an array of CTA objects (and then use a flatMap in your loop below).

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -102,9 +97,65 @@
     margin-top: $standard-card-base-spacing/2;
     @include Text(
       $color: $color-brand-primary,
-      $style: italic);
+      $style: italic);
```
tmeyer2115: Remove extra tab.

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: This is for a function that returns the array of cta config objects, are you saying just include this portion inside resolveCTAMapping and have resolveCTAMapping take in an array?

tmeyer2115: resolveCTAMapping wouldn't take in an array, it would output an array. Yeah, I think we should make that change, seems a bit cleaner. In the other cases, resolveCTAMapping would just return an array with one element. You can use a flatMap down below in your map(...)

oshi97: Ah ok that's really smart does this work?

tmeyer2115 (approved):

PR #492, opened 2020-02-10
Add footer to vertical results

oshi97:
TEST=manual


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -50,6 +51,45 @@ class VerticalResultsConfig {
      * @type {Object}
      */
     this.card = config.card || {};
+
+    /**
+     * Config for the footer at the bottom of the vertical results
+     * @type {Object}
+     */
+    this.footer = {
```
tmeyer2115: I think the footer will also be used for UniversalResults, so it could make sense to have a FooterComponent that can be reused.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -50,6 +51,45 @@ class VerticalResultsConfig {
...
+       */
+      url: 'https://yext.com',
+
+      /**
+       * Whether the logo should open the link in a new window
+       * @type {boolean}
+       */
+      newWindow: false,
+
+      ...config.footer
```
tmeyer2115: This syntax obscures what's going on. It makes it seem like we're hardcoding the various config values above.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -50,6 +51,45 @@ class VerticalResultsConfig {
...
+    /**
+     * Config for the footer at the bottom of the vertical results
+     * @type {Object}
+     */
+    this.footer = {
+      /**
+       * Either the name of an icon to display or a url to an image.
+       * @type {string}
+       */
+      logo: 'yext',
```
tmeyer2115: The spec called for the ability to do an in-line SVG. Are we not doing that for security reasons (which is totally fine)?

tmeyer2115 (changes requested):

src/ui/components/results/verticalresultscomponent.js
oshi97: Yeah I put a comment in the spec about not doing that anymore

src/ui/components/results/verticalresultscomponent.js
oshi97: Updated!

src/ui/components/results/verticalresultscomponent.js
oshi97: Updated

src/ui/components/utils/footercomponent.js
```diff
@@ -0,0 +1,57 @@
+/** @module VerticalResultsComponent */
```
tmeyer2115: Update.

src/ui/components/utils/footercomponent.js
```diff
@@ -0,0 +1,57 @@
...
+     * Click through url for the logo
+     * @type {string}
+     */
+    this._config.url = config.url || 'https://yext.com';
+
+    /**
+     * Whether the logo should open the link in a new window
+     * @type {boolean}
+     */
+    this._config.newWindow = config.newWindow;
```
tmeyer2115: Do we need a default for this?

src/ui/components/utils/footercomponent.js
```diff
@@ -0,0 +1,57 @@
...
+     * Icon name of footer image
+     * @type {string}
+     */
+    this._config.icon = Icons[logo] && logo;
+
+    /**
+     * Url of footer image
+     * @type {string}
+     */
+    this._config.image = !this._config.icon && logo;
```
tmeyer2115: The logic for setting this._config.icon and this._config.image is confusing. this._config.icon is always going to be true here, isn't it? Shouldn't we set this._config.icon = Icons[logo]?

src/ui/templates/cards/card.hbs
```diff
@@ -1,2 +1,2 @@
-<div data-component="{{cardType}}" data-prop="result">
-</div>
+<li data-component="{{cardType}}" data-prop="result">
```
tmeyer2115: Why is this an li now? Should card.hbs be scoped to only one result?

src/ui/templates/results/verticalresults.hbs
```diff
@@ -15,5 +15,10 @@
         data-prop="results">
       </ul>
     </section>
+    {{#if results}}
+      <div data-component='Footer'
+        data-opts='{{json _config.footer}}'
```
tmeyer2115, line 20: How come we need a json helper here? Isn't _config.footer a normal object without function values?

src/ui/components/utils/footercomponent.js
oshi97: updated!

src/ui/components/utils/footercomponent.js
oshi97: Felt like || false was redundant and can lead to confusion when || true doesn't work as expected but will add it

src/ui/components/utils/footercomponent.js
oshi97: updated

src/ui/templates/cards/card.hbs
oshi97: I mean, I think it should be scoped to one result but this change has 0 effect on the ui with the current styling

tmeyer2115 (approved):

PR #493, opened 2020-02-11
No Results: add models to convert api response

alexisgrow:
Parse alternativeVerticals and allResultsForVertical. These will appear
in the response in no-results situations

TEST=manual,auto


src/core/models/alternativeverticals.js
```diff
@@ -0,0 +1,38 @@
...
+     * @type {Section}
+     */
+    this.alternativeVerticals = data || [];
+
+    /**
+     * The current state of the search, used to render different templates before, during,
+     * and after loading
+     * @type {string}
+     */
+    this.searchState = data.searchState || SearchStates.SEARCH_COMPLETE;
```
tmeyer2115: What's the reasoning for having a searchState with AlternativeVerticals? Couldn't we use the searchState for the vertical results to inform whether or not alternativeVerticals have been loaded?

src/core/models/alternativeverticals.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Create alternative verticals from server data
+   * @param {Object} response The server response
+   * @param {Object.<string, function>} formatters The field formatters to use
+   */
+  static from (response, formatters) {
+    return new AlternativeVerticals(Section.from(response.alternativeVerticals, {}, formatters));
```
tmeyer2115, line 20: Do we want a Section per blob in the alternativeVerticals array? Each blob represents a different vertical.

src/core/models/verticalresults.js
```diff
@@ -17,16 +26,38 @@ export default class VerticalResults {
     const merged = { ...this };
     merged.results = this.results.concat(results.results);
     merged.map.mapMarkers = this.map.mapMarkers.concat(results.map.mapMarkers);
+    merged.resultsContext = this.resultsContext;
```
tmeyer2115: Not sure how this would happen, but what if the two sets of results had a different context? Should we throw an AnswersError here indicating those two kinds of results shouldn't be merged?

src/core/models/verticalresults.js
```diff
@@ -17,16 +26,38 @@ export default class VerticalResults {
...
     return new VerticalResults(merged);
   }
 
+  /**
+   * Form response as if the results from `allResultsForVertical` were the actual
+   * results in `results`
+   * @param {Object} response The server response
+   */
+  static formResponseWithAlternateResults (response) {
```
tmeyer2115: Minor nit, but I think the name should include allResultsForVertical or allResults in some way as that's the source of the results. Also minor nit, but I think we should use from instead of with in the name.

src/core/models/verticalresults.js
```diff
@@ -17,16 +26,38 @@ export default class VerticalResults {
...
    * @param {Object.<string, function>} formatters The field formatters to use
    */
   static from (response, formatters) {
-    return new VerticalResults(Section.from(response, null, formatters));
+    const hasResults = response.results && response.results.length > 0;
+    const resultsContext = hasResults ? ResultsContext.NORMAL : ResultsContext.NO_RESULTS;
+
+    return new VerticalResults(
+      Section.from(
+        hasResults ? response : VerticalResults.formResponseWithAlternateResults(response),
```
tmeyer2115: Could we make formResponseWithAlternateResults a private static method? It feels like something internal to the class that we wouldn't want others using.

src/core/models/alternativeverticals.js
alexisgrow: Yes, we do! It's not really named in an intuitive way, but Section.from() actually returns an array of sections, so the param being passed in is one section per blob

src/core/models/verticalresults.js
alexisgrow: Ah true, updating ✨

src/core/models/verticalresults.js
alexisgrow: agree! updating 🔥

src/core/models/verticalresults.js
alexisgrow: Yes definitely ✅

src/core/models/alternativeverticals.js
alexisgrow: Yeah, we definitely could use the searchState from vertical results. My concern was that the alternativeVerticals may not necessarily need to be tied to vertical results--but I suppose it should always be tied to some type of results (universal or vertical), so it could/should use the searchState of the results component regardless

src/core/models/alternativeverticals.js
tmeyer2115: Ugh, what confusing naming.

src/core/models/alternativeverticals.js
tmeyer2115: Yeah, I think we should key off the searchState of VERTICAL_RESULTS for now since this is a by-product of a vertical query. If they ever made allResultsForVertical a first class thing that we could hit an endpoint for, we'd definitely want it to have it's own searchState.

tmeyer2115 (approved):

PR #496, opened 2020-02-12
Navigation: update component to pull from vertical pages config

alexisgrow:
Alter the navigation section in answers config. Remove tabs key,
add a hideInNavigation flag, rename navigation to verticalPages

TEST=manual,auto


src/core/models/verticalpagesconfig.js
```diff
@@ -38,21 +44,21 @@ export class TabConfig {
   validate () {
   }
 
-  static from (tabs) {
-    let tabConfigs = [];
-    if (tabs === undefined) {
-      return tabConfigs;
+  static from (pages) {
```
tmeyer2115, line 53: Would this be better suited as a static method on VerticalPagesConfig as it generates an array of VerticalPageConfig?

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -217,7 +215,7 @@ describe('navigation component configuration', () => {
       {
         label: 'This this is my second label',
         url: '/tab1/',
-        configId: 'tab1'
+        verticalKey: 'tab1'
```
tmeyer2115, line 225: I think it would be good to add a test case ensuring hideInNavigation is correctly handled by the NavComponent too.

oshi97 (approved):
lgtm! just wondering whether there's a readme we need to update

src/core/models/verticalpagesconfig.js
alexisgrow: Yeah that sgtm! 🔥

tests/ui/components/navigation/navigationcomponent.js
alexisgrow: good point, i'll add 🆕

alexisgrow: added to test case that ends on line 105!

tmeyer2115 (approved):

PR #497, opened 2020-02-13
No Results: UI

alexisgrow:


src/core/models/verticalpagesconfig.js
```diff
@@ -39,6 +39,12 @@ export class VerticalPageConfig {
      * @type {boolean}
      */
     this.isActive = config.isActive || false;
+
+    /**
+     * name of an icon from the default icon set
+     * @type {string}
+     */
+    this.icon = config.icon || false;
```
tmeyer2115: Should this default to star?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+ */
+export class VerticalSearchSuggestion {
+  constructor (config) {
+    /**
+     * The name of the vertical that is exposed for the link
+     * @type {string}
+     */
+    this.label = config.label;
+    if (typeof this.label !== 'string') {
+      throw new AnswersComponentError('label is a required configuration option for verticalPage.', 'AlternativeVerticalsComponent');
```
tmeyer2115: I should figure out a way to add line length to the linter, but let's break up this line. The way we do things in Java world is if not all function params fit on the same line, break them onto one line each.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+    this.url = config.url;
+    if (typeof this.url !== 'string') {
+      throw new AnswersComponentError('url is a required configuration option for verticalPage.', 'AlternativeVerticalsComponent');
+    }
+
+    /**
+     * The base URL used for constructing the URL with params
+     * @type {string}
+     */
+    this.baseUrl = config.url;
```
tmeyer2115: I'm not sure I understand the distinction between baseUrl and url

src/core/models/verticalpagesconfig.js
oshi97: does this just mean default is no icon?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+import { AnswersComponentError } from '../../../core/errors/errors';
+import Component from '../component';
+import StorageKeys from '../../../core/storage/storagekeys';
+
+/**
+ * The VerticalSearchSuggestion is a model that is used to power the search
+ * suggestions info box. It's initialized through the configuration provided
+ * to the component.
+ */
+export class VerticalSearchSuggestion {
```
oshi97: could we have VerticalSearchSuggestion under models instead of here in components?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -129,6 +155,14 @@ export default class VerticalResultsComponent extends Component {
         ...opts
       };
       return super.addChild(data, type, newOpts);
+    } else if (type === AlternativeVerticalsComponent.type) {
+      data = this.core.globalStorage.getState(StorageKeys.ALTERNATIVE_VERTICALS);
+      const newOpts = Object.assign({}, {
```
oshi97: nit: I personally prefer using the spread syntax instead of Object.assign, totally up to you just writing this in case you didn't have a preference

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+     */
+    this.resultsCount = config.resultsCount;
+  }
+
+  /**
+   * from will construct an array of {VerticalSearchSuggestion} from
+   * a configuration file
+   * @param {object} tabsConfig the configuration to use
+   */
+  static from (alternativeVerticals, verticalsConfig) {
```
tmeyer2115: I think we should make this a private method of the AlternativeVerticalsComponent since it creates on a collection of VerticalSearchSuggestion, which is a slightly higher level of abstraction than in this class.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+    let verticals = [];
+
+    for (let alternativeVertical of alternativeVerticals) {
+      const verticalKey = alternativeVertical.verticalConfigId;
+
+      const matchingVerticalConfig = verticalsConfig.find(config => {
+        return (config.verticalKey === verticalKey) ||
+          (!config.verticalKey && !verticalKey);
```
tmeyer2115: Is this to account for the universal case? I think verticalKey should always be populated here since it's coming from alternativeVerticals.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+      });
+
+      if (!matchingVerticalConfig) {
+        continue;
```
tmeyer2115: What would you think about throwing an error in this case? I'm not sure if we'd want an error here or not.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,165 @@
...
+    return 'results/alternativeverticals';
+  }
+
+  static areDuplicateNamesAllowed () {
+    return true;
+  }
+
+  setState (data) {
+    for (let vertical of this.verticalSuggestions) {
+      vertical.url = vertical.baseUrl + window.location.search;
```
tmeyer2115: Could we move this logic to the from method where we populate this.verticalSuggestions? That way, we could just have a single url in the VerticalSearchSuggestions class and not have to maintain baseUrl as well (unless it's used elsewhere).

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -84,7 +108,9 @@ export default class VerticalResultsComponent extends Component {
       mapConfig: this._config.mapConfig,
       eventOptions: this.eventOptions(),
       universalUrl: this._universalUrl ? this._universalUrl + window.location.search : '',
-      showNoResults: results.length === 0,
+      showNoResults: results.length === 0 || data.resultsContext === ResultsContext.NO_RESULTS,
+      showResults: displayResultsIfExist && results.length !== 0,
+      showEnhancedNoResults: this._config._showEnhancedNoResults,
```
tmeyer2115: Is there a way to simplify this a bit? It's a bit complex with these 4 booleans. The variables have similar sounding names too. Would it be possible to do something like: resultsPresent, resultsContext, and showEnhancedNoResults? We could use resultsContext to switch behavior depending on the source of the results.

src/core/models/verticalpagesconfig.js
alexisgrow: Ah good catch, the default for icons overall is star, so I didn't even think of it.

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: updated 🔥

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: oh that's fair, removing

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: Yeah, I think we would probably want to error. If the config doesn't match what the API is sending there's probably something off in their config.

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: That sounds good to me, I'll update!

src/ui/components/results/verticalresultscomponent.js
alexisgrow: Updated to this, what do you think?
```
      resultsPresent: displayResultsIfExist && results.length !== 0,
      showNoResults: data.resultsContext === ResultsContext.NO_RESULTS,
      isEnhancedNoResultsEnabled: this._config._showEnhancedNoResults,
```

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: commenting on thread on line 123! (edit: line number after pr fixes)

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: I like that, do we have a library/pattern we use that removes the URL params from a url?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,178 @@
...
+   */
+  static _buildVerticalSuggestions (alternativeVerticals, verticalsConfig) {
+    let verticals = [];
+
+    for (let alternativeVertical of alternativeVerticals) {
+      const verticalKey = alternativeVertical.verticalConfigId;
+
+      const matchingVerticalConfig = verticalsConfig.find(config => {
+        return (config.verticalKey === verticalKey) ||
+          (!config.verticalKey && !verticalKey);
```
tmeyer2115: Should this be removed?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,178 @@
...
+        throw new AnswersComponentError(
+          'no matching verticalPages config entry for ' + verticalKey,
+          'AlternativeVerticalsComponent'
+        );
+      }
+
+      verticals.push(new VerticalSearchSuggestion({
+        label: matchingVerticalConfig.label,
+        url: matchingVerticalConfig.url,
+        baseUrl: matchingVerticalConfig.url,
```
tmeyer2115: So, here, could we just set url to be matchingVerticalConfig.url + window.location.search? Would that allow us to get rid of baseUrl in the VerticalSearchSuggestion class? (Doesn't seem like we use baseUrl anywhere but to set the url)

tmeyer2115: We're not relaxing any filters on the back-end right now to return allResultsForVertical, right? If that starts happening, I guess we can't just add the query params to the baseUrl, we'd have to figure out another way to generate the url here.

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: Oof, yes 🙈 copy-pasted function, then modified the version that i deleted. updated

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: Ah yeah, that works!

src/core/models/verticalpagesconfig.js
alexisgrow: oh true, updated default to be ⭐️

src/ui/components/results/alternativeverticalscomponent.js
alexisgrow: updated 👾

src/ui/components/results/verticalresultscomponent.js
alexisgrow: 🔥 i like that, updating!

tmeyer2115 (approved):

PR #498, opened 2020-02-13
verticalresults readme

oshi97:
update readme with new verticalresults config
 - update template mappings to take in functional attributes
 - accept object eventOptions in addition to json string


README.md
```diff
@@ -478,10 +478,192 @@ ANSWERS.addComponent('VerticalResults', {
...
+  //       'RSVP'
+  analyticsEventType: 'CTA_CLICK',
+  // Whether the click should open in a new window, defaults to false
+  newWindow: false,
+  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. defaults to null
+  eventOptions: `{ "verticalKey": "credit-cards", "entityId": "123123", "searcher":"UNIVERSAL", "ctaLabel": "cards"}`
+}]
+```
+
+2. as a custom field name. For each field name, if the entity in the result data has this field name,
```
tmeyer2115: This is not currently supported yet, right? If so, I'd leave it out of the README for now.

README.md
```diff
@@ -478,10 +478,192 @@ ANSWERS.addComponent('VerticalResults', {
...
+  url: "https://maps.google.com",
+  analyticsEventType: "CTA_CLICK",
+  newWindow: false,
+  icon: "phone",
+  eventOptions: `{ "verticalKey": "credit-cards", "entityId": "${item._raw.id}", "searcher":"UNIVERSAL", "ctaLabel": "cards"}`
+}]
+```
+
+Each individual field in a CTA config can also be a function that operates on the result item, and all
+options above can be mixed and matched like so.
```
tmeyer2115: I don't think we should allow them to mix and match like this. Even if the code supports it, we shouldn't advertise it in a README. Did Rose ask that we support this?

README.md
```diff
@@ -478,10 +478,192 @@ ANSWERS.addComponent('VerticalResults', {
...
+      url: "https://yext.com",
+    }, 'c_customCTAField']
+  }
+  /* ...other vertical results config... */
+})
+```
+
+## Template Mappings
+Card config outside of the cardType and callsToAction is defined in a templateMappings attribute that either is
+a function that returns a templateMappings object or a static templateMappings object.
```
tmeyer2115: I think we should make clear that templateMappings defines how the card's attributes, such as title and details, will be rendered.

README.md
oshi97: oh I thought I did add support for it in resolveCTAMapping in the last case, just want to double check

```
resolveCTAMapping (result, ...ctas) {
    return ctas.map(ctaMapping => {
      if (typeof ctaMapping === 'function') {
        return ctaMapping(result);
      }
      if (typeof ctaMapping === 'object') {
        const ctaObject = { ...ctaMapping };
        for (let [ctaAttribute, attributeMapping] of Object.entries(ctaMapping)) {
          if (typeof attributeMapping === 'function') {
            ctaObject[ctaAttribute] = attributeMapping(result);
          }
        }
        return ctaObject;
      }
      // If cta is specified through an API Field Name
      if (typeof ctaMapping === 'string') {
        return { ...JSON.parse(result[ctaMapping]) };
      }
    });
```

README.md
oshi97: No I just thought we may as well write it, I guess you don't want to promise this?

README.md

README.md
oshi97: \* also updated

tmeyer2115: Nah, I don't want us to promise the ability to mix and match. Even if we can handle it now, I don't want to commit to it in perpetuity. There could be some future world where a new method of specifying CTAs is added that would make supporting mixed formats really difficult.

tmeyer2115 (approved):

PR #499, opened 2020-02-13
QA vertical results: styling changes

oshi97:
Update styling to fix issues in qa sheet/discussed over


src/ui/sass/modules/_StandardCard.scss
```diff
@@ -85,9 +79,24 @@
...
+    justify-content: space-between;
+    align-self: flex-end;
+    width: 100%;
+  }
+
   &-details {
-    flex-basis: $standard-card-text-width;
+    min-width: $standard-card-text-min-width;
+    flex-basis: 0;
     flex-grow: 1;
```
tmeyer2115: Since yxt-StandardCard-details and yxt-StandardCard-ctas are under the same container div (and are the only two things in that div), do they both need flex-basis and flex-grow if the values are the same for both?

oshi97: moved up to parent div

src/ui/sass/_icons.scss
```diff
@@ -0,0 +1,14 @@
+.yxt-Icons {
```
alexisgrow: since this isn't for icons overall, it's a class that's being added in the VerticalResults component, I think it should live in the vertical results CSS. Being in an icons.scss makes me think that this applies to all icons

src/ui/templates/cards/standard.hbs
```diff
@@ -1,39 +1,47 @@
...
-      {{#if _config.subtitle}}  
-        <div class="yxt-StandardCard-subtitle">
-          {{_config.subtitle}}
-        </div>
-      {{/if}}
-      <div class="yxt-StandardCard-details">
-        {{details}}
+  <div class="yxt-StandardCard-wrapper">
+    {{#if _config.title}}
+      <a class="yxt-StandardCard-title"
```
alexisgrow: it looks like this is on its own line with display: block. that means the clickable target is going to have a lot of white space, adding a < div > wrapper around the < a > tag is a common way to fix this

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -98,16 +107,18 @@
...
     @include Text(
       $color: $color-brand-primary,
       $style: italic);
     @include TextButton($padding: 0);
   }
 
   &-ctas {
-    margin-left: $standard-card-base-spacing/2;
+    margin-left: $standard-card-base-spacing;
+    min-width: $standard-card-cta-width;
```
alexisgrow: instead setting `min-width: X` and `max-width: X` on the line below, can use `width: X`?

src/ui/sass/_icons.scss
oshi97: I actually thought it might be nice to have some general icon css! Just for general things like rotating the caret icon to be up or down or changing the size. lmk if this isn't the right approach for things

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -85,8 +79,23 @@
     );
   }
 
+  &-content {
+    display: flex;
+    flex-basis: 100%;
+    flex-grow: 1;
+    flex-wrap: wrap;
+    justify-content: space-between;
+    align-self: flex-end;
```
alexisgrow: instead of using align self here and on its children, is it possible to use justify content on the parent element?

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -85,8 +79,23 @@
...
+    width: 100%;
```
alexisgrow, line 86: since the flex direction is row here, i think this width and the flex-basis above do the same thing, is that right?

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -85,8 +79,23 @@
...
+  }
+
   &-details {
-    flex-basis: $standard-card-text-width;
+    min-width: $standard-card-text-min-width;
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    margin-bottom: $standard-card-base-spacing / 2;
+    flex-basis: 0;
```
alexisgrow, line 95: since there's a min-width a few lines above, would `flex-basis: auto` be better?

src/ui/sass/_icons.scss
```diff
@@ -0,0 +1,14 @@
...
+  &-collapseUp svg {
```
alexisgrow: for both of these, is svg needed? does
```
  &-collapseDown {
    transform: rotate(90deg);
  }
```
do effectively the same thing?

src/ui/sass/_icons.scss
alexisgrow: Oh yeah, that would be great. This CSS works, I think the only potentially confusing thing is that these selectors aren't in the Icon Component, so it may be worth adding them there / allowing a caller to pass in a selector

src/ui/sass/_icons.scss
oshi97: Tried it without svg and I think svg has to be there because otherwise just the container gets rotated which doesn't change the actual icon

src/ui/sass/_icons.scss
oshi97: ah ok yeah that's a great idea!

src/ui/sass/modules/_StandardCard.scss
oshi97: updated, don't think flex-basis was doing anything

src/ui/sass/modules/_StandardCard.scss
oshi97: yes that's definitely better! But I messed around a bit and think this should be flex-start anyways

src/ui/sass/modules/_StandardCard.scss
oshi97: Oh hmm I'm probably not following best css practices or something but it breaks if I change this

src/ui/sass/modules/_StandardCard.scss
oshi97: yes, updated!

src/ui/templates/cards/standard.hbs
oshi97: ah ok, didn't even realize this was happening

src/ui/components/icons/iconcomponent.js
```diff
@@ -42,7 +48,7 @@ export default class IconComponent extends Component {
     }
 
     if (this.iconUrl) {
-      return `<img src="${this.iconUrl}" alt="" class="Icon-image">`;
+      return `<img src="${this.iconUrl}" alt="" class="Icon-image ${this.classNames}>`;
```
tmeyer2115: I think you're missing a closing "

src/ui/components/icons/iconcomponent.js
```diff
@@ -25,11 +25,17 @@ export default class IconComponent extends Component {
...
     /**
      * the url to a custom image icon
      * @type {null}
      */
     this.iconUrl = opts.iconUrl || null;
+
+    /**
+     * additional css classnames to give to the icon
+     */
+    this.classNames = opts.classNames || null;
```
tmeyer2115, line 39: How is this formatted? Is it a list of strings, a single string with class names separated by spaces?

src/ui/sass/modules/_StandardCard.scss
```diff
@@ -54,14 +46,15 @@
     }
   }
 
-  &-textWrapper {
+  &-wrapper {
+    display: flex;
+    flex-direction: column;
+    flex-basis: 0;
```
tmeyer2115, line 52: I don't think you need to specify a flex-basis of 0. I think this should be the same as omitting it.

src/ui/sass/modules/_StandardCard.scss
tmeyer2115: It seems like we would want flex-basis: auto here. That way, the initial size computed by the flex-basis is consistent with the size property of the element.

src/ui/sass/modules/_StandardCard.scss
oshi97: I think the default is auto, which starts the flex growing at the default size of the content instead of at 0 but isn't what we want here

src/ui/sass/modules/_StandardCard.scss
oshi97: In that case the -details div grows too much and crowds out the ctas to the right of it plus has extra whitespace, I think we would rather have the flexbox start growing at 0 and have the text wrap in cases when it meets with the ctas

src/ui/components/icons/iconcomponent.js
oshi97: updated comment documentation

src/ui/components/icons/iconcomponent.js
oshi97: thanks for the catch!

src/ui/templates/cards/standard.hbs
```diff
@@ -1,39 +1,44 @@
...
-        <a class="yxt-StandardCard-title"
-          {{#if _config.url}} 
-            href="{{_config.url}}"
-            target={{#if _config.newWindow}}"_blank" rel="noopener noreferrer nofollow"{{else}}"_self"{{/if}}
-          {{/if}}
-        >
+  <div class="yxt-StandardCard-wrapper">
+    {{#if _config.title}}
+      <div class="yxt-StandardCard-title">
+        <a {{#if _config.url}}
```
alexisgrow, line 10: i think this may be out of scope of the current PR (bc you didn't actually change this line), but we'll want to change this slightly in the future. If _config.url is falsy here then the <a> tag will cause WCAG issues (because it's not actually a link if there's no href). The logic should probably be something like
```
{{#if _config.url}}
<a> 
{{else}}
<div>
{{/if}}
```

alexisgrow (approved):

tmeyer2115 (approved):

src/ui/templates/cards/standard.hbs
oshi97: o 😮

PR #500, opened 2020-02-13
No Results: add styling

alexisgrow:


src/ui/sass/modules/_AlternativeVerticals.scss
```diff
@@ -0,0 +1,124 @@
...
+      font-weight: $font-weight-normal;
+    }
+
+    @media (min-width: $breakpoint-mobile-max) {
+      font-size: $font-size-md-lg;
+    }
+  }
+
+  &-suggestionsList {
+    // Overriding user agent stylesheet, TODO ask about this
```
tmeyer2115: Yeah, we've had to battle against user agent styling in other components too. At one point, I added a global CSS reset that targeted things like this. But, the specificity was high enough that it clobbered some of consulting's styling. Since then, we've just been adding resets like this as needed at lower specificity.

src/ui/icons/pin.js
```diff
@@ -1,6 +1,6 @@
 import SVGIcon from './icon.js';
 export default new SVGIcon({
   name: 'pin',
-  viewBox: '0 0 13 18',
+  viewBox: '5 0 9 18',
```
tmeyer2115, line 4: For my own edification, was this to make the pin icon appear smaller? Not very familiar with viewBox.

tmeyer2115 (approved):

src/ui/sass/modules/_AlternativeVerticals.scss
alexisgrow: Ah fair, that makes sense, it'd probably be annoying if we're wiping out user agent styling they rely on. Removing TODO

src/ui/icons/pin.js
alexisgrow: It's centering the pin within the SVG's viewbox

e.g. before it looked like [this](https://drive.google.com/file/d/1-H3vO3trd8kUiV1AI34JRO_WhfLskYKY/view?usp=sharing), which was not aesthetically pleasing because centering the icon wouldn't center the pin. This change moves it basically [here](https://drive.google.com/file/d/1NKF9dd3Ahp3BECcS9CPAcWacXaZZbbXO/view?usp=sharing) so it lines up better.

PR #501, opened 2020-02-15
VerticalResults: Applied Filters

oshi97:
This commit adds filter metadata to all filters in vertical results,
which are then displayed in a new AppliedFiltersComponent, along
with results count.

TEST=manual


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -62,6 +62,30 @@ class VerticalResultsConfig {
...
+     * If present, show the filters that were ultimately applied to this query
+     * @type {boolean}
+     */
+    this.showAppliedFilters = config.showAppliedFilters === undefined ? true : undefined;
+
+    /**
+     * Display the count of results at the very top of the results
+     * @type {boolean}
+     */
+    this.showResultsCount = config.showResultsCount === undefined ? true : undefined;
```
alexisgrow, line 76: can this be simplified to `this.showResultsCount = config.showResultsCount === undefined` or does it need to be undefined when falsy? same comment for this.showAppliedFilters

src/ui/sass/_base.scss
```diff
@@ -13,3 +13,7 @@ body
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
+
+.yxt-DummyContainer {
+  display: contents;
```
alexisgrow, line 18: IE11 doesn't support display: contents iirc 😞

alexisgrow: https://caniuse.com/#feat=css-display-contents

src/ui/templates/results/appliedfilters.hbs
```diff
@@ -0,0 +1,17 @@
+<span class='yxt-Results-resultsCount'>
+  {{pageStart}} - {{pageEnd}} of {{pageEnd}}
```
alexisgrow, line 2: i think i may be reading this wrong, but it seems to me like having pageEnd twice is wrong, i would expect
``` 
{{pageStart}} - {{pageEnd}} of {{resultCount}} 
```
e.g. [showing results] 1 - 5 of 500. does that make sense for this use case?

src/ui/templates/results/appliedfilters.hbs
```diff
@@ -0,0 +1,17 @@
...
+  {{/each}}
+{{/if}}
+{{#if appliedFiltersString}}
+  <span class='yxt-Results-filterValue'>-</span>
+  <span class="yxt-Results-filterValue">{{appliedFiltersString}}</span>
+{{/if}}
+{{#if sortBysString}}
+  <span class='yxt-Results-filterValue'>-</span>
+  <span class="yxt-Results-filterValue">{{sortBysString}}</span>
+{{/if}}
```
alexisgrow, line 17: nit: since this is its own hbs file / component, I would probably expect the selectors to have a .yxt-AppliedFilters prefix instead of .yxt-Results, that would mean its styling would also go in its own _AppliedFilters.scss file to stop the results one from getting too big

src/ui/templates/results/resultssectionheader.hbs
```diff
@@ -21,21 +21,16 @@
...
-    {{#if _config.showChangeFilters}}
-      {{#if verticalURL}}
-        <a class="yxt-Results-changeFilters" href="{{verticalURL}}"
-           data-eventtype="FILTERING_WITHIN_SECTION"
-           data-eventoptions='{{eventOptions}}'>
-          change
-        </a>
-      {{/if}}
+<div class="yxt-Results-filters" >
+  <div class="yxt-DummyContainer" data-component="AppliedFilters" data-opts='{{json _config.appliedFiltersOpts}}'></div>
```
alexisgrow, line 25: nit: instead of dummy container here, maybe .yxt-AppliedFilters?

src/ui/sass/modules/_Results.scss
```diff
@@ -102,30 +105,35 @@
...
 
   &-filterValue
   {
+    margin-left: $base-spacing/2;
     @include Text(
       $results-filters-text-font-size,
       $results-filters-text-line-height,
       $results-filters-text-font-weight,
-      $color: $results-filters-text-color
+      $color: $results-filters-text-color,
```
alexisgrow, line 136: 🎨 does the trailing comma work w/ our scss compiler? same comment for the filterSeparator and resultsCount above

src/ui/components/results/verticalresultscomponent.js
oshi97: ahh that's so much better

oshi97:
will be breaking this pr into 3 chunks

src/ui/sass/_base.scss
oshi97: oh, darn, I thought this might be a good workaround for the dummy div containers we have whenever we make a child component

src/ui/templates/results/appliedfilters.hbs
oshi97: Yeah you're absolutely right

PR #504, opened 2020-02-18
consolidate filter/facet/sorting getstate logic for vertical search

oshi97:
TEST=manual


src/core/core.js
```diff
@@ -111,16 +109,33 @@ export default class Core {
...
+    // If no verticalKey is set try to get verticalKey from the common search config
+    if (!verticalKey) {
+      verticalKey = this.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
+    }
+
+    // Get filters and facet to send in the request
+    const allFilters = this.globalStorage.getAll(StorageKeys.FILTER);
+    const totalFilter = allFilters.length > 1
+      ? Filter.and(...allFilters)
+      : allFilters[0];
```
tmeyer2115: Is this safe, what if allFilters is empty?

src/core/core.js
```diff
@@ -280,7 +288,7 @@ export default class Core {
         direction: option.direction
       };
     });
-    this.globalStorage.set(StorageKeys.SORT_BYS, JSON.stringify(sortBys));
+    this.globalStorage.set(StorageKeys.SORT_BYS, sortBys || []);
```
tmeyer2115: Should we just not call globalStorage.set if sortBys is empty?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -285,21 +285,11 @@ export default class FilterBoxComponent extends Component {
...
-
-    const facetFilter = this.core.globalStorage.getAll(StorageKeys.FACET_FILTER)[0];
-
     this.core.persistentStorage.delete(StorageKeys.SEARCH_OFFSET);
     this.core.globalStorage.delete(StorageKeys.SEARCH_OFFSET);
     this.core.verticalSearch(this._verticalKey, {
-      input: query,
-      filter: JSON.stringify(totalFilter),
-      facetFilter: JSON.stringify(facetFilter)
+      input: query
```
tmeyer2115: let's put everything on the previous line now that the other attributes are gone

tmeyer2115 (commented):
Were you able to test each component you modified to make sure it still worked as expected? In future commits, I'd suggest adding a bit more detail about what you did, testing, etc. Follow the format here: https://sites.google.com/a/yext.com/engineering/orientation/style/commit-messages

src/core/core.js
oshi97: Yes !

src/ui/components/filters/filterboxcomponent.js
oshi97: Done

src/core/core.js
oshi97: Updated

tmeyer2115 (approved):

PR #506, opened 2020-02-20
Use FilterView instead of Filter

oshi97:
This commit implements FilterView and FacetView,
which are container classes for Filter and Facets
which contain additional metadata
TEST=manual
Created a test site testing that FilterOptions, FilterBox, Facets,
DateRangeFilter, GeolocationFilter, DateRangeFilter, RangeFilter,
and GeoLocationFilter still filtered properly and also contained
desired metadata


src/core/models/filterview.js
```diff
@@ -0,0 +1,213 @@
+/** @module FilterView */
+
+import Filter from './filter';
+
+/**
+ * Contains a {@link Filter} and associated metadata.
+ */
+export default class FilterView {
```
tmeyer2115, line 9: I think we should leave the Filter class as is, with all those static methods. Those methods are all at the same level of abstraction as the Filter class, they create a Filter. I don't think the FilterView class should provide static methods for you to create Filters. It operates at a higher level of abstraction than a Filter.

src/core/models/facet.js
```diff
@@ -9,31 +9,32 @@
  *
  * @see {@link Filter}
  */
-export default class Facet {
-  constructor (data = {}) {
-    Object.assign(this, data);
+export default class FacetView {
```
tmeyer2115: How come we didn't add a new class for FacetView?

src/core/core.js
```diff
@@ -322,13 +323,13 @@ export default class Core {
   }
 
   /**
-   * Stores the given filter into storage, to be used for the next search
+   * Stores the given filter view into storage, to be used for the next search
```
tmeyer2115: Stores the given filter view into storage. The view's filter data object will be used in the next search.

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -132,26 +132,30 @@ export default class DateRangeFilterComponent extends Component {
...
+    const filterView = this._buildFilterView();
     if (this._storeOnChange) {
-      this.core.setFilter(this.name, filter);
+      this.core.setFilterView(this.name, filterView);
     }
     this.core.persistentStorage.set(`${this.name}.min`, this._date.min);
     this.core.persistentStorage.set(`${this.name}.max`, this._date.max);
 
-    this._onChange(filter);
+    this._onChange(filterView);
```
tmeyer2115: Seems like we probably want to call _onChange with the actual filter. At the very least, for backwards compatibility in case anyone has written onChange methods for this component already. Plus, I don't think folks will care about the metadata object in any onChange method they create.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -272,12 +271,12 @@ export default class FilterOptionsComponent extends Component {
...
   updateListeners () {
-    const filter = this._buildFilter();
+    const filterView = this._buildFilterView();
     if (this.config.storeOnChange) {
-      this.core.setFilter(this.name, filter);
+      this.core.setFilterView(this.name, filterView);
     }
 
-    this.config.onChange(filter);
+    this.config.onChange(filterView);
```
tmeyer2115: Same comment about onChange

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -105,14 +105,14 @@ export default class GeoLocationComponent extends Component {
      * The filter to use for the current query
      * @type {Filter}
      */
-    this.filter = this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || {};
-    if (typeof this.filter === 'string') {
+    this.filterView = this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || {};
+    if (typeof this.filterView === 'string') {
```
tmeyer2115, line 110: This is kinda weird. When would the type of filterView be string? Is this what happens when the filter is seeded from persistentStorage?

tmeyer2115 (changes requested):

src/ui/components/filters/geolocationcomponent.js
oshi97: Yeah the current code stores the filter as json in the url for a few but not all filters

src/ui/components/filters/daterangefiltercomponent.js
oshi97: included metadata in additional parameter because we need it for filterbox

src/core/core.js
```diff
@@ -115,18 +115,16 @@ export default class Core {
...
-    const allFilters = this.globalStorage.getAll(StorageKeys.FILTER);
-    let totalFilter = {};
-    if (allFilters.length > 0) {
-      totalFilter = allFilters.length > 1
-        ? Filter.and(...allFilters)
-        : allFilters[0];
-    }
-    const facets = this.globalStorage.getAll(StorageKeys.FACET_FILTER);
-    const facetFilter = facets.length > 0 ? facets[0] : {};
+    const allFilters = this.globalStorage.getAll(StorageKeys.FILTER_VIEW) || [];
```
tmeyer2115: allFilterViews?

src/core/models/facet.js
```diff
@@ -17,17 +19,18 @@ export default class Facet {
 
   /**
    * Create a facet filter from a list of Filters
+   * @param {Array<string>} availableFieldIds array of expected field ids
    * @param  {...Filter} filters The filters to use in this facet
    * @returns {Facet}
    */
-  static fromFilters (availableFieldIds, ...filters) {
+  static fromFilterViews (availableFieldIds, ...filters) {
```
tmeyer2115: Is ...filters a list of Filters or FilterViews? If the former, we should continue calling this fromFilter. I think having this take in a list of Filter would make the most sense, as Facet should not be built with anything that is purely a presentation concern (like the Metadata in FilterView).

src/core/models/facetview.js
```diff
@@ -0,0 +1,34 @@
...
+export default class FacetView {
+  constructor (facet, metadatas) {
+    /**
+     * The exactly formatted filter to send to the backend.
+     * @type {Facet}
+     */
+    this.facet = facet;
+
+    /**
+     * Metadata for the filter.
```
tmeyer2115: facet?

src/core/models/facetview.js
```diff
@@ -0,0 +1,34 @@
...
+
+import Facet from './facet';
+
+/**
+ * Contains a {@link Facet} and associated metadata.
+ */
+export default class FacetView {
+  constructor (facet, metadatas) {
+    /**
+     * The exactly formatted filter to send to the backend.
```
tmeyer2115: Not sure what's meant by exactly formatted. Should we be referring to it as a filter?

src/core/models/filterview.js
```diff
@@ -0,0 +1,81 @@
...
+     * Display string for the field name.
+     * @type {string}
+     */
+    this.displayField = displayField;
+
+    /**
+     * Display values for the filter's metadata.
+     * @type {Array<string>}
+     */
+    this.displayValues = [ displayValues ].flat();
```
tmeyer2115: Do we also want to have a separator here for the displayValues?

src/core/models/filtermetadata.js
```diff
@@ -0,0 +1,22 @@
+/** @module FilterMetadata */
+
+/**
+ * Metadata helper for {@link Filter} or {@link Facet}
+ */
+export default class FilterMetadata {
```
tmeyer2115, line 6: Even though it will be simple, let's fill out this class a bit. 
constructor(metadata) {
   this._metadata = metadata;
}

get() {
  return this._metadata;
}

src/core/models/filtermetadata.js
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * Combines several filter metadata
+   * @param  {Array<Object>} metadataArray
```
tmeyer2115: Array of FilterMetadata

src/core/models/filtermetadata.js
```diff
@@ -0,0 +1,22 @@
...
+   */
+  static combine (metadataArray) {
+    return metadataArray.reduce((group, metadata) => {
```
tmeyer2115: Let's update this to work with the filled-out FilterMetadata class.

src/core/models/filterview.js
```diff
@@ -0,0 +1,49 @@
...
+     */
+    this.metadata = metadata;
+    Object.freeze(this);
+  }
+
+  /**
+   * Ands together multiple filterViews.
+   * @param  {...FilterView} filterViews
+   */
+  static and (...filterViews) {
```
tmeyer2115: I think we should just expose one static method for combining filters. I think it's a bit difficult for someone reading the code to understand the difference between combineFilterViews and the and method. We could make and a private static method by renaming it _and. Anything that calls _and would be updated to call combineFilterViews. Do we also need to handle the case of or'ing filter views together? Could we add a param to combineFilterViews that indicates how they should be combined?

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -132,26 +133,29 @@ export default class DateRangeFilterComponent extends Component {
...
+    const filterView = this._buildFilterView();
     if (this._storeOnChange) {
-      this.core.setFilter(this.name, filter);
+      this.core.setFilterView(this.name, filterView);
     }
     this.core.persistentStorage.set(`${this.name}.min`, this._date.min);
     this.core.persistentStorage.set(`${this.name}.max`, this._date.max);
 
-    this._onChange(filter);
+    this._onChange(filterView.filter, filterView.metadata);
```
tmeyer2115, line 143: I think we should maintain the existing contract we have with onChange. It is only passed back the actual Filter data. I don't think people should be writing onChange handlers that depend on any FilterMetadata. The FilterMetadata is pretty internal to the SDK.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -203,14 +204,14 @@ export default class FilterBoxComponent extends Component {
           showReset: this.config.resetFilter,
           resetLabel: this.config.resetFilterLabel,
           showExpand: this.config.expand,
-          onChange: (filter) => {
-            this.onFilterChange(i, filter);
+          onChange: (filter, metadata) => {
```
tmeyer2115, line 207: Same thing about onChange

src/core/models/filterview.js
oshi97: Or-ing together filters is only used in filteroptions (and by extension filterbox which uses it and facet with uses filterbox) so or-ing together filter views currently isn't needed

src/ui/components/filters/daterangefiltercomponent.js
oshi97: This is needed by FilterBox to update the FilterViews in global storage, it shouldn't have any impact on preexisting onChange functions because they should only be passing back one parameter anyways

src/core/models/filtermetadata.js
oshi97: I think it makes sense to add Object.assign(this, metadata) as a constructor and wrap any metadata with this class but anything beyond that feels like just adding get() calls everywhere without really having cleaner code or anything

If we need to add additional properties in the future I don't think it would hard to refactor things

tmeyer2115 (approved):

Issue #508, opened 2020-02-24
Uncaught ReferenceError: google is not defined


creotutar:
On dist-v.0.12.1

When navigating on universal results to a search with a map component, getting error
"Uncaught ReferenceError: google is not defined" causing the maps to not load

This is pointing to references of the global variable "google" in
src/ui/components/map/providers/googlemapprovider.js

When setting the timeout to a bigger number (1000)
src/ui/components/map/providers/googlemapprovider.js#100
the maps load correctly again

alexisgrow:
Thanks @creotutar! I'm investigating this, I'll post here when I have an update.

alexisgrow:
Code has been pushed to fix this issue. We're cutting release v0.12.2 right now.

PR #510, opened 2020-02-24
Maps Component: Enforce order of loadJS and init

alexisgrow:
Github Issue: #508
TEST=auto,manual

Added the maps component to my local answers experience. Verified
that the component loads properly.


src/ui/components/map/mapcomponent.js
```diff
@@ -57,11 +57,10 @@ export default class MapComponent extends Component {
 
   onCreate () {
     this._map = this.getProviderInstance(this._mapProvider);
-    this._map.loadJS();
   }
 
   onMount () {
-    this._map.onLoaded(() => {
+    this._map.loadJS(() => {
```
creotutar: The javadoc for _onLoaded says: "Callback to invoke once the Javascript is loaded"

Is this not what we want? A look at the loadJS function for googlemapprovider looks like this callback will be called if a #yext-map-js element exists otherwise it won't be called.

creotutar: Sorry took a jump there, onLoaded is using `    this._onLoaded = cb;`, that's where my first statement is coming from

alexisgrow: Eek you right, updating

tmeyer2115 (approved):

PR #511, opened 2020-02-26
AppliedFiltersComponent

oshi97:
This commit adds the AppliedFiltersComponent,
which is used internally only within vertical results
to display results count, applied filters, and applied
sorts in the header.

TEST=manual
Created a test site with all filters except FilterSearch,
checked that filters and styling worked as expected.
There are issues with facets not updating correctly or
at least weirdly from the url, however this issue was present
prior to this change.


src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+import Component from '../component';
+
+import StorageKeys from '../../../core/storage/storagekeys';
+import FilterMetadata from '../../../core/models/filtermetadata';
+import { DOM } from '../..';
+
+export default class AppliedFiltersComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+    this.moduleId = StorageKeys.VERTICAL_RESULTS;
```
tmeyer2115: Are applied filters used with universal results as well?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+     * Display the count of results at the very top of the results
+     * @type {boolean}
+     */
+    this._config.showResultsCount = config.showResultsCount === undefined ? true : config.showResultsCount;
+
+    /**
+     * If present, show the filters that were ultimately applied to this query
+     * @type {boolean}
+     */
+    this._config.showAppliedFilters = config.showAppliedFilters === undefined ? true : config.showAppliedFilters;
```
tmeyer2115: Why do we have this boolean in this class? If this is false, shouldn't we just not render the AppliedFiltersComponent? Is this component doing something else besides showing the applied filters?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+     * @type {boolean}
+     */
+    this._config.showFieldNames = config.showFieldNames || false;
+
+    /**
+     * If showAppliedFilters is true, this is list of filters
+     * that should not be displayed (common one is entity type)
+     * @type {Array<string>}
+     */
+    this._config.hiddenFields = config.hiddenFields || [];
```
tmeyer2115: Should entity type be a default here?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * If showResultsCount and showAppliedFilters are true,
+     * display this a separator between the result count and the applied query filters
+     * @type {string}
+     */
+    this._config.resultsCountSeparator = config.resultsCountSeparator || '|';
```
tmeyer2115: What would you think about making a ResultsHeader component that invokes the AppliedQueryFilter component? The ResultsHeader would be responsible to rendering results count.

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+   * because currently we cannot loop through an object in handlebars, only arrays.
+   * @returns {Array<Object>}
+   */
+  getAppliedFilters () {
+    const facets = this.core.globalStorage.getAll(StorageKeys.FACET_FILTER_VIEW);
+    const facetMetadata = facets.length > 0 ? facets[0].metadata : {};
+    const filterViews = this.core.globalStorage.getAll(StorageKeys.FILTER_VIEW);
+    const filterMetadata = filterViews.map(fv => fv.metadata);
+    const metadata = FilterMetadata.combine([...filterMetadata, facetMetadata]);
+    return Object.entries(metadata).map(([displayField, displayValues]) => ({
```
tmeyer2115: Does Object.entries(metadata) not return an array of displayField, displayValues pairs?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Combines {@link FilterView} and {@link FacetView}
+   * into an object of displayField to displayValues.
+   * This has to be parsed into an array of objects with displayField and displayValues
+   * because currently we cannot loop through an object in handlebars, only arrays.
+   * @returns {Array<Object>}
+   */
+  getAppliedFilters () {
```
tmeyer2115, line 123: I might have missed this in a previous CR, but where are we pulling in the NLP filters from the back-end? It seems like these filters might violate our assumption that only one filter can be applied to a field. In the spec, it seems like multiple NLP filters can be applied to a field.

src/ui/templates/results/appliedfilters.hbs
```diff
@@ -0,0 +1,45 @@
...
+    {{#each appliedFilters}}
+      {{#if ../_config.showFieldNames}}
+        <div class="yxt-AppliedFilters-displayField">{{this.displayField}}:</div>
+      {{/if}}
+      {{#each this.displayValues}}
+        <div class="yxt-AppliedFilters-displayValue">{{this}}</div>
+      {{/each}}
+    {{/each}}
+
+    {{#each appliedQueryFilters}}
```
tmeyer2115: What's the difference between appliedQueryFilters and appliedFilters?

tmeyer2115 (changes requested):

src/ui/components/results/appliedfilterscomponent.js
oshi97: It does but I think we prefer to have an array of objects not array of arrays, so that in the handlebars we dont have to do some weird loop.  maybe the syntax I used is unclear, if so let me know, but it's creating objects of
`{
displayField: 'alkdjlfasdj',
displayValues: ['1','2','3312413']
}
`

src/ui/components/results/appliedfilterscomponent.js
oshi97: I asked rose what we should do about these. These can be combined pretty easily with the other filter metadata, but looking into the later mocks if we want to be able to click an applied filter in the header and remove it for the search that obviously wouldn't work for any nlp filters

src/ui/templates/results/appliedfilters.hbs
oshi97: these are the nlp filters from the backend, and are also what they are called in the response. I felt uncomfortable with the current naming but also uncomfortable with renaming it to be inconsistent with the response

src/ui/components/results/appliedfilterscomponent.js
oshi97: Yes

src/ui/components/results/appliedfilterscomponent.js
oshi97: Created ResultsHeaderComponent

src/ui/components/results/appliedfilterscomponent.js
oshi97: created!

src/ui/components/results/appliedfilterscomponent.js
oshi97: was being used to get the appliedQueryFilters (nlpFilters) but probably better to pass them in directly

src/ui/components/results/appliedfilterscomponent.js
oshi97: Rose says we don't need to worry about it

src/core/models/filtermetadata.js
```diff
@@ -2,27 +2,48 @@
 
 /**
  * Metadata helper for {@link Filter} or {@link Facet}
+ * Metadata structure is
+ * {
+ *  c_birdTypes: {
+ *    displayLabel1: [displayValue1, displayValue2],
+ *    displayLabel2: [displayValue3, displayValue4],
+ *    'Your Favorite Bird': ['rooster', 'penguin', 'turkey', 'chikn', 'dog']
```
tmeyer2115, line 8: Why are there multiple display labels and lists of display values under c_birdTypes? Won't there only be one display name for the filter and one list of display values for it? Something like:

c_birdTypes: {
   displayName: 'Name';
   displayValues: [...]
}

src/core/models/section.js
```diff
@@ -83,6 +84,7 @@ class AppliedQueryFilter {
   constructor (appliedQueryFilter) {
     this.key = appliedQueryFilter.key || appliedQueryFilter.displayKey;
     this.value = appliedQueryFilter.value || appliedQueryFilter.displayValue;
+    this.fieldId = Filter.getFilterKey(appliedQueryFilter.filter);
```
tmeyer2115, line 87: Where's fieldId used here?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,164 @@
...
+     * @type {Array<string>}
+     */
+    this._config.hiddenFields = config.hiddenFields || [ 'builtin.entityType' ];
+
+    /**
+     * If showResultsCount and showAppliedFilters are true,
+     * display this a separator between the result count and the applied query filters
+     * @type {string}
+     */
+    this._config.resultsCountSeparator = config.resultsCountSeparator || '|';
```
tmeyer2115: Do we still need this here? Could we not put this logic in the ResultsHeader component?

src/ui/components/results/appliedfilterscomponent.js
tmeyer2115: What does she mean by that? Like we'll deal with it later? Does the back-end currently send us these NLP filters back?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -0,0 +1,92 @@
...
+     * display this a separator between the result count and the applied query filters
+     * @type {string}
+     */
+    this._config.resultsCountSeparator = config.resultsCountSeparator || '|';
+
+    /**
+     * Results count
+     * @type {number}
+     */
+    this._config.resultsCount = config.resultsCount || 0;
```
tmeyer2115: Why is resultsCount part of the config? Isn't that data we can just pull from the globalStore?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -0,0 +1,92 @@
...
+
+    this._config.appliedFiltersOpts = {
+      /**
+       * NLP filters sent back from the backend
+       * @type {Array<AppliedQueryFilter}
+       */
+      appliedQueryFilters: config.appliedQueryFilters || [],
```
tmeyer2115: Same thing here, these don't seem like config options. Wouldn't we just get this from the globalStore?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -0,0 +1,92 @@
...
+       * @type {boolean}
+       */
+      showFieldNames: config.showFieldNames || false,
+
+      /**
+       * If showAppliedFilters is true, this is list of filters
+       * that should not be displayed (common one is entity type)
+       * @type {Array<string>}
+       */
+      hiddenFields: config.hiddenFields || [ 'builtin.entityType' ]
```
tmeyer2115: I would assume that Rose is aware of this, but seems kinda difficult for customers to use the hiddenFields if they have to know what the API names are. Not sure of a way around this though

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -58,17 +58,56 @@ class VerticalResultsConfig {
...
+       * Display the count of results at the very top of the results
+       * @type {boolean}
+       */
+      showResultsCount: config.showResultsCount === undefined ? true : config.showResultsCount,
+
+      /**
+       * If present, show the filters that were ultimately applied to this query
+       * @type {boolean}
+       */
+      showAppliedFilters: config.showAppliedFilters === undefined ? true : config.showAppliedFilters,
```
tmeyer2115, line 111: Could we make the appliedFilterOpts object here? Then the ResultsHeader wouldn't have to do it and would be able to pass it straight through.

src/ui/components/results/appliedfilterscomponent.js
oshi97: The backend sends us back these NLP filters in the response for vertical search. She said we shouldn't worry about them filtering on the same fields as a static filter, we can just display them separately

src/core/models/filtermetadata.js
oshi97: Yeah you're right the user should not be adding multiple filters for the same fieldId with different labels, I still think it would be good to support it in the code with for loops or something

src/core/models/section.js
oshi97: Oh this is only used for removing hidden fields in the AppliedFiltersComponent, since otherwise that information is lost

src/ui/components/results/verticalresultscomponent.js
oshi97: updated!

src/ui/components/results/appliedfilterscomponent.js
oshi97: removed

src/ui/components/results/resultsheadercomponent.js
oshi97: ah so I wanted to avoid listening on global store since there's universal and also vertical results using this component

src/ui/components/results/resultsheadercomponent.js
oshi97: moving this to data

src/ui/components/results/resultsheadercomponent.js
oshi97: Yeah not sure what the best way around this is

oshi97:
CLOSING this pull request until the higher priority card/section items for hitchhikers are completed

PR #514, opened 2020-02-26
show ordinal option for cards

oshi97:


src/ui/sass/modules/_StandardCard.scss
```diff
@@ -23,13 +23,34 @@
...
 
+  &-ordinalWrapper {
+    margin-right: $standard-card-ordinal-size;
+  }
+
+  &-ordinal {
+    padding: $standard-card-ordinal-size * 0.2 $standard-card-ordinal-size * 0.35;
+    border-radius: $standard-card-ordinal-size;
+    width: $standard-card-ordinal-size * 0.6;
+    height: $standard-card-ordinal-size * 0.8;
```
alexisgrow, line 44: what is the desired behavior if content is bigger than this width/height? if it looks OK from a UI perspective, a standard option is to set a width and a min-height (instead of exact height)

alexisgrow (approved):

src/ui/sass/modules/_StandardCard.scss
oshi97: ah this is basically the index of the card in the search results (1,2,3,4...) so I think something like min-width and height would be good. It definitely does break with larger numbers

PR #515, opened 2020-02-26
qa styling changes for vertical results

oshi97:


src/ui/sass/modules/_Results.scss
```diff
@@ -154,10 +157,13 @@
     list-style: none;
     display: flex;
     flex-flow: row wrap;
+    margin-right: -1 * $results-cards-margin;
```
alexisgrow, line 160: ooh is there a way to do this without negative margins? generally it's good to avoid them, they make things overlap and behave in weird ways

alexisgrow (approved):

src/ui/sass/modules/_Results.scss
oshi97: I tried to think about it for a bit and I think this is the only way? Because I don't think the flexboxes have any way of knowing if they are on the edge of the screen or not, unless we either use javascript or change this to display: grid

PR #517, opened 2020-02-26
Polyfills: Move polyfills out of global namespace

alexisgrow:
Move away from babel/polyfill, upgrade core-js to version 3.

TEST=auto,manual

Pull client sites and test that on IE11 the polyfill works.
Initialize answers on dev test sites using all components.


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -71,15 +71,15 @@ function legacyBundle () {
             '@babel/preset-env',
             {
               'loose': true,
-              'modules': false,
-              'corejs': 3,
-              'useBuiltIns': 'entry'
```
alexisgrow, line 76: note to self: 
"useBuiltIns and @babel/plugin-transform-runtime are mutually exclusive. Both are used to add polyfills: the first adds them globally, the second one adds them without attaching them to the global scope."

so removing this in favor of the transform-runtime plugin below 

src: https://github.com/babel/babel/issues/10271#issuecomment-528379505

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -71,15 +71,15 @@ function legacyBundle () {
             '@babel/preset-env',
             {
               'loose': true,
-              'modules': false,
-              'corejs': 3,
```
alexisgrow, line 75: note to self:

stdout says the `corejs` option is useless/irrelevant w/o the `useBuiltIns` option, removing both

package.json
```diff
@@ -40,7 +41,7 @@
     "rollup": "^1.4.1",
     "rollup-plugin-babel": "^4.3.2",
     "rollup-plugin-commonjs": "^9.2.1",
-    "rollup-plugin-insert": "^1.1.0",
+    "rollup-plugin-insert": "1.1.0",
```
alexisgrow: note to self: this is a separate dependency thing that is breaking the circleci build, it should not be merged as part of the polyfill item

conf/gulp-tasks/library.gulpfile.js
alexisgrow: second source, maintainer saying docs need to be updated https://github.com/babel/babel/issues/10121#issuecomment-505005586

package.json
```diff
@@ -4,18 +4,20 @@
...
     "handlebars": "^4.1.2",
+    "regenerator-runtime": "^0.13.3",
     "template-helpers": "^1.0.1"
   },
   "devDependencies": {
     "@babel/core": "^7.3.4",
     "@babel/plugin-proposal-object-rest-spread": "^7.5.5",
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@babel/plugin-transform-arrow-functions": "^7.2.0",
+    "@babel/plugin-transform-classes": "^7.8.3",
```
alexisgrow: i actually don't think this is relevant/necessary any more

package.json
tmeyer2115: Should this line be removed, then?

alexisgrow: Yeah I was going to ask you about that! It started breaking our CircleCI build, so we fixed w/ this PR (https://github.com/yext/answers/pull/505). The fix is merged into the v13.0 branch, but is still an issue on this v12.2 branch. Should we pull https://github.com/yext/answers/pull/505 into this (v12.2) branch?

b1lly:
LGTM if LGTY @tmeyer2115

package.json
alexisgrow: Removing!

tmeyer2115 (approved):

PR #518, opened 2020-02-27
Add AccordionCardComponent

oshi97:
This commit adds the AccordionCardComponent which is
a card template that looks like an accordion.
Also re-adds resultsaccordion.hbs which should not have been deleted and is still used by universal results.
TEST=manual
Tested that card template loaded correctly, has proper
cardMappings, ctas fires analytics events correctly and goes to url,
opening and closing accordion fires analytics event and changes
aria-expanded for toggle and aria-hidden for content, and that animation is smooth.


src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -0,0 +1,136 @@
...
+  .yxt-CTA .Icon 
+  {
+    height: $accordion-card-cta-svg-size;
+    width: $accordion-card-cta-svg-size;
+    margin-right: $accordion-card-spacing * 0.3;
+  }
+}
+
+
+.yxt-Card {
```
oshi97: This is icky but there's I couldn't find another way to reference whether a certain AccordionCard component was the last one, since it has to be wrapped in a container div thanks to how data-component works, and also can't just remove the yxt-AccordionCard container in accordion.hbs and change the data-component container class to yxt-AccordionCard because we need be able to select it inside the javascript

Just having border: border-default without some kind of logic like this results in a double border and I just KNOW somebody is going to yell at me for that

README.md
```diff
@@ -484,7 +484,7 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 3,
   // The card used to display each individual result, see [Cards](#Cards) section for more details,
   card: {
-    // Optional: The type of card, currently only 'Standard' is supported, defaults to 'Standard'
+    // Optional: The type of card, currently only 'Standard' and 'Accordion' ard supported, defaults to 'Standard'
```
creotutar: typo

README.md
```diff
@@ -687,7 +687,26 @@ const cardMappings = {
...
+```js
+const cardMappings = {
+  // Title for the card, defaults to the name of the entity
+  title: item.title,
+  // Subtitle, defaults to null
+  subtitle: `Department: ${item._raw.name} `,
+  // Details, defaults to the entity's description
+  details: item._raw.description,
+  // Whether the accordion should be open on page load, defaults to false
+  expanded: false
```
creotutar, line 710: does this mean only the first will be expanded or all will be expanded on load?

creotutar:
@oshi97 I'm seeing a commit for the QA outside of adding AccordionCard, is that intentional? possible to split into two different PRs?

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -134,6 +121,7 @@ export default class StandardCardComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(new StandardCardConfig(config), systemConfig);
     this.hideExcessDetails = this._config.showToggle;
+    console.log(this);
```
oshi97: Ah shoot I'll remove this

README.md
oshi97: Ah so cardMappings are on a per card basis, would "Whether this cardMapping's accordion should be..." work better?

creotutar: Ah no, what you have current makes sense!

src/ui/components/cards/cardcomponent.js
```diff
@@ -61,10 +67,34 @@ export default class CardComponent extends Component {
...
+  /**
+   * Used by children card components like StandardCardComponent to
+   * apply given template mappings as config.
+   * @param {Result} result
+   * @param {Object|Function} cardMappings
+   */
+  static applyCardMappings (result, cardMappings) {
+    const config = {};
+    if (typeof cardMappings === 'function') {
+      cardMappings = cardMappings(result);
```
creotutar, line 99: is this meant to be `config=cardMappings(result)`?

creotutar: if that's the case, we may want to go through the entries and add to config as you're doing below

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -18,23 +19,9 @@ class StandardCardConfig {
...
-    if (typeof cardMappings === 'object') {
-      Object.entries(cardMappings).forEach(([attribute, value]) => {
-        if (typeof value === 'function') {
-          this[attribute] = value(this.result);
-        } else {
-          this[attribute] = value;
-        }
-      });
-    }
+    Object.assign(this, CardComponent.applyCardMappings(this.result, cardMappings));
```
creotutar: do we want to give the user control over the `this` variables in this component? if i'm reading this correctly if i give the cardMapping

```
{
  result: 'hacked result!'
}
```

then the result variable should be overridden and changed for the other lines relying on this.result?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -78,6 +71,17 @@ export default class VerticalResultsComponent extends Component {
     this.moduleId = StorageKeys.VERTICAL_RESULTS;
     this.results = [];
     this.placeholderCount = 0;
+
+    /**
+     * verticalKey used for analytics and passed to children
+     * @type {string}
+     * @private
+     */
+    this.verticalKey = this._config.verticalConfigId;
```
creotutar: if this is a private variable (from js doc above), do we want to follow the `_` naming convention?

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -0,0 +1,137 @@
...
+
+  .yxt-CTA .Icon 
+  {
+    height: $accordion-card-cta-svg-size;
+    width: $accordion-card-cta-svg-size;
+    margin-right: $accordion-card-spacing * 0.3;
+  }
+}
+
+.yxt-Card
```
creotutar: why do we need this parent class? is there a situation when we have a `.yxt-AccordionCard` without a `.yxt-Card` parent?

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -0,0 +1,137 @@
...
+    @include Text(
+      $size: $accordion-card-details-size,
+      $color: $accordion-card-details-color,
+      $weight: $accordion-card-details-weight,
+      $line-height: $accordion-card-details-line-height,
+    );
+  }
+
+
+  .yxt-CTA
```
creotutar: in the ideal modularized scss world, only AccordionCard module styling would live in this file. if we needed a CTA that had specific stying /only/ for AccordionCard, then we would have a CTA variant etc: `.yxt-CTA--accordionCard` which would live in the `CTA.scss`

this makes it easier to trace scss styling as all module styles live in one file

this may be difficult to achieve as i dont think the library is set up to pass around classes given some contexts. do you have any insight here?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,159 @@
...
+
+export default class AccordionCardComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(new AccordionCardConfig(config), systemConfig);
+
+    /**
+     * Whether the accordion is collapsed or not.
+     * @type {boolean}
+     */
+    this.isExpanded = false;
```
creotutar: this is always false? where does the config option come to play?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,159 @@
...
+   * it doesn't rip the whole component off of the page and remount it.
+   * Also reports an analytics event.
+   * @param {HTMLElement} toggleEl the toggle element
+   * @param {HTMLElement} iconEl the icon element
+   * @param {HTMLElement} contentEl the content element
+   * @param {HTMLElement} accordionEl the root accordion element
+   */
+  handleClick (toggleEl, iconEl, contentEl, accordionEl) {
+    this.isExpanded = !this.isExpanded;
+    iconEl.classList.toggle('yxt-AccordionCard-icon-expanded');
```
creotutar: if we are following SUIT naming conventions, this class should be `yxt-AccordionCard-icon--expanded` (double hyphens to indicate a modifier)

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,159 @@
...
+    const event = new AnalyticsEvent(this.isExpanded ? 'ROW_COLLAPSE' : 'ROW_EXPAND')
+      .addOptions({
+        verticalKey: this._config.verticalKey,
+        entityId: this._config.result._raw.id
+      });
+    this.analyticsReporter.report(event);
+  }
+
+  onMount () {
+    const toggleEl = DOM.query(this._container, '.yxt-AccordionCard-toggle');
```
creotutar: if we are querying DOM elements in the JS, we should be targeting `js-*` classes to separate concerns of styling and js selector classes

creotutar (changes requested):
Looks great! A few minor things about element selectors. Also want to make sure we aren't letting user options unintentionally override our class variables.

src/ui/components/cards/standardcardcomponent.js
oshi97: OH wow that's a really cool catch, no they shouldn't be allowed to mess with the result

src/ui/components/results/verticalresultscomponent.js
oshi97: Yes, updated!

src/ui/sass/modules/_AccordionCard.scss
oshi97: Ok I think this is a really good suggestion for the long run to make styling easier, I tried to do this by passing in a context config option to the CTAs and then inserted that as an additional class like .yxt-CTA-AccordionCard

src/ui/sass/modules/_AccordionCard.scss
oshi97: this was for the :first-child selector

src/ui/components/cards/cardcomponent.js
oshi97: soo I had it as `cardMappings = cardMappings(result)` instead of `config = cardMappings(result); } else if...` just for the case where somebody could function cardMappings that returned an object with functions inside it. 

Tom convinced me that isn't behavior we want to encourage, but I still feel like it would be good to handle it just in case?

oshi97: I think I should remove the `if (typeof cardMappings === 'object') {` check, which might make the code a little cleaner and maybe things should error if your function cardMappings doesn't return an object?

src/ui/components/cards/accordioncardcomponent.js
oshi97: ah ok makes sense

src/ui/components/cards/accordioncardcomponent.js
oshi97: updated!

src/ui/components/cards/accordioncardcomponent.js
oshi97: updated!

src/ui/sass/modules/_AccordionCard.scss
oshi97: updated so that it uses .yxt-AccordionCard--{{index}} selectors

oshi97:
Sorry I didn't realize this would close when I deleted vr/qa

PR #519, opened 2020-02-27
Remove templateUrl option from the SDK.

tmeyer2115:
This PR removes the templateUrl option from the SDK. This functionality led
to a security vulnerability as we were dynamically adding a <script> tag
to the page with whatever the templateUrl was. With this change, a client is
still able to supply a custom bundle to Answers, but it must be done using the
templateBundle attribute. If no templateBundle is supplied, we will fetch the
default set of templates for the current release.

TEST=manual

Made sure that I could supply a custom bundle via templateBundle. Also made
sure that if I did not supply a custom bundle, the default was fetched from
the CDN.


src/ui/rendering/defaulttemplatesloader.js
```diff
@@ -57,16 +45,10 @@ export default class TemplateLoader {
...
       });
-
       DOM.append('body', script);
-    })
-      .then((response) => {
-      // TODO(billy) Implmenet error handling here (e.g. request could fail)
-        console.log('Templates loaded successfully!');
-      });
-    return this;
+    });
```
alexisgrow, line 51: Do you think it would be helpful to add error handling if the request fails here? I'm not sure what the current/default behavior looks like.

README.md
```diff
@@ -67,7 +67,6 @@ Below is a list of configuration options that can be used during initialization.
 | onReady  | function     | Invoked when the Answers component library is loaded/ready | required  |
 | onStateChange | function | Invoked when the sate changes | optional |
 | useTemplates | boolean   | default: `true`.  If false, don't fetch pre-made templates. Only use this if you plan to implement custom renders for every component!  | optional  |
```
alexisgrow, line 69: Does an implementer need both `useTemplates` and the `templateBundle`? 

From testing, my understanding is that `useTemplates` doesn't do anything. 
-> Adding `useTemplates` without specifying `templateBundle` (or `templateUrl`) fetches the templates
-> Adding `templateBundle` without `useTemplates` doesn't fetch templates and uses provided bundle

tmeyer2115: useTemplates is kinda wonky. If someone sets useTemplates to false, we assume they provide a custom render function for every component. So, we won't fetch any templates for them. Technically, and this is an existing bug, they could specify useTemplates as false and provide a templateBundle and we'd actually use their template bundle. This is confusing behavior and we should update it at some point. If useTemplates is false, we should ignore any bundle that's provided too. 

If someone provides a templateBundle, we will not fetch anything from the CDN, regardless of the value of useTemplate.

src/ui/rendering/defaulttemplatesloader.js
tmeyer2115: It would be useful to have some error handling here. In the previous TemplateLoader, there was no error handling. I can chat with Rose about what we might want to do here.

README.md
alexisgrow: Woah, that is weird. Haha. Yes I agree that we should fix, but seems like it's out of scope for this. LGTM!

src/ui/rendering/defaulttemplatesloader.js
alexisgrow: Oh sweet, sgtm as a follow-up.

alexisgrow (approved):

PR #520, opened 2020-02-28
Vr/qa

oshi97:
This PR should have everything required of Vertical Results for v0.13.0 EXCEPT for 1. AppliedFiltersComponent + ResultsHeaderComponent and 2. Accordion Card


README.md
```diff
@@ -538,7 +538,7 @@ Note: A CTA without both a label and icon will not be rendered.
 const callsToAction = [{
   // Label below the CTA icon, default null
   label: 'cta label',
-  // Icon name for the CTA that is one of the SDK icons
+  // Icon name for the CTA that is one of the SDK icons, default to undefined (no icon)
```
tmeyer2115, line 543: Is the spec out of date? It says the default should be star if the icon does not map?

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -64,11 +65,28 @@ export default class CTACollectionComponent extends Component {
...
+        // Assign default event options if none exist.
+        ctaObject.eventOptions = ctaObject.eventOptions || this.defaultEventOptions(result);
         return ctaObject;
       }
     });
   }
 
+  defaultEventOptions (result) {
+    const verticalKey = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
+    if (!verticalKey) {
```
tmeyer2115: It's possible they didn't add a SEARCH_CONFIG to their init, right? They could have just specified verticalKey when creating the SearchBar component? It's not part of result, is it?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -62,13 +63,21 @@ class VerticalResultsConfig {
      * @type {Object}
      */
     this.footer = config.footer || {};
+
+    /**
+     * Vertical/Universal results
```
tmeyer2115: Universal results? Seems weird that VerticalResultsConfig would potentially contain Universal results too. Is this referring to the fact that the UniversalResults component uses this too?

tmeyer2115: Seems a bit odd to be passing results as part of the config.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -62,13 +63,21 @@ class VerticalResultsConfig {
...
+     */
+    this.results = [];
   }
 }
 
 export default class VerticalResultsComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(new VerticalResultsConfig(config), systemConfig);
     this.moduleId = StorageKeys.VERTICAL_RESULTS;
+    this.results = [];
```
tmeyer2115, line 88: Do we need both this and the initialization of this.results in the Config class?

src/ui/templates/cards/standard.hbs
```diff
@@ -1,44 +1,60 @@
...
   <div class="yxt-StandardCard-wrapper">
-    {{#if _config.title}}
-      <div class="yxt-StandardCard-title">
-        <a {{#if _config.url}} 
-          href="{{_config.url}}" target={{#if _config.newWindow}}"_blank" rel="noopener noreferrer nofollow"{{else}}"_self"{{/if}}
-        {{/if}}>
-          {{_config.title}}
-        </a>
+    {{#if _config.showOrdinal}}
+      <div class="yxt-StandardCard-ordinalWrapper">
```
tmeyer2115: Do we need a wrapper div here since there is only one thing inside?

src/ui/templates/cards/standard.hbs
```diff
@@ -1,44 +1,60 @@
...
       </div>
     {{/if}}
-    {{#if _config.subtitle}}  
-      <div class="yxt-StandardCard-subtitle">
-        {{_config.subtitle}}
+    {{#if _config.image}}
+      <div class="yxt-StandardCard-imgWrapper">
+        <img src="{{_config.image}}">
+        {{#if _config.tagLabel}}
+          <div class="yxt-StandardCard-imgTagWrapper">
```
tmeyer2115, line 14: Same question about needing a wrapper div

src/ui/templates/results/resultsaccordion.hbs
```diff
@@ -0,0 +1,66 @@
+{{#if isSearchComplete}}
```
tmeyer2115, line 1: How come we needed to add all this accordion stuff?

README.md
oshi97: Yes this was another hidden spec update

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: stashing away the defaultEventOptions change for now since we decided to focus more on having all other feature be incredibly solid

src/ui/components/results/verticalresultscomponent.js
oshi97: removed!

src/ui/templates/cards/standard.hbs
oshi97: yes but some tweaks were still needed

src/ui/components/results/verticalresultscomponent.js
oshi97: only need this one, updated

src/ui/templates/cards/standard.hbs
oshi97: it's needed at least the way I did the scss - spent some time trying to think of a cleaner way but didn't want to spend too long on this

src/ui/templates/results/resultsaccordion.hbs
oshi97: I accidentally erased this when ripping out the ACTUALLY unused results item templates

src/ui/templates/utils/footer.hbs
```diff
@@ -1,13 +1,20 @@
...
+      target={{#if _config.newWindow}}"_blank"
+      rel="noopener noreferrer nofollow"{{else}}"_self"{{/if}}
+      aria-label="Vertical Results footer link"
+    >
   {{/if}}
->
+
   {{#if _config.image}}
-    <img src='{{_config.image}}'>
+    <img src='{{_config.image}}' alt='Vertical Results footer image'>
```
creotutar: If this image is for iconography or a logo, we usually omit the alt-tag as this is a purely visual part of the page and does not give meaning to someone unable to see the image

to keep this general, I would recommend to have an option for the alt tag instead of having a baked in alt tag. what do you think?

creotutar: nit: i am a big fan of using `"` for attributes and `'` when you have to

creotutar: for other WCAG purposes: we are required to have text for all links (we cannot have just images) so I would recommend having an `sr-only` text span here describing what the image link does. it may be best to have this passed in by the user because this image can be variable

oshi97: that sounds like a great way to do things! let me know if this looks okay

README.md
```diff
@@ -557,8 +561,10 @@ const callsToAction = [{
   analyticsEventType: 'CTA_CLICK',
   // Whether the click should open in a new window, defaults to false
   newWindow: false,
-  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. defaults to null
-  eventOptions: `{ "verticalKey": "credit-cards", "entityId": "123123", "searcher":"UNIVERSAL", "ctaLabel": "cards"}`
+  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. defaults to the below,
+  // Where a value in double brackets like {{VERTICAL_KEY_OF_SEARCH}} means the vertical key of the search, like 'Location',
+  // Not the literal string {{VERTICAL_KEY_OF_SEARCH}}
+  eventOptions: `{ "verticalKey": {{VERTICAL KEY OF SEARCH IF NOT UNIVERSAL}}, "entityId": {{ENTITY ID}}, "searcher":"VERTICAL"}`
```
tmeyer2115: Are we including this default in the QA changes? I thought we'd discussed doing this as a separate item. This functionality was communicated by Rose a couple of weeks ago, right?

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -116,7 +116,13 @@ class StandardCardConfig {
      * to contain CTA configuration.
      * @type {Array<string>}
      */
-    this.callsToActionFields = config.callsToActionFields || [];
+    this.callsToActionFields = this.callsToActionFields || [];
```
tmeyer2115, line 119: I have to say, I'm not a huge fan of the Object.assign(this, config) pattern. It makes it really unclear what all is in "this" and leads to some confusing lines of code like this or "this.subtitle = this.subtite". For lines like the latter, we have duplicate code because we want to communicate what's actually in "this". Can we just not use Object.assign(this, config) and instead have things like: this.subtitle = config.subtitle.

src/ui/components/ctas/ctacomponent.js
```diff
@@ -39,13 +39,14 @@ class CTAConfig {
     this.newWindow = config.newWindow;
 
     /**
-     * The eventOptions needed for the event to fire, passed as a string from templateMappings.
+     * The eventOptions needed for the event to fire, passed as a string or Object
+     * from cardMappings.
      * @type {Object}
      */
     if (typeof config.eventOptions === 'string') {
       this.eventOptions = JSON.parse(config.eventOptions);
```
tmeyer2115, line 47: Is it necessary to parse the config.eventOptions? In reportAnalyticsEvent, we just addOptions of _config.eventOptions, we don't seem to reference this.eventOptions

src/ui/components/ctas/ctacomponent.js
```diff
@@ -56,7 +57,7 @@ export default class CTAComponent extends Component {
 
   onMount () {
     const el = DOM.query(this._container, `.js-yxt-CTA`);
-    if (el) {
+    if (el && this._config.eventOptions) {
```
tmeyer2115, line 60: Should reporting of the analytics event be guarded by the presence of eventOptions in the config? If someone didn't specify any options, couldn't we still send an analytics event?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -78,16 +84,47 @@ export default class VerticalResultsComponent extends Component {
...
+      this.placeholderCount = placeholderCount;
+      this.setState(this._state.get());
+    }
+  }
+
   static get duplicatesAllowed () {
     return true;
   }
 
+  getPlaceholderCount () {
```
tmeyer2115: Is this stuff still accurate? I know you mentioned things changed with your maxNumberOfColumns PR

src/ui/templates/utils/footer.hbs
```diff
@@ -1,13 +1,25 @@
...
+    {{#if _config.screenReaderText}}
+      <span class="sr-only">{{_config.screenReaderText}}</span>
+    {{/if}}
   {{/if}}
->
+  
+
   {{#if _config.image}}
-    <img src='{{_config.image}}'>
+    <img src="{{_config.image}}" {{#if _config.altText}} alt="{{_config.altText}}" {{/if}}>
```
tmeyer2115: As Christian noted, an alt tag is needed only if this is not a purely visual part of the page. If it is purely visual, they won't lose any information if they are unable to see it. Is this image supposed to represent a logo in the footer?

src/ui/components/cards/standardcardcomponent.js
oshi97: There's additional base component config that has to be copied over, which was the main reason in the past I tried to apply default config after calling super

src/ui/components/ctas/ctacomponent.js
oshi97: this current line of code is within CTAConfig class so it ends up as this._config

src/ui/components/ctas/ctacomponent.js
oshi97: I think we still need some idea of a default event options then

src/ui/components/results/verticalresultscomponent.js
oshi97: that's my bad - I didn't want to close this pr, but the columns stuff came up later and I thought would be cleaner in a separate PR. That was reason I didn't post this specific PR as a ptal, only the other 3 (Accordion, filters, vertical results columns), but I should have been more explicit.

README.md
oshi97: removed from readme

src/ui/templates/utils/footer.hbs
oshi97: ah ok I took it out then, I think the idea is its just a logo

tmeyer2115 (approved):

PR #522, opened 2020-02-28
maxNumberOfColumns column resizing

oshi97:
This commit adds a resize handler for vertical results columns. While this isn't perfect it's better than trying to calculate each time the component is mounted whether it's necessary to add placeholder elements for purely styling purposes, and then calling setstate again.
TEST=manual
Tested that the yxt-Results-items-${numColumns} classname was being set correctly with different vertical results widths, and that the page resizes correctly.


src/ui/components/cards/cardcomponent.js
```diff
@@ -49,10 +49,14 @@ export default class CardComponent extends Component {
   }
 
   setState (data) {
+    let cardType = this._config.cardType;
+    if (!cardTypes[cardType]) {
+      cardType = 'Standard';
```
tmeyer2115: I think defaulting to 'Standard' is the right move here. I also think it would be good to throw an AnswersError here letting someone know they tried to use a card that's not valid.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -69,6 +69,18 @@ class VerticalResultsConfig {
...
+     * Min-width of the card, used when calculating number of columns needed.
+     * @type {number}
+     */
+    this._cardMinWidth = config._cardMinWidth || 210;
+
+    /**
+     * Breakpoint width for desktop, used when calculating number of columns needed.
+     * @type {number}
+     */
+    this._desktopBreakpoint = config._desktopBreakpoint || 400;
```
tmeyer2115: I might be wrong, but these don't seem like things that should be in the config. This seems like something somebody would configure with CSS styling.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -87,40 +100,52 @@ export default class VerticalResultsComponent extends Component {
...
-      if (!itemsEl || !cardEl) {
-        return 0;
-      }
-      const totalWidth = itemsEl.scrollWidth + 32;
-      const cardWidth = cardEl.scrollWidth;
-      const numColumns = Math.floor(totalWidth / cardWidth);
-      const numPlaceholders = (numColumns - resultsCount % numColumns) % numColumns;
-      return numPlaceholders;
+    const totalWidth = this._container.scrollWidth;
+    if (totalWidth <= this._config._desktopBreakpoint) {
```
tmeyer2115: Just want to confirm, on mobile we always have one column?

src/ui/components/results/verticalresultscomponent.js
tmeyer2115: Would it be possible to get this info from the DOM somehow? I think having this as a config object could introduce some complexity. I'm assuming somebody could alter the desktopBreakpoint or cardMinWidth using CSS (if that's not the case, lemme know). If they do that, they'd need to change it here too.

src/ui/components/cards/cardcomponent.js
```diff
@@ -49,10 +49,14 @@ export default class CardComponent extends Component {
   }
 
   setState (data) {
+    let cardType = this._config.cardType;
+    if (!cardTypes[cardType]) {
```
creotutar, line 54: do we want this to fail silently if an incorrect card type is given?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -87,40 +100,52 @@ export default class VerticalResultsComponent extends Component {
...
+      return 1;
+    }
+    const numColumns = Math.floor(totalWidth / this._config._cardMinWidth);
```
tmeyer2115: This seems a bit dangerous to me. Since cardMinWidth is a config object, someone could set it in such a way that numColumns is > 4.

tmeyer2115 (commented):
I'm really not sure what the best way to get cardMinWidth and desktopBreakpoint is. It seems odd having those in the config. They seem like CSS styling to me. I suggested trying to pull these values from the DOM somehow. Not sure how possible that would be.

src/ui/components/results/verticalresultscomponent.js
oshi97: Yes!

src/ui/components/results/verticalresultscomponent.js
oshi97: They would also have to set maxNumberOfColumns > 4 so I think it's okay (at least with regards to point)

src/ui/components/results/verticalresultscomponent.js
oshi97: Hmm originally these were hardcoded constants (the same way it's being done in NavigationComponent) but I thought maybe if consulting wants to do something different it might be easier for them to change this hidden hack variable. It's definitely possible we don't ever want to have this hack available or at least not until it's needed. 

The desktop-breakpoint can't be pulled from the dom but I also think it's okay since this is a static value given by Jeremy. 

For the card min width this could be pulled using getComputedStyle and getting the minWidth but I don't think this is actually better than just having the value somewhere in the js

src/ui/components/cards/cardcomponent.js
oshi97: Actually hmm this is doing a bit more than just defaulting to Standard but also using Standard if there are any typos. Thinking about it a bit more we should probably just throw an error here and not be graceful in case they made a typo or god forbid I make another typo in the code.

oshi97: Throwing an error is probably good here

src/ui/components/cards/cardcomponent.js
oshi97: No, I think throwing an error is a great idea

oshi97:
image
oops

src/ui/components/cards/cardcomponent.js
```diff
@@ -49,10 +50,17 @@ export default class CardComponent extends Component {
   }
 
   setState (data) {
+    const cardType = this._config.cardType;
+    if (!cardTypes[cardType]) {
+      const validCards = `["${Object.keys(cardTypes).join('", "')}"]`;
+      const msg = `Card type "${cardType}" is not recognized as a valid build-in card type.` +
```
creotutar: nit: built-in

src/ui/templates/results/verticalresults.hbs
```diff
@@ -10,7 +10,7 @@
              data-prop="map">
         </div>
       {{/if}}
-      <ul class="yxt-Results-items yxt-Results-items-{{_config.maxNumberOfColumns}}">
+      <ul class="yxt-Results-items {{#if numColumns}}yxt-Results-items-{{numColumns}}{{/if}}">
```
creotutar: nit: can we remove the space between the items and the if? the way we have it now, there will always be a trailing space after `yxt-Results-items`

oshi97: updated!

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -84,40 +97,52 @@ export default class VerticalResultsComponent extends Component {
...
+
+  handleResize () {
+    if (this._handleResizeTimer) {
+      clearTimeout(this._handleResizeTimer);
     }
-    return this._config.maxNumberOfColumns;
+
+    this._handleResizeTimer = setTimeout(() => {
+      const numColumns = this.getNumColumns();
+      if (this.numColumns !== numColumns && numColumns !== 1) {
```
tmeyer2115: What happens if numColumns is 1? Wouldn't we want to perform this logic? Or, is the styling handled by a media query in the CSS? If it is handled by a media query, could we just handle it this way? Would there be downsides to that?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -136,7 +162,8 @@ export default class VerticalResultsComponent extends Component {
       universalUrl: this._universalUrl ? this._universalUrl + window.location.search : '',
       showNoResults: this.results.length === 0,
       query: this.core.globalStorage.getState(StorageKeys.QUERY),
-      placeholders: new Array(this.placeholderCount)
+      placeholders: new Array(this._config.maxNumberOfColumns - 1),
```
tmeyer2115, line 169: Could you run me through the logic for creating max - 1 placeholders again? I can't remember it off the top of my head.

oshi97: the maximum number of cards in a row, at any point in time, is the `maxNumberOfColumns`. BUT placeholders are only required on the last row if there is at least one card, otherwise the placeholders don't do anything. 
i.e. we only need placeholders when there is at least one card on the last row.

src/ui/components/results/verticalresultscomponent.js
oshi97: no this needs to be performed when numColumns is 1, this was supposed to be if maxNumberOfColumns is 1 -> don't bother with any resize logic

tmeyer2115 (approved):

PR #524, opened 2020-02-28
No Results: UI updates

alexisgrow:
Make small front end changes found when testing the backend no-results response with the front end logic.

TEST=manual,auto

Change config to test multiple configurations, test using several queries.


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -171,10 +171,13 @@ export default class VerticalResultsComponent extends Component {
       };
       return super.addChild(data, type, newOpts);
     } else if (type === AlternativeVerticalsComponent.type) {
+      let results = this.core.globalStorage
```
creotutar: can we make this a const? or do you anticipate this reference changing?

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -116,5 +116,11 @@ export default class HandlebarsRenderer extends Renderer {
         ? ''
         : JSON.stringify(name);
     });
+
+    this.registerHelper('plural', function (number, singularText, pluralText) {
```
creotutar, line 120: really useful function helper :)

src/ui/components/results/verticalresultscomponent.js
alexisgrow: yes, def! updating 👾

creotutar (approved):
🚀 🚀 🚀 LGTM

src/ui/templates/results/alternativeverticals.hbs
```diff
@@ -1,18 +1,23 @@
...
       No results found
     </em>
-    in {{currentVerticalLabel}}. Showing
-    <em class="yxt-AlternativeVerticals-noResultsInfo--emphasized">
-      all {{currentVerticalLabel}}
-    </em>
-    instead.
+    in {{currentVerticalLabel}}.
+    {{#if isShowingResults}}
+      Showing
```
creotutar, line 8: how much of this text do we want hardcoded in the `.hbs` vs setting it in the javascript? keeping the text in the JavaScript lets us swap it out and keep it portable for if someone wants to override this text in the future (ie this text becomes configurable by the client)

(eg looking at the questionsubmission.hbs, we're keeping strings set in the javascript)

alexisgrow: Ahh so my understanding with the questionsubmission component is that the spec said that all of that text should be set by the user in the config. In this case, the only configurable things per the spec are a `displayAllResults` option and then a `template` option (so that a user could override this .hbs entirely). 

I'm not sure what the plan for these is moving forward, I can loop in Rose and ask how much of this default templates should be configurable?

tmeyer2115: I think we should stick to the spec for now. We can still reach out to Rose and ask if she wants this to be made configurable. If so, we can make an item in the backlog.

tmeyer2115 (approved):

src/ui/templates/results/alternativeverticals.hbs
alexisgrow: Sounds good to me. I'll merge this and then talk to Rose about this during her QA of the component.

PR #527, opened 2020-03-02
v0.12.2

tmeyer2115:
# Version 0.12.2

## Changes

* Expanded unit test coverage.
* Polyfills are no longer stored in the global namespace.

## Fixes

* Fixed the active navigation tab border in Safari. 
* Fixed a race condition preventing multiple maps from getting initialized on the same page.


alexisgrow (approved):
LGTM

PR #528, opened 2020-03-02
No Results: Update alternative response data model

alexisgrow:
The alternativeVerticals portion of the server response is now
an object instead of an array. Update parsing logic get data from
the alternativeVerticals.modules array.

TEST=manual

Use mock API response provided by back end teams and project manager.
Test by running answers experience locally and hardcoding the mock
response. UI behaves as expected.


src/core/models/alternativeverticals.js
```diff
@@ -17,6 +17,11 @@ export default class AlternativeVerticals {
    * @param {Object.<string, function>} formatters The field formatters to use
    */
   static from (response, formatters) {
-    return new AlternativeVerticals(Section.from(response.alternativeVerticals, {}, formatters));
+    const alternativeVerticals = response.alternativeVerticals;
+    if (!alternativeVerticals || !alternativeVerticals.modules) {
+      return {};
```
tmeyer2115: Do we want to return {} here or new AlternativeVerticals() ?

alexisgrow: Oh good point, that works, updating!

tmeyer2115 (approved):

PR #530, opened 2020-03-03
Implement AccordionCardComponent

oshi97:
This commit implements the AccordionCardComponent.
The opening and closing of the Accordion was implemented
without calling setState, because I believe that, in general,
setState should only be called when the actual data sent to the
template changes, not for purely presentational changes like
opening and closing the accordion.

Setting the height of the content el by setting style.height directly
equal to the scrollHeight, instead of something like height: 100%,
was done to facilitate smoother animation.

In order to better follow the SUIT css pattern, as well as better
co-locate styling, additional _context and _isSolo config options
were created for CTAComponent, and a _index config option was created
for the StandardCard and AccordionCard.

TEST=manual
Checked that accordion card template renders correctly, has proper
cardMappings, ctas fires analytics events correctly and goes to url,
opening and closing accordion fires analytics event and changes
aria-expanded, and that animation is smooth. Also checked that
the StandardCard still renders properly, with multi column support.


oshi97:
Ok wait one second I'm trying to rebase on top of the default-event-options branch and it's being dangusy!!!

oshi97:
Ok it's been rebased, tested that both standard and accordion cards render, have correct styling, and their ctas use default eventOptions when eventOptions are not specified

README.md
```diff
@@ -694,6 +704,23 @@ const cardMappings = {
...
+## Accordion Card
+
+The card mapping for an accordion card has these attributes
+
+```js
+const cardMappings = {
+  // Title for the card, defaults to the name of the entity
+  title: item.title,
+  // Subtitle, defaults to null
+  subtitle: `Department: ${item._raw.name} `,
```
creotutar, line 718: i dont see this being used in the hbs

creotutar: ah, looking at the wrong hbs,
are we not removing the previous AccordionResults component?

src/ui/templates/cards/accordion.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="yxt-AccordionCard js-yxt-AccordionCard
+  {{#if isExpanded}} yxt-AccordionCard--expanded {{/if}} yxt-AccordionCard--{{_config._index}}"
+  id="{{id}}">
+  <div class="yxt-AccordionCard-toggle js-yxt-AccordionCard-toggle"
```
creotutar: can we make this a button? best to have an indication that this is an interactive element

src/ui/templates/cards/accordion.hbs
```diff
@@ -0,0 +1,24 @@
...
+    aria-controls="{{id}}"
+    tabindex="0"
+  >
+    <div class="yxt-AccordionCard-title">
+      {{_config.title}}
+    </div>
+    <div class="yxt-AccordionCard-icon js-yxt-AccordionCard-icon {{#if isExpanded}} yxt-AccordionCard-icon--expanded {{/if}}"
+      data-component="IconComponent"
+      data-opts='{"iconName": "chevron"}'
+      data-prop="icon"></div>
```
creotutar: nit: closing tag on another line please

README.md
oshi97: Hmmm technically it's still being used in universal results, and while I think it would be an easy change to go into UniversalResults and change the logic surrounding useAccordion I don't think it's something we realllly want to get into testing since it's going to go away VERY soon with the universal sections.

src/ui/templates/cards/accordion.hbs
oshi97: updatedd!

src/ui/templates/cards/accordion.hbs
oshi97: yeah! Used to TextButton mix-in we have which looks super cool

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    const contentEl = DOM.query(this._container, '.js-yxt-AccordionCard-content');
+    const accordionEl = DOM.query(this._container, '.js-yxt-AccordionCard');
+
+    DOM.on(toggleEl, 'click', () => this.handleClick(toggleEl, iconEl, contentEl, accordionEl));
+  }
+
+  /**
+   * For passing functions to the config of children {@link CTACollectionComponent}
+   */
+  addChild (data, type, opts) {
```
creotutar, line 162: does this need to follow the new interface?

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -24,13 +38,36 @@ export default class CTACollectionComponent extends Component {
...
+      if (!cta.label && !cta.url) {
+        console.warn('Call to Action:', cta, 'is missing both a label and url attribute and is being automatically hidden');
+      } else if (!cta.label) {
+        console.warn('Call to Action:', cta, 'is missing a label attribute and is being automatically hidden');
+      } else if (!cta.url) {
+        console.warn('Call to Action:', cta, 'is missing a url attribute and is being automatically hidden');
+      } else {
+        return {
+          eventOptions: this.defaultEventOptions(this.result),
+	      _context: this._config._context,
```
creotutar: nit: extra tab?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,201 @@
...
+     * @type {Object}
+     */
+    const rawResult = result._raw || {};
+
+    /**
+     * The templateMappings attribute of the config
+     * is either a function that returns additional config for
+     * a card or an object that is the additional config.
+     */
+    let cardMappings = config.cardMappings || {};
```
creotutar: can this be a const?

src/ui/components/cards/accordioncardcomponent.js
oshi97: Yes, but we should do that in a separate commit

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,201 @@
...
+    this.verticalKey = config.verticalKey;
+
+    /**
+     * Title for the card
+     * @type {string}
+     */
+    this.title = this.title || result.title || rawResult.name || '';
+
+    /**
+     * Subtitle
```
creotutar: I'm not sure how useful these jsdoc descriptions are. Do we want to omit the ones that are just "this is a <variable name>"

src/ui/components/cards/accordioncardcomponent.js
oshi97: yes updated!

src/ui/components/cards/accordioncardcomponent.js
oshi97: I'll make them a little more informative

src/ui/templates/cards/accordion.hbs
```diff
@@ -0,0 +1,25 @@
+<div class="yxt-AccordionCard js-yxt-AccordionCard
+  {{#if isExpanded}} yxt-AccordionCard--expanded {{/if}} yxt-AccordionCard--{{_config._index}}"
+  id="{{id}}">
+  <button class="yxt-AccordionCard-toggle js-yxt-AccordionCard-toggle"
+    aria-expanded="{{#if isExpanded}}true{{else}}false{{/if}}"
+    aria-controls="{{id}}"
+    tabindex="0"
+  >
+    <div class="yxt-AccordionCard-title">
```
creotutar: because this is a title heading, it would ideally have an h-level element tag (e.g. h3 in the previous accordion card). however wcag states that we don't skip h-levels so it would require we always have an h1 and h2 parent. in the current configuration, we have h1 as the searchbar title and the h2 as the heading for the results. possible to get this to be an h3?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,202 @@
...
+   * @param {HTMLElement} toggleEl the toggle element
+   * @param {HTMLElement} iconEl the icon element
+   * @param {HTMLElement} contentEl the content element
+   * @param {HTMLElement} accordionEl the root accordion element
+   */
+  handleClick (toggleEl, iconEl, contentEl, accordionEl) {
+    this.isExpanded = !this.isExpanded;
+    iconEl.classList.toggle('yxt-AccordionCard-icon--expanded');
+    contentEl.classList.toggle('yxt-AccordionCard-content--expanded');
+    contentEl.classList.remove('yxt-AccordionCard--initExpanded');
```
creotutar: is this supposed to be `yxt-AccordionCard-content--initExpanded`?

creotutar: how is this different from its `--expanded` counterpart? is it not possible to put the height:auto on the expanded class?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,202 @@
...
+   * it doesn't rip the whole component off of the page and remount it.
+   * Also reports an analytics event.
+   * @param {HTMLElement} toggleEl the toggle element
+   * @param {HTMLElement} iconEl the icon element
+   * @param {HTMLElement} contentEl the content element
+   * @param {HTMLElement} accordionEl the root accordion element
+   */
+  handleClick (toggleEl, iconEl, contentEl, accordionEl) {
+    this.isExpanded = !this.isExpanded;
+    iconEl.classList.toggle('yxt-AccordionCard-icon--expanded');
```
creotutar: ideally, you would only need to put the `--expanded` modifier on the parent module class (ie AccordionCard) and then in the scss you can determined expanded through that parent class.

e.g.
```
.yxt-AccordionCard--expanded
{
  .yxt-AccordionCard-icon
  {
    //icon styling for expanded only
  }
}
```

src/ui/components/ctas/ctacomponent.js
```diff
@@ -47,6 +47,18 @@ class CTAConfig {
...
+     * Context for the cta
+     * @type {string}
+     */
+    this._context = config._context;
+
+    /**
+     * Whether the cta is the only one in its CTACollectionComponent
+     * @type {boolean}
+     */
+    this._isSolo = config._isSolo || false;
```
creotutar, line 61: if we are only using this property for `only-child` styling, can we remove this property and instead rely on the `:only-child` selector?

src/ui/components/cards/cardcomponent.js
```diff
@@ -35,6 +35,12 @@ class CardConfig {
      */
     this.callsToActionFields = config.callsToActionFields || [];
 
+    /**
+     * The index of the card.
+     * @type {number}
+     */
+    this._index = config._index || 0;
```
creotutar, line 42: why are we adding this as part of the AccordionCardComponent? it doesn't seem to be used

creotutar: can you remind me again why we needed the index? i vaguely remember this is for the grid styling but i don't remember its need if we added placeholders

src/ui/components/cards/cardcomponent.js
```diff
@@ -90,12 +96,37 @@ export default class CardComponent extends Component {
...
     return super.addChild(updatedData, type, newOpts);
   }
 
+  /**
+   * Used by children card components like StandardCardComponent to
+   * apply given template mappings as config.
+   * @param {Result} result
+   * @param {Object|Function} cardMappings
+   */
+  static applyCardMappings (result, cardMappings) {
```
creotutar, line 113: also have a similar question here: why is this being added as part of AccordionCardComponent? shouldn't this be part of some previous CardComponent PR?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -0,0 +1,202 @@
...
+    Object.assign(this, CardComponent.applyCardMappings(result, cardMappings));
+
+    /**
+     * Vertical key for the card, added to analytics events sent by this component.
+     * @type {string}
+     */
+    this.verticalKey = config.verticalKey;
+
+    /**
+     * Title for the card, located on the accordion toggle.
```
creotutar: I'm not sure we want these js docs tied to the relative UI layout. It feels to me like the layout could change and it would be very easy to forget to change this comment.

src/core/models/verticalresults.js
```diff
@@ -24,6 +24,7 @@ export default class VerticalResults {
    * Create vertical results from server data
    * @param {Object} response The server response
    * @param {Object.<string, function>} formatters The field formatters to use
+   * @param {string} verticalKey The verticalKey for the search
```
creotutar: a bit confused about why this is in the AccordionCardComponent PR

src/ui/components/cards/accordioncardcomponent.js
oshi97: when accordion card initialized as expanded there was some lag the first time you tried to close it - I think I fixed this by moving height out of the yxt-AccordionCard-content class and only setting to 0 if not expanded

src/ui/components/ctas/ctacomponent.js
oshi97: I would like to but the way the divs get nested when you add child components means it's either something like this or using 'parentclass:first-child > childclass' selectors

src/ui/components/cards/cardcomponent.js
oshi97: before what was being done in verticalresults.hbs, in order to loop through all results and create a card component for each, was to pass the array of results as a data-prop. data-prop was written so that if the value it points to is an array (e.g. if I have data-prop="results" and results is an array in my component's state) then it will automatically iterate through all values and create a child component for each.
It does all of this INSIDE the data-component container, and doesn't let you add anything additional inside that container, like placeholders.

src/ui/components/cards/cardcomponent.js
oshi97: This method used to be standardcardcomponent's, since this PR adds the second card, I thought it made sense to move it to a shared place here

src/ui/components/cards/accordioncardcomponent.js
oshi97: updated to remove self-explanatory jsdocs

src/core/models/verticalresults.js
oshi97: updated!

src/ui/components/cards/accordioncardcomponent.js
oshi97: updated!

src/ui/templates/cards/accordion.hbs
oshi97: updated!

src/ui/components/cards/cardcomponent.js
creotutar: ah makes sense!

src/ui/components/ctas/ctacomponent.js
creotutar: talked offline: going to try a `ctaModifiers` approach vs two separate isSolo and AccordionCard properties

oshi97: just renamed it from _context to _ctaModifiers

creotutar (approved):
🎉 🎉 🎉 LGTM!

PR #531, opened 2020-03-03
VerticalResults: Add default event options for CTAs

oshi97:
This commit adds default event options for CTAs
created by the VerticalResultsComponent.

The verticalConfigId field was added to the VerticalResults model.
It is called the more legacy name verticalConfigId instead of
verticalKey, because the VerticalResults model is a wrapper over the section model,
which is given a verticalConfigId when used by universal search
(but not when used by vertical search). However outside of the model
I felt like it made more sense to try to call it verticalKey where possible.
In the future we would like to rename all instances of verticalConfigId
to verticalKey, including in analytics events since they are currently
both supported on the backend.

Adding verticalKey to the Result model instead of the VerticalResults model
would have allowed the code to avoid passing the verticalKey down through many
layers, however it felt odd to have verticalKey in the Section for universal search
and in the Result for vertical search, and similarly odd to duplicate it in both
models just to appear symmetrical.

TEST=manual
Checked that clicking a CTA without any event options set in the callsToAction
config had the default analytics event, that CTAs with set custom event options
in the config still fired, and that callsToActionFields without event options also
used the default.


README.md
```diff
@@ -559,9 +559,10 @@ const callsToAction = [{
   analyticsEventType: 'CTA_CLICK',
   // Whether the click should open in a new window, defaults to false
   newWindow: false,
-  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. Defaults to undefined, meaning
-  // no event will be fired
-  eventOptions: `{ "verticalKey": 'people', "entityId": 12312312, "searcher":"VERTICAL"}`
+  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. defaults to the below,
```
tmeyer2115: To avoid confusion, lets not use the word list here. It doesn't need to be a list, right? It's any valid JSON or JS object?

src/core/core.js
```diff
@@ -122,7 +122,7 @@ export default class Core {
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
         sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS)
       })
-      .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters))
+      .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))
```
tmeyer2115, line 125: I don't suppose the response has the verticalKey, does it?

src/ui/components/cards/cardcomponent.js
```diff
@@ -41,6 +41,12 @@ class CardConfig {
      * @type {Array<string>}
      */
     this.callsToActionFields = config.callsToActionFields || [];
+
+    /**
+     * Vertical key for the search.
+     * @type {string}
+     */
+    this.verticalKey = config.verticalKey;
```
tmeyer2115: I feel like this should be passed in as part of config.data. The verticalKey is not really a configuration parameter, it's on the same level of abstraction as the results.

src/core/models/verticalresults.js
```diff
@@ -4,8 +4,14 @@ import Section from './section';
 import SearchStates from '../storage/searchstates';
 
 export default class VerticalResults {
-  constructor (data = {}) {
+  constructor (data = {}, verticalKey) {
```
tmeyer2115: Can we include verticalKey in data somehow? In the static from method, assuming the response doesn't already have the verticalKey info, we could do something like:

return new VerticalResults(Section.from(Object.assign({}, response, { verticalConfigId: ... }, null, formatter), ...)

Here, then we don't need to change anything because we'll just use the verticalConfigId already in the Section.

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -123,6 +123,12 @@ class StandardCardConfig {
      * @type {boolean}
      */
     this.showOrdinal = this.showOrdinal || false;
+
+    /**
+     * Vertical key for the search.
+     * @type {string}
+     */
+    this.verticalKey = config.verticalKey;
```
tmeyer2115: Same thing about including in data. By having verticalKey as part of the config, someone might think they could pass a different verticalKey than the one actually corresponding to the result.

tmeyer2115 (changes requested):
I think we're almost there! Made a few suggestions which I think will streamline further. Great detail in the commit message!

src/core/core.js
oshi97: Nope, only for universal search

src/ui/components/cards/cardcomponent.js
oshi97: updated! I really don't like perpetuating the lie that data really has any different behavior than config but that's more an issue of this data-prop pattern I guess

README.md
oshi97: Yes, updated!

src/ui/components/cards/standardcardcomponent.js
oshi97: updated!

README.md
```diff
@@ -559,9 +559,10 @@ const callsToAction = [{
   analyticsEventType: 'CTA_CLICK',
   // Whether the click should open in a new window, defaults to false
   newWindow: false,
-  // The list of eventOptions needed for the event to fire. Either a valid json string or an object. Defaults to undefined, meaning
-  // no event will be fired
-  eventOptions: `{ "verticalKey": 'people', "entityId": 12312312, "searcher":"VERTICAL"}`
+  // The eventOptions needed for the event to fire. Either a valid json string or an object. defaults to the below,
```
tmeyer2115: Defaults to the below, where

oshi97: Now reads  
```
  // result => {
  //   const eventOptions = {
  //     verticalKey: this.verticalKey,
  //     searcher: 'VERTICAL'
  //   };
  //   if (result._raw.id) {
  //     eventOptions.entityId = result._raw.id;
  //   }
  //   return eventOptions;
  // }
  eventOptions: item => `{ "verticalKey": "people", "entityId": "${item._raw.id}", "searcher":"VERTICAL"}`
```

oshi97: instead of weird pseudo code that would probably confuse someone

creotutar: I'm not sure who we are catering to for these READMEs but the HHs are going to be reading this and im not sure this is the most understandable example of the eventOptions (as they're looking for an object default not what looks like an arrow function?)

src/ui/components/cards/cardcomponent.js
```diff
@@ -59,18 +71,23 @@ export default class CardComponent extends Component {
...
       ...data,
-      result: this._config.result,
+      result: this.result,
       cardType: cardTypes[cardType]
     });
   }
 
   addChild (data, type, opts) {
-    return super.addChild(data, type, {
+    const newData = {
```
creotutar: nit: is this really `newData` in the same way the below is `newOpts`? it looks like this is a wrapper over the data

oshi97: Would `updatedData` be better?

README.md
oshi97: oh so the default is basically an arrow function though, rose wrote it in the qa sheet as 
(item) =>  item._raw.id ? `{ "verticalKey": "{{VERTICAL_KEY}}", "searcher":"VERTICAL", "entityId": "${item._raw.id}"}` : `{ "verticalKey": "{{VERTICAL_KEY}}", "searcher":"VERTICAL"}` but I thought what I had above was more readable? What do you think we should write here

src/ui/components/cards/cardcomponent.js
oshi97: changed to updatedData!

README.md
oshi97: updated to 

```
// By default, if no event options are specified the SDK will try to add verticalKey, entityId, and searcher options
// to the analytics event.
eventOptions: result => ({
  // The vertical key for the CTA. If unspecified, this defaults to the vertical key this cta is a part of
verticalKey: "people",
  // The entity id of the result this cta is a part of, defaults to the entityId field in Knowledge Graph
entityId: result._raw.id,
  // If the CTA is inside a vertical search, defaults to the value "VERTICAL",
  // if is inside a universal search, defaults to the value "UNIVERSAL"
searcher: "VERTICAL"
})
```

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -30,7 +44,12 @@ export default class CTACollectionComponent extends Component {
      * The computed calls to action array
      * @type {Array<Object>}
      */
-    this.callsToAction = this.resolveCTAMapping(this._config.result, callsToActionFields, ...callsToAction);
+    this.callsToAction = this.resolveCTAMapping(this.result, callsToActionFields, ...callsToAction);
+
+    this.callsToAction = this.callsToAction.filter(cta => cta.url && cta.label).map(cta => ({
```
creotutar: is this saying that both CTA url and CTA label are required? if it is, do we want to indicate to the user that one of their CTAs is missing either a url or a label?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -52,7 +52,10 @@ export default class UniversalResultsComponent extends Component {
     if (childOpts.useAccordion === true) {
       return super.addChild(data, AccordionResultsComponent.type, childOpts);
     }
-    return super.addChild(data, type, childOpts);
+    return super.addChild(data, type, {
+      isUniversal: true,
+      ...opts
```
creotutar: we want to use opts instead of childOpts?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -154,6 +160,7 @@ export default class VerticalResultsComponent extends Component {
      * @type {Array<Result>}
      */
     this.results = data.results || [];
+    this.verticalKey = data.verticalConfigId;
```
creotutar, line 150: why would verticalKey change outside of the constructor?

oshi97: I kept the bit of code where we get verticalKey from the config because it was there before and I was thinking universal results might need it. BUT ACTUALLY I don't think we ever need to be getting vertical key from the config, since it's always in the data now for universal search and vertical search.

oshi97: this only does not work if we need vertical key before results are loaded in

oshi97: I thought about it for a bit and at first wasn't comfortable with a VerticalResultsComponent not knowing what it's vertical key is until a search has been run. But if we aren't letting them have a verticalKey in the config, then I think it actually makes sense to not initialize it to anything until search data is returned.

oshi97: Sooooo I took out the bit where verticalKey is set by the config because that bit of code is basically a lie that pretends like you can set verticalKey from the config

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: Soo Rose said she just wanted was for it to not display but she said it kind of randomly as part of QA so I'll ask her what she thinks of your idea

src/ui/components/results/universalresultscomponent.js
oshi97: updated!

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: back again! so talked with rose a bit and since ctas can take in functions to set their config they might want to set their label to a field that is sometimes blank, and in those cases we would want to automatically hide their CTAs for them, without crashing their page. I felt like a console.error might be a little too harsh, especially since this might be intentional behavior, but a console.warn feels like a good middle ground

tmeyer2115 (approved):

PR #532, opened 2020-03-04
Assorted security fixes needed to open source the SDK

tmeyer2115:
- Removing templateUrl as an option from the SDK.
- Updating XSS vulnerabilities across multiple HBS templates.


src/core/constants.js
```diff
@@ -1,7 +1,7 @@
 /** @module */
 
 /** The current lib version, reported with errors and analytics */
-export const LIB_VERSION = 'v0.13';
+export const LIB_VERSION = 'v0.12';
```
alexisgrow, line 4: Is this change intended?

tmeyer2115: Yep, updating the LIB_VERSION should be the last commit of the release. Otherwise, we will be attempting to fetch the default set of compiled templates from a location that doesn't exist in the CDN: https://assets.sitescdn.net/answers/v0.13/answerstemplates.compiled.min.js

alexisgrow: Ahh, gotcha, that makes sense!

alexisgrow (approved):

PR #533, opened 2020-03-04
No results

alexisgrow:


tmeyer2115 (commented):
Is the failing test due to the rollup includes stuff? If so, I can just lg!

tmeyer2115 (approved):

src/ui/templates/icons/icon.hbs
```diff
@@ -2,6 +2,7 @@
   {{#if iconUrl}}
     <img src="{{iconUrl}}" alt="" class="Icon-image">
   {{else}}
+    {{iconName}}
```
tmeyer2115: Is this printing the string iconName before the actual icon?

src/ui/components/icons/iconcomponent.js
```diff
@@ -62,7 +62,6 @@ export default class IconComponent extends Component {
     return super.setState(Object.assign(data, {
       iconUrl: this.iconUrl,
       iconName: this.iconName,
-      name: this.iconName ? this.iconName : 'custom',
```
tmeyer2115: If we remove this, does the {{name}} reference in icon.hbs also need to be removed?

src/ui/templates/icons/icon.hbs
```diff
@@ -2,7 +2,6 @@
   {{#if iconUrl}}
     <img src="{{iconUrl}}" alt="" class="Icon-image">
   {{else}}
-    {{iconName}}
```
tmeyer2115: Whoops, was looking at the commits one by one, ignore my comment about this

src/ui/components/icons/iconcomponent.js
alexisgrow: Ahh, you're right, this one should not be removed!

PR #534, opened 2020-03-05
Navigation Component: Allow vertical config to be passed directly to …

alexisgrow:
…component


oshi97:
hi Alexis is there any extra context for this PR? Do we just want them to give them the choice to pass verticalPages either at the top or in the nav component?

alexisgrow (commented):
Yes exactly! The plan is that, ideally, a component should be able to get all the config it needs from its individual `addComponent`, but it will fall back to the current top-level config if it currently exists

oshi97 (approved):
LGTM!

PR #537, opened 2020-03-05
Pagination Component: Prefer search options from component config

alexisgrow:
Accept verticalKey and limit as options in the component's config,
if not present, fall back to existing search config.

TEST=manual,auto

Test using query with multiple pages of results, add pagination
component and correctly page through results. Ensure it prefers
components config by setting it as something different than search
config; test that it works without component config, test that it
works without search config.


alexisgrow (commented):
The main issue with this approach is that the component's limit isn't respected on page load. So, if someone sets a defaultInitialSearch or links to a page with a query in the URL, it doesn't respect that limit until a user clicks to the next page.

PR #539, opened 2020-03-05
UniversalResults: Use new StandardSection instead of VerticalResults

oshi97:
This component changes UniversalResultsComponent to use a new
StandardSectionComponent instead of VerticalResultsComponent
for rendering results. The includeMap option for UniversalResults
will remain until the MapSection is implemented. The hiddenFields
config option for UniversalResults will remain unused until
the AppliedFiltersComponent is implemented, at which point the
Section components will switch to using that instead of the
current solution for displaying filters in the section.

One strange note is that the way sections are specified
in universal results config is that all attribute names
that are not designated as other config are assumed to be
configuration for a section with veritcalKey equal to the
attribute name. This can result in having names clash with
built-in config options.

It was also discussed that if no config for a veritcal section
is specified, the default section should be provided. In this
way, the only mechanism for hiding a vertical is to change back-end
config. The reasoning for this would be that hiding on the front-end,
but still receiving that data from the backend, would make analytics
much more difficult to do.
TEST=manual
Create a test site with Universal Results, check that results
load onto the page and look like the mocks. Check that custom
inline template can be passed into each section. Check that both
accordion and standard cards can be specified, and that ctas render
as expected and fire correct analytics events (default if unspecified
or the specified eventOptions), with searcher: 'UNIVERSAL' not 'VERTICAL'.


oshi97:
Okay should be ready for review I believe I've double checked everything for obvious stupidity

oshi97:
I realized it was possible to implement the hidden fields option without the filtermetadata changes from the applied filters pr since universal search does not ever use any static filters

oshi97:
Waiting on Tom and Rose to discuss whether sections should be a first-class concept in the sdk

oshi97:
Moving to new pull request since specs have changed

PR #542, opened 2020-03-09
Change results count to uses Section.resultsCount

oshi97:
Change results count to uses Section.resultsCount, which is the total…mount of results the query has in the backend, instead of results.length which is only the number of results sent in the response


src/ui/templates/results/resultsheader.hbs
```diff
@@ -5,7 +5,7 @@
       <span>-</span>
       {{resultsCountEnd}}
       <span>of</span>
-      {{resultsLength}} 
+      {{resultsCount}}
```
alexisgrow: If resultsCount is zero, do we want to display the "of {{resultsCount}}"? Or will `_config.showResultsCount` handle that case?

oshi97: right now the results header doesn't render if there are no results, but you're right it's better check that in the header as well since they're separate components

oshi97: updated!

src/ui/templates/results/resultsheader.hbs
```diff
@@ -1,14 +1,16 @@
...
-      {{resultsCountEnd}}
-      <span>of</span>
-      {{resultsLength}} 
-    </div>
-    {{#if showResultSeparator}}
-      <div class="yxt-ResultsHeader-resultsCountSeparator">{{_config.resultsCountSeparator}}</div>
+    {{#if resultsCount}}
+      <div class="yxt-ResultsHeader-resultsCount">
+        {{resultsCountStart}}
+        <span>-</span>
```
alexisgrow: nit: Can this span have a class so that if implementers wanted to style the separator by itself (like make it bold or something) they could?

src/ui/templates/results/resultsheader.hbs
```diff
@@ -1,14 +1,16 @@
...
+        {{resultsCountEnd}}
+        <span>of</span>
```
alexisgrow: Same nit here 🙈

alexisgrow (approved):
LGTM, some changes from this commit (`dd0ba05f6a078053ace3a760c023e9055c2f7328`) are being picked up in this PR, I think a rebase would fix

src/ui/templates/results/resultsheader.hbs
oshi97: Ah okay might as well go all the way if we're doing this stuff

oshi97: everything has a class now!

alexisgrow (approved):

PR #544, opened 2020-03-10
Better errors in component manager when component type is not found

oshi97:
This commit gives users a list of similarly named components
when they try to add a component with a type that is not found.
TEST=manual
check that trying to addComponent a weird name gives new error message


src/ui/components/componentmanager.js
```diff
@@ -94,6 +94,15 @@ export default class ComponentManager {
     };
 
     let componentClass = COMPONENT_REGISTRY[componentType];
+    if (!componentClass) {
+      let similarComponents = Object.keys(COMPONENT_REGISTRY).filter(type => type.startsWith(componentType.substring(0, 2)));
```
tmeyer2115: Minor nit, but do a new line for 'type => type.startsWith(componentType.substring(0, 2))'.

tmeyer2115: Again, minor thing but let's factor some of this out into another method. Let's make a method called getSimilarComponents or something whose responsibility is to return similar components (or all if none are found). How did you come up with the substring(0, 2) metric?

oshi97: It's somewhat arbitrary, but in general the first two letters of each component name form some kind of sound, and because of this I imagined if a typo would occur it would occur in the later portions of the component name

oshi97: updated!

oshi97: updated with new method

oshi97: there are definitely better ways to do get similar words but given the extremely limited pool of component names it didn't really seem worth the effort

PR #552, opened 2020-03-11
remove the footer component

oshi97:


oshi97:
will be part of jambo layout

PR #554, opened 2020-03-11
use old resultssectionheader for universal results

oshi97:


oshi97:
TEST=manual
check that universal results uses old header
check that vertical results still uses new header when showResultsCount or showAppliedFilters are true

PR #559, opened 2020-03-11
Remove callsToActionFields config option for card

oshi97:
update readme


oshi97:
TEST=manual
Test that my Oliver Shi person entity in roses dev account does not use the cta custom field with
callsToActionFields specified

oshi97:
TEST=manual
Test that my Oliver Shi person entity in roses dev account does not use the cta custom field with
callsToActionFields specified

PR #565, opened 2020-03-11
V13/qa

oshi97:
v0.13.0 QA for vertical results
1. Fix results count to read 0-20 of 1039 instead of 0-20 of 20
2. Filter out ctas without url and label before adding additional css classname modifiers
3. In StandardCard, do not style the title as a link if there is no url
4. Columns work without a resize handler now. I've never felt so stupid and smart at the same time.
5. Add additional error step in componentmanager.js to check if component type is recognized. If it is not, send an error message that sends valid components types with similar names and asks if you made a typo.
6. Always use absolute urls in cards and CTAs. Rose said we can assume nobody would ever want a relative url.
7. Fix ie11 styling. Whopee! Also a number of the other styling qa items.
8. Remove the footer component from vertical/universal results - will be part of jambo layout in the future.
9. Allow config.details: null for cards, and to use no details instead of using defaults.
10. Use old resultssectionheader for universal results.
11. Fix includeMap option for vertical/universal results. Needed to wrap data in { map: data }
12. Add TITLE_CLICK analytics event for standard card. Also add SEARCHER analytics option in accordion card. Fix ROW_COLLAPSE and ROW_EXPAND events being flipped.
13. Update all instances of the newWindow config option for cards and CTAs to the target config option. This allows users to set any target attribute instead of only allowing _self or _blank.
14. Rename showResultsCount to showResultCount in vertical results config. I like showResultsCount more but whatever.
15. Remove callsToActionFields config option for cards. Rose said they didn't want this anymore, sad.
16. Unescape title, subtitle, details config options in accordion/standard card handlebars and label config option in cta.hbs. Rose said it should be okay and has been discussed with Jon Hyler.
17. Do not render CTACollectionComponent container when there are no CTAs with both label and url field. This was done because the container has it's own flex-grow which was causing spacing styling issues.
18. Add word-break: break-all to CTA.scss. I was against this for a while but don't care anymore.
19. Don't register TITLE_CLICK analytics event handler in standard card when there is no title.
20. Add aria-label alt text for image in the Standard Card.
21. Change the behavior of the expanded config option for Accordion Cards to only expand the card if it is the first one in the results.
22. Update the readme for all these changes.

This list is consistent with the order of commits below, excluding updating the readme which was done throughout.


src/ui/templates/results/resultsheader.hbs
```diff
@@ -1,22 +1,25 @@
...
-      <span>-</span>
-      {{resultsCountEnd}}
-      <span>of</span>
-      {{resultsLength}} 
-    </div>
-    {{#if showResultSeparator}}
-      <div class="yxt-ResultsHeader-resultsCountSeparator">{{_config.resultsCountSeparator}}</div>
+    {{#if resultsCount}}
+      <div class="yxt-ResultsHeader-resultsCount">
+        <span class="yxt-ResultsHeader-resultsCount-start">{{resultsCountStart}}</span>
```
creotutar: ok i LOVE that you are adding the descriptive class names (esp. for the dash, tysm)
but to follow SUIT, we should only have one single dash outside of the `yxt-`
so `yxt-ResultsHeader-resultsCountStart` or `yxt-ResultsHeader-resultsCountEnd`

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -46,31 +46,28 @@ export default class CTACollectionComponent extends Component {
...
     this._config._ctaModifiers = this._config._ctaModifiers;
 
-    /**
-     * The computed calls to action array
-     * @type {Array<Object>}
-     */
-    this.callsToAction = this.resolveCTAMapping(this.result, callsToActionFields, ...callsToAction);
-
-    this.callsToAction = this.callsToAction.map(cta => {
+    callsToAction = this.resolveCTAMapping(this.result, callsToActionFields, ...callsToAction);
```
creotutar: will this `...` work if callsToAction is a function?

src/ui/components/ctas/ctacollectioncomponent.js
```diff
@@ -46,31 +46,28 @@ export default class CTACollectionComponent extends Component {
...
-        return {
-          eventOptions: this.defaultEventOptions(this.result),
-          _ctaModifiers: _ctaModifiers,
-          ...cta
-        };
       }
     });
+    callsToAction = callsToAction.filter(cta => cta.url && cta.label);
+    this.callsToAction = callsToAction.map(cta => {
+      const _ctaModifiers = this._config._ctaModifiers;
```
creotutar: nit: is there a reason we use `_` syntax for this const if its not an instance variable?

src/ui/components/ctas/ctacollectioncomponent.js
creotutar: maybe this is just an opinion: im not sure this is a case when want to use `let`. i think variables should represent a single concept, either the callsToAction we get from the config or the resolution of those callsToAction. maybe split this out into two separate variables?

src/ui/templates/cards/standard.hbs
```diff
@@ -19,13 +19,23 @@
...
-            href="{{_config.url}}" target={{#if _config.newWindow}}"_blank" rel="noopener noreferrer nofollow"{{else}}"_self"{{/if}}
-          {{/if}}>
+        {{#if _config.url}}
+          <div class="yxt-StandardCard-title">
+            {{#if _config.newWindow}}
+              <a href="{{_config.url}}" target="_blank" rel="noopener noreferrer nofollow">
+                {{_config.title}}
+              </a>
+            {{else}}
+              <a href="{{_config.url}}" target="_blank" "_self">
```
creotutar: i think you meant target="_self" here?

creotutar: i kinda liked what you were doing with the `target={{#if`
saves you from writing out/changing _config.url and _config.title twice

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -38,17 +39,20 @@
     width: 100%;
     display: flex;
     cursor: pointer;
-    height: $accordion-card-spacing * 2.75;
     background: none;
-    cursor: pointer;
```
creotutar, line 43: im surprised you're removing this
do we not want this to indicate a clickable element?

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -114,6 +121,7 @@
   &-ctas 
   {
     margin-top: $accordion-card-spacing / 2;
+    visibility: hidden;
```
creotutar: ~is there a reason we are setting visibility to hidden vs removing the elements from the hbs?~

i see below we are bringing back visibility

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -133,10 +141,21 @@
...
     overflow: hidden;
     transition: all 400ms ease;
   }
+
+  .yxt-AccordionCard-ctas
+  {
+    visibility: visible;
+  }
+
+  .yxt-CTACollection-item:not(:last-child) {
```
creotutar: nit: `{` on new line please

src/ui/sass/modules/_AccordionCard.scss
creotutar: if we want the elements still in the DOM, would vote to use `display: none;` so that spacing isnt reserved for the elements

README.md
```diff
@@ -504,19 +504,6 @@ ANSWERS.addComponent('VerticalResults', {
     // see [Calls To Action](#Calls-To-Action) and [Calls To Action Fields](#Calls-To-Action-Fields)
     callsToAction: () => [],
     callsToActionFields: [ "c_primaryCTA","c_secondaryCTA" ]
-  },
-  // Config for the footer at the bottom of the results
-  footer: {
```
creotutar, line 509: 👋

src/ui/components/cards/standardcardcomponent.js
```diff
@@ -183,6 +196,9 @@ export default class StandardCardComponent extends Component {
         this.setState();
       });
     }
+
+    const titleEl = DOM.query(this._container, '.js-yxt-StandardCard-title');
+    DOM.on(titleEl, 'click', this.handleTitleClick);
```
creotutar: why is this a click event instead of being handled in the hbs template? from what i've seen, we only do it in the JS when the eventType is dynamic (e.g. accordionresults and directanswers) but opt to keep it in the DOM otherwise

src/ui/templates/cards/standard.hbs
```diff
@@ -8,7 +8,7 @@
   {{/if}}
   <div class="yxt-StandardCard-wrapper">
     {{#if _config.image}}
-      <div class="yxt-StandardCard-img" style="background-image: url('{{_config.image}}');">
+      <div class="yxt-StandardCard-img" style="background-image: url('{{_config.image}}');" role="img" aria-label="{{imgAlt}}">
```
creotutar: im curious: why is this a div instead of an ``?

src/ui/templates/cards/standard.hbs
creotutar: seems to have been resolved with `_target`

README.md
oshi97: 💀

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: Uhh no real reason I just wanted the names the match up since I call it _ctaModifiers everywhere else

src/ui/sass/modules/_AccordionCard.scss
oshi97: Ah yeah this was so that the CTAs would not be tabbed into if the accordion was not expanded. Display none is better!

src/ui/sass/modules/_AccordionCard.scss
oshi97: sorry bad habits :'(

src/ui/templates/cards/standard.hbs
oshi97: yeah rose pulled a fast one on me haha

src/ui/sass/modules/_AccordionCard.scss
oshi97: I had it twice for some reason, probably due to a merge conflict from before

src/ui/templates/cards/standard.hbs
oshi97: I couldn't figure out how to get object-position styling for ie11

src/ui/templates/results/resultsheader.hbs
oshi97: updated!

src/ui/components/ctas/ctacollectioncomponent.js
oshi97: tried to refactor this to be cleaner

src/ui/components/cards/standardcardcomponent.js
oshi97: updated to use in-line events

oshi97:
Added no results QA as the latest commit
1. Add noResults config to vertical results component readme
2. Remove filters from results header, and from readme
3. Change verticalpagesconfig.js to be semi-immutable,
	to avoid issue with navigation component mutating the config
4. refactor models/verticalresults.js
5. add no_results_config to storagekeys.js, so that pagination
	displays only when no results are configured to displayAllResults
6. Fix not having search query in alternativeVertical's universalUrl
7. Change universalUrl in vertical results to try to pull from
	the global vertical pages config first, using the first
	element that does not have a verticalKey
8. Styling changes/fixes for no results
9. Fix navigation component not setting verticalKey in navigation analytics events
10. Change default noresults icon to null instead of star
11. Change no results to always display on vertical, instead of falling
	back to old no results

oshi97:
WIP until finished with Rose's new ideas after seeing qa branch

oshi97:
Finished remaining second round QA from rose

oshi97:
1. In standard/accordion card, ternary operator for details was flipped
2. don't use headshot.alternateText for alternate image text in standard card
3. in ctacollectioncomponent.js, should filter out ctas that don't have a url and label AFTER applying cta mappings
4. update text for aria-label in resultsheader.hbs results count
5. add bottom border to no results when on mobile
6. use breakpoint-mobile-max instead of container-desktop-base for mobile card styling
7. add back applied filters to results header (talked with rose on this kinda sucks we have to go back and forth)
8. remove black border from accordion card :focus (another back and forth guy)
9. improve standard card styling based on rose's complaints/suggestions
10. fix weird bug in rose's test cases where a vertical search using standard cards with one result and no details led to the whole width of vertical results being very thin. fixed by adding placeholder elements
11. for universal results' verticalURL, try to use the VERTICAL_PAGES_CONFIG to get the url first before looking in the setState data (which is a models/section.js)

src/ui/components/results/paginationcomponent.js
```diff
@@ -69,10 +70,18 @@ export default class PaginationComponent extends Component {
     return 'results/pagination';
   }
 
+  shouldShowControls (results, limit) {
+    const hasResults = results.searchState === 'search-complete' && results.resultsCount > limit;
+    const noResultsConfig = this.core.globalStorage.getState(StorageKeys.NO_RESULTS_CONFIG) || {};
```
creotutar, line 75: does this make it so that if we want this function to function properly, we must addComponent('VerticalResults') before we addComponent('Pagination')

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -209,7 +210,7 @@ export default class VerticalResultsComponent extends Component {
       data = this.core.globalStorage.getState(StorageKeys.ALTERNATIVE_VERTICALS);
       const newOpts = {
         template: this._config.noResultsTemplate,
```
creotutar, line 212: is this allowing us to default to our normal template if a noResultsTemplate is not defined as the README indicates? any reason not to use `this.noResultsTemplate`?

src/ui/sass/modules/_AlternativeVerticals.scss
```diff
@@ -7,8 +7,9 @@
   border-right: $border-default;
   border-left: $border-default;
 
+  $alternative-verticals-emphasized-font-weight: 500;
```
creotutar: can we use one of the font-weight variables in _fonts.scss? if not, can we add this to that file?

src/ui/templates/navigation/navigation.hbs
```diff
@@ -4,8 +4,13 @@
       <li>
         <a class="js-yxt-navItem yxt-Nav-item{{#if isActive}} is-active{{/if}}"
           href="{{url}}"
-          data-eventtype="{{#if @first}}ALL_TAB_NAVIGATION{{else}}VERTICAL_TAB_NAVIGATION{{/if}}"
-          data-eventoptions='{"verticalConfigId": "{{configId}}"}'
+          {{#if @first}}
+            data-eventtype="ALL_TAB_NAVIGATION"
+            data-eventoptions='{"verticalKey": "all"}'
```
creotutar: possible for this to be the "configId" that is passed into the hbs? worried about hardcoding this,
does this also imply that verticals in the future cannot be named "all"?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -132,6 +124,15 @@ export default class VerticalResultsComponent extends Component {
     return true;
   }
 
+  getUniversalUrl () {
+    const universalConfig = this._verticalsConfig.find(config => !config.verticalKey) || {};
+    let universalUrl = this._universalUrl;
```
creotutar: im not able to trace where `config.universalUrl` is coming from. it's not in the README and i don't see it being passed in anywhere.

if we're given the scenario where `universalUrl` is null, and we only have vertical configs with non-null `verticalKey`, is it possible we will return a non-working universal url below?

creotutar: im not understanding why this lives in the verticalresultscomponent. should we perhaps be doing this URL construction in some common URL file that is always referring to some source of truth? how do we get the URLs with query parameters for navigation tabs when we click on them?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -133,6 +133,13 @@ export default class VerticalResultsComponent extends Component {
     return `${universalUrl}?query=${this.query || ''}`;
   }
 
+  getVerticalURL (data) {
+    const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
+    const verticalURL = verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
+    console.log(this._verticalsConfig, verticalConfig, data, this.verticalKey);
```
creotutar: console.log

src/ui/sass/modules/_Card.scss
```diff
@@ -16,8 +16,16 @@
     }
   }
 
+  &-child--universal {
+    margin-bottom: 0;
+  }
+
+  & + & .yxt-Card-child--universal {
+    border-top: 0;
```
creotutar, line 22: why is the border between cards put on the yxt-Card-child instead of the yxt-Card?

creotutar (commented):
LGreatTM

What do you think about a common place for putting URL construction logic based on the verticalPages?

src/ui/components/results/paginationcomponent.js
oshi97: nope order doesn't matter! every time a new StorageKeys.VERTICAL_RESULTS is updated the pagination component does a setState, which should update the no results config again

src/ui/components/results/verticalresultscomponent.js
oshi97: the weird config class pattern we have is probably not the best, but this.noResultsTemplate is part of the config class, not the component, so it ends up in this._config

src/ui/sass/modules/_AlternativeVerticals.scss
oshi97: yes, good idea!

src/ui/templates/navigation/navigation.hbs
oshi97: hmmmm yeah the "ALL_TAB_NAVIGATION" event really doesn't need a verticalKey since it's pretty implied. I'll ask Rose, she had just told me to hardcode it to "all"

src/ui/sass/modules/_Card.scss
oshi97: it let me make the columns flex-basis percentages really precise, but I think putting the borders and margin on yxt-Card makes more sense

src/ui/components/results/verticalresultscomponent.js
oshi97: for URLs there's only a few places but it looks like we do something like

```
      let query = inputEl.value;
      let params = new SearchParams(window.location.search.substring(1));
      params.set('query', query);
```

oshi97: for the navigation component they do some complicated stuff that doesn't seem obvious to me, but it's all contained within the component

src/ui/templates/navigation/navigation.hbs
oshi97: yeah she said we don't need it!

oshi97:
thanks for the suggestions christian! I think the code is in a much better place now

src/ui/components/results/verticalresultscomponent.js
oshi97: updated to use urlutils.js

oshi97: added new method addQueryToUrl to urlutils.js

src/ui/components/results/verticalresultscomponent.js
oshi97: yes this still uses normal template if it is not defined

oshi97: *it being the noResultsTemplate

oshi97:
1. separate vertical/universal url with query string logic into urlutils.js
2. remove unused this._universalUrl in verticalresultscomponent.js
3. add font-weight-medium = 500 to _fonts.scss, used in alternativeverticals.scss
4. update card.scss to put border and margin on yxt-Card instead of yxt-Card-child.
   in order to do this flex-basis of for yxt-Card had to be updated for columns
5. remove veritcalKey: all from navigation.hbs event options

src/core/utils/urlutils.js
```diff
@@ -31,3 +31,12 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
...
     : 'https://answers.yext-pixel.com';
 }
+
+/**
+ * Returns the passed in url with the query appended to it.
+ * @param {string} url
+ * @returns {string}
+ */
+export function addQueryToUrl (url, query) {
+  return url.split('?')[0] + '?query=' + query;
```
creotutar: if the url already has query parameters, do we want to drop this? (e.g. client has analytics information in the search parameters)

do we want to use SearchParams.set?

oshi97: oh that's a good point, I just assumed anything extra in there was unwanted

src/core/utils/urlutils.js
```diff
@@ -31,3 +31,12 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
     ? 'https://sandbox-answers.yext-pixel.com'
     : 'https://answers.yext-pixel.com';
 }
+
+/**
+ * Returns the passed in url with the query appended to it.
+ * @param {string} url
+ * @returns {string}
+ */
+export function addQueryToUrl (url, query) {
```
creotutar: do we want to follow the naming convention with a "getQueryUrl"?

creotutar: maybe construct?

oshi97: I changed it to addParamsToUrl, to make it more generic, does that work?

src/core/utils/urlutils.js
oshi97: updated to only add params

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -122,14 +122,14 @@ export default class VerticalResultsComponent extends Component {
   getUniversalUrl () {
     const universalConfig = this._verticalsConfig.find(config => !config.verticalKey) || {};
     if (universalConfig.url) {
-      return addQueryToUrl(universalConfig.url, this.query);
+      return addParamsToUrl(universalConfig.url, { query: this.query });
```
creotutar, line 125: if query is null or empty, will this add an empty query to the url? is this a bad thing if they have allowEmptyQuery set to true?

oshi97: Did you mean set to false? I tried looking around and I've heard of that option a few times but can't find it anywhere in the codebase or readme

oshi97: I think it should be fine

oshi97: universalUrl shouldn't show up if they don't allow empty queries and there's an empty query

PR #569, opened 2020-03-16
Universal Results Styling Updates

oshi97:
This commit implements universal results styling
updates, and new config options for this styling.
Universal results uses multiple vertical results
to render each individual section in universal results.
TEST=manual
Went one by one down the config options list in oobx
spec and made sure they worked as expected, and defaults
were being applied.


README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    // Optional: A custom handlebars template for this section
+    template: "<div> Custom section template </div>",
+    // The title of the vertical
+    // Defaults to the vertical key, in this example 'people'
+    title: "People",
+    // Icon to display to the left of the title. Must be one of our built-in icons, defaults to 'star'
+    icon: 'star',
+    // The url for both the viewMore link and the change-filters link. Defaults to '/{{VERTICAL_KEY}}.html',
+    // in this case that is '/people.html'
+    url: '/people/about.html',
```
creotutar: are we listing examples or defaults in these values?

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
-      }
+  people: {
+    card: {
+      // Configuration for the cards in this vertical, see [Cards](#Cards)
+    },
+    // Optional: A custom handlebars template for this section
+    template: "<div> Custom section template </div>",
+    // The title of the vertical
+    // Defaults to the vertical key, in this example 'people'
+    title: "People",
```
creotutar: very insignificant nit: can we stick to one style of quote? `'` vs `"`

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
-      // If includeMap is true, mapconfig that contains a mapProvider and apiKey is required
-      mapConfig: {
-        // Either 'mapbox' or 'google'
-        mapProvider: 'google',
-        // Api key for the map provider
-        apiKey: '<<< enter your api key here >>>',
-      }
+  people: {
+    card: {
+      // Configuration for the cards in this vertical, see [Cards](#Cards)
```
creotutar: does this link work inside of a code block?

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    },
+    // Optional: A custom handlebars template for this section
+    template: "<div> Custom section template </div>",
```
creotutar: is this only html or can we also provide a Handlebars fn?

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    // If showAppliedFilters is true, whether to display the field name of an applied filter, e.g.
+    // if a filter on 'Location' by the value 'Virginia', display 'Location: Virginia' if true,
+    // otherwise display just 'Virginia'. Defaults to false.
+    showFieldNames: false,
+    // If showAppliedFilters is true, this is list of filters that should not be displayed.
+    // By default, builtin.entityType will be hidden
+    hiddenFields: ['builtin.entityType'],
+    // If true, adds a map to the vertical using the provided mapConfig. Default is false
+    includeMap: true,
+    // Configuration for a map, needed if includeMap is true.
```
creotutar: im surprised this is /required/ per se. is map config not in the top level?

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    // Requires a mapProvider and apiKey, where mapProvidder is either 'google' or 'mapBox'
```
creotutar: Providder

creotutar: `mapbox` vs `mapBox`?

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    // The url for both the viewMore link and the change-filters link. Defaults to '/{{VERTICAL_KEY}}.html',
+    // in this case that is '/people.html'
+    url: '/people/about.html',
+    // Whether to display a view more link at the bottom of the vertical, defaults to true
+    viewMore: true,
+    // The text for the view more link, if viewMore is true. Defaults to 'View More'
+    viewMoreLabel: 'View More!',
+    // Whether or not to display the change-filters link, which links to the url config option
+    changeFilters: true,
+    // If true, show any applied back-end filters that were applied to the universal search. defaults to false
```
creotutar: If period, can we make `Defaults`

README.md
```diff
@@ -384,114 +384,74 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    template: "<div> Custom section template </div>",
+    // The title of the vertical
+    // Defaults to the vertical key, in this example 'people'
+    title: "People",
+    // Icon to display to the left of the title. Must be one of our built-in icons, defaults to 'star'
+    icon: 'star',
+    // The url for both the viewMore link and the change-filters link. Defaults to '/{{VERTICAL_KEY}}.html',
+    // in this case that is '/people.html'
+    url: '/people/about.html',
+    // Whether to display a view more link at the bottom of the vertical, defaults to true
```
creotutar: Do we want the relative UI in the documentation?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -1,69 +1,100 @@
...
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
     this.moduleId = StorageKeys.UNIVERSAL_RESULTS;
-    this._limit = opts.limit || 10;
-  }
 
-  static get type () {
-    return 'UniversalResults';
-  }
+    const { container, isTwin, render, template, ...sectionConfigs } = config;
```
creotutar: this worries me. isn't any component configuration option fair game? e.g.: name, analyticsOptions, class, transformData

even outside of accepted configuration, couldn't a user put a random key here and that would be read as a sectionConfig?

what was the reasoning behind putting the verticals on the top-level UniversalResults configuration?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -1,69 +1,100 @@
...
+  // Also links to verticalURL
+  changeFilters: false,
+  // Whether to show the applied filters.
+  showAppliedFilters: true,
+  // Whether to show field names in the applied filters, e.g. 'Location: Virginia' vs just 'Virginia'.
+  showFieldNames: false,
+  // Field ids to hide in the applied filters.
+  hiddenFields: ['builtin.entityType'],
+  // Whether to show a result count.
+  showResultCount: false
```
creotutar, line 95: if we are including false values, did we want `includeMap` here too?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -1,69 +1,100 @@
...
-  getChildConfig (configId) {
-    const defaultConfig = {
-      verticalConfigId: configId,
-      isUniversal: true
-    };
-    let config = this._config.config;
-    if (config === undefined) {
-      return defaultConfig;
+      /**
+       * @type {Section}
```
creotutar: javadoc in addChild?

README.md
oshi97: Nope, changed to just Cards

README.md
oshi97: I'm not sure what a Handlebars fn is, but the input must be a string understood by the handlebars compiler. This is what is happening internally in component.js, maybe it can answer your question?
```
  /**
   * compile a handlebars template so that it can be rendered,
   * using the {Handlebars} compiler
   * @param {string} template The template string to compile
   */
  compile (template) {
    if (typeof template !== 'string') {
      return '';
    }
    return this._handlebars.compile(template);
  }
```

README.md
oshi97: It looks like all the double quotes were me, and I prefer single quotes anyways so I changed all of them.

README.md
oshi97: defaults where possible, but for defaults that aren't static values I tried to put some sane example.

README.md
oshi97: updated

README.md
oshi97: It's not in the top level, we're just going with putting it per vertical for now.

README.md
oshi97: Either, since it changes it all to lower case, but the error message we give back if you pass an invalid provider is
```
throw new Error('MapComponent: Invalid Map Provider; must be `google` or `mapBox`');
```

So i thought may as well match it

src/ui/components/results/universalresultscomponent.js
oshi97: So, I noted this on the spec sheet and Rose pretty much said "That sucks, but it's okay for now."

To answer your question, there's nothing wrong with a random key being interpreted as sectionConfig, since the associated config is only read if there is a section with that vertical key.

There **IS** an issue if a user tries to have a vertical key with the same name as a special word, like transformData, because the user would not be able to configure BOTH transformData the keyword and transformData the vertical key.

Anyways, if it really bothers you we can try asking her again

src/ui/components/results/universalresultscomponent.js
oshi97: I like my self documenting code

oshi97: I made it squatter if that's better

src/ui/components/results/universalresultscomponent.js
oshi97: yeah, thanks for the catch!

README.md
oshi97: removed

oshi97: I honestly don't think it's terrible to tell them what the oobx behavior is, but sure

README.md
creotutar: very interesting we chose that. mapbox is one word

src/ui/components/results/universalresultscomponent.js
oshi97: Ok, it's been updated to use a verticals config option!

oshi97: Originally in the spec it actually said config.{VERTICAL_CONFIG_ID}.section, I was confused and thought config meant the actual top level config, not an attribute in the config (called config)

src/ui/sass/modules/_Card.scss
```diff
@@ -51,3 +51,10 @@
     }
   }
 }
+
+.yxt-UniversalResults .yxt-Card {
```
creotutar: if this is UniversalResults specific styling, can we give the card a modifier?
we dont want to use `--universal` above?

src/ui/sass/modules/_Results.scss
```diff
@@ -70,90 +65,112 @@
...
+
+  &-items {
+    margin-top: 0;
+    margin-bottom: 0;
+    padding: 0;
+    list-style: none;
     display: flex;
-    align-items: center;
+    flex-flow: row wrap;
+    margin-right: -1 * $results-cards-margin;
```
creotutar, line 81: why do we need a negative margin?

oshi97: when cards are in columns, they need margin between each other, so I did margin-bottom and margin-right. But, this leaves a bit of extra margin on the right edge of the screen which Rose didn't like. Since we want to page to be resizable, I couldn't think of an easy way to say "hey, make the cards on the right edge of the screen not have a margin-right", so I settled on this instead.

src/ui/sass/modules/_Card.scss
oshi97: updated!

src/ui/templates/results/verticalresults.hbs
```diff
@@ -1,38 +1,68 @@
...
-  {{#if resultsPresent}}
-    <section class="yxt-Results yxt-Results--{{_config.modifier}}">
-      {{#if _config.isUniversal}}
-        {{> results/resultssectionheader}}
+{{/inline}}
+
+{{#*inline "header"}}
+  {{#if _config.isUniversal}}
+    <div class="yxt-Results-title">
+      {{#if iconIsBuiltIn}}
```
creotutar, line 23: is this necessary? iirc, in iconcomponent, if you supply an `iconUrl` then it uses the url with an img tag, o/w uses the iconName

(historically, ive seen iconName and iconUrl passed in and if iconUrl is null/empty it is fine and falls back to the iconName

src/ui/sass/modules/_Card.scss
```diff
@@ -1,53 +1,72 @@
...
+
+.yxt-Card--placeholder
+{
+  visibility: hidden;
+  height: 0;
+  margin: 0;
+  padding: 0;
+  min-width: $cards-min-width;
+
+  *
```
creotutar: using a `*` is usually not good practice as it makes it hard if we want something /except one thing/ to have this styling. this is probably not a problem for this case, just making a note

also, any reason we dont want the display: none on just `yxt-Card--placeholder`? should hide the the card and its contents just fine instead of having the css engine have to apply a style to all descendants

creotutar (approved):
🚀 LGTM

src/ui/sass/modules/_Card.scss
oshi97: display:none would make it not take up any space, but we need it to take up space for the columns

oshi97: these placeholder cards should be empty anyways though, updated the pr

src/ui/templates/results/verticalresults.hbs
oshi97: in this case they're the same variable though

PR #571, opened 2020-03-17
Fix aria-label for search bar.

oshi97:


oshi97:
J=SPR-1817
TEST=manual
Used Wave chrome browser extension to see that wcag error is gone after change

alexisgrow (approved):

PR #577, opened 2020-03-19
Applied filters

oshi97:
This PR is separated into 4 sections:

1. consolidate filter/facet/sorting getstate logic for vertical search
    - There was a lot of repeated logic surrounding getting all filters before a search
2. Use FilterView instead of Filter
    - This refactors all filter classes to use FilterView instead of Filter, which contains a Filter and
FilterMetadata
3. Make Filters work on Page Reload
    - We've known about this problem for a while: filters don't repopulate themselves on page
load even though they pretend like they are storing data in the url to do that
4. Display applied static, facet, and sort filters in the results header


README.md
```diff
@@ -496,15 +496,24 @@ ANSWERS.addComponent('VerticalResults', {
   // Set a maximum number of columns that will display at the widest breakpoint.
   // Possible values are 1, 2, 3 or 4. Defaults to 1
   maxNumberOfColumns: 3,
-  // Whether to display the total number of results. Defaults to true
-  showResultCount: true,
-  // The card used to display each individual result, see the Cards section for more details,
+  // Whether to display the count of results at the very top of the results, default true.
+  showResultsCount: true,
```
creotutar: is this a breaking change?

src/core/core.js
```diff
@@ -97,8 +97,6 @@ export default class Core {
    * @param {string} verticalKey vertical ID for the search
    * @param {object} query The query details
    * @param {string} query.input The input to search for
-   * @param {string} query.filter The filter to use in the search
```
creotutar, line 100: I don't see these being removed in this PR. Was this just not being used before?

src/core/core.js
```diff
@@ -111,16 +109,37 @@ export default class Core {
       this.globalStorage.set(StorageKeys.LOCATION_BIAS, {});
     }
 
+    // If no verticalKey is set try to get verticalKey from the common search config
```
creotutar: I think this is self-explanatory and doesn't need a comment

oshi97: sure

src/core/core.js
oshi97: Ah yeah they're being removed in this commit. All places that call core.verticalSearch should not be passing their own filters in anymore.

README.md
oshi97: No, sorry Rose pulled a fast one on me during the v13 qa to rename this from showResultsCount to showResultCount, and I wrote the rough draft for this code before that particular qa item.

oshi97:
rebased v0.14.0 on top of master

src/core/models/filtermetadata.js
```diff
@@ -0,0 +1,50 @@
+/** @module FilterMetadata */
+
+/**
+ * Metadata helper for {@link Filter} or {@link Facet}
+ * Metadata structure is
+ * {
+ *  c_birdTypes: {
+ *    'Your Favorite Bird': ['rooster', 'penguin', 'turkey', 'chikn', 'dog']
```
tmeyer2115: It's been a while since I looked at this. Was the consensus that we can have multiple filters per field? If so, do we also need to include some notion of how they are combined here? Could we have two filters on the same field that are and-ed together? How would we represent that metadata here?

src/core/core.js
```diff
@@ -111,16 +109,34 @@ export default class Core {
       this.globalStorage.set(StorageKeys.LOCATION_BIAS, {});
     }
 
+    if (!verticalKey) {
+      verticalKey = this.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
```
tmeyer2115: Out of curiosity, why would someone be calling verticalSearch without a verticalKey? That seems like a prerequisite.

src/core/storage/storagekeys.js
```diff
@@ -13,10 +13,10 @@ export default {
   ALTERNATIVE_VERTICALS: 'alternative-verticals',
   AUTOCOMPLETE: 'autocomplete',
   DIRECT_ANSWER: 'direct-answer',
-  FILTER: 'filter',
+  FILTER_VIEW: 'filter-view',
   QUERY: 'query',
   QUERY_ID: 'query-id',
-  FACET_FILTER: 'facet-filter',
+  FACET_FILTER_VIEW: 'facet-filter-view',
```
tmeyer2115: Again, apologies because it's been a while. Are we storing filter views in persistent storage as well?

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -100,6 +102,8 @@ export default class DateRangeFilterComponent extends Component {
     DOM.delegate(this._container, '.js-yext-date', 'change', (event) => {
       this._updateRange(event.target.dataset.key, event.target.value);
     });
+
+    this.core.setFilterView(this.name, this._buildFilterView());
```
tmeyer2115, line 122: How come we weren't setting the filter before?

src/ui/components/filters/facetscomponent.js
```diff
@@ -148,6 +148,7 @@ export default class FacetsComponent extends Component {
      * @private
      */
     this._filterbox = null;
+    this.needsInitialSearch = true;
```
tmeyer2115, line 151: What is needsInitialSearch?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -203,24 +204,25 @@ export default class FilterBoxComponent extends Component {
           showReset: this.config.resetFilter,
           resetLabel: this.config.resetFilterLabel,
           showExpand: this.config.expand,
-          onChange: (filter) => {
-            this.onFilterChange(i, filter);
+          isDynamic: true,
```
tmeyer2115: What's isDynamic?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -263,43 +266,25 @@ export default class FilterBoxComponent extends Component {
...
   /**
    * Trigger a search with all filters in storage
    */
-  _search () {
-    const allFilters = this.core.globalStorage.getAll(StorageKeys.FILTER);
-    const totalFilter = allFilters.length > 1
-      ? Filter.and(...allFilters)
-      : allFilters[0];
-
+  search () {
```
tmeyer2115, line 281: Should this still be _search()? Is it functionally used as a private method on the class?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -2,69 +2,165 @@
 
 import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
+import { DOM } from '../..';
```
tmeyer2115: Does there need to be a file name in this import?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -2,69 +2,165 @@
...
+  appliedQueryFilters: []
 };
 
 export default class ResultsHeaderComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super({ ...DEFAULT_CONFIG, ...config }, systemConfig);
-
-    const data = config.data || {};
-
+    this.data = Object.assign({}, DEFAULT_DATA, config.data);
```
tmeyer2115: Do we need to default the data passed in to this component? I'm not sure of that behavior. Seems like a bit of a slippery slope.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -2,69 +2,165 @@
...
+    this.isExpanded = false;
     /**
-     * Total number of results.
+     * Line height of the results header.
      * @type {number}
      */
-    this.resultsCount = data.resultsCount || 0;
+    this.lineHeight = 0;
```
tmeyer2115, line 29: Is there a spec for the ResultsHeaderComponent? I want to understand why we need the lineHeight in JS and the handleResize and handleOverflow.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -2,69 +2,165 @@
...
+        appliedFilters[label].push(...displayValues);
+      });
     });
-    // Has to be parsed into an array because our handlebars can only loop through arrays, not objects.
-    return Object.keys(groupedFilters).map(label => ({
-      label: label, displayValues: groupedFilters[label]
+
+    // Also add nlp filters from the backend
+    this.data.appliedQueryFilters
+      .filter(f => !this._config.hiddenFields.includes(f.fieldId))
```
tmeyer2115, line 111: What would happen if someone specifies a field as hidden, but also makes a Filter component from it?

src/core/core.js
oshi97: Right now verticalKey is optional config for the geolocation filter. If the geolocation filter has searchOnChange = true, but no verticalKey, it will try to make vertical search without verticalKey. 

Since this looks funny to you, and after checking the code only the geolocation filter has this issue, I'll just update the readme for the geolocation filter and say that verticalKey is required if searchOnChange is true.

src/core/models/filtermetadata.js
oshi97: Yes we can have multiple filters per field. Up to this current commit, combining filtermetadata loses one-to-one pairing between the original filters and metadata, lumping them into one big filter and metadata. 
We need this information, though, to be able to remove filters by clicking on their tag in the results header in a later item. I'm updating the code the be able to do this.

src/core/storage/storagekeys.js
oshi97: yup!

src/ui/components/filters/daterangefiltercomponent.js
oshi97: The date range filter WAS initializing its own state from persistent storage correctly. It was just not also updating its globally stored filter, if there was some state carried over in persistent storage.

oshi97: I just asked the PMs on what exactly the behavior should be about populating the filters from persistentStorage. 
Currently the code just always does it, however that will force the user to always have a filter set, if it is on the page. I actually think this makes since, but it's possible this was not their intention and they have some grander ideas.

src/ui/components/filters/facetscomponent.js
oshi97: When the Facets are created, we need to run a search to apply the initialized facets. This cannot be done in the constructor/onCreate of the FacetsComponent, because the FilterBox that contains and creates the actual filters are created onMount, which happens after both of these. It also cannot be done inside the FilterBox, because that will trigger all of the Facets to update, recreating the FilterBox, which then will do another search, infinitely.

I think a better alternative might be to store the FacetView in persistentStorage, and onCreate set globalStorage from that. However, this has potential issues with hitting the url character limit or 2048.

src/ui/components/filters/filterboxcomponent.js
oshi97: The FacetComponent sends an isDynamic config option to the FilterBox component, and I needed a way to tell the FilterOptions component to not set it's filter in global storage if it was part of a facet. Looking back, the storeOnChange option already does this, so I will update the code to use that instead.

src/ui/components/filters/filterboxcomponent.js
oshi97: I needed to call this function from outside of this component, inside of the FacetsComponent, so I thought it made sense to make it public.

src/ui/components/results/resultsheadercomponent.js
oshi97: Technically this works fine (it was done by my auto-import thing) but I can change it

src/ui/components/results/resultsheadercomponent.js
oshi97: Yes, if the contents of the header overflow past one line, we want there to be a dropdown button that will expand to show the rest of the contents.
Essentially we want it to be an accordion, but only when things would overflow past one line.

src/ui/components/filters/daterangefiltercomponent.js
oshi97: We decided on having null be a special value to mean the range value (min or max) should be unset.

src/ui/components/filters/facetscomponent.js
oshi97: I tried this, as well as a few changes to cut down on the amount of noise in the URL, but if I clicked a sufficient amount of facets I was still able to go over the limit. I think there is definitely a better way, but after some thinking I haven't really thought of one and don't want to drag my feet any longer on it.

src/core/storage/storagekeys.js
oshi97: On second thought, we should be trying to minimize what we store in persistent storage. I changed my PR to try to only store the minimum needed state for each filter component. There are a few cases where the filter is received as a response from the backend, in which case I was forced to store the whole filter (but still did not store the filter view).

src/ui/components/results/resultsheadercomponent.js
oshi97: I guess you're worried about relying on defaults when other code is not working properly? I'll remove this. I wanted to be explicit about what was being passed in as data but I guess it's just extraneous code.

src/ui/components/results/resultsheadercomponent.js
oshi97: I mean, it would be hidden? Isn't that a good thing?

oshi97: Actually I thought about it more, and you have a good point. The current behavior is what is outlined in the spec (this is list of filters that should not be displayed) however it probably makes sense to only hide filters from the backend? I'll ask the PMs when we meet to talk about the removable filters.

PR #578, opened 2020-03-20
Navigation Component: Empty or undefined verticalKey has same behavio…

alexisgrow:
…r as null

If an implementer passed in a verticalKey that was empty or undefined, its
entry would be ignored in the navigation component. Changing the null check
so any falsy verticalKey is treated the same.

TEST=manual

Test on local answers instance that universal verticalPages config is
used when omitting vertical key.


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -99,7 +99,7 @@ export class Tab {
 
       // For tabs without config ids, map their URL to the configID
       // to avoid duplication of renders
-      if (tab.verticalKey === null && tabs[tab.verticalKey] === undefined) {
+      if (!tab.verticalKey && !tabs[tab.verticalKey]) {
```
oshi97: I know it was like this in the old code, but do we need the `&& !tabs[tab.verticalKey]` ?

alexisgrow: Oh good point, is this the logic we should have: 
`if (!tab.verticalKey && !tabs[tab.url])`?

oshi97: yes that looks good to me!

oshi97 (approved):
lgtm!

PR #579, opened 2020-03-20
onVerticalSearch + onUniversalSearch analytics hooks

oshi97:
Consulting needs to fire analytics events when searches are run.
They cannot rely on current life-cycle methods like setState,
or onMount, because these life-cycle methods are also used for
things like javascript interaction, and are also often called
multiple times (per component) for even just one user interaction,
like a new search.

Instead, I think it makes sense to give consulting (and hitchhikers)
a hook onto both vertical and universal search. They will be presented
with several exposed variables related to a search, and asked to
write a method that returns their desired analytics event. Presenting
them their own designated interface, instead of simply giving
them all of the search response data as well as the analytics reporter,
allows us to make changes to either in the future without breaking
their code.

TEST=manual
Check that only vertical fires on vertical and only universal fires on universal
Check that _globalOptions are being applied
Check that params are accessible from method
Check that “this” context within method is the window, and does not expose stuff
Check that validation works for both options, and error message says the current typeof
Check that you can add arbitrary event options for both cases
Check that _globalOptions queryId still changes when changing the query through the searchbar
Check all expected exposedParams are exposed in the onSearch functions


src/answers-umd.js
```diff
@@ -254,6 +259,14 @@ class Answers {
...
       throw new Error('Missing required `experienceKey`. Type must be {string}');
     }
+
+    if (config.onVerticalSearch && typeof config.onVerticalSearch !== 'function') {
+      throw new Error('onVerticalSearch must be a function. Current type is: ' + typeof config.onVerticalSearch);
+    }
+
+    if (config.onUniversalSearch && typeof config.onUniversalSearch !== 'function') {
+      throw new Error('onUniversalSearch must be a function. Current type is: ' + typeof config.onUniversalSearch);
+    }
```
alexisgrow, line 269: I don't think `businessId` is in our README as part of the config, but it is required for analytics. Would it be possible to add to the README with a note that it's required for analytics and/or add a warning message here saying that analytics are not going to be registered because there is no businessId specified?

oshi97: That's a great idea! Updated the readme.

alexisgrow (approved):
LGTM!

README.md
```diff
@@ -116,6 +119,80 @@ ANSWERS.registerHelper('noop', function(options) {
...
+     * The total number of results found.
+     * @type {number}
+     */
+    const resultsCount = searchParams.resultsCount;
+
+    /**
+     * Either 'normal' or 'no-results'.
+     * @type {string}
+     */
+    const resultsContext = searchParams.resultsCount
```
creotutar, line 158: this says `searchParams.resultsCount`

README.md
```diff
@@ -116,6 +119,80 @@ ANSWERS.registerHelper('noop', function(options) {
...
+    const analyticsEvent = {
+      type: 'ANALYTICS_EVENT_TYPE',
+      label: 'Sample analytics event',
+      query: queryString
+    };
+    return analyticsEvent;
```
creotutar, line 164: does returning do anything in the lib?

creotutar: i see. this fires a library event. do we want to make it clear that an object return is optional?

PR #583, opened 2020-03-23
Add additional label text configuration to QA and Pagination components.

tmeyer2115:
This PR adds the following label text configuration options:
     - networkErrorText to the QASubmission component. This text is shown
       when the submission request results in an error.
     - pageLabel to the Pagination component. This is the text indicating a
       page of results.

TEST=manual

Made sure both new configuration options worked. Also made sure that their
defaults worked.


creotutar (approved):
🙏 🙏 🙏 LGTM

src/ui/templates/results/pagination.hbs
```diff
@@ -24,7 +24,7 @@
             data-opts='{ "iconName": "chevron" }'></span>
     </button>
 
-    <span class="yxt-Pagination-page">Page {{pageNumber}}</span>
+    <span class="yxt-Pagination-page">{{pageLabel}} {{pageNumber}}</span>
```
benmcginnis, line 27: i think for proper i18n you need both of these as one string.  This is probably ok for now but for Japan for example we've had a lot of issues with how they actually phrase things.  I'm not really certain this is a particular instance but in soy i18n we'd pass the whole thing with the pageNumber variable included.  Not sure how to do that in hbs

benmcginnis (approved):

PR #584, opened 2020-03-25
noresults: allow messaging to be configured

benmcginnis:


src/ui/templates/results/noresults.hbs
```diff
@@ -1,20 +1,32 @@
...
+        {{_config.noResultsMessage}}
+      {{else}}
+        Your search - <span class="yxt-NoResults-query">{{query}}</span> - did not match any answers we have.
+      {{/if}}
     </p>
     {{#if universalUrl}}
       <p class="yxt-NoResults-returnLinkWrapper">
-        Try refining your query or <a href="{{universalUrl}}" class="yxt-NoResults-returnLink">search all content.</a>
+        {{#if _config.noResultsTryUniversalMessage}}
+          {{{_config.noResultsTryUniversalMessage}}}
```
creotutar, line 13: why triple?

creotutar: i guess if someone adds html to this config option, they're not trying to sabotage the site

src/ui/templates/results/noresults.hbs
```diff
@@ -1,20 +1,32 @@
 <div class="yxt-NoResults">
   <div class="yxt-NoResults-wrapper">
     <p class="yxt-NoResults-info">
-      Your search - <span class="yxt-NoResults-query">{{query}}</span> - did not match any answers we have.
+      {{#if _config.noResultsMessage}}
+        {{_config.noResultsMessage}}
```
creotutar, line 5: will a span work here?

creotutar (approved):
🚀 🚀 🚀 LGTM

src/ui/templates/results/noresults.hbs
benmcginnis: yeah the purpose of this message is to prompt a user to visit the main universal results page, so it has to accept html

src/ui/templates/results/noresults.hbs
benmcginnis: No, the query can't be passed in. Jon Hyler would execute me even if we're going to be responsible and pre-escape things, someone else might not.

tmeyer2115 (approved):

PR #586, opened 2020-03-25
Gulpfile: Add answers-umd build to support node_module import

alexisgrow:


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -90,6 +90,53 @@ function legacyBundle () {
     .pipe(dest('dist'));
 }
 
+function legacyBundleUMD () {
```
tmeyer2115, line 57: How about we make a legacyBundle() function that takes in an output config? Then we can have legacyBundleUMD and legacyBundleIIFE call into that with their respective configs. This can help us prevent a future mistake where an update is needed for both tasks, but someone only adds it to one and not the other.

tmeyer2115 (commented):
Looks like the CircleCI config just copies whatever is in dist to AWS/CDN. So, shouldn't need to make any changes there!

conf/gulp-tasks/library.gulpfile.js
alexisgrow: Ah yeah, I like that! Updating now

tmeyer2115 (dismissed):

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -127,11 +147,12 @@ function watchCSS (cb) {
...
   series(bundle, minifyJS),
-  series(legacyBundle, minifyLegacy),
+  series(legacyBundleIIFE, minifyLegacy),
+  series(legacyBundleUMD, minifyLegacyUMD),
   series(compileCSS)
 );
 
 exports.dev = parallel(
-  series(bundle, legacyBundle, watchJS),
+  series(bundle, legacyBundleIIFE, watchJS),
```
tmeyer2115: legacyBundleUMD too?

alexisgrow: I'll update! Didn't think we would need it in dev, but you're right that we may. I'll fix!

tmeyer2115 (approved):

PR #587, opened 2020-03-26
FieldFormatters: implement formatRichText

oshi97:
In order to display rich text from knowledge graph,
we need to parse it into html. This commit use MarkdownIt
to parse markdown into raw html, putting that functionality
in ANSWERS.formatRichText().

Underline is not part of the commonmark specification, however
it has been implemented in our other services as '++'.
A new MarkdownIt rule called yext_underline was written and
added to formatRichText to be able to parse out underlines.

TEST=manual, auto
Tested that both standard and accordion card can display
rich text field from my entity (Oliver Shi) in roses test account.
The rich text stored in that entity was also used to create the
last test of the unit tests added in this commit.


oshi97:
Increased the size limit to 250kb because MarkdownIt pushes the bundle size to 127.66KB. Using the minified version of MarkdownIt pushed the bundle size down to 127.65KB. We could reduce the bundle size by only including MarkdownIt rules we need, instead of all rules, but I don't think we have time for that right now.

.size-limit.js
```diff
@@ -1,7 +1,7 @@
 module.exports = [
   {
     path: 'dist/answers.min.js',
-    limit: '100 KB'
+    limit: '250 KB'
```
tmeyer2115: What's the exact bundle size now that we're including markdown-it as a dependency? I think that 250 KB may be too big. Webpack suggests a size limit of ~240 KB for a bundle. Larger than that and performance can be negatively impacted. @bryanreed, I looked at commonmark-js and both it and markdown-it have a large unpacked size (430 KB and 570 KB, respectively). I think we may have a couple options here:

1. Bite the bullet and appreciably increase the SDK's bundle size.
2. Move this to the theme, which will have a much smaller bundle size.

src/answers-umd.js
```diff
@@ -422,4 +423,5 @@ function initScrollListener (reporter) {
 }
 
 const ANSWERS = new Answers();
+ANSWERS.formatRichText = formatRichText;
```
tmeyer2115: We should expose formatRichText the same way we do Component. In the constructor, have something like:

this.formatRichText = formatRichText.

src/core/markdown/richtext.js
```diff
@@ -0,0 +1,19 @@
+import MarkdownIt from 'markdown-it/dist/markdown-it';
+import yextUnderline from './yext_underline';
+
+function formatRichText (markdown) {
+  if (typeof markdown !== 'string') {
+    console.warn(`Rich text "${markdown}" needs to be of type string. Currently is of type ${typeof markdown}`);
```
tmeyer2115: I think this should be wrapped in an AnswersError.

src/core/markdown/yext_underline.js
```diff
@@ -0,0 +1,36 @@
+export default function (state, silent) {
```
tmeyer2115: Instead of us having to define a new markdown-it rule, could we use this existing plugin: https://www.npmjs.com/package/markdown-it-plugin-underline? Ostensibly, it does what we need. It knows to parse ++blah++ as underlined.

tests/core/markdown/yext_underline.js
```diff
@@ -0,0 +1,71 @@
...
+      '* ++I am underline list++\n\n' +
+      '1. ++I am number list++\n' +
+      '2. ++[I am link to site](http://olivershi.io)++\n\n' +
+      '++[url link](http://google.com)++\n\n' +
+      '++[phone link](tel:+17326183404)++\n\n' +
+      '++[email link](mailto:oshi@yext.com)++\n';
+
+    const html = formatRichText(richText);
+    const expectedResult =
+`<p><strong>I AM BOLD</strong> now I am not <em>I AM ITALICS</em> now I am not <u>BRASAAAAAP</u></p>
```
tmeyer2115: Indent this block

.size-limit.js
oshi97: New size is 127KB I can make it 150KB limit maybe? I thought I remembered you said webpack recommended a size of 250KB when Alexis was going through the core-js stuff but I guess I remember wrong.

tmeyer2115: Or 140 KB. We want to keep this as low as possible. Webpack suggests a maximize bundle size of 244 KB, or between 240-250 KB.

src/core/markdown/yext_underline.js
oshi97: updated, thanks for finding this!

.size-limit.js
oshi97: updated to 150KB

README.md
```diff
@@ -800,7 +800,7 @@ const dataMappings = item => {
     image: item._raw.headshot ? item._raw.headshot.url : '',
     // Url for the title/subtitle, defaults to the entity's website url
     url: item.link || item._raw.website,
-    // Character limit to hide remaining details and display a show more button, defaults to 350
+    // Character limit to hide remaining details and display a show more button, defaults to not having a limit.
```
tmeyer2115, line 803: Was this something Rose said was needed for RTF?

src/core/markdown/richtext.js
```diff
@@ -0,0 +1,19 @@
+import MarkdownIt from 'markdown-it/dist/markdown-it';
+import underline from 'markdown-it-plugin-underline';
+import { AnswersCoreError } from '../errors/errors';
+
+function formatRichText (markdown) {
```
tmeyer2115: Minor nit, but export default function formatRichText is more succinct.

src/core/markdown/richtext.js
```diff
@@ -0,0 +1,19 @@
+import MarkdownIt from 'markdown-it/dist/markdown-it';
```
tmeyer2115: Maybe we should make this a class that exposes a method called formatRichText. As part of the class constructor, we could create an instance of markdown-it and register the plugin. This would save us from having to do all that work each time we call formatRichText.

tmeyer2115: This could probably live in core/utils as it's a markdown utility function and there's only one file in core/markdown.

tests/core/markdown/underline.js
```diff
@@ -0,0 +1,71 @@
+import formatRichText from '../../../src/core/markdown/richtext';
```
tmeyer2115: This can just be a test for our new class.

src/core/markdown/richtext.js
oshi97: I thought it should live in its own folder since we are probably adding extensions in the future.

oshi97: Eh okay I changed my mind

README.md
oshi97: Put a message in slack this morning, I don't think it makes any sense to have a character count limit for rich text fields. Currently there's no way to specify no limit, since the default would always be used. Alternatively, we could have null be no limit but I don't think that's as intuitive.

oshi97: I guess we can have them just set it manually to Infinity for now.

oshi97: Rose just responded she said it sounds good to her to have it default to no limit

tests/core/utils/richtext.js
```diff
@@ -0,0 +1,73 @@
+import RichText from '../../../src/core/utils/richtext';
```
tmeyer2115, line 1: Let's add one more test to make sure we're escaping any kind of HTML in the RTF, such as a div tag

tmeyer2115 (approved):

PR #588, opened 2020-03-26
SearchBar: Change default icon to new yext-animated-icon

oshi97:
This commit changes the default icon for the searchbar
to the new animated icon created by Jeremy, that morphs
from a yext logo to search bar when the search bar is focused
on back to a yext logo on blur.

In order for the two svgs Jeremy gave to not clash
with each other namespace wise, as well as not have
multiple icons on the searchbar clash with each other,
The complexContents property of icons was updated to
be a function if needed, and can receive a complexContentsParams
parameter if needed.

The werid nth-of-type css was done to extract the styling
from jeremy's svgs, which when he posted them in the slack
channel were embedded individually within the <g> elements.

TEST=manual
Test that the new icon animates as expected, is used by default,
can be overriden by customIconUrl or submitIcon config options


oshi97:
rose has demo at https://answers-js-demo.netlify.com/sites/rose/icon.html?query=

src/ui/icons/yext_animated_forward.js
```diff
@@ -0,0 +1,208 @@
...
+      <rect x="-4" y="4" fill="#fff" width="720" height="720" />
+    </mask>
+    <mask id="${id}_forward_Mask-19">
+      <rect fill="#fff" width="720" height="720" />
+      <path d="M432.6 310.9c0 0-24.2 25.3-24.2 25.3c0 0 63.3 21.5 63.3 21.5c0 0 59.6-41.1 59.6-41.1c0 0 10.6-53.8 10.6-53.8c0 0-38.2 .6-38.2 .6c0 0-71.1 47.5-71.1 47.5m-71-79.8c0 0 20.3 90.4 20.3 90.4c0 0 29-29.5 29-29.5c0 0 69.8-70.1 69.8-70.1c0 0 39.3-31.9 39.3-31.9c0 0-65.9-30.5-65.9-30.5c0 0-92.5 71.6-92.5 71.6" />
+    </mask>
+    <mask id="${id}_forward_Mask-20">
+      <rect fill="#fff" width="720" height="720" />
+    </mask>
+    <clipPath id="${id}_forward_ClipPath-1">
```
benmcginnis: using Id's here could be a problem if there are multiple search bars on the page, e.g. because of a clients header

benmcginnis (approved):
lgtm!

src/ui/icons/yext_animated_forward.js
oshi97: I was hoping that in that scenario each SearchBar would get its own name. I think since SearchBar has areDuplicateNamesAllowed false (just the default value) an error shooould be thrown when you try to instantiate two searchbars with the same name?

oshi97: oh, maybe I shouldn't have called this id. This isn't the id of the svg element it's a passed in value from the component

oshi97: Updated to be called iconPrefix

benmcginnis: So the main problem is actually going to occur when there are two search bars initialized by two separate instances of answers (client header example) if the client doesn’t iframe us we’d have an issue with the JS not hooking properly.  I’d suggest converting all of these selectors to classes rather than ids

PR #589, opened 2020-03-26
Updates to support NJ Coronavirus site

tmeyer2115:
This PR includes various updates to the SDK. These updates were initially made to support the NJ Coronavirus site. However, because they are applicable to any Answers site, we are folding them into an official release. The changes are as follows:

* Updates to the QASubmission, Pagination, and NoResults partials to allow more of the hard-coded text to be configurable. 
* Updated the question submission route.
* Allow some attributes of the QASubmission component to accept raw HTML in their configuration.

TEST=manual

All changes have been verified manually.


MattCBowman (approved):
As long as this doesn't leave us open to getting hackt, lgtm

PR #590, opened 2020-03-26
VerticalResults: use old no results when noResults config option is n…

oshi97:
…ot set

TEST=manual
Tested that when there is no noResults config option set, will
use the old no results template. Otherwise, will use the new one.


alexisgrow (commented):
Instead of branching in the hbs, what do you think about having the logic in JS?

oshi97:
> Instead of branching in the hbs, what do you think about having the logic in JS?

sure that makes sense

alexisgrow (approved):
LGTM! 🔥

alexisgrow (commented):
Also! Talked to Rose and this should be merged into v0.13.1 🎉 
(because the fancy no results is still behind a feature flag, so can't be default behavior yet)

PR #593, opened 2020-03-27
Cards: Use result._raw rather than result for data mappings

alexisgrow:
TEST=manual

Test with client site, see console errors when using result, change to result._raw, everything works like v0.12.


oshi97 (commented):
This looks great (I think I should have done it this way to begin with), could we just check with Rose to make sure this is okay?  Cause the spec for cards has tons of _raw examples in it. We'd also want to update the readme too, since there are a few examples of _raw there as well.

We'd also need to update the dataForRender method in the hitchhiker theme (just writing this for later).

alexisgrow (commented):
Just got confirmation in slack answers-js-internal!

oshi97 (approved):
sweet haha

alexisgrow (commented):
Oh shoot I almost forgot, need to update the README before I ship 🏄‍♀

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -31,7 +31,7 @@ class AccordionCardConfig {
      * a card or an object that is the additional config.
      */
     const dataMappings = config.dataMappings || {};
-    Object.assign(this, CardComponent.applyDataMappings(result, dataMappings));
+    Object.assign(this, CardComponent.applyDataMappings(rawResult, dataMappings));
```
tmeyer2115, line 34: If we're operating on raw result, will that negatively impact anyone trying to access item.title (or something like that)? There are fields in result that aren't in rawResult, right?

tmeyer2115 (commented):
Is AEB providing the raw profile to the various formatters?

src/ui/components/cards/accordioncardcomponent.js
alexisgrow: It seems like the data mappings are to pull data from `_raw` into the top-level fields: ```title, details, link, subtitle, modifier, bigDate, image```. 

All of the NJ mappings errored when given the full `result` because the formatters try things like result.address without null checking, so it seems like they only want `result._raw`

tmeyer2115 (approved):

PR #595, opened 2020-03-27
Cards: Backwards Compatibility - CTAs, ordinals

alexisgrow:


src/ui/components/cards/cardcomponent.js
```diff
@@ -86,6 +86,7 @@ export default class CardComponent extends Component {
       result: data
     };
     const newOpts = {
+      showOrdinal: this._config.showOrdinal,
```
tmeyer2115, line 89: Prior to this, were we always showing ordinals on cards? There was no way to configure that?

src/ui/components/ctas/ctacomponent.js
```diff
@@ -30,7 +30,7 @@ class CTAConfig {
      * Analytics event that should fire:
      * @type {string}
      */
-    this.analyticsEventType = config.analyticsEventType || 'CTA_CLICK';
+    this.analyticsEventType = config.analytics || 'CTA_CLICK';
```
tmeyer2115, line 33: Do we need to update the README?

src/ui/templates/ctas/cta.hbs
```diff
@@ -1,12 +1,16 @@
-<a class='yxt-CTA js-yxt-CTA{{#each _config._ctaModifiers}} yxt-CTA--{{this}}{{/each}}'
+<a class='yxt-CTA js-yxt-CTA{{#each _config._ctaModifiers}} yxt-CTA--{{this}}{{/each}} yxt-Result-cta'
   href="//{{_config.url}}"
   target="{{_config.target}}"
+  data-eventtype="{{#if _config.analyticsEventType}}{{_config.analyticsEventType}}{{else}}CTA_CLICK{{/if}}"
```
tmeyer2115: The JS already has logic for handling CTA analytics. I think we should move this there. Otherwise, we might get two events firing. The JS logic does not issue a request if there are no eventOptions in the config. That may be an issue. Potentially we'll need to add a default if those aren't present in the config.

src/ui/components/cards/cardcomponent.js
alexisgrow: Prior to this, we weren't passing the config to the child, so this was always falling back to default (`false`): https://github.com/yext/answers/blob/v0.13.1/src/ui/components/cards/standardcardcomponent.js#L113

src/ui/components/ctas/ctacomponent.js
alexisgrow: Yes, I'll update!

src/ui/templates/ctas/cta.hbs
alexisgrow: Ah interesting, I'm unfamiliar with that, I'll look now!

alexisgrow: updated!

src/ui/components/ctas/ctacomponent.js
alexisgrow: updated

src/ui/components/cards/cardcomponent.js
oshi97: showOrdinal is expected to be part of dataMappings, is this line here for aeb?

src/ui/components/ctas/ctacomponent.js
```diff
@@ -69,14 +69,16 @@ export default class CTAComponent extends Component {
...
     }
   }
 
   reportAnalyticsEvent () {
     const analyticsEvent = new AnalyticsEvent(this._config.analyticsEventType);
-    analyticsEvent.addOptions(this._config.eventOptions);
+    if (this._config.eventOptions) {
+      analyticsEvent.addOptions(this._config.eventOptions);
+    }
     this.analyticsReporter.report(analyticsEvent);
```
tmeyer2115: @oshi97 I think we'd talked about this. We decided not to report an analytics event if there were no event options. That would correspond to sending an analytics event with no info (no entity id, vertical key, etc.). Should we have a sane default here for config.eventOptions?

oshi97: Part of the default event options v0.13.0 QA item was changing CTAs to have a default value for event options if none were specified. This is done in CTACollectionComponent line 48. Were analytics not firing without these changes @alexisgrow ?

alexisgrow: Hmm, the default should be working. I'll check now and report back, but I think we can remove this check that I added.

src/ui/components/cards/cardcomponent.js
alexisgrow: Wait so it should be `card: { dataMappings: { showOrdinal: true } }`? `showOrdinal` isn't really a data mapping though, right?

README.md
```diff
@@ -579,7 +579,7 @@ const callsToAction = [{
   //       'EMAIL',
   //       'BOOK_APPOINTMENT',
   //       'RSVP'
-  analyticsEventType: 'CTA_CLICK',
+  analytics: 'CTA_CLICK',
```
oshi97, line 582: I did a quick test locally and don't think the eventType option in analytics is correctly set if we do it like this. Could you double check?

oshi97: I tried analyticsEventType: "BOOK_APPOINTMENT" and also  analytics: "BOOK_APPOINTMENT"

src/ui/components/cards/cardcomponent.js
oshi97: well it couuuuld be a function haha

README.md
alexisgrow: Yeah I can check, I'll post back here with an update.

src/ui/components/ctas/ctacomponent.js
alexisgrow: Just checked, it works, removing the changes here.

src/ui/components/cards/cardcomponent.js
alexisgrow: ah gotcha so it could depend on the result somehow, I can update the AEB and test without this SDK change

src/ui/components/ctas/ctacomponent.js
alexisgrow: updated!

README.md
alexisgrow: hmm that works for me: https://drive.google.com/file/d/1wCsXw8yg_qbbABGjux8QSJo7mCP5eRaV/view?usp=sharing

this is my HTML file: https://drive.google.com/file/d/1VBb5sSg24QULQ2sFc4D4ELhJqNRtlpJl/view?usp=sharing

tmeyer2115 (approved):

README.md
oshi97: ok sweet!

src/ui/components/cards/cardcomponent.js
alexisgrow: I asked Rose about this and now there is a thread in Slack!  Her preference is that `showOrdinal` be a direct child of the cards section, e.g. `cards: { showOrdinal: true }`, so this change accurately reflects what we are going for. Resolving with no changes!

PR #596, opened 2020-03-27
Legacy Card: Add legacy card

alexisgrow:
Adapted from v0.12's 'resultsitem.hbs' to preserve classnames
and div structure. Copied styling exactly from v0.12's result scss.

TEST=manual

Test on client site using standard card vs legacy card, notice
client styling is preserved with this legacy card.


src/ui/templates/cards/legacy.hbs
```diff
@@ -0,0 +1,62 @@
...
+    {{/if}}
+
+    <div class="yxt-Result-infoCtasWrapper">
+      <div class="yxt-Result-info">
+        <h3 class="yxt-Result-title">
+          {{#if _config.url}}
+            <a class="yxt-Result-titleLink"
+               href="{{_config.url}}"
+               data-eventtype="TITLE_CLICK"
+               data-eventoptions='{{eventOptions}}'
```
tmeyer2115, line 33: Do these eventoptions need to be defined in the legacy card's JS?

alexisgrow: Yikes yes good catch ✨ , updated!

tmeyer2115 (approved):

PR #597, opened 2020-03-27
Create configuration options for remaining hard-coded strings.

tmeyer2115:
I recently did a survey of the SDK's templates to determine all the instances
of hard-coded text. All text in a template needs to be configurable to support
internationalization. This means adding configuration options to the
corresponding component(s) to drive these strings. For each new configuration
option, I added a sane default that was the existing, hard-coded value.

J=SPR-1903
TEST=manual

I went through each instance of hard-coded text I'd noted and made sure the
text was now coming from a component's configuration.


oshi97:
*nit but remaining is spelled wrong

alexisgrow (approved):
🔥 🔥 LGTM

PR #598, opened 2020-03-27
CTAs: Allow cta data mapping functions to return more than one cta, update warnings

alexisgrow:
Currently, clients are using a CTA data mapping function that iterates
through a custom field from KG and returns multiple CTAs. Update
resolveCTAMapping so that it can support that behavior. Also, move warnings
so that they are accurate with CTA data mapping functions as well as objects.

TEST=manual

Run with client implementation locally with adjusted code.


oshi97 (approved):
lgtm!

PR #600, opened 2020-03-30
Legacy Card: Do not truncate and show ellipsis

alexisgrow:
The Legacy card doesn't have config options to `hideExcessDetails`, so remove that logic.

TEST=manual

Run experience locally, see details show up properly.


oshi97 (approved):
lgtm!

PR #601, opened 2020-03-30
SearchBar: do not show autocomplete after search

oshi97:
This commit removes the focusInputElement() call after a search is run.
This was the behavior prior to the icon change #588.

TEST=manual
autocomplete does not display after a search is run
autocomplete still displays on load if autocompleteOnLoad is true
Animated icon returns to yext logo state after search, since input element
  Is no longer focused


alexisgrow (approved):
🔥

PR #603, opened 2020-03-30
SearchBar: Fix autocomplete

oshi97:
Autocomplete uses the same css selector as
the search input, and relies on doing form.querySelector
instead of DOM.query(this._container, this._inputEl)

TEST=manual
test that both clicking an autocomplete option, as well
as using the arrow keys and pressing enter, search
with the selected option


alexisgrow (approved):
✨

PR #604, opened 2020-03-30
SearchBar: Do not animate icon on load

oshi97:
The aniamted searchbar icon should only
begin animating once the user clicks into the searchbar

TEST=manual
Test that on page load the icon is static


src/ui/templates/search/search.hbs
```diff
@@ -37,7 +37,7 @@
               data-opts='{ "iconUrl": "{{_config.customIconUrl}}", "classNames": "Icon--lg" }'
           ></div>
         {{else}}
-          <div class="js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon"
+          <div class="js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon yxt-SearchBar-AnimatedIcon--paused"
```
alexisgrow, line 40: nit: make this a `js-` class since it's being targeted in JS

alexisgrow (approved):
overall LGTM, left one comment

src/ui/templates/search/search.hbs
oshi97: That seems a little weird to me becuase yxt-SearchBar-AnimatedIcon--paused is used purely to apply styling. I'm using js-yxt-AnimatedForward as the js-selector so I think this is fine?

alexisgrow: You definitely have more context since this is a piece of the larger thing you're doing, so totally your call! The reason I said that is because in this PR there is `forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');` in the search component's JS. If it's never being added by the JS, but it is removed by the JS, then it could have a `js-` prefix since it's a class that the JS has a dependency on. In this case you're definitely right that it is used mostly for styling, and I don't really have strong opinions on it, hahaha feel free to merge

oshi97: o ok just wanted to make sure!

oshi97: thank you for the explanation!

PR #605, opened 2020-03-30
SearchBar: use 16x16 icon size for non-animated icons

oshi97:
Rose decided that 32x32 looked weird for non-animated
search bar icons.

TEST=manual
when customIconUrl or submitIcon are specified in the config,
  size is 16x16 and searchBar height is 40px


alexisgrow (approved):
LGTM!

PR #607, opened 2020-03-30
Filters: set filters on page load for range filters

oshi97:
Prior to this change, range filters stored inside
of persistent storage (the url) were not being applied
to the search on page load.

TEST=manual
Test that changing the filters, then clicking search
  still applies filters to request
Test that refreshing the page after that results
  in initial search using previous filters


alexisgrow (approved):
LGTM!!

PR #611, opened 2020-03-30
If allResultsForVertical is present, use the included facets.

tmeyer2115:
This PR updates the logic in the DynamicFilters.from method. Now, if the
provided response includes allResultsForVertical, we extract the facets from
there. For the sake of consistency, I added similar logic to the
VerticalResults.from method. I added a TODO to the SearchDataTransformer. I
think it would ultimately be best to have SearchDataTransformer create a
parsed version of the response and pass that down. This parsed version would
have results, resultsCount, and facets pulled up from allResultsForVertical
(if applicable).

TEST=auto, manual

Updated the VerticalResults unit tests. Manually tested the following:

- Performed a query that had results. Saw all the included facets in the
  response correctly populated in the UI.
- Performed a query that had no results. Saw all the facets from
  allVerticalResults appear in the UI.


src/core/search/searchdatatransformer.js
```diff
@@ -31,6 +31,7 @@ export default class SearchDataTransformer {
   }
 
   static transformVertical (data, formatters, verticalKey) {
+    // TODO (tmeyer): SPR-1923, centralize the allResultsForVertical parsing here.
```
alexisgrow, line 34: Ah sweet glad this is happening (in future), this was my main comment on this PR!

alexisgrow (approved):

PR #612, opened 2020-03-30
SortOptions: apply sort from url on load

oshi97:
This commit fixes the sort options component to
apply a sort stored in the url on the initial search
that runs when a page is loaded. Previously, the
sort option was being set visibly but was not being
applied to the request.

Also fix a bug where the assignDefaults config function would
try to use this.core.globalStorage, which it doesn't have.

TEST=manual
- choose sort option
- check that it is applied to request
- refresh page
- check that is applied to request


src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -14,11 +14,17 @@ import Filter from '../../../core/models/filter';
 export default class SortOptionsComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(assignDefaults(config), systemConfig);
+    this._config.verticalKey = config.verticalKey || this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
```
tmeyer2115, line 18: Let's add a TODO here. I don't think it's great that we keep having to add this logic to different files. We need to figure out a way to centralize this.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -14,11 +14,17 @@ import Filter from '../../../core/models/filter';
...
     this.options = this._config.options;
     this.selectedOptionIndex = parseInt(this.core.globalStorage.getState(this.name)) || 0;
     this.options[this.selectedOptionIndex].isSelected = true;
-    this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex <= this._config.showMoreLimit;
+    this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex < this._config.showMoreLimit;
```
tmeyer2115, line 22: Was having it be <= also a bug?

src/ui/components/filters/sortoptionscomponent.js
oshi97: Sure, I'll make an item!

src/ui/components/filters/sortoptionscomponent.js
oshi97: Yeah, noticed it during testing.

tmeyer2115 (approved):

PR #614, opened 2020-03-30
Filters/Facets: fix filters/facets not applying on page load

oshi97:
This commit changes FilterBox to store the current facet in persistent storage.
Then, on page reload, all filters and facets are automatically placed into
global storage from persistent storage in ANSWERS.init, which calls
globalStorage.setAll(persistentStorage.getAll()).

The two storage systems were already designed to work like this prior to
this commit, however since persistent storage can only store string values
(since it stores them in the url), and because globalStorage.setAll was
just directly copying values over, this was not working properly.
This was fixed by calling JSON.parse for FILTER and FACET_FILTER storage
keys, which should always be JSON strings.

This commit also fixes  SearchBar's search method to send facets to core.verticalSearch,
before it would never send any facets at all.

TEST=manual
- choose facets
- click apply
- check that they are applied in the request
- reload page, check that they are applied in the request and checked off visually


src/core/storage/globalstorage.js
```diff
@@ -32,10 +32,19 @@ export default class GlobalStorage {
...
     for (const [key, val] of Object.entries(data)) {
+      const isFilterKey =
+        key.startsWith(StorageKeys.FACET_FILTER) || key.startsWith(StorageKeys.FILTER);
       if (key === StorageKeys.QUERY) {
         continue;
+      } else if (isFilterKey && typeof val === 'string') {
+        try {
+          this.set(key, JSON.parse(val));
+        } catch (e) {
+          console.error(e);
```
tmeyer2115: AnswersError here

oshi97: updated!

tmeyer2115 (approved):

PR #615, opened 2020-03-31
remove legacyBundleUMD from npm run dev

oshi97:
It should be fine to remove this since
everybody develops off of answer.js as far
as I know (since that was the only option a few week ago)
TEST=manual
- delete answers.js
- see that npm run dev still generates it
- see that my test page still loads


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -140,7 +140,7 @@ function compileCSS () {
 function watchJS (cb) {
   return watch(['./src/**/*.js'], {
     ignored: './dist/'
-  }, parallel(bundle, legacyBundleIIFE, legacyBundleUMD));
+  }, parallel(bundle, legacyBundleIIFE));
```
alexisgrow: can we remove the answer-modern.js `bundle` as well?

alexisgrow (commented):
i like this idea!! it looks like this PR was borked in a rebase maybe?

conf/gulp-tasks/library.gulpfile.js
oshi97: Yes!!! 👌

alexisgrow (approved):
I have this locally lol, it is amazing!

oshi97:
> I have this locally lol, it is amazing!

Awesome!

PR #617, opened 2020-03-31
SearchBar: manually animate icon on search

oshi97:
This commit makes the manually trigger a forward icon animation
after a search. In firefox, triggering a search by clicking on
search icon keeps the explicitOriginalTarget on the icon, which
is not the case for other browsers.

TEST=manual
for firefox, ie11, safari, chrome, ios safari (my phone not browserstack)
1. for both autofocus on and off
- click on search bar
1a. use arrow keys to select autocomplete option
1b. tab to search icon, press enter,
1c. tab to clear button, press enter
1d. click search icon
1e. click autocomplete option


alexisgrow (approved):
nice 👾

PR #619, opened 2020-03-31
Legacy Card: Use default icon height and width on mobile

alexisgrow:
This was forcing the icon wrapper to be smaller than the icon, resulting in alignment issues. Remove styling fallback to the default icon height/width.

TEST=manual

Visually inspect when running locally


oshi97 (approved):
nice one!!

PR #621, opened 2020-03-31
SPR-1931: set timeout and maximum age for geolocation.getCurrentPosition

oshi97:
The geolocation api allows a maximum age option,
which sets the maximum age allowed to use a cached
geolocation. A maximum age of 5 minutes was added
 for all 3 usages of geolocation.getCurrentPosition in the SDK.
Even though the SDK stores geolocation into global storage,
this storage does not persist across sessions, so setting
maximum age should improve performance in a few cases.

The timeout option sets a maximum time allowed for
the browser to compute getCurrentPosition. After this
point, the error callback will be called.
A timeout of 2 seconds was added to the search component,
for searches with "near me" intent. After this point,
the search will run as a normal search.

J=SPR-1931
TEST=manual
Test that setting timeout of 1 results in no location or radius
get parameters in the query request, and a large timeout
does result in location and radius being present in the request.
Test that with timeout of 2000, searches that caused the
"can this browser use your location" popup did not last longer
than 2 seconds.
New incognito tabs were used for each test.


oshi97:
https://yexttest.atlassian.net/browse/SPR-1931

bryanreed (approved):

tmeyer2115 (commented):
If we timeout fetching this info, do we prompt the user again for their location info on the next near me query? If so, we should check with Rose about that behavior. Did we talk with the PMs about setting a maximumAge? That seems fine to me, but we may want to give them the heads up.

oshi97:
> If we timeout fetching this info, do we prompt the user again for their location info on the next near me query? If so, we should check with Rose about that behavior. Did we talk with the PMs about setting a maximumAge? That seems fine to me, but we may want to give them the heads up.

On chrome, the browser remembers the user's last choice for the "use my location" popup. On firefox, you are prompted again. 

Either way, in the situation when the getCurrentPosition times out, what do you think of running another getCurrentPosition in the background, but with no timelimit? While this second getCurrentPosition is running, any additional near me queries would just do a regular search (which is currently what happens if getCurrentPosition fails) until the second getCurrentPosition finishes.

tmeyer2115:
> > If we timeout fetching this info, do we prompt the user again for their location info on the next near me query? If so, we should check with Rose about that behavior. Did we talk with the PMs about setting a maximumAge? That seems fine to me, but we may want to give them the heads up.
> 
> On chrome, the browser remembers the user's last choice for the "use my location" popup. On firefox, you are prompted again.
> 
> Either way, in the situation when the getCurrentPosition times out, what do you think of running another getCurrentPosition in the background, but with no timelimit? While this second getCurrentPosition is running, any additional near me queries would just do a regular search (which is currently what happens if getCurrentPosition fails) until the second getCurrentPosition finishes.

On Chrome, what happens in the following scenario:

- User makes "near me" query. They allow access to their location.
- The getCurrentPosition call times out. A request is issued without their location info.
- The user makes another "near me" query. 

They are not prompted again, but, in the SDK, will another attempt be made to get the current position of the device? Will the timed out response be cached? Bowman mentioned the same idea about making two requests (one in the background with no timeout). That is something we'd need to talk with the PMs about. Previously, there was functionality like that in the SDK. For "near me" queries, two requests could be issued (one attempting to pass geolocation info, the other not). This resulted in a flashing behavior which they found undesirable. They actually had me remove the duplicate call in favor of what's there today.

oshi97:
Yes another attempt is made, no the timed out response is not cached.

oshi97:
To avoid any flashing the duplicate call should only update the StorageKeys.GEOLOCATION value in global storage, instead of updating the geolocation AND issuing a new search

*Edit this wouldn't work on firefox, since it prompts the user a second time after the timeout. We would have to find a way to achieve this with only a single getCurrentPosition call. EVEN then, it may not be the desired experience. At least from my perspective, it would feel very buggy to search for "bobs near me,"  and have the result without any near me context, then search for it again and have the same search return different results.

oshi97:
Closing this for now, until we decide on what is the correct behavior.

PR #625, opened 2020-04-01
Filter: Add getFilterKey helper method

oshi97:
TEST=manual, auto
test that page with facets still can use facets in search


src/core/models/filter.js
```diff
@@ -10,6 +10,17 @@ export default class Filter {
     Object.freeze(this);
   }
 
+  /**
+   * A filter should have exactly ONE key. That key is EITHER the field name to filter by, or
+   * a special string such as $or or $and.
+   * @type {string}
+   */
+  static getFilterKey (filter) {
```
tmeyer2115: You are accessing the data of a particular filter instance, this feels more like an instance method than a static method to me. Another benefit of making this an instance method is you won't need to import the Filter class to use this method.

oshi97: I did it that way originally, however sometimes we get Filters from the backend as json, in which case we have to wrap them in the Filter class. That probably makes sense, though.

tmeyer2115: I think it makes sense to wrap them in the Filter class. We'd have to do that at some point, right?

oshi97: Yeah! It's definitely something that makes certain other things easier as well.

tmeyer2115 (approved):
LGTM! I think the addition of the empty and from methods is a definite improvement. Filters should always be represented in the SDK by an instance of the Filter class.

PR #626, opened 2020-04-01
Cards: No right margin if only one column

alexisgrow:
TEST=manual

Visually inspect running locally with one and more than one column.


oshi97 (approved):
speeeeed

PR #627, opened 2020-04-01
Accordion Result: Add width to prevent overflow in IE11

alexisgrow:
TEST=manual

Test running locally using browserstack.


tmeyer2115 (approved):

alexisgrow:
don't need this actually, I borked a preset locally

PR #629, opened 2020-04-01
CTAs: Update falsy check to apply to functions and objects

alexisgrow:


oshi97 (approved):
oooh true

oshi97:
So I guess before if they had a function that returned null it would break?

alexisgrow (commented):
Yeah, or a list of nulls/undefined, or just didn't return anything 😮🙈

PR #631, opened 2020-04-01
AccordionResult: flex-shrink 0 to toggle icon container

oshi97:
The size of the icon should not change.
All creds to Alexis!

TEST=manual
test that adding to the staging site for bankatfirst fixes strange
accordion toggle icon issues


oshi97:
We should be fixing AccordionCard not AccordionResult

PR #633, opened 2020-04-01
Autocomplete: Use cached live api endpoint

alexisgrow:
TEST=manual

Run locally, test autocomplete triggers in the UI and view correct requests in the network request Chrome inspector.


src/core/utils/urlutils.js
```diff
@@ -1,5 +1,6 @@
 import { PRODUCTION, SANDBOX } from '../constants';
 import SearchParams from '../../ui/dom/searchparams';
+import { AnswersCoreError } from '../errors/errors';
```
tmeyer2115: I think we can remove this import now that we're not throwing if env is Sandbox.

src/core/search/autocompleteapi.js
```diff
@@ -74,6 +76,13 @@ export default class AutoCompleteApi {
      * @private
      */
     this._environment = config.environment;
+
+    /**
+     * The base url for the autocomplete API request
+     * @type {string}
+     * @private
+     */
+    this._baseUrl = getCachedLiveApiUrl(PRODUCTION);
```
tmeyer2115: We should take in this._environment instead of hard-coding to PRODUCTION.

src/core/search/autocompleteapi.js
```diff
@@ -83,6 +92,7 @@ export default class AutoCompleteApi {
       apiKey: this._apiKey,
       version: this._version,
       environment: this._environment,
+      baseUrl: this._baseUrl,
```
tmeyer2115: Talked with Mike Dunn, we should not update the filtersearch stuff to use the cached LiveApi. Only vertical/universal autocomplete.

alexisgrow: Ahh good to know! i'll update!

src/core/search/autocompleteapi.js
alexisgrow: ahahah yes sgtm!

src/core/utils/urlutils.js
alexisgrow: removed!

src/core/search/autocompleteapi.js
alexisgrow: updated!

tmeyer2115 (approved):

PR #634, opened 2020-04-01
AccordionCard: Update spacing

oshi97:
TEST=manual
load page with AccordionCard, make sure it looks alright


oshi97:
I know we said that this would go in 13.2, but I already wrote the code and think it looks alright! If this can save some heartache in the future I would rather get it out right now.

oshi97:
For later

PR #639, opened 2020-04-01
Component: consolidate vertical search logic into component.js

oshi97:
…earch

Components outside of the parent Component class should interact with
core as little as possible. For running vertical searches, there
was a lot of repeated logic for pulling data necessary for a search,
that could be consolidated into a single method.

In the future, I think it would be good to not have core.verticalSearch
interact with globalStorage, and instead have it be purely functional
based on the parameters it's given. Instead we can move any logic around
globalStorage into component.verticalSearch. I made a todo and item in
the backlog for this.

TEST=manual
Created test page with Facets, SortOptions, LocationBias, GeoLocationFilter,
and FilterSearch.
For Facets, FilterSearch, SortOptions, GeoLocationFilter, tested that
  changing their values would affect the search, and refreshing the page
  kept these values and also applied them to the initial search
For LocationBias, checked that clicking "Update my location" would
  refresh the search, with location data added to the query


src/ui/components/component.js
```diff
@@ -395,6 +397,36 @@ export default class Component {
     return el.innerHTML;
   }
 
+  /**
+   * Common interface between components to trigger a vertical search.
+   * TODO(oshi) SPR-1937: move all usages of globalStorage out of core.verticalSearch
+   * into Component.verticalSearch
+   */
+  verticalSearch (options = {}) {
```
tmeyer2115: I don't think we should make a top-level method called verticalSearch part of Component.js. This method is not relevant to all components.  I don't have much of an issue with Components having a reference to core. Components interacting with global and persistent store directly is another matter. Giving any component a way to initiate a verticalSearch seems like something that could be abused to me.

oshi97: I mean, aren't we giving them a way to initiate a search through core anyways?

oshi97: I was thinking it would be nice to have some kind of layer in between a component making a search and core.verticalSearch, so we can separate the gathering of search settings from global storage and the actual search, but I guess that's already done between core.js and searchapi.js, so with regards to that anyway I think you're right

src/core/core.js
```diff
@@ -114,27 +114,39 @@ export default class Core {
...
+
     return this._searcher
       .verticalSearch(verticalKey, {
         limit: this.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit,
         geolocation: this.globalStorage.getState(StorageKeys.GEOLOCATION),
+        input: this.globalStorage.getState(StorageKeys.QUERY) || '',
         ...query,
+        filter: JSON.stringify(totalFilter),
+        facetFilter: JSON.stringify(facet),
+        offset: this.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0,
```
tmeyer2115, line 149: Is this going to cause an issue? Before, an offset would be present only if someone included it with their query. Now, though, we will always include an offset, provided there's one in globalStorage.

src/core/core.js
```diff
@@ -114,27 +114,39 @@ export default class Core {
...
+    const totalFilter = allFilters.length > 1
+      ? Filter.and(...allFilters)
+      : allFilters[0];
+    const facet = useFacets ? this.globalStorage.getAll(StorageKeys.FACET_FILTER)[0] : {};
+
     return this._searcher
       .verticalSearch(verticalKey, {
         limit: this.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit,
         geolocation: this.globalStorage.getState(StorageKeys.GEOLOCATION),
+        input: this.globalStorage.getState(StorageKeys.QUERY) || '',
```
tmeyer2115, line 145: Is there any scenario in which a caller of this method was trying to issue a query with something different than what was in globalStorage?

src/core/core.js
tmeyer2115: I guess, to verify this, we need to be sure that any call of this method that was passing query.offset was using the current value in global store. And, those callers of this method that did not provide an offset deleted the key from storage to begin with. Assuming there's no difference between an offset of 0 and not providing one

src/core/core.js
oshi97: There is no scenario where `input` should be different than the query stored in global storage. All calls of core.verticalSearch are refactored in this PR, and the only instance where `input` is not directly taken from globalStorage is in searchcomponent.js. Looking at the code, it looks like the passed in query should also always be equal to what is stored in globalStorage as well.
To be on the safe side I did not change its behavior, and instead still pass `input` into core.verticalSearch's `query` param, where it should override the value in globalStorage.

src/core/core.js
oshi97: While testing this PR, I made sure that passing an offset of 0 had the same results as passing no offset param. I also made sure that all searches that were deleting the offset are now given the `resetPagination` flag.

oshi97: Any calls of verticalSearch that did not touch the search offset are not given the resetPagination flag, and don't touch it.

tmeyer2115: We need to know that any callers of this method that did not include an offset did so because they deleted it beforehand. If they did not include an offset and left it totally unchanged in global storage, we could run into a problem here. Now, they would get the unchanged offset in their verticalSearch request, whereas the expected behavior would be a request with no offset.

oshi97: Yes that is the case. All callers of core.verticalSearch (every filter component) that did not include an offset did so because they deleted it before hand. All callers that did include an offset either took that value from globalStorage (locationbias, searchcomponent) or set it immediately beforehand (pagination).

tmeyer2115 (approved):

PR #640, opened 2020-04-02
Geolocation timeout for near me queries

oshi97:
This commit adds a timeout for navigator.geolocation.getCurrentPosition calls after receiving a "near me" query.

Using the builtin timeout functionality of the geolocation api is too naive for our use case. After the timeout is reached, the geolocation api will completely give up on the geolocation call instead of doing some timeout handling while continuing the call in the background. Any subsequent "near me" searches will also kick off new geolocation calls, which will likely timeout (since the first one timed out) and cause a perpetually timed out search experience.

What we really want is to calculate the geolocation in the background, as well as a window.setTimeout search, simultaneously, and have them race each other. The first one to win gets to issue the search, and the loser should not do any searching. And no matter what, the geolocation call needs to update the geolocation stored in global storage. Any subsequent "near me" queries will use this stored value instead.

Even while using Promise.race, there are 3 cases to handle if the user initiates another search while these two are running.
1. both geolocation and timeout are still running -> don't do anything
2. timeout has finished, geolocation is still running -> issue another timeout search
3. geolocation has finished, timeout is still running -> this new search should use the stored geolocation, and should not go into this code pathway to begin with.
These situations were handled using condition variables, this.geolocationTimeoutInProgress and  this.geolocationInProgress.

J=SPR-1931
TEST=manual
Wrapped geolocation in a very long window.setTimeout, so that it would never complete, and added console logs to ensure that clicking search while the timeout was still running would not result in another search
Wrapped geolocation in a window.setTimeout of 3 seconds, so that the timeout would always finish before the geolocation. Checked that no extra search was ran after the geolocation finished.


oshi97:
closing this for now since 1. no longer high priority 2. has not gotten signed off by PMs and it looks like they have a different idea of what they want

PR #641, opened 2020-04-02
Revert parsing facets/filters from persistent storage (#607, #608, #612, #614)

oshi97:


oshi97:
This could potentially be "fixed" by doing JSON.parsing persistentStorage, and then calling core.setFacetFilter inside the Facets component. But I there still might be similar issues, if one page has a facet with the default name, and another page also has a facet with the same name but different back-end configuration.

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #644, opened 2020-04-02
Analytics: Allow analytics raw object

alexisgrow:
Previously, implementers could not send analytics events using JS,
the only supported way to do that would be to use the data attributes 
`eventType` and `eventOptions`. This PR updates the JS class to
accept a raw analytics object, not just instances of the built-in
class.

TEST=manual

Test using a local test site, confirm analytics are firing.


src/core/analytics/analyticsreporter.js
```diff
@@ -73,15 +72,10 @@ export default class AnalyticsReporter {
       throw new AnswersAnalyticsError('Tried to enable conversion tracking without including ytag');
     }
 
-    if (!(event instanceof AnalyticsEvent)) {
-      throw new AnswersAnalyticsError('Tried to send invalid analytics event', event);
-    }
-
-    event.addOptions(this._globalOptions);
-
+    const apiEvent = Object.assign(event, this._globalOptions);
```
oshi97, line 75: Could we do Object.assign({}, event, this._globalOptions)? Otherwise, this will also overwrite the original event (all the way back to the component!)

oshi97 (commented):
mostly lgtm! It doesn't look like analytics event does anything special except hold data. In the future if we wanted any behavior per analytics event, I think we could just do that inside analyticsreporter.js

src/core/analytics/analyticsreporter.js
alexisgrow: True true! So the analytics JS actually modifies the original object, which is why I did it like this. 

When I was writing this, I originally did `Object.assign({}, event, this._globalOptions)` and the test cases failed so I went back to this (I don't remember if my manual tests worked). I'll test with it locally again to see if I can get it to work, bc you're right it is better to make a new object

oshi97: oooh okay I get it now. This is probably fine then!

oshi97 (approved):
lgtm!

oshi97:
Hi! I rebased v0.14.0 onto v0.13.2 but noticed it messed up your PR so rebased this for you as well

PR #645, opened 2020-04-03
Update SDK to use the @yext/rtf-converter package for formatRichText …

tmeyer2115:
…method.

This PR updates the existing Answers.formatRichText method to use a new npm
package, @yext/rtf-converter. This replaces the existing richtext.js util
class.

J=SPR-1920
TEST=manual

Tested the following scenarios in IE11 and Chrome:

- For the details attribute of a card, I used formatRichText on an RTF field.
  Saw the correct HTML displayed.
- For the details attribute of a card, passed in 'undefined' to formatRichText,
  saw the expected error thrown in the console.

Made sure the HTML displayed in the first scenario matched the HTML rendered by
the previous formatRichText implementation.


oshi97 (approved):
beautiful

PR #646, opened 2020-04-03
SearchBar: Fix clear button not showing up w/o typing into searchbar

oshi97:
The clear button was not showing up, if the user clicked an
autocomplete option, and on page load there was not already
a query string.

In the past, this was being done by calling an empty setState,
which would replace the enter component with a freshly mounted
version. This was removed when implementing the
default animated icon.

Also remove a showClearButton variable that was being passed
into the handlebars template, but not used at all in the .hbs

T=https://yextops.zendesk.com/agent/tickets/317215
J=SPR-1942
TEST=manual
1. clearButton config option off, checked that using an autocomplete
option as well as typing and searching did not render a clear button
2. clearButton config option on
- on page load with no query, clicking the searchbar,
then choosing an autocomplete option still had a clear button that
clears the text when clicked
- on page load, with no query, there no clear button until a search is made
- on page load with a query, searchbar renders with a clear button that works
- searching with blank query hides the clear button


src/ui/components/search/searchcomponent.js
```diff
@@ -559,7 +563,6 @@ export default class SearchComponent extends Component {
       submitIcon: this.submitIcon,
       submitText: this.submitText,
       clearText: this.clearText,
-      showClearButton: this._showClearButton,
```
tmeyer2115: How are we currently disabling the clear button? We are supposed to provide a config option that lets someone opt out from this

src/ui/components/search/searchcomponent.js
```diff
@@ -358,6 +358,10 @@ export default class SearchComponent extends Component {
 
       inputEl.blur();
       DOM.query(this._container, '.js-yext-submit').blur();
+      if (this.clearButton && this.query) {
```
tmeyer2115: Why is this logic under DOM.on(form, 'submit', ...)? I think it should be outside this, in the top-level of initSearch

src/ui/components/search/searchcomponent.js
oshi97: If config.clearButton is off, we set this.clearButton to false, and never initialize the clearButton DOM listeners. The clearButton has display: none by default

oshi97: I think a better approach would be to not render the clearButton at all, updated the PR and re ran previous tests

src/ui/components/search/searchcomponent.js
oshi97: That won't work because we need this to display the clear button when a search is run for a non-blank query.

src/ui/templates/search/search.hbs
```diff
@@ -13,18 +13,20 @@
...
-              data-opts='{ "iconName": "close" }'
-              data-prop="icon"
-      >
-        <span class="yxt-SearchBar-clearButtonText sr-only">
-          {{clearText}}
-        </span>
-      </button>
+      {{#if showClearButton}}
+        <button type="button"
+                class="js-yxt-SearchBar-clear yxt-SearchBar-clear"
```
oshi97: Change here is to only render the clearButton is config.clearButton is true, and removed display: none (yxt-SearchBar-hidden) by default.

src/ui/components/search/searchcomponent.js
tmeyer2115: I'm not sure I follow. Is it here because we need to display the clear button when we refresh the page with a non-empty query param?

src/ui/components/search/searchcomponent.js
```diff
@@ -358,6 +358,14 @@ export default class SearchComponent extends Component {
...
+        const button = DOM.query(this._container, '.js-yxt-SearchBar-clear');
+        if (this.query) {
```
tmeyer2115, line 364: Is this nested inside the if (this.clearButton && this.query)? If it is, this.query will always be truthy and we will never hit the else case below.

src/ui/components/search/searchcomponent.js
tmeyer2115: Please undo this change. I don't want to add more changes. If it was working before and the logic for this was unchanged in 13.1, we can just leave it as is.

src/ui/components/search/searchcomponent.js
oshi97: If the user clicks an autocomplete option without changing any input, inputEl on change handler will not run. Prior to this commit, on input change and on mount were the only places the clear button was being switched from display: none to display: whatever.

tmeyer2115 (approved):

PR #648, opened 2020-04-03
Spellcheck, Autocomplete: Correctly pass in inverted param to handlebars helper

alexisgrow:
TEST=manual

Visually inspect when running locally.


oshi97 (approved):
Could we also update the highlightvalue helper in autocomplete.hbs?

alexisgrow (commented):
Yes! I will update in this PR. That one is truthy so it doesn't cause a bug but is definitely not "correct" per se

oshi97 (approved):
nice one!

PR #654, opened 2020-04-06
Check in package-lock.json

oshi97:
Checked-in package-lock.json after npm install


tmeyer2115 (commented):
I'm not sure why this is included in the .gitignore. We want to be including the package-lock.json (if modified) in commits. Can you remove it from .gitignore? Was the package-lock supposed to be checked in with an earlier commit?

oshi97:
> I'm not sure why this is included in the .gitignore. We want to be including the package-lock.json (if modified) in commits. Can you remove it from .gitignore? Was the package-lock supposed to be checked in with an earlier commit?

O wow I just assumed it wasn't in the gitignore - I very specifically remember asking Christian (during v0.13.0 QA) what is this and whether it should be checked in. I ran a new npm install after pulling this morning and found this in my changed files.

tmeyer2115 (approved):
I think you saw these changes because you have a different version of npm than the person who checked this in last.

PR #655, opened 2020-04-06
Accordion 13.2 qa

oshi97:
Accordion Card 13.2 QA:
- improve animation to match AccordionResult, by adding a container around contents
- add padding to button title to match AccordionResult
- bold title
- move margin-bottom from yxt-AccordionCard--expanded to padding-bottom of yxt-AccordionCard-content

TEST=manual
load page with AccordionCard, test animation, inspect DOM for padding and bold title


src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -122,17 +122,17 @@ export default class AccordionCardComponent extends Component {
    * it doesn't rip the whole component off of the page and remount it.
    * Also reports an analytics event.
    * @param {HTMLElement} toggleEl the toggle element
-   * @param {HTMLElement} contentEl the content element
+   * @param {HTMLElement} bodyEl the body element
```
alexisgrow: nit: "the body element" makes me think the HTML < body > element, what do you think?

alexisgrow (approved):

src/ui/components/cards/accordioncardcomponent.js
oshi97: yeah that's a good point I can rename it

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -88,14 +88,21 @@
   &-icon svg
   {
     transform: rotate(90deg);
-    transition: transform 400ms ease-in;
+    transition: all 300ms ease-in;
```
tmeyer2115: 400ms is the duration used in AccordionResults. Do we need to alter the duration here and below, or just change ease to ease-in below?

tmeyer2115 (changes requested):
What is the motivation for adding a new wrapper div around container?

oshi97:
> What is the motivation for adding a new wrapper div around container?

Before, yxt-AccordionCard-content was the one with the height being animated. But the margin/padding had to be set to 0, otherwise it would stick out of the accordion, which made the animation have a weird "clicking" feeling the old animation didn't have. AccordionResult avoided this by animating a container around contents.

src/ui/sass/modules/_AccordionCard.scss
oshi97: 400ms is fine

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -20,7 +20,7 @@
 
   $accordion-card-details-size: $font-size-md !default;
   $accordion-card-details-weight: $font-weight-normal !default;
-  $accordion-card-details-line-height: $line-height-md !default;
+  $accordion-card-details-line-height: 20px !default;
```
tmeyer2115: We should try to express this as a proportion of an existing line-height variable. We want to avoid hard-coding whenever we can. That way, if someone updates the corresponding variable, this is automatically updated as well, preserving the proportion between $accordion-card-details-line-height and whatever the $line-height-var is.

tmeyer2115:
Are we still bolding the title?

oshi97:
> Are we still bolding the title?

nope, it's one of the few things the mocks are pretty clear about

src/ui/sass/modules/_AccordionCard.scss
oshi97: updated!

src/ui/sass/_fonts.scss
```diff
@@ -15,6 +15,7 @@ $font-size-xxlg: 40px !default;
 $line-height-xs: 1 !default;
 $line-height-sm: 1.2 !default;
 $line-height-md-sm: 4/3 !default;
+$line-height-md-md: 1.43 !default;
```
tmeyer2115: I think someone reading the code would be confused about the difference between line-height-md and line-height-md-md.

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -20,7 +20,7 @@
 
   $accordion-card-details-size: $font-size-md !default;
   $accordion-card-details-weight: $font-weight-normal !default;
-  $accordion-card-details-line-height: $line-height-md !default;
+  $accordion-card-details-line-height: $line-height-md-md !default;
```
tmeyer2115: I didn't necessarily mean making a new variable in line heights. You can use an existing variable and prefix it with a numerical proportion, i.e. 5/4 * $line-height-xs

tmeyer2115 (approved):

PR #656, opened 2020-04-06
Models: Create FilterView, FacetView, and FilterMetadata models

oshi97:
In order to facilitate storage and processing of FilterMetadata,
instead of directly storing backend filters in global storage,
we now want to store some kind of filter/facet view object, which
can contain extra metadata.

Filters themselves can be classified into two main types,
filters that are a composed of a single filter, and filters
that are composed of multiple filters. Trying to store both
kinds of filters in the same FilterView class is difficult
and clumsy, so BasicFilterView and CombinedFilterView were
created as implementations of a FilterView interface,
which has two methods, getFilter and getBasicFilterViews.
Implementing filter views in a manner that keeps the information
of which FilterMetadata corresponds to which portion of which
filter makes it much easier and more intuitive to implement
removable filters in the future.

TEST=NONE
This commit was a portion of a larger PR, in the past,
for implementing applied filters in the results header.
In this commit, these files are currently not used anywhere.


src/core/models/filterview.js
```diff
@@ -0,0 +1,18 @@
...
+ * FilterView is an interface for storing and handling filter data and metadata.
+ *
+ * @interface
+ */
+export default class FilterView {
+  /**
+   * Returns the filter data, formatted to send in a query.
+   * @returns {Filter} Filter to send in query.
+   */
+  getFilter () {}
```
tmeyer2115, line 11: Do we need a getMetadata() in this interface as well? When we talked (albeit like 3 months ago), I think we'd discussed the FilterView consisting of metadata and a Filter.

src/core/models/basicfilterview.js
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Metadata for the filter, and is an object of
+     * field display name to array of field values.
+     * @type {FilterMetadata}
+     */
+    this._metadata = new FilterMetadata(metadata);
+    Object.freeze(this);
+  }
+
+  getBasicFilterViews () {
```
tmeyer2115: I think it's a little weird that BasicFilterView has a method called getBasicFilterViews. It seems like this is here because the composite class needs it. What is the use case for the composite filer view class needing this array? Will the composite class still need this array if, instead, it has the correctly combined FilterMetadata?

src/core/models/facetview.js
```diff
@@ -0,0 +1,48 @@
+/** @module FacetView */
+
+import Facet from './facet';
```
tmeyer2115, line 3: At its core, is a Facet any different than a combination of filters? I'm having trouble understanding the distinction. Is this it:
- A combination filter is built from multiple filters, each corresponding to a potentially different field.
- A facet corresponds to a collection of filters, all on the same field.

src/core/models/facetview.js
```diff
@@ -0,0 +1,48 @@
...
+     * A list of all the filter views in this facet.
+     * @type {Array<FilterView>}
+     */
+    this._filterViews = filterViews;
+
+    /**
+     * List of available field ids used in this facet.
+     * @type {FilterMetadata}
+     */
+    this._availableFieldIds = availableFieldIds;
```
tmeyer2115, line 20: Isn't there one fieldId associated with a Facet?

src/core/models/basicfilterview.js
oshi97: The idea is that you can call getBasicFilterViews on any FilterView (so, combined or basic) and it will return you all the constituent BasicFilterViews and you don't have to care about what type you're calling

src/core/models/facetview.js
oshi97: The only thing different about it (right now at least) is that we store it under it's own storage key, and when we send the information to the backend it is expected in a different format

src/core/models/facetview.js
oshi97: No, facets are composed of multiple filters, which each have at least one field id

src/core/models/filterview.js
oshi97: I didn't end up needing it when I was implementing removable filters, since it's so easy to get them from getBasicFilterViews(). If it ends up being needed I don't think anybody should hesitate in adding it though!

src/core/models/basicfilterview.js
oshi97: The idea of combining FilterMetadata won't work for the direction we want to go, since we need to maintain information about which filter metadata corresponds to which filter. Or at the very least keep them separate so we can assign callbacks or something.

src/core/models/facetview.js
oshi97: After the discussion we had today maybe it would be better to call this model "Facets" instead of just "Facet", since a Facet can have multiple fieldIds.

src/core/models/filterview.js
oshi97: updated filterview to have getMetadata instead of getBasicFilterViews, plan is to use filterIdentifier in the future for this kind of behavior

oshi97:
Closing this in favor of FilterRegistry architecture discussed over zoom zoom

PR #666, opened 2020-04-08
Facets/FilterOptions: Show selected facets first by default

oshi97:
This commit floats selected checkboxes/radio buttons for FilterOptions within a Facets component. Facets rebirth a new FilterBox (which births new FilterOptions) every time a new search is returned, so applying this all that is needed. Adding it as a listener to clicking the apply button actually creates a weird shuffling effect which is undesired, since what would happen is 1. user clicks apply button 2. selected options float to the top 3. new search is returned with new options

J=SPR-1970
TEST=manual
Tested that Facets are still applied on searches,
and after clicking the apply button (but not the search button or
changing any of the options) the filter options will be sorted.


README.md
```diff
@@ -1019,6 +1019,8 @@ ANSWERS.addComponent('FilterSearch', {
   autoFocus: true,
   // Optional, redirect search query to url
   redirectUrl: 'path/to/url',
+  // Optional, whether to float selected options to the top, default true.
+  floatSelected: true,
```
alexisgrow: In the Searchable Facets spec, I think it says that selected facets should always appear first and it should not be configurable. My understanding is this should not be a config option, is that right?

oshi97: Yes you're right! updated stuff

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -208,6 +208,9 @@ export default class FilterBoxComponent extends Component {
           }
         }));
       component.mount();
+      if (this.config.isDynamic && typeof component.floatSelected === 'function') {
```
tmeyer2115, line 210: Whenever a user submits a query, there can be newly selected facet options. Either because a user manually checked them or they were inferred from the back-end. These newly selected facet options need to bubble to the top of their respective FilterOption component. Is it enough, then, to just have this logic in the Filterbox onMount?

oshi97: Yup it's enough. Every time a query is submitted, the Facets component recreates its FilterBox component from scratch, which will call onMount again. 

I initially also had a few lines inside of the Apply button's handler to bubble up the selected options as well, however this resulted in the FilterOptions reordering themselves for a second, then the new facets being loaded in and replacing everything. So click apply -> selected options float to the top -> facet recreates filter box and all of the options change, which isn't what we want

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -208,6 +208,9 @@ export default class FilterBoxComponent extends Component {
           }
         }));
       component.mount();
```
alexisgrow: Is this `component.mount()` then the new `floatSelected()` (on line 212) causing this to render twice? Is the floatSelected/option-ordering-logic something that we should consider putting before the `component.mount()`?

oshi97:
This probably makes more sense to be part of the searchable-facets branch

src/ui/components/filters/filterboxcomponent.js
oshi97: Yes it is! You're right, if we put this before component.mount() we don't need the setState call

alexisgrow (approved):
LGTM!

PR #667, opened 2020-04-08
README: update navigation param to verticalPages

nbramblett:
The navigation param in the base config is called verticalPages now, and
is an array rather than an object.

TEST=none


oshi97:
I think we normally try to add commits onto a version branch instead of directly onto master. 
Right now we're doing what I like to call semantic versioning + 1, where going from v0.13 to v0.14 would be considered a breaking upgrade. Once we go public and have major version 1, we will switch to doing true semantic versioning, so only going from v1 to v2 would be breaking.

This would probably go into either v0.14.0 or v0.13.3 (which doesn't exist yet, but you could make it!)

oshi97 (approved):
nice!

PR #668, opened 2020-04-09
Always use CustomEvent polyfill

oshi97:
This commit changes DOM.trigger to always use our custom event
polyfill, instead of looking in the window first to see if
CustomEvent exists as a function.

Consulting was running into issues with ie11 saying that
it doesn't know how to use the Event constructor. This likely
could have also been fixed by changing `return new Event` to
`return new CustomEvent`, however it feels safer (to me)
to always rely on our own polyfill, instead of potentially
using somebody else's polyfill.

J=SPR-1979
TEST=manual
Went on ie11, checked that adding a CustomEvent function onto the
window manually through the dev tools causes the error consulting
was seeing. This was checked by typing into the searchbar,
since the searchcomponent uses DOM.trigger.
Then, removed code and checked that there were no errors,
and autocomplete was displaying values that I could click on
to run a search. Also checked on chrome.


alexisgrow (approved):
LGTM!

PR #669, opened 2020-04-09
Fix FilterSearch Readme

Mxs2019:
`sectioned` should not be part of the array. It should be part of the `searchParameters` object. I tested and confirmed this is the way it works.


MattCBowman (dismissed):
If this is correct, LGTM

tmeyer2115 (changes requested):
We can't merge this directly into master. It should be merged into one of the release branches.

oshi97:
cherry-picking into v1.4.0

PR #670, opened 2020-04-09
Add first TestCafe acceptance test for the SDK.

tmeyer2115:
This PR adds the infrastructure needed to write TestCafe acceptance tests for
the SDK. It also creates the first such test. This acceptance test is
relatively simple and verifies a basic flow on a universal search page.
Essentially, it types a manual query, clears it, picks an autocomplete option
and verifies that the expected VerticalResults sections are there. This test
is a jumping off point and is by no means exhaustive. We should look to add
more to this test, test other flows, and make sure we have Block models for
every component. Right now, tests are only run in Safari and Chrome. Firefox
support and IE11 support we come in later PRs.

J=SPR-1963
TEST=auto

Made sure the new acceptance test could pass. Watched as the test ran in both
Safari and Chrome to make sure the test matched what I thought it should be
doing.


MattCBowman (approved):
☕ lgtm

PR #671, opened 2020-04-09
Analytics: Track middle-clicks on link and CTA elements

nbramblett:
The middle-mouse button opens links in a new tab, and should trigger
an analytics event. For other clickable elements, like buttons, the
middle-mouse button is not a valid click option, and so we don't care
to track it. To simplify things, I've added a middleclick element in the
dataset for elements that should track the middle click. It also simply
assumes that CTA components should support middle-click tracking, as
their actual content/behavior is mostly up to the user.

J=SPR-1657
TEST=manual, auto

Manually built and verified that the service was firing analytics events
on a middle-click for links. Ran enzyme tests.


src/ui/components/component.js
```diff
@@ -433,13 +433,16 @@ export default class Component {
     const dataset = domComponent.dataset;
     const type = dataset.eventtype;
     const label = dataset.eventlabel;
+    const middleclick = dataset.middleclick;
```
oshi97, line 436: Do we need this middleclick? It looks like it's not really doing too much.

README.md
```diff
@@ -1381,6 +1381,7 @@ ANSWERS.addComponent('SortOptions', {
 # Analytics
 
 Answers will track some basic interaction analytics automatically, such as search bar impressions and Call-To-Action clicks. You may add additional, custom analytic events to templates using certain data attributes, explained below. You may also send analytics from external code with the below interface.
+If you do not include a business ID in your base config, no analytics will be tracked.
```
alexisgrow, line 1384: 😍YAY thank you for adding!

tests/ui/components/component.js
```diff
@@ -77,7 +79,7 @@ describe('attaching analytics events', () => {
     const wrapper = mount(component);
     expect(domOn).toHaveBeenCalledTimes(1);
 
-    wrapper.find('#test').simulate('click');
+    wrapper.find('#test').getDOMNode().dispatchEvent(new MouseEvent('mousedown', { button: 0 }));
```
oshi97, line 82: Nice! Do we still need tests for regular click or does this also work for that.

oshi97 (changes requested):

alexisgrow (approved):
LGTM!

src/ui/components/component.js
nbramblett: Not every component needs or wants the middleclick to be tracked. Adding this in lets us flag the elements that should track middleclicks.

tests/ui/components/component.js
nbramblett: All component.js objects use mousedown for analytics tracking now, since it's a superset of click.

oshi97 (approved):
lgtm!

PR #672, opened 2020-04-09
Add FilterRegistry, FilterNode, FilterView and supporting classes

oshi97:
This commit adds FilterRegistry, FilterNode, and FilterView, which
will be a new way to store filters/facets in the SDK, and replace our
current usage of global storage (only for filters/facets).

This should allow us to store additional metadata and callbacks
next to a filter, without sacrificing information on how the
filter was originally constructed.

TEST=auto
wrote unit tests for FilterRegistry, FilterView, and FilterNode
testing that basic functionality works, and that we don't have
issues with more nested structures


src/core/filters/filtermetadata.js
```diff
@@ -0,0 +1,23 @@
+/** @module FilterView */
+
+export default class FilterMetadata {
+  constructor (metadata = {}) {
+    const { fieldId, fieldName, displayValue } = metadata;
+
+    /**
+     * @type {string}
```
tmeyer2115, line 12: Just add a little JSDoc blurb for these attributes.

src/core/filters/filtermetadata.js
```diff
@@ -0,0 +1,23 @@
+/** @module FilterView */
+
+export default class FilterMetadata {
```
tmeyer2115, line 6: JSDoc describing this class. Also, update the module name

src/core/filters/filternode.js
```diff
@@ -0,0 +1,64 @@
+/** @module FilterNode */
+
+import Filter from '../models/filter';
+import FilterView from './filterview';
+import FilterCombinators from './filtercombinators';
+
+export default class FilterNode {
```
tmeyer2115, line 11: JSDoc

src/core/filters/filternode.js
```diff
@@ -0,0 +1,64 @@
...
+  constructor (filterNode = {}) {
+    const { filterView, combinator, children } = filterNode;
+
+    /**
+     * @type {FilterView}
+     */
+    this.filterView = filterView ? new FilterView(filterView) : null;
```
tmeyer2115: Can't we assume the filterView we get passed in is a FilterView?

src/core/filters/filternode.js
```diff
@@ -0,0 +1,64 @@
...
+    this.combinator = combinator;
+
+    /**
+     * @type {Array<FilterNode>}
+     */
+    this.children = children || [];
+  }
+
+  getFilter () {
+    if (this.filterView) {
```
tmeyer2115: Is there a filterView only of this is not a combination filter? Combination filters only contain a combinator and children? If that's the case, we need to make that clearer via documentation. Potentially we should also enforce that in the FilterNode's constructor.

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,74 @@
+/** @module FilterRegistry */
+
+import Filter from '../models/filter';
+import Facet from '../models/facet';
+import FilterView from './filterview';
+
+export default class FilterRegistry {
```
tmeyer2115, line 12: JSDoc

nbramblett:
Looks like your unit tests are failing but a +1 from me otherwise.

oshi97:
> Looks like your unit tests are failing but a +1 from me otherwise.

they're working locally, I mean they're like 50% of the PR :'(

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ jobs:
   # generate bundles
   build:
     docker:
-      - image: circleci/node:7.10
```
oshi97: Updated the node image version because flatMap() and Object.values() do not work on the node 7.10 . Even though we were already using flatMap in facet.js, this did not come up in the past because it was not in our test coverage.

tmeyer2115: What is the current version of node? I'm concerned that if we change the version of node we build the SDK CDN bundles with, there could be unintended, negative consequences. Especially if there are breaking changes between node:current and node:7.10 (not sure if they follow semantic versioning). If we make this change, we will need to thoroughly QA 13.3 to make sure we exercise all the functionality of the bundles.

src/core/filters/filternode.js
```diff
@@ -0,0 +1,64 @@
...
+  constructor (filterNode = {}) {
+    const { filterView, combinator, children } = filterNode;
+
+    /**
+     * @type {FilterView}
+     */
+    this.filterView = filterView ? new FilterView(filterView) : null;
+
+    /**
+     * @type {string}
```
tmeyer2115: Please add JSDocs for these fields.

src/core/filters/filternode.js
```diff
@@ -0,0 +1,64 @@
...
+
+  static and (...filterNodes) {
+    return FilterNode._combine(FilterCombinators.AND, filterNodes);
+  }
+
+  static or (...filterNodes) {
+    return FilterNode._combine(FilterCombinators.OR, filterNodes);
+  }
+
+  static _combine (combinator, filterNodes) {
```
tmeyer2115: Please add JSDocs to these methods

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,74 @@
...
+  constructor (filterNodeMap = {}, facetFilterViews = {}, availableFieldIds = []) {
+    /**
+     * Map of filter namespace to FilterNode.
```
tmeyer2115: What is the filter namespace?

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,74 @@
...
+     * to have to worry about one set.
+     * @type {Array<FilterView>}
+     */
+    this.facetFilterViews = facetFilterViews;
+
+    /**
+     * Used when building the request facet.
+     * @type {Array<string>}
+     */
+    this.availableFieldIds = availableFieldIds;
```
tmeyer2115, line 25: Does this represent the list of fieldIds corresponding to the facets in facetFilterViews? If so, could we consolidate the two somehow? It feels weird to separate them out like this.

tmeyer2115 (commented):
I want to make sure I understand the mechanics. The FilterRegistry will maintain a map of identifier to FilterNode. What would the map look like if I have the following filters:

{ $and: [{$or: [{foo}, {bar}]}, {$or: [{blah}, {baz}]}]},

where foo, bar, blah, baz are some kind of $eq condition. Will the filterIdentifier be part of the FilterMetdata? Are combined filters not supposed to have a FilterView? I think it would be nice, conceptually, if they could. If we're saying a FilterView is a visual representation of a filter, combined filters do have a visual representation in our UI (i.e. Specialties: Korean, Japanese). When combining filters, will we need to update the identifiers of the filters involved to indicate the additional nesting?

src/core/filters/filterregistry.js
oshi97: Prior to this commit, when creating a facet we include all availableFieldIds from the FilterBox, not just the ones that have filters. I'm not sure if this is necessary (some light testing makes it seem like it isn't) but I would like to preserve this behavior for now.

.circleci/config.yml
oshi97: Updated to use a polyfill for now

src/core/filters/filternode.js
oshi97: Yeah, filterViews should only be given to the leaves of the tree.

src/core/filters/filterregistry.js
oshi97: updated!

src/core/filters/filternode.js
oshi97: I tried to update the code to use a pseudo-interface, instead of throwing all of the logic into one class

src/core/filters/filternode.js
oshi97: updated!

oshi97:
> I want to make sure I understand the mechanics. The FilterRegistry will maintain a map of identifier to FilterNode. What would the map look like if I have the following filters:
> 
> { $and: [{$or: [{foo}, {bar}]}, {$or: [{blah}, {baz}]}]},
> 
> where foo, bar, blah, baz are some kind of $eq condition. Will the filterIdentifier be part of the FilterMetdata? Are combined filters not supposed to have a FilterView? I think it would be nice, conceptually, if they could. If we're saying a FilterView is a visual representation of a filter, combined filters do have a visual representation in our UI (i.e. Specialties: Korean, Japanese). When combining filters, will we need to update the identifiers of the filters involved to indicate the additional nesting?

That kind of filter would come from some filter component (in this case a FilterOptions or FilterBox). That filter component would set one key in the FilterRegistry, probably the same key it sets right now in global storage (minus the StorageKeys.FILTER prefix). The FilterRegistry would only worry about the top level $and node.

Right now my mental model is that a filter node's visual representation comes from its getFilterViews(). In order to get the visual representation of ALL filters in the FilterRegistry, FilterRegistry would just call each of its top-level nodes' getFilterViews().

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Sets the specified {@link FilterNode} under the given key.
+   * Will replace a preexisting node if there is one.
+   * @param {string} key
+   * @param {Object|FilterNode} filterNode
+   */
+  setFilterNode (key, filterNode) {
+    this.filterNodeMap[key] = FilterNode.from(filterNode);
```
tmeyer2115: Will the interactions with global and persistent store come in a later PR?

src/core/filters/filternode.js
```diff
@@ -0,0 +1,163 @@
...
+   */
+  getFilterViews () {
+    return this.children.flatMap(node => node.getFilterViews());
+  }
+}
+
+/**
+ * LeafNodes represent a single, atomic filter.
+ */
+class _LeafNode extends FilterNode {
```
tmeyer2115: Let's make separate files for LeafNode and BranchNode. I think it's okay for those to be public. Maybe call the public classes SimpleFilterNode and CombinedFilterNode (or something like that). I think this file has too much going on and it would help to break it up. The static constructors that create a LeafNode instance or BranchNode instance should live in the corresponding, new files.

oshi97: Updated!

src/core/filters/filterregistry.js
oshi97: I was originally thinking we would stop using global storage for filters, but it makes more sense to still use global storage. For persistent store I would rather not change any behavior yet, and keep any logic currently in the components, since we haven't decided on a good way to handle persistent storage yet.

oshi97: Updated the PR to use global storage underneath instead of a regular object

src/core/filters/combinedfilternode.js
```diff
@@ -0,0 +1,95 @@
...
+
+import Filter from '../models/filter';
+import FilterCombinators from './filtercombinators';
+import SimpleFilterNode from './simplefilternode';
+
+/**
+ * A CombinedFilterNode represents a combined filter. They do not have their own filterView,
+ * and instead are defined by their children.
+ */
+export default class CombinedFilterNode {
```
tmeyer2115: extends FilterNode?

src/core/filters/combinedfilternode.js
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @type {Array<FilterNode>}
+     */
+    this.children = children || [];
+    Object.freeze(this);
+  }
+
+  /**
+   * Returns the filter associated with this node. Since this is node has children
```
tmeyer2115: Since this node has children?

src/core/filters/combinedfilternode.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Creates a branch filter node with children
+   * @param {string} combinator
+   * @param {Object|FilterNode} filterNodes
+   * @returns {FilterNode}
+   */
+  static _combine (combinator, filterNodes) {
+    if (!filterNodes.length) {
+      return new SimpleFilterNode();
```
tmeyer2115: I don't think a method like this should live at the CombinedFilterNode level. This method makes a decision whether or not you want a SimpleFilterNode or CombinedFilterNode. That's a higher level of abstraction than this class. Any static constructors of this class should return only instances of this class. It's a bit confusing otherwise. Maybe we can add a static method to FilterNode that makes the SimpleFilterNode/ComplexFilterNode decision and calls the appropriate class's static constructor.

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,103 @@
...
+import FilterView from './filterview';
+import StorageKeys from '../storage/storagekeys';
+
+/**
+ * FilterRegistry is a structure that manages {@link Filter}s and {@link Facet}s.
+ *
+ * Filters are stored in a tree structure, to allow combining of filters with
+ * combinators $and and $or.
+ *
+ * Facets are stored as an array of {@link FilterView}s/
```
tmeyer2115: Trailing /

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Sets the specified {@link FilterNode} under the given key.
+   * Will replace a preexisting node if there is one.
+   * @param {string} key
+   * @param {FilterNode} filterNode
+   */
+  setFilterNode (key, filterNode) {
+    this.globalStorage.set(`${StorageKeys.FILTER}.${key}`, filterNode);
```
tmeyer2115, line 81: Do we need to update persistentStorage as well?

src/core/filters/simplefilternode.js
```diff
@@ -0,0 +1,52 @@
+/** @module SimpleFilterNode */
+
+import FilterView from './filterview';
+
+/**
+ * A SimpleFilterNode represents a single, atomic filter.
+ */
+export default class SimpleFilterNode {
```
tmeyer2115: extends FilterNode

src/core/filters/simplefilternode.js
```diff
@@ -0,0 +1,52 @@
+/** @module SimpleFilterNode */
+
+import FilterView from './filterview';
+
+/**
+ * A SimpleFilterNode represents a single, atomic filter.
```
tmeyer2115, line 8: Update doc to be clearer about what an atomic filter is

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,103 @@
...
+     */
+    this.availableFieldIds = availableFieldIds;
+  }
+
+  /**
+   * Get all of the active {@link FilterView}s. Each {@link FilterView} corresponds
+   * to a single atomic filter.
+   * @returns {Array<FilterView>}
+   */
+  getFilterViews () {
```
tmeyer2115: I think the currency of the FilterRegistry should be FilterNodes. The need for FilterViews is a concern of the AppliedQueryFilters component. That component can extract the views from the nodes.

src/core/filters/filterregistry.js
```diff
@@ -0,0 +1,103 @@
...
+    return this.globalStorage.getAll(StorageKeys.FILTER).flatMap(node => node.getFilterViews());
+  }
+
+  /**
+   * Get all of the active {@link FilterView}s used for facets. Each {@link FilterView} corresponds
+   * to a single atomic filter.
+   * @returns {Array<FilterView>}
+   */
+  getFacetFilterViews () {
+    return this.globalStorage.getState(StorageKeys.FACET_FILTER) || [];
```
tmeyer2115, line 41: Are facets also stored using the FilterNode?

src/core/filters/filternode.js
```diff
@@ -0,0 +1,23 @@
...
+   * @returns {Filter}
+   */
+  getFilter () {}
+
+  /**
+   * Returns the filter views associated with this node.
+   * Overriden by implementing classes below.
+   * @returns {Array<FilterView>}
+   */
+  getFilterViews () {}
```
tmeyer2115: It feels weird to say that FilterNode is a tree structure and then not have a method in the interface like:

getChildren() returns Array<FilterNode>

Could we have something like:

class FilterNode {
  getFilterView() {}
  getChildren() {...}
}

The CombinedFilterNode would then use getChildren().map(node => node.getFilterView()) to construct it's filterView. With that interface, we acknowledge that every FilterNode has a visual representation (which even the combined ones do).

src/core/filters/filterregistry.js
oshi97: No, because not all filter components do that

src/core/filters/combinedfilternode.js
oshi97: I like your idea of adding the static helpers into filternode.js, since that lets them return different implementation of filter node without being confusing. Since doing this requires filternode.js to import SImpleFilterNode and CombinedFilterNode, I could either forego the extends FilterNode (which is effectively a fancy comment + mini unit test) or separate the interface from the static helpers. I'm not sure which one would be better? I feel kind of hesitant to add even more files and I'm not quite sure what to call the class full of static helpers.

oshi97: I guess I can just call it FilterNodeFactory, that kind of makes sense since it's creating filter nodes?

src/core/filters/combinedfilternode.js
oshi97: moved in to FilterNodeFactory

src/core/filters/simplefilternode.js
oshi97: updated!

src/core/filters/filterregistry.js
oshi97: That makes a lot of sense, updated!

src/core/filters/filterregistry.js
oshi97: I was thinking the minimal amount of information we need for facets is just the list of filter views, but it's much more orderly to just have the filter registry (and I believe also the filter components) to just deal with filternodes as you said. That should also make it more obvious that facets aren't nearly that special

src/core/filters/filternode.js
oshi97: Updated the interface so now every filternode is defined as something that
1. has an array of children (empty for simplefilternode)
2. has a filter
3. has a list of filter views that correspond to the filter

oshi97: I'm not entirely happy about the whole list of filter views thing, instead of just a single object we can point to and say "hey this is the display information for this node". It probably makes more sense that if somebody wants all of the filterviews at the leaves of the tree they can do that work themselves.

oshi97: Looking back I guess I'm just rewording what you're saying, but I kind of had to sit back for a minute and imagine at the interface

oshi97: updated so each filternode has
1. children(can be [])
2. filter
3. filterview (can be null)

src/core/filters/simplefilternode.js
```diff
@@ -0,0 +1,46 @@
...
+  getChildren () {
+    return [];
+  }
+
+  /**
+   * Returns the filter view of this node.
+   * @returns {Array<FilterView>}
+   */
+  getFilterView () {
+    return this.filterView;
```
tmeyer2115: Update JSDoc here.

src/core/filters/filternode.js
```diff
@@ -0,0 +1,29 @@
...
+   * Returns this node's filter.
+   * @returns {Filter}
+   */
+  getFilter () {}
+
+  /**
+   * Returns this node's filter view.
+   * @returns {FilterView}
+   */
+  getFilterView () {}
```
tmeyer2115: Could this be getFilterMetadata() and we just get rid off the FilterView class all together? Because getFilterView() has a getFilter() method of it's own, it feels somewhat redundant to have both a getFilter() and getFilterView() method here. Because FilterView contains a Filter, it also feels strange to say that a CombinedFilterNode has no FilterView, given that it has a Filter. But, the CombinedFilterNode does have no FilterMetadata. So maybe it would be simpler to scrap FilterView all together and have getMetadata() instead of getFilterView() here? How does that feel to you?

oshi97: Yeah I think that makes sense! FilterView doesn't do anything that FilterNode wouldn't want to. In the future having the onRemove method go on the FilterNode also probably makes more sense.

tmeyer2115 (approved):

tmeyer2115:
LGTM!

PR #673, opened 2020-04-10
Facets: Make facets searchable

alexisgrow:
TEST=manual


src/ui/components/filters/facetscomponent.js
```diff
@@ -202,6 +214,8 @@ export default class FacetsComponent extends Component {
         resetFilterLabel: this.config.resetFacetLabel,
         resetFiltersLabel: this.config.resetFacetsLabel,
         isDynamic: true,
+        isSearchable: this.config.isSearchable,
+        placeholderText: this.config.searchPlaceholderText,
```
tmeyer2115: Are the field-level overrides coming in a later PR? I think we could reduce the work required for that PR by making placeholderText and isSearchable filter-level configuration objects in this PR. For now, the isSearchable and placeHolderText in each filter config would be the same across all the fields.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -230,15 +231,61 @@ export default class FilterOptionsComponent extends Component {
 
     // show more/less button
     if (this.config.showMore) {
+      const showLessEl = DOM.query(this._container, '.js-yxt-FilterOptions-showLess');
+      const showMoreEl = DOM.query(this._container, '.js-yxt-FilterOptions-showMore');
+      const optionsOverLimitEls = DOM.queryAll(this._container, '.js-yxt-FilterOptions-aboveShowMoreLimit');
```
tmeyer2115, line 235: +2, this seems like a much more straightforward way of showing/hiding!

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -230,15 +231,61 @@ export default class FilterOptionsComponent extends Component {
...
+        DOM.query(this._container, '.js-yxt-FilterOptions-filter'),
+        'keyup',
+        event => {
+          const filter = event.target.value;
+          filterContainerEl.classList.add('yxt-FilterOptions-container--searching');
+
+          for (let filterOption of filterOptionEls) {
+            const labelEl = DOM.query(filterOption, '.js-yxt-FilterOptions-optionLabel');
+            const labelText = labelEl.textContent || labelEl.innerText;
+            if (!filter) {
```
tmeyer2115, line 272: Do they need to manually clear the filter from the input after they've made their selection(s)? Did we receive any guidance from UX about some kind of "Clear filter" button for the input?

alexisgrow: Ah true, I don't think there is a Clear Filter button. This is loop/function is called on every `keyup` event, including the backspace key, so this branch of the `if` is for when they delete the entire query, we want to remove all of the filtering styling that we added!

src/ui/components/filters/facetscomponent.js
alexisgrow: Oooh yes true, they are coming in a later PR but I agree. I'll move to the filter-level!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Just went through the spec again, it looks like the only case when they'd want to clear the filters is when someone does not have a query in the search input, so we should be good here!

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -103,6 +103,10 @@ export default class HandlebarsRenderer extends Renderer {
       return (arg1 !== arg2) ? options.fn(this) : options.inverse(this);
     });
 
+    this.registerHelper('ifgte', function (arg1, arg2, options) {
```
oshi97, line 106: :O nice!

README.md
```diff
@@ -1376,7 +1376,6 @@ ANSWERS.addComponent('SortOptions', {
   // Optional: The label to use for the reset button, defaults to 'reset'
   resetLabel: 'reset',
   // Optional: Allow collapsing excess filter options after a limit, defaults to true
-  // Note: screen readers will not read options hidden by this flag, without clicking show more first
```
oshi97: (I think you meant to update FilterOptions not SortOptions!!!)

src/ui/components/filters/facetscomponent.js
```diff
@@ -96,6 +96,18 @@ class FacetsConfig {
      */
     this.fieldControls = config.fieldControls || {};
 
+    /**
+     * The placeholder text used for the filter option search input
+     * @type {string}
```
oshi97, line 101: Could we add these new config options to the readme for Facets? I think we probably want them in the readme for FilterOptions and FilterBox as well (I mean the work is there at any rate)

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -230,15 +231,61 @@ export default class FilterOptionsComponent extends Component {
...
+      const showMoreEl = DOM.query(this._container, '.js-yxt-FilterOptions-showMore');
+      const optionsOverLimitEls = DOM.queryAll(this._container, '.js-yxt-FilterOptions-aboveShowMoreLimit');
       DOM.on(
-        DOM.query(this._container, '.yxt-FilterOptions-showToggle'),
+        showLessEl,
         'click',
         () => {
-          this.allShown = !this.allShown;
-          this.setState();
+          showLessEl.classList.add('hidden');
```
oshi97, line 240: Would it be possible to just add some class to the container to do all of this? Or does that make the css really annoying

README.md
alexisgrow: Eek good catch! The Filter Options portion of the readme doesn't have this note, so just going to undo this change.

src/ui/components/filters/facetscomponent.js
alexisgrow: Updated!

src/ui/components/filters/facetscomponent.js
alexisgrow: I'm going to update this in the later PR, I looked into it and it is slightly complicated. The spec wants these config options to apply across all fields; an implementer could override them for a field by specifying a field config, but if they want the same behavior for all, they would just specify the two config options here.

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Yes it is possible, but also yes it makes the CSS annoying haha! I liked this way of doing it because it's very clear what classes are required to make things work so it is easier to maintain. The way I'm thinking it would be done with a class on the container would have styling that targets HTML elements (e.g. `.container button div {}` or similar), which would be risky if anyone adjusts the DOM structure in the future.

oshi97: Ah okay got it!

src/ui/components/filters/facetscomponent.js
tmeyer2115: Does the spec call for FilterBox to take in these two, new config options? If not, we could achieve the same result by having the Facets component pass Filterbox a filters array where each filter has isSearchable and placeolderText set (to the value provided to the Facets config). This will do most of the plumbing for us when we need to do field-level overrides, then.

alexisgrow: Ah nice, ok true yes updated!

README.md
```diff
@@ -960,7 +960,7 @@ ANSWERS.addComponent('FilterBox', {
   // The label to show on the apply button
   applyLabel: 'apply',
   // Optional, whether or not this filterbox contains dynamic filters, default false
-  isDynamic: true
+  isDynamic: true,
```
tmeyer2115: Looks like there's a trailing comma.

src/ui/templates/controls/filteroptions.hbs
```diff
@@ -34,17 +34,20 @@
...
 
-    <div class='yxt-FilterOptions-options'>
+    {{#if searchable}}
+      <input type="text" class="js-yxt-FilterOptions-filter" placeholder="{{placeholderText}}">
+    {{/if}}
+    <ul class='yxt-FilterOptions-options'>
       {{#each options}}
         {{#if ../isSingleOption}}
-          <div class="singleoption-option yxt-FilterOptions-option">
+          <li class="singleoption-option yxt-FilterOptions-option">
```
tmeyer2115: Could we change this back to a div (and the ul back to a div)? I'm most likely being overly paranoid. If we change these, it alters the DOM structure for people without searchable facets as well. More than likely this would be totally fine and not break any customer's CSS. I think, as a team, we need to have a discussion about what kinds of alterations we can make to existing DOM structures.

alexisgrow: Yeah, I'll change back!

README.md
alexisgrow: Updating!

alexisgrow: updated

src/ui/templates/controls/filteroptions.hbs
```diff
@@ -56,33 +59,38 @@
...
               {{#if selected}}checked{{/if}}>
-            <label class="yxt-FilterOptions-optionLabel" for="js-yext-checkbox-{{../name}}-{{@index}}">{{label}} {{#if countLabel}}({{countLabel}}){{/if}}</label>
+            <label class="yxt-FilterOptions-optionLabel js-yxt-FilterOptions-optionLabel" for="js-yext-checkbox-{{../name}}-{{@index}}">
+              {{label}} {{#if countLabel}}({{countLabel}}){{/if}}
+            </label>
           </div>
         {{/if}}
       {{/each}}
-    </div>
+    </ul>
```
tmeyer2115: div?

alexisgrow: oof 🙈

alexisgrow: lol fixed

tmeyer2115 (approved):

PR #674, opened 2020-04-10
Add setGeolocation to ANSWERS object.

nbramblett:
Allows the user to override the geolocation value in global storage.

TEST=manual

Called setGeolocation and verified that the location value in the search
request was the expected value.


oshi97 (approved):
lgtm!!

PR #675, opened 2020-04-10
Pagination: Miscellaneous updates

nbramblett:
Adds miscellaneous quality-of-life changes to pagination:

- Icons can be overwritten with a user-supplied url
- Booleans for first/last button have been reworked (and the deprecated booleans might just be scrapped before release)
- New onPaginate function which is called on paginate

TEST=manual

Verified the new features worked nicely on my local page.


oshi97 (approved):

oshi97 (commented):
nice!

src/ui/components/results/paginationcomponent.js
```diff
@@ -24,18 +24,37 @@ export default class PaginationComponent extends Component {
...
      */
-    this._lastPageButtonEnabled = config.showLast === undefined ? true : config.showLast;
+    this._lastPageButtonEnabled = config.showLast === undefined ? config.showFirstAndLastButton : config.showLast;
+
+    /**
+     * If true, displays the first and last page buttons
+     * @type {boolean}
+     * @private
+     */
+    this._showFirstAndLastPageButtons = config.showFirstAndLastButton !== false;
```
tmeyer2115, line 45: For consistency's sake, let's adopt the === undefined ? true : config.showFirstAndLastButton conversion here. I think it would simplify things to have this._showFirstAndLastPageButtons computed before this._firstPageButtonEnabled and this._lastPageButtonEnabled. Then, the latter two booleans could use this in their default. That would save the || on line 163 and 164.

src/ui/components/results/paginationcomponent.js
```diff
@@ -113,7 +139,7 @@ export default class PaginationComponent extends Component {
   }
 
   updatePage (offset) {
-    this.scrollToTop();
+    this._onPaginate();
```
tmeyer2115, line 142: According to the spec, onPaginate is supposed to take 3 arguments: newPage, currentPage, and totalPageCount.

src/ui/templates/results/pagination.hbs
```diff
@@ -7,12 +7,18 @@
         data-eventoptions='{"furthermost":true, "pageNumber":{{pageNumber}} }'
         aria-label="Go to the first page of results"
       >
-        <span class="yxt-Pagination-doubleChevron--left"
-              data-component="IconComponent"
-              data-opts='{ "iconName": "chevron" }'></span>
-        <span class="yxt-Pagination-chevron--left"
```
tmeyer2115, line 13: This is confusing to me. Do you know why we include both spans in this case? Wouldn't the previous button take care of this second span?

src/ui/templates/results/pagination.hbs
```diff
@@ -7,12 +7,18 @@
...
+          <span class="yxt-Pagination-doubleChevron--left"
+                data-component="IconComponent"
+                data-opts='{ "iconName": "chevron" }'></span>
+          <span class="yxt-Pagination-chevron--left"
+                data-component="IconComponent"
+                data-opts='{ "iconName": "chevron" }'></span>
+        {{else}}
+          <span class="yxt-Pagination-doubleChevron--left"
+                data-component="IconComponent"
+                data-opts='{ "iconUrl": "{{icons.firstButtonIcon}}" }'></span>
```
tmeyer2115, line 20: My reading of the spec makes me think that the string here can either be a URL or the name of one of our built-in icons. If it's the name of a built-in icon, we'd use the iconName property. I'd double check with Rose to be sure this (and the other icons) can be either a name or a url.

src/ui/components/results/paginationcomponent.js
nbramblett: Adopting the undefined ? true. The other part has already been resolved.

src/ui/components/results/paginationcomponent.js
nbramblett: Rip my life. Resolving in a separate pr.

src/ui/templates/results/pagination.hbs
nbramblett: There's no double-chevron icon, they just make the chevron twice and put them close together.

src/ui/templates/results/pagination.hbs
nbramblett: Are those icon names publicly documented at all? Seems a bit weird to me. I can ask.

nbramblett: I think if we're going to support icon names as well then it would need to be a separate parameter unless we decided we were going to get clever with regex (I think other components also have separate params for iconName and iconUrl as well). Also the doubleChevron might be a bit tricky since it's actually the same icon twice :/

PR #680, opened 2020-04-13
Searchable Facets: Add highlight on match and overrides

alexisgrow:
TEST=manual


src/ui/components/filters/facetscomponent.js
```diff
@@ -92,6 +92,9 @@ class FacetsConfig {
     /**
      * The controls to use for each field. Each type of filter has a default
      * $eq : multioption (checkbox)
+     *
+     * DEPRECATED: prefer putting this in config.fields
+     *
      * @type {Object}
      */
     this.fieldControls = config.fieldControls || {};
```
tmeyer2115, line 100: Does the spec say we should deprecate fieldControls? Could we just continue using fieldControls instead of adding a new config object called fields? I'm wary of us continuing to add more and more config options/synonyms to these components. I think it causes confusion.

src/ui/components/filters/facetscomponent.js
```diff
@@ -196,11 +207,14 @@ export default class FacetsComponent extends Component {
     }
 
     filters = filters.map(f => {
+      const fieldOverrides = this.config.fields[f.fieldId] || {};
+      const control = fieldOverrides.control || this.config.fieldControls[f.fieldId] || 'multioption';
```
tmeyer2115: This kind of logic is precisely what I want to avoid peppering our code with, where we have to check multiple objects for the same piece of info. Hopefully the PMs are cool with just continuing to use fieldControls.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -311,8 +321,13 @@ export default class FilterOptionsComponent extends Component {
     }
   }
 
-  _doesOptionMatchFilter (option, filter) {
-    return option && filter && option.toLowerCase().indexOf(filter.toLowerCase()) > -1;
+  _getMatchedSubstrings (option, filter) {
```
tmeyer2115: getMatchedSubstring since this only returns an array of one?

src/ui/templates/controls/filteroptions.hbs
```diff
@@ -56,7 +56,14 @@
               value="{{value}}"
               data-index="{{@index}}"
               {{#if selected}}checked{{/if}}>
-            <label class="singleoption-label" for="js-yext-radio-{{../name}}-{{@index}}">{{label}} {{#if countLabel}}({{countLabel}}){{/if}}</label>
+            <label class="singleoption-label" for="js-yext-radio-{{../name}}-{{@index}}">
+              <span class="yxt-FilterOptions-optionLabel--label js-yxt-FilterOptions-optionLabel--label">
```
tmeyer2115: Could we come up with a different suffix than --label? It's a bit confusing to see something called optionLabel--label, makes me think we have an unnecessary layer of nesting in our DOM. Maybe optionLabel--name and optionLabel--count?

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Updating!

src/ui/templates/controls/filteroptions.hbs
alexisgrow: Yeah for sure! Updating

src/ui/components/filters/facetscomponent.js
alexisgrow: Talked offline, removing this

src/ui/components/filters/facetscomponent.js
alexisgrow: Talked offline, removing this as well!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: updated 🥳

src/ui/templates/controls/filteroptions.hbs
alexisgrow: updated! ✨

src/ui/components/filters/facetscomponent.js
```diff
@@ -196,11 +201,13 @@ export default class FacetsComponent extends Component {
...
 
     filters = filters.map(f => {
+      const fieldOverrides = this.config.fields[f.fieldId] || {};
       return Object.assign({}, f, {
         type: 'FilterOptions',
         control: this.config.fieldControls[f.fieldId] || 'multioption',
         searchable: this.config.searchable,
-        placeholderText: this.config.placeholderText
+        placeholderText: this.config.placeholderText,
+        ...fieldOverrides
```
tmeyer2115, line 213: Sorry if I caused confusion earlier, I think we should allow people to specify control in fieldOverrides as well. I think the DEPRECATED comment you had over fieldControls was totally fine. I was initially hoping that we could just get rid of fields and use fieldControls, but because that's not the case, I think we want to make it clear that field supports everything fieldControls did (and more) and that fieldControls is deprecated. We should update the README with that info too

alexisgrow: Ah nice ok I will update the comment and the README! 

I think specifying the fieldOverrides last (like this) will override anything above it and it's a little cleaner looking than creating a variable like I had earlier. What do you think?

tmeyer2115: Sounds good to me!

tmeyer2115 (approved):

PR #681, opened 2020-04-13
Searchable Facets: Implement levenshtein distance matching

alexisgrow:
Allow searchable facets to support a one-character typo if the search string is longer than three characters, using the levenshtein distance.

TEST=manual

Tested on Chrome and IE11 using our demo account. The current minimum query length for typos is 3, so tested with a 4 character string, with and without a 1 character typo. Verify that results come back correctly.


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -269,14 +270,15 @@ export default class FilterOptionsComponent extends Component {
 
           for (let filterOption of filterOptionEls) {
             const labelEl = DOM.query(filterOption, '.js-yxt-FilterOptions-optionLabel--label');
-            const labelText = labelEl.textContent || labelEl.innerText;
+            let labelText = labelEl.textContent || labelEl.innerText || '';
```
tmeyer2115, line 273: Do we need the default to '' here?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -269,14 +270,15 @@ export default class FilterOptionsComponent extends Component {
...
+            labelText = labelText.trim();
```
tmeyer2115, line 274: Just to centralize some of the logic, could we make a sanitizeSting method that trims and toLowerCases? We could call that on both the filter and labelText.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -322,14 +324,50 @@ export default class FilterOptionsComponent extends Component {
...
+        offset: offset
       }];
     }
   }
 
+  _calcLevenshteinDistance (a, b) {
+    return levenshtein(a, b);
+  }
+
+  _getOffset (option, filter) {
```
tmeyer2115, line 397: Add some JSDocs to _getOffset, _calcLeven, and _getMatchedSubstring

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -322,14 +324,50 @@ export default class FilterOptionsComponent extends Component {
   }
 
   _getMatchedSubstrings (option, filter) {
-    if (option && filter && option.toLowerCase().indexOf(filter.toLowerCase()) > -1) {
+    const arbitraryStringLength = 3; // Defined in spec
+    const arbitraryLevenshteinMax = 1; // Defined in spec
```
tmeyer2115: I don't think we need the arbitrary prefix in these variable names, the values were chosen deliberately. Maybe:
minFilterSizeForLevenshtein and maxLevenshteinDistance?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -322,14 +324,50 @@ export default class FilterOptionsComponent extends Component {
...
+    let offset = this._getOffset(option, filter);
+
+    if (filter.length > arbitraryStringLength) {
```
tmeyer2115: Before we do this work, should we just check if offset > -1? If so, return and we don't have to go into this branch at all.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -322,14 +324,50 @@ export default class FilterOptionsComponent extends Component {
...
+      // Break option into X filter.length size substrings
+      let substrings = [];
+      for (let start = 0; start <= (option.length - filter.length); start++) {
+        substrings.push(option.substr(start, filter.length));
+      }
+
+      // Find the substring that is the closest in levenshtein distance to filter
+      let minLevDist = filter.length;
+      let minLevSubstring = filter;
+      for (let substring of substrings) {
```
tmeyer2115, line 361: Not necessary, but I think there's another optimization you can make here. You don't necessarily need to compute the levenshtein distance for each substring. If one substring has a distance >= 3, you should be able to skip the next substring, since the distance of that substring can, at best, be 2.

tmeyer2115 (commented):
Could you add some detail in the commit message, including how info about testing? Was this tested in IE11 as well? Whenever we add a 3rd party library, I think it's good to make sure it works as expected in IE11.

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Ah yes, updating 👾

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Yes totally, updating!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: 💯 updated

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: on it! 🚀

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: I think so if we are going to `.trim()` right after this, just in case `textContent` and `innerText` are null/undefined for some reason

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Yeah I agree that it is sort of weird that the white space and casing manipulation is not in one place. I'm not really sure the best way to fix that because the .trim() and toLowerCase are both kind of distinct usages

This is the only place `.trim()` happens, and everything that uses labelText needs it trimmed. All usages but except this one do not want casing manipulated:
- The lower-cased labelText and filter are used to find a text match in the `let matchedSubstrings = this._getMatchedSubstrings(labelText.toLowerCase(), filter.toLowerCase());`.

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: looking into this!!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: updated!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Talked offline - Moving the _getOffset call so it is only used when the levenshtein distance is < 1 will be a larger performance gain! We are going to do that and come back to this if we need further performance enhancements.

alexisgrow: Talked offline - Moving the _getOffset call so it is only used when the levenshtein distance is < 1 will be a larger performance gain! We are going to do that and come back to this if we need further performance enhancements.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -321,13 +323,79 @@ export default class FilterOptionsComponent extends Component {
...
+   * (string) filter "match".
+   *
+   * "Match" is defined as an exact text match, or -- if the length of filter
+   * is greater than the `minFilterSizeForLevenshtein` -- a "match" can occur if
+   * any "n length" substring of option (where "n length" is the length of filter)
+   * is within the `maxLevenshteinDistance` levenshtein distance of the filter.
+   *
+   * Note: this is case sensitive.
+   *
+   * @returns {Array}
```
tmeyer2115: Minor nit, let's have this return an Object and wrap it in an [..] when we call HighlightedValue. This method is only meant to return at most one matchedSubstring.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -269,14 +270,15 @@ export default class FilterOptionsComponent extends Component {
...
+            let labelText = labelEl.textContent || labelEl.innerText || '';
+            labelText = labelText.trim();
             if (!filter) {
               filterContainerEl.classList.remove('yxt-FilterOptions-container--searching');
               filterOption.classList.remove('hiddenSearch');
               filterOption.classList.remove('displaySearch');
               labelEl.innerHTML = labelText;
             } else {
-              let matchedSubstrings = this._getMatchedSubstring(labelText, filter);
+              let matchedSubstrings = this._getMatchedSubstring(labelText.toLowerCase(), filter.toLowerCase());
```
tmeyer2115: matchedSubstring =

alexisgrow: updating!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: updating!

tmeyer2115 (approved):

PR #682, opened 2020-04-14
Pagination: Page number update and tests

nbramblett:
Adds page numbers to the Pagination component, per the spec.

J=SPR-1969
TEST=manual,auto

Manually verified the pagination component had correct behavior. Wrote
unit tests.


src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,75 @@ export default class PaginationComponent extends Component {
     document.body.scrollTop = 0;
   }
 
+  /**
+   * Pagination should evenly allocate between front and back, unless one side hits the limit,
+   * in which case the remaining side should be allocated up to the max.
```
oshi97: Could we maybe include an example of what this does, or find a better word than allocate here (I know its hard I can't really think of any at the top of my head)? It was a little confusing to figure out what this does.

README.md
```diff
@@ -883,6 +883,12 @@ ANSWERS.addComponent('Pagination', {
   container: '.pagination-component',
   // Must be assigned either here or in the search config.
   verticalKey: '',
+  // The maximum number of pages visible to non-mobile users.
```
oshi97, line 886: I think it could be nice to have the default values included here too!

src/ui/templates/results/pagination.hbs
```diff
@@ -38,12 +38,62 @@
...
+      data-eventoptions='{"currentPage":{{pageNumber}}, "newPage":1, "totalPageCount":{{maxPage}} }'
+    >
+      1
+    </a>
+
+    <span class="yxt-Pagination-page{{#unless ellipses.desktopBack}} desktop-hidden{{/unless}}{{#unless ellipses.mobileBack}} mobile-hidden{{/unless}}">
+      ...
+    </span>
+
+    {{#for 1 pageNumber 1}}
```
oshi97: I was a little confused by this #for helper too, since the regular #for helper looks through all values in a given variable I assumed this was doing something like {{#for [1, pageNumber, 1]}} but that turned out to not be the case. Could we either call this helper something else, or just use a regular handlebars #for? If we do that, we could have it be an array of objects with additional attributes like desktopHidden and mobileHidden so we could take out some of the lte and gt logic, and even combine the below anchor tags into a single inline template.

tests/ui/components/results/paginationcomponent.js
```diff
@@ -0,0 +1,150 @@
+import { mount } from 'enzyme';
```
oshi97, line 1: yay unit tests!

src/ui/templates/results/pagination.hbs
nbramblett: Handlebars's 'each' block mimics the for-each structure, but this helper mimics a more standard for loop of startValue - limit - increment. Breaking out the logic into the backend makes sense though, and fits better with the logic-less standards of Handlebars, so I'll see what I can do.

README.md
nbramblett: For the most part, the values I put in the readme match the defaults. I'll make sure to add a note when this is not the case.

src/ui/components/results/paginationcomponent.js
nbramblett: Will do!

README.md
oshi97: Oh I think that's pretty reasonable in general! it's just that the rest of the readme doesn't always do that

README.md
```diff
@@ -883,11 +883,17 @@ ANSWERS.addComponent('Pagination', {
...
+  maxVisiblePagesDesktop: 1,
```
oshi97, line 887: Doesn't this default to 0, not 1?

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,102 @@ export default class PaginationComponent extends Component {
...
+    var desktopLimits = this._allocate(pageNumber, maxPage, this._maxVisiblePagesDesktop);
+    const { ellipses, pageNumbers } = this._createPageNumbers(pageNumber, maxPage, mobileLimits, desktopLimits);
+
     return super.setState({
       showControls: this.shouldShowControls(results, limit),
-      firstPageButtonEnabled: this._firstPageButtonEnabled || this._showFirstAndLastPageButtons,
-      lastPageButtonEnabled: this._lastPageButtonEnabled || this._showFirstAndLastPageButtons,
+      firstPageButtonEnabled: this._firstPageButtonEnabled,
+      lastPageButtonEnabled: this._lastPageButtonEnabled,
       pageNumber: pageNumber + 1,
```
oshi97: I know it was like this before, but it feels kinda weird to call it `pageNumber` when it's really closer to `pageIndex`.  Could we make pageNumber = offset / limit + 1? Then in your allocate method you could change the minPage from 0 to 1, and you might not need this maxPage = Math.trunc weirdness (which kind of just looks like they were whacking bugs into the ground until they stopped coming up)

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,102 @@ export default class PaginationComponent extends Component {
...
+  }
+
   setState (data) {
     const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
     let offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
     const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
     const pageNumber = offset / limit;
     const isMoreResults = results.resultsCount > offset + limit;
     const maxPage = Math.trunc((results.resultsCount - 1) / limit);
+    var mobileLimits = this._allocate(pageNumber, maxPage, this._maxVisiblePagesMobile);
```
oshi97: Could these two limit arrays go inside _createPageNumbers? I think we like to use const where we can as well

README.md
nbramblett: Oof, good catch. It's supposed to default to 1!

oshi97 (approved):
lgtm!!

src/ui/components/results/paginationcomponent.js
```diff
@@ -44,6 +58,13 @@ export default class PaginationComponent extends Component {
      */
     this._showFirstAndLastPageButtons = config.showFirstAndLastButton !== false;
 
+    /**
+     * If true, always displays the page numbers for first and last page.
+     * @type {boolean}
+     * @private
+     */
+    this._pinFirstAndLastPage = !!config.pinFirstAndLastPage;
```
tmeyer2115, line 66: It may not be the best way to do the conversion to a Boolean, but we do the === undefined ? DEFAULT_VALUE : CONFIG_VALUE approach everywhere else in our codebase. I think we should do that here for consistency's sake. I'm certainly open to a different way of casting, but we'd want it consistent everywhere.

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
   }
 
+  /**
+   * Pagination should evenly add page numbers in the "forward" and "backward" directions, unless
+   * one side has reached the max/min value, in which case the remaining side should be the only
+   * one to get more pages.
+   * @param {number} pageNumber the current page's number
+   * @param {number} maxPage the highest page number, acts as the upper bound
+   * @param {number} limit the maximum total number of pages that are allocated
+   * @returns {Array<number>} the backLimit and frontLimit, respectively
```
tmeyer2115, line 186: I think we should return an object here. With an object, we can add attributes with descriptive names. If we return an array, I think it could be difficult for consumers of this method to keep track of which array element corresponds to what.

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+   */
+  _allocate (pageNumber, maxPage, limit) {
+    var backLimit = pageNumber;
+    var frontLimit = pageNumber;
+    for (var i = 0; i < limit; i++) {
```
tmeyer2115, line 191: Not necessary, but I think we can simplify this code somewhat. We'd maintain a leftOffset and rightOffset, much like what you have here. We'd initialize them to be limit/2 or (limit-1)/2 (not sure if limit includes the current page or not). If pageNumber + rightOffset > maxPage, we subtract the delta from rightOffset and add it to leftOffset. Else if pageNumber - leftOffset < 0, we subtract the delta from left Offset and add it to the rightOffset. Then, our backLimit is pageNumber - leftOffset and frontLimit is pageNumber + rightOffset.

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+        }
+        if (i <= desktopBackLimit || i > desktopFrontLimit) {
+          num.desktopHidden = true;
+        }
+      }
+      pageNumbers.push(num);
+    }
+
+    return {
+      ellipses: {
```
tmeyer2115, line 236: Do we only need ellipses if pinFirstAndLast is true?

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+      }
+    }
+
+    return [backLimit, frontLimit];
+  }
+
+  /**
+   * Creates an object representing the view state of the page numbers and ellipses
+   * @param {number} pageNumber refers to the page number, not the page index
+   * @param {number} maxPage
```
tmeyer2115, line 213: We should add some jsdoc about maxPage

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+   * @returns {Object}
```
tmeyer2115, line 214: Add some JSDoc about this

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+   */
+  _createPageNumbers (pageNumber, maxPage) {
```
tmeyer2115, line 216: _createPageNumberViews? That indicates we're returning some kind of view model.

src/ui/components/results/paginationcomponent.js
```diff
@@ -151,24 +176,98 @@ export default class PaginationComponent extends Component {
...
+    }
+
+    return {
+      ellipses: {
+        mobileBack: mobileBackLimit > 0,
+        mobileFront: mobileFrontLimit < maxPage,
+        desktopBack: desktopBackLimit > 0,
+        desktopFront: desktopFrontLimit < maxPage
+      },
+      pageNumbers
```
tmeyer2115, line 242: pageNumberViews?

src/ui/components/results/paginationcomponent.js
nbramblett: Done

src/ui/components/results/paginationcomponent.js
nbramblett: I tried something similar previously, but had some annoyances with off-by-one errors (thanks trunc) and the issue where _both_ sides were exceeding the limit. It's definitely doable, but I'm not convinced it's cleaner.

src/ui/components/results/paginationcomponent.js
nbramblett: Yep, rose confirmed they should only be visible if there's a "page number skip", which is equivalent to pinfirstAndLastPage=true. This has been changed in pr #688  to reflect that.

PR #683, opened 2020-04-14
Use FilterRegistry for filter storage and retrieval.

oshi97:
This commit changes all filter components to use FilterRegistry
for storing filternodes and facet filternodes. StorageKeys.FILTER
and StorageKeys.FACET_FILTER were changed to StorageKeys.FILTER_NODE and and StorageKeys.FACET_FITLER_NODE respectively.

Each filter component also now gives each FilterNode it stores a
FilterMetadata object.

There is currently an issue with facets when no results are returned.
Even if displayAllResults is not set to true, or if using the old
no results, the facets behave as if all results are being returned.
This is likely due to #611, and was not changed as of this commit.

TEST=manual
Created test page with Filter(Options, Box, Search), Facets,
DateRange and Range Filters, Geolocation filter.
Each component was first tested by individually on the page.
Later on all of them were added to the page, and applied to
the search at the same time.
Checked that the expected request was created (by checking
the network tab), and that expected metatdata was created (by
manually checking ANSWERS.core.filterRegistry in the console)
When all were turned on together, checked that there were 6
top level filter nodes in the registry.
Checked that removing filters removed them from the request.
Currently, that results in setting that filter component's node
to an empty filter node, as opposed to clearing it.


src/core/models/filter.js
```diff
@@ -70,26 +70,30 @@ export default class Filter {
   }
 
   /**
-   * OR filters with the same keys, then AND the resulting groups
-   * @param  {...Filter} filters The filters to group
+   * Helper method for creating a range filter
+   * @param {string} field field id of the filter
+   * @param {*} min minimum value, can be falsy
```
nbramblett: This can be (number | undefined), which is more explicit

src/core/models/filter.js
```diff
@@ -70,26 +70,30 @@ export default class Filter {
...
+  static range (field, min, max, isExclusive) {
+    if (!min && !max) {
+      return null;
+    } else if (min && !max) {
+      return isExclusive
+        ? Filter.greaterThan(field, min)
+        : Filter.greaterThanEqual(field, min);
+    } else if (max && !min) {
+      return isExclusive
+        ? Filter.lessThan(field, max)
```
nbramblett, line 91: Would it be possible to have exclusivity be a parameter in the filter instead of a totally separate object?

src/core/models/filter.js
```diff
@@ -70,26 +70,30 @@ export default class Filter {
...
-    for (const field of Object.keys(groups)) {
-      groupFilters.push(groups[field].length > 1 ? Filter.or(...groups[field]) : groups[field][0]);
+  static range (field, min, max, isExclusive) {
+    if (!min && !max) {
+      return null;
+    } else if (min && !max) {
+      return isExclusive
+        ? Filter.greaterThan(field, min)
+        : Filter.greaterThanEqual(field, min);
+    } else if (max && !min) {
```
nbramblett: If you're returning then you don't need the else's, unless you just like how it looks I guess.

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -132,26 +143,53 @@ export default class DateRangeFilterComponent extends Component {
...
+        fieldName: this._title
+      });
     }
-    return this._isExclusive
-      ? Filter.exclusiveRange(this._field, this._date.min, this._date.max)
-      : Filter.inclusiveRange(this._field, this._date.min, this._date.max);
+    let displayValue;
+    if (min && !max) {
+      displayValue = this._isExclusive
+        ? `After ${min}`
```
nbramblett, line 177: This doesn't look like the user would be able to configure it, meaning it isn't i18n-able. That ok?

src/core/models/filter.js
```diff
@@ -70,26 +70,30 @@ export default class Filter {
...
-      groups[key].push(filter);
-    }
-
-    const groupFilters = [];
-    for (const field of Object.keys(groups)) {
-      groupFilters.push(groups[field].length > 1 ? Filter.or(...groups[field]) : groups[field][0]);
+  static range (field, min, max, isExclusive) {
+    if (!min && !max) {
+      return null;
+    } else if (min && !max) {
```
nbramblett: Can the values being filtered ever potentially be negative? If so, it's not possible to check a "less than/greater than 0" with this setup, because min=0/max=0 will get falsey-ed.

src/ui/components/filters/rangefiltercomponent.js
```diff
@@ -128,21 +139,55 @@ export default class RangeFilterComponent extends Component {
...
+   */
+  _buildFilterMetadata () {
+    const { min, max } = this._range;
+
+    if (!min && !max) {
+      return new FilterMetadata({
+        fieldName: this._title
+      });
+    }
+    // TODO allow range filter to have exclusive ranges
```
nbramblett: Seems like this has been implemented, based on the work done in filters.js. Remove?

oshi97: I meant the range filter component itself, not the filter model

src/core/models/filter.js
oshi97: The filter model is a little weird in that instead of constructing the format the backend wants at the time of request, and letting us do whatever we want in between, it does that immediately upon construction. Part of the whole FilterNode business was NOT doing that for combined filters ($or of $and). In the future a similar, and much less complicated since there is no nesting, thing could be done for "simple" filters.

Anyways until we refactor that it has to be a separate object.

src/core/core.js
```diff
@@ -360,8 +362,12 @@ export default class Core {
     this.globalStorage.set(`${StorageKeys.FILTER}.${namespace}`, filter);
   }
 
-  setFacetFilter (namespace, filter) {
```
tmeyer2115, line 363: Does the setFilter method need to be deleted as well?

src/core/storage/storagekeys.js
```diff
@@ -14,6 +14,7 @@ export default {
   AUTOCOMPLETE: 'autocomplete',
   DIRECT_ANSWER: 'direct-answer',
   FILTER: 'filter',
```
tmeyer2115: Do we still need FILTER or FACET_FILTER?

src/ui/components/filters/daterangefiltercomponent.js
tmeyer2115: Most likely, this needs to be "i18n"-able. We'll see what Rose says.

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -132,26 +143,53 @@ export default class DateRangeFilterComponent extends Component {
...
+  }
+
+  /**
+   * Helper method for creating a date range filter metadata
+   * @returns {FilterMetadata}
+   */
+  _buildFilterMetadata () {
+    const { min, max } = this._date;
+
+    if (!min && !max) {
```
tmeyer2115, line 169: Would this happen when they have cleared both input fields of the filter? In that case, wouldn't the FilterNode be deleted and we wouldn't need the metadata?

src/core/core.js
```diff
@@ -360,8 +362,12 @@ export default class Core {
...
-    this.globalStorage.set(`${StorageKeys.FACET_FILTER}.${namespace}`, filter);
+  setFacetFilterNodes (availableFieldids = [], filterNodes = []) {
```
tmeyer2115, line 355: The previous setFacetFilter included a namespace. Is that needed here?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -303,21 +301,46 @@ export default class FilterOptionsComponent extends Component {
...
+      }
+      fieldIdToFilterNodes[fieldId].push(fn);
+    }
+
+    const totalFilterNodes = [];
+    for (const sameIdNodes of Object.values(fieldIdToFilterNodes)) {
+      totalFilterNodes.push(FilterNodeFactory.or(...sameIdNodes));
+    }
+
+    return FilterNodeFactory.and(...totalFilterNodes);
```
tmeyer2115, line 340: Since this is the FilterOptions component, wouldn't we just want a top-level or here?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -190,7 +191,7 @@ export default class GeoLocationComponent extends Component {
       onSubmit: (query, filter) => {
         this.query = query;
         this.filter = Filter.fromResponse(filter);
-        this._saveDataToStorage(query, this.filter);
+        this._saveDataToStorage(query, this.filter, query.split(',')[0]);
```
tmeyer2115: What is query here? What would an example displayValue for this component be? "McLean, VA" or something like that?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -178,6 +179,14 @@ export default class FilterSearchComponent extends Component {
         this.core.persistentStorage.set(`${StorageKeys.QUERY}.${this.name}`, this.query);
         this.core.persistentStorage.set(`${StorageKeys.FILTER}.${this.name}`, this.filter);
         this.core.setFilter(this.name, this.filter);
```
tmeyer2115: Should we still be calling setFilter here?

src/ui/components/filters/daterangefiltercomponent.js
oshi97: yes great catch T_T

src/core/models/filter.js
oshi97: eh I guess that's shorter haha

src/core/models/filter.js
oshi97: yeah that's a great point! I'll add unit tests for these edge cases after this commit

tmeyer2115 (commented):
Can we run those same tests in IE11 as well? I find it a bit weird that the FieldNodeRegistry only interacts with global storage and it's up to each component to make sure the filter state persists across refreshes. Can we think of some way that the FieldNodeRegistry can handle this interaction with persistentStorage as well? Are these the main blockers to that:

- Different filters serialize their data in persistent storage differently.
- We don't store enough info in persistent storage to reconstruct the filter's metadata.

src/core/core.js
oshi97: I want to keep it around for now, the geolocation filter and filtersearch both listen on global storage for StorageKeys.FILTER updates, and I didn't want to deal with those for now.

src/core/storage/storagekeys.js
oshi97: still need FILTER, can rename FACET_FILTER to FACET_FILTER_NODE

src/ui/components/filters/daterangefiltercomponent.js
oshi97: the current behavior (in this pr) is that _updateRange is called when one of the values is updated. This ends up setting a new FilterNode, with newly generated metadata. It could be changed to delete the FilterNode as well, I just thought this might be simpler. Which do you prefer? I kind of like it this way because it keeps the responsibilities slightly separated.

src/core/core.js
oshi97: nope. The previous code for pulling facets from global storage and adding them to the request only ever pulled the first available value. No components ever listened on FACET_FILTER in global storage, and as we very well know facets are not stored in persistent storage.

src/ui/components/filters/filteroptionscomponent.js
oshi97: FilterOptions has always ORed together all filters of the same fieldId, then ANDed the resulting filters.

src/ui/components/filters/geolocationcomponent.js
oshi97: Yup! 'McLean, Virginia, United States of America' is an exact example. The geolocation filter and filter search are missing from the spec so I should ask the PMs about it.

src/ui/components/search/filtersearchcomponent.js
oshi97: Yes this was very intentional! Do you think it's a bad idea?

oshi97:
> Can we run those same tests in IE11 as well? I find it a bit weird that the FieldNodeRegistry only interacts with global storage and it's up to each component to make sure the filter state persists across refreshes. Can we think of some way that the FieldNodeRegistry can handle this interaction with persistentStorage as well? Are these the main blockers to that:
> 
> * Different filters serialize their data in persistent storage differently.
> * We don't store enough info in persistent storage to reconstruct the filter's metadata.

Yeah those are some of the blockers, each component really just follows its own beat right now.

There's an issue of unique namespace across pages as well. Even if we assume that every component has a unique name, that doesn't stop people from giving two completely different components on two separate tabs the same name. normally at worst this won't do anything but make the page look slightly misconfigured, but with regards to the 13.1 facets recut this can end up breaking the page. 

In addition to that, we don't have a good idea of how persistent storage should work. We can't just throw pure json into the URL forever, since we have a limit of 2048 characters in ie11, and what exactly is supposed to go inside persistent storage. In my mind, persistent storage is for storing the current state of all the components on the page. Then, when the page reloads, we can reseed the components from that state. If this were react, we could quite literally store the state, but this very clearly is not react, and our component._state is way too bulky to store in the url, even if we find some way to compress things. We would have to refactor at least our filter components so that state really is just the minimum information to derive state from, not everything our handlebars templates felt like having inside of them.

So basically it's very frustrating but I don't think there's a good programmatic way to do this right now without ripping up lots and lots of code.

oshi97:
Tested in ie11. Everything looked good except for the geolocation filter, which always returns a "This site does not have permission to use the Geolocation API." This occurs both in browserstack and my home windows machine, and was present in v0.13.2 as well. For my home machine, I tried changing the internet options for ie11, as well as enabling location tracking in my windows 10 settings and restarting the computer, and even reinstalling ie11 to no avail.

oshi97:
Tested on safari and firefox, things are working as expected (had to open html file directly into safari instead of through local dev server, otherwise apparently safari blocks geolocation on insecure protocols)

README.md
```diff
@@ -1178,6 +1178,24 @@ ANSWERS.addComponent('DateRangeFilter', {
   initialMax: '2019-09-01',
   // If true, this filter represents an exclusive range, rather than an inclusive one
   isExclusive: false,
+  // A function that generates the text to show when displaying a less than filter
```
tmeyer2115: I'm not sold on making these configuration options. My objections are:

1) We don't know if NJ or the other Covid sites will get AppliedQueryFilters. 
2) I don't think we should give people all these options, even in English. They can override the template if they want. We need to iron out that line with the PMs between what constitutes a config option and what should be done by overriding the template.

src/ui/components/search/filtersearchcomponent.js
tmeyer2115: If possible, I think we should remove all references to setFilter, StorageKeys.FILTER, or StorageKeys.FACET. If we keep those things in, it feels like we're in a hybrid place where some things rely on the FieldRegistry to manage their global state and other filters manage it themselves. Everything in globalStorage related to facets or filters should be a FilterNode, right?

README.md
oshi97: It's not as simple as just overriding the template, since these are stored inside filter metadata. But I agree that this isn't a good solution, I don't think we want to add a billion config options for things text overrides and don't affect functionality outside of that.

tmeyer2115:
> > Can we run those same tests in IE11 as well? I find it a bit weird that the FieldNodeRegistry only interacts with global storage and it's up to each component to make sure the filter state persists across refreshes. Can we think of some way that the FieldNodeRegistry can handle this interaction with persistentStorage as well? Are these the main blockers to that:
> > 
> > * Different filters serialize their data in persistent storage differently.
> > * We don't store enough info in persistent storage to reconstruct the filter's metadata.
> 
> Yeah those are some of the blockers, each component really just follows its own beat right now.
> 
> There's an issue of unique namespace across pages as well. Even if we assume that every component has a unique name, that doesn't stop people from giving two completely different components on two separate tabs the same name. normally at worst this won't do anything but make the page look slightly misconfigured, but with regards to the 13.1 facets recut this can end up breaking the page.
> 
> In addition to that, we don't have a good idea of how persistent storage should work. We can't just throw pure json into the URL forever, since we have a limit of 2048 characters in ie11, and what exactly is supposed to go inside persistent storage. In my mind, persistent storage is for storing the current state of all the components on the page. Then, when the page reloads, we can reseed the components from that state. If this were react, we could quite literally store the state, but this very clearly is not react, and our component._state is way too bulky to store in the url, even if we find some way to compress things. We would have to refactor at least our filter components so that state really is just the minimum information to derive state from, not everything our handlebars templates felt like having inside of them.
> 
> So basically it's very frustrating but I don't think there's a good programmatic way to do this right now without ripping up lots and lots of code.

Agreed, we don't have a clear sense about the relationship between filters, facets, and persistent storage. We should make separate item(s) to address:

1. Figuring out how we can persist facets/filters correctly across page reload.
2. Standardizing how filters are serialized into persistent storage.
3. Investigating alternatives to just shoving everything in the URL.

src/ui/components/search/filtersearchcomponent.js
oshi97: Updated!

src/ui/components/filters/geolocationcomponent.js
oshi97: Still waiting on spec from PMs, have reminded them one time so far

src/core/core.js
oshi97: Updated to remove these

src/core/core.js
```diff
@@ -360,8 +362,12 @@ export default class Core {
...
+    this.filterRegistry.setFacetFilterNodes(availableFieldids, filterNodes);
+  }
+
+  setFilterNode (namespace, filterNode) {
```
tmeyer2115: Can we get rid of setFilter too?

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -81,6 +77,10 @@ export default class DateRangeFilterComponent extends Component {
     };
   }
 
+  static defaultTemplateName () {
+    return 'controls/date';
```
tmeyer2115, line 81: Weird this wasn't here before, nice catch!

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -202,30 +201,30 @@ export default class FilterBoxComponent extends Component {
...
-          this._saveFiltersToStorage();
+          this._saveFilterNodesToStorage();
           this._search();
         });
       }
     }
 
     // Initialize reset button
-    if (this.config.resetFilters) {
+    if (this.config.resetFilters && !this.config.searchOnChange) {
```
tmeyer2115: Was this conditional incorrect previously?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -262,18 +261,15 @@ export default class FilterBoxComponent extends Component {
...
-  _saveFiltersToStorage () {
-    const validFilters = this._filters.filter(f => f.getFilterKey());
+  _saveFilterNodesToStorage () {
+    const validFilterNodes = this._filterNodes.filter(fn => fn.getFilter().getFilterKey());
 
     if (this.config.isDynamic) {
       const availableFieldIds = this.config.filterConfigs.map(config => config.fieldId);
-      const combinedFilter = Facet.fromFilters(availableFieldIds, ...validFilters);
-      this.core.setFacetFilter(this.name, combinedFilter || {});
+      this.core.setFacetFilterNodes(availableFieldIds, validFilterNodes);
```
tmeyer2115, line 268: Do we need to do any kind of conversion from validFilterNodes to a Facet here or is that handled elsewhere?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -303,21 +301,46 @@ export default class FilterOptionsComponent extends Component {
     this._applyFilter();
   }
 
+  _buildFilter (option) {
+    return option.filter ? option.filter : Filter.equal(option.field, option.value);
+  }
+
+  _buildFilterMetadata (option) {
+    return new FilterMetadata({
+      fieldName: this.config.label,
```
tmeyer2115, line 311: Do we need fieldId here?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -303,21 +301,46 @@ export default class FilterOptionsComponent extends Component {
...
+      .map(o => FilterNodeFactory.from({
+        filter: this._buildFilter(o),
+        metadata: this._buildFilterMetadata(o)
+      }));
 
     this.core.persistentStorage.set(this.name, this.config.options.filter(o => o.selected).map(o => o.label));
-    return filters.length > 0
-      ? Filter.group(...filters)
-      : Filter.empty();
+    const fieldIdToFilterNodes = {};
```
tmeyer2115: Minor nit, but I think we could do this a little more functionally: https://gomakethings.com/a-vanilla-js-equivalent-of-lodashs-groupby-method/

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -108,17 +109,17 @@ export default class GeoLocationComponent extends Component {
...
     /**
-     * The filter to use for the current query
-     * @type {Filter}
+     * The filter node to use for the current query
+     * @type {FilterNode}
      */
-    this.filter = this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || {};
-    if (typeof this.filter === 'string') {
+    let filterNode = this.core.globalStorage.getState(`${this.name}.${StorageKeys.FILTER_NODE}`) || {};
+    if (typeof filterNodeData === 'string') {
```
tmeyer2115: Where is filterNodeData defined?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -108,17 +109,17 @@ export default class GeoLocationComponent extends Component {
...
 
     /**
-     * The filter to use for the current query
-     * @type {Filter}
+     * The filter node to use for the current query
+     * @type {FilterNode}
      */
-    this.filter = this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || {};
-    if (typeof this.filter === 'string') {
+    let filterNode = this.core.globalStorage.getState(`${this.name}.${StorageKeys.FILTER_NODE}`) || {};
```
tmeyer2115: Are these swapped?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -108,17 +109,17 @@ export default class GeoLocationComponent extends Component {
...
+    if (typeof filterNodeData === 'string') {
       try {
-        this.filter = JSON.parse(this.filter);
+        filterNode = JSON.parse(filterNode);
       } catch (e) {}
     }
 
-    this.core.globalStorage.on('update', `${StorageKeys.FILTER}.${this.name}`, f => { this.filter = f; });
```
tmeyer2115, line 121: Was this redundant?

tmeyer2115 (commented):
"A new storage key, FILTER_NODE, was added to allow backwards compatibility
with any components that had listeners on a StorageKeys.FILTER. These
components call both core.setFilter, as well as the newer core.setFilterNode.
For components that do not listen to changes to global storage, this
was not necessary."

I'm not sure I really follow this. Why do we still need things to call core.setFilter? Bear with me, I haven't made it all the way through this PR yet! It's a big boy. I still need to review the various tests, but I thought I'd submit the feedback I had so far.

src/ui/components/filters/daterangefiltercomponent.js
oshi97: Oh it was setting _templateName directly! But this felt better for setting up the unit tests

src/ui/components/filters/filterboxcomponent.js
oshi97: Yeah, caught by the unit tests!

src/core/core.js
oshi97: Sorry I should have updated the original commit message, we no longer have set filter or any of the hybrid system thing!

src/ui/components/filters/filterboxcomponent.js
oshi97: That's done in FilterRegistry.getRequestFacet! I think it's a little nicer to only do that processing at the very end step, so we can just deal with filternodes everywhere else.

src/ui/components/filters/filteroptionscomponent.js
oshi97: I took it out since after getting rid of FilterView it felt very strange to have that duplicated information across FilterMetadata and Filter.

src/ui/components/filters/filteroptionscomponent.js
oshi97: Yeah that makes a lot of sense! This identical logic is used when grouping the filters again in the results header, would it make sense to put it in a utils/arrayutils.js?

src/ui/components/filters/geolocationcomponent.js
oshi97: Nope this was on purpose! Since FilterRegistry looks at all keys in global storage that start with StorageKeys.FILTER_NODE, but we don't want this to affect filter registry's storage, we have to make sure the key doesn't start with StorageKeys.FILTER_NODE. I didn't swap the order for StorageKeys.QUERY because this isn't a problem for QUERY. I'll change it for QUERY as well since this seems confusing.

src/ui/components/filters/geolocationcomponent.js
oshi97: Yeah, I checked the rest of the code to make sure nothing outside of the geolocation component was trying to set its filter in global storage.

src/ui/components/filters/geolocationcomponent.js
oshi97: Shoot sorry I renamed it a few times

src/ui/components/filters/geolocationcomponent.js
oshi97: > "A new storage key, FILTER_NODE, was added to allow backwards compatibility
> with any components that had listeners on a StorageKeys.FILTER. These
> components call both core.setFilter, as well as the newer core.setFilterNode.
> For components that do not listen to changes to global storage, this
> was not necessary."
> 
> I'm not sure I really follow this. Why do we still need things to call core.setFilter? Bear with me, I haven't made it all the way through this PR yet! It's a big boy. I still need to review the various tests, but I thought I'd submit the feedback I had so far.

Let me know when you're done with this review! I'll push a commit with the changes after that. Thanks for looking at this I know it feels like a giant brick.

src/ui/components/filters/geolocationcomponent.js
oshi97: spec here! https://docs.google.com/document/d/1BOn6MLV4ExNp54zdtQS5IGPpwtVDqUqyWfiXKpJ1gdo/edit#heading=h.3s5krrq5fizs

.circleci/config.yml
```diff
@@ -62,6 +62,7 @@ jobs:
   unit_test:
     docker:
       - image: circleci/node:7.10
+    resource_class: large
```
tmeyer2115, line 65: Big boi

src/core/filters/filterregistry.js
```diff
@@ -30,15 +30,15 @@ export default class FilterRegistry {
    * @returns {Array<FilterNode>}
    */
   getFilterNodes () {
-    return this.globalStorage.getAll(StorageKeys.FILTER);
+    return this.globalStorage.getAll(StorageKeys.FILTER_NODE);
   }
 
   /**
    * Get all of the active {@link FilterNode}s used for facets.
```
tmeyer2115: Update docs, FacetFilterNode is it's own thing, right?

src/core/models/filter.js
```diff
@@ -70,26 +70,37 @@ export default class Filter {
...
+    const falsyMin = min === null || min === undefined || min === '';
+    const falsyMax = max === null || max === undefined || max === '';
+    if (falsyMin && falsyMax) {
+      return Filter.empty();
     }
-
-    const groupFilters = [];
-    for (const field of Object.keys(groups)) {
-      groupFilters.push(groups[field].length > 1 ? Filter.or(...groups[field]) : groups[field][0]);
+    if (falsyMax) {
```
tmeyer2115: else if?

src/core/models/filter.js
```diff
@@ -70,26 +70,37 @@ export default class Filter {
...
+      return isExclusive
+        ? Filter.greaterThan(field, min)
+        : Filter.greaterThanEqual(field, min);
     }
-
-    return groupFilters.length > 1 ? Filter.and(...groupFilters) : groupFilters[0];
+    if (falsyMin) {
```
tmeyer2115: Same

src/core/models/filter.js
```diff
@@ -70,26 +70,37 @@ export default class Filter {
   }
 
   /**
-   * OR filters with the same keys, then AND the resulting groups
-   * @param  {...Filter} filters The filters to group
+   * Helper method for creating a range filter
```
tmeyer2115, line 73: I think we should aim to refactor this class eventually. The vast majority of this class is devoted to static factory methods. I think those belong in a dedicated class(es)

src/core/utils/arrayutils.js
```diff
@@ -0,0 +1,18 @@
+
+/**
+ * Groups an array by a given key function.
+ * @param {string} url
+ * @param {params} Object
+ * @returns {string}
+ */
+export function groupArray (arr, keyFunc) {
```
tmeyer2115, line 8: What's this needed for? Is it because of our older Node version?

tests/acceptance/acceptancesuite.js
```diff
@@ -44,3 +45,62 @@ test('pagination flow', async t => {
   const pageNum = await paginationComponent.getActivePageLabelAndNumber();
   await t.expect(pageNum).eql('Page 2');
 });
+
+test.page`http://localhost:9999/tests/acceptance/fixtures/html/facets`(
```
tmeyer2115: Is this short-hand for creating a fixture with the given page, then specifying a test?

tests/acceptance/blocks/facetscomponent.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+export default class FacetsComponentBlock {
+  constructor () {
+    this._filterBox = new FilterBoxComponentBlock('.js-yxt-Facets');
+  }
+
+  /**
+   * Return the child filter box component's block.
+   */
+  async getFilterBox () {
```
tmeyer2115: Do all these methods need "async" qualifiers? I don't think this one does at least.

tests/acceptance/blocks/facetscomponent.js
```diff
@@ -0,0 +1,48 @@
...
+    return this._filterBox;
+  }
+
+  /**
+   * Toggle the given option for the given filter.
+   */
+  async toggleOption (filterIndex, optionIndex) {
```
tmeyer2115: I think we could make this more user-friendly by allowing someone to select a filter by name and an option by name.

tests/acceptance/blocks/filterboxcomponent.js
```diff
@@ -0,0 +1,88 @@
...
+    return Selector(`${this._container} ${selector}`);
+  }
+
+  /**
+   * Gets the child filter component of the given index.
+   * Currently assume all filter components are FilterOptions, since that is
+   * By far the primary use case
+   * @param {number} index
+   */
+  getFilter (index) {
```
tmeyer2115: Is this function "async"? It's being awaited above.

tests/acceptance/blocks/filterboxcomponent.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+export default class FilterBoxComponentBlock {
+  constructor (container) {
+    this._container = container || '';
+  }
+
+  /**
+   * Toggle the given option for the given filter.
+   */
+  async toggleOption (filterIndex, optionIndex) {
```
tmeyer2115: Would it make more sense for toggle to be an option on the FilterOptions block and the FilterBox block has a method to get a FilterOptions instance by name?

tests/acceptance/blocks/filteroptionscomponent.js
```diff
@@ -0,0 +1,75 @@
...
+  async isExpanded () {
+    return this._Selector('.yxt-FilterOptions--collapsed').exists;
+  }
+
+  /**
+   * Return a selector within this block's container context.
+   * @param {string} selector selector for desired element
+   */
+  _Selector (selector) {
+    return Selector(`${this._container} ${selector}`);
```
tmeyer2115: You can also create a private var this._selector = Selector(container) in the constructor and then do this._selector.find(selector) here. That will scope it too.

tests/acceptance/blocks/filteroptionscomponent.js
```diff
@@ -0,0 +1,75 @@
+import { Selector, t } from 'testcafe';
+
+/**
+ * This class models user interactions with the {@link FilterOptionsComponent}.
+ */
+export default class FilterOptionsComponentBlock {
+  constructor (container) {
+    this._container = container || '';
```
tmeyer2115: Do we want to supply a sane default for container here?

tests/acceptance/blocks/filterboxcomponent.js
```diff
@@ -0,0 +1,88 @@
...
+    filterOptions.toggleOption(optionIndex);
+  }
+
+  /**
+   * Get the number of results associated with a particular option.
+   * This value is enclosed within parenthesis at the end of an option's
+   * label.
+   * @param {number} index
+   */
+  async getOptionCount (filterIndex, optionIndex) {
```
tmeyer2115: I don't think we need this convenience method. It's mixing levels of abstraction. Filterbox can give you a FilterOptions and from FilterOptions you can get this count.

tests/ui/components/filters/daterangecomponent.js
```diff
@@ -0,0 +1,376 @@
...
+      mockCore,
+      DateRangeFilterComponent.defaultTemplateName()
+    );
+
+    defaultConfig = {
+      container: '#test-component'
+    };
+  });
+
+  it('renders correctly for default values', () => {
```
tmeyer2115, line 44: For this and the other Filter components, do we have full test coverage for the different display string scenarios?

src/core/utils/arrayutils.js
oshi97: You suggested this a month ago remember!!!!

oshi97: this one 
image

tests/acceptance/acceptancesuite.js
oshi97: yup! Had to look around a bit to find this again but https://devexpress.github.io/testcafe/documentation/guides/basic-guides/organize-tests.html

tests/acceptance/blocks/filterboxcomponent.js
oshi97: the name for a FilterOptions is `${this.name}.filter${i}` where this.name is the FilterBox name, and the container div for FilterOptions (in a FilterBox) is `.js-yext-filterbox-filter${i}`.
In other words both of which are derived from the index, not the name, of the FilterOptions. I think it would make the tests a good bit more readable if we could assign names to the FilterOptions, but it seems difficult with how FilterBox currently works

oshi97: Ah nevermind I think you probably just meant by things like the title headers or other text on the screen right? Not the literal this.name of the component. We can do that

src/core/filters/filterregistry.js
oshi97: hmm I looked at this a little more and calling one set of FilterNodes FacetFilterNodes and the other just FilterNodes seems pretty confusing so I renamed FILTER_NODE to STATIC_FILTER_NODE and the same with any similarly named methods. Also changed getRequestFacetFilter => getFacetFilterPayload and same for static filters

tests/ui/components/filters/daterangecomponent.js
oshi97: Added coverage for exclusive range for date range filter, geolocation, and filtersearch
I believe it should be 100% or very close now!

tests/acceptance/blocks/facetscomponent.js
oshi97: Yeah I wasn't very clear on how these Selector promises (?) were working (still a little foggy). My current understanding is declaring a selector doesn't need async but using any of its methods that inspect the dom does?

tests/acceptance/acceptancesuite.js
oshi97: think it makes more sense for my test to just match the format of the other ones, at least for now

oshi97:
rebased on top of nik's pagination update

package.json
```diff
@@ -71,6 +72,13 @@
     "setupFilesAfterEnv": [
       "./tests/setup/setup.js"
     ],
+    "moduleFileExtensions": [
+      "js"
+    ],
+    "moduleDirectories": [
+      "node_modules",
+      "<rootDir>"
+    ],
```
tmeyer2115, line 81: What are these two attributes used for?

src/ui/components/filters/rangefiltercomponent.js
```diff
@@ -77,19 +79,24 @@ export default class RangeFilterComponent extends Component {
...
     this._maxLabel = config.maxLabel || null;
+  }
 
-    /**
-     * The template to render
-     * @type {string}
-     * @private
-     */
-    this._templateName = `controls/range`;
+  getFirstValidValue (...values) {
```
tmeyer2115, line 84: Was the previous defaulting logic incorrect? If not, I'd say we should stick with what was there before. It's a bit more succinct. This method doesn't do too much.

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -89,27 +90,42 @@ export default class FilterSearchComponent extends Component {
      * Optionally provided
      * @type {string}
      */
-    this.query = config.query || this.core.globalStorage.getState(`${StorageKeys.QUERY}.${this.name}`) || '';
-    this.core.globalStorage.on('update', `${StorageKeys.QUERY}.${this.name}`, q => {
+    this.query = config.query || this.core.globalStorage.getState(`${this.name}.${StorageKeys.QUERY}`) || '';
+    this.core.globalStorage.on('update', `${this.name}.${StorageKeys.QUERY}`, q => {
```
tmeyer2115: What caused the order of these two keys to be switched? Was the existing order incorrect or buggy? We should avoid switching keys around whenever possible. If we change the keys we're expecting in the URL, it can break links someone may have saved.

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -89,27 +90,42 @@ export default class FilterSearchComponent extends Component {
...
+    if (typeof filter === 'string') {
       try {
-        this.filter = JSON.parse(this.filter);
+        filter = JSON.parse(filter);
       } catch (e) {}
     }
 
-    this.searchParameters = buildSearchParameters(config.searchParameters);
+    let filterNode = this.core.globalStorage.getState(`${this.name}.${StorageKeys.STATIC_FILTER_NODE}`) || {};
+    if (typeof filterNode === 'string') {
```
tmeyer2115: Is it possible that a filter node can be stringified in the global storage?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -89,27 +90,42 @@ export default class FilterSearchComponent extends Component {
...
+      try {
+        filterNode = JSON.parse(filterNode);
+      } catch (e) {}
+    }
 
-    this.core.globalStorage.on('update', `${StorageKeys.FILTER}.${this.name}`, f => { this.filter = f; });
+    if (filter) {
+      this.filterNode = FilterNodeFactory.from({
+        ...filterNode,
```
tmeyer2115: Is it possible that there could no filter node in global storage but a config.filter? Would that cause issues here?

tests/acceptance/blocks/filterboxcomponent.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Apply the filters in this filter box.
+   */
+  async applyFilters () {
+    const applyButton = await this._selector.find('.js-yext-filterbox-apply');
+    await t.click(applyButton);
+    // TODO more sophisticated waiting mechanism, docs suggest waiting for a specific DOM change
+    await t.wait(1000);
```
tmeyer2115: Let's add the improved waiting mechanism now. Waiting for a specific amount of time can lead to flaky/unpredictable tests.

tests/acceptance/blocks/filteroptionscomponent.js
```diff
@@ -0,0 +1,69 @@
...
+
+  async getLabel (label) {
+    return this._selector.find('label').withText(label);
+  }
+
+  /**
+   * Return a selector for the option with the given label.
+   * @param {String} label
+   */
+  async getOption (label) {
```
tmeyer2115: Minor nit, let's make this a private method (_getOption) since it's only used internally in this class. I think most consumers of this class will only want either the count or to toggle anyway.

tests/acceptance/pageobjects/facetspage.js
```diff
@@ -0,0 +1,38 @@
+import SearchComponentBlock from '../blocks/searchcomponent';
+import VerticalResultsComponentBlock from '../blocks/verticalresultscomponent';
+import FacetsComponentBlock from '../blocks/facetscomponent';
+
+/**
+ * A model of a universal search page, containing block representations
```
tmeyer2115: A model of a vertical search page with facets...

src/ui/components/search/filtersearchcomponent.js
oshi97: Yeah, when the page loads everything from persistent storage is dumped into global storage as a string

src/ui/components/filters/rangefiltercomponent.js
oshi97: Yeah, it doesn't work when the max is 0

package.json
oshi97: this lets you do something like 
`import FilterNodeFactory from 'src/core/filters/filternodefactory';`
instead of
`import FilterNodeFactory from '../../../../src/core/filters/filternodefactory';`
in the jest tests

src/ui/components/search/filtersearchcomponent.js
oshi97: Tried this and the filternode will be correctly created (seen through console logging), so there's no functional issues with this. I think I can remove setting this.filterNode in the constructor, though, because in order to store the filternode the autocomplete has to be submitted, and when that happens a new filternode is generated.

tests/acceptance/blocks/filteroptionscomponent.js
oshi97: done

tests/acceptance/pageobjects/facetspage.js
oshi97: done

src/ui/components/search/filtersearchcomponent.js
oshi97: FilterRegistry currently uses globalStorage for storing and retrieving filternodes. In order to get all current filternodes it calls this.globalStorage.getAll(StorageKeys.STATIC_FILTER_NODE), which gets all items stored in globalStorage that start with StorageKeys.STATIC_FILTER_NODE. This has the side-effect of not letting anything else store things in globalStorage that start with StorageKeys.STATIC_FILTER_NODE.

We can get around this by appending another prefix for things FIlterRegistry stores, or not using globalStorage. I think it still makes sense to use globalStorage, so I'll just append a prefix if that sounds okay.

package.json
oshi97: the default is just ["node_modules"]

oshi97:
rebased on top of the applied-filters branch, which was rebased on top of the new 1.4.0 (which is 0.14.0 rebased onto 1.3.0). whee

tests/acceptance/blocks/filterboxcomponent.js
oshi97: First tried to use RequestLogger, but apparently testcafe already has built-in waiting for xhr requests. Removed this t.wait and wasn't seeing the waiting issues I was having before, so just removed it.

package.json
tmeyer2115: Does this also work for the source code as well? That could be nice.

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -98,18 +99,17 @@ export default class FilterSearchComponent extends Component {
     /**
      * The filter string to use for the provided query
      * Optionally provided
+     * TODO(oshi): config.filter is not in the readme. Do we need this?
      * @type {string}
      */
-    this.filter = config.filter || this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || '';
-    if (typeof this.filter === 'string') {
+    this.filter = config.filter || '';
```
tmeyer2115: It seems like the only way the config.filter was used previously was to maybe drive the initial render (via it's usage in setState). But, looking at the Handlebars template of this component, the filter from setState is not actually used anywhere. So, it seems to me like the config.filter option was not actually hooked up to anything. It seems safe to remove, what do you think?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -102,25 +102,12 @@ export default class GeoLocationComponent extends Component {
...
-     * @type {FilterNode}
-     */
-    let filterNode = this.core.globalStorage.getState(`${this.name}.${StorageKeys.STATIC_FILTER_NODE}`) || {};
-    if (typeof filterNode === 'string') {
-      try {
-        filterNode = JSON.parse(filterNode);
-      } catch (e) {}
-    }
-
-    this.filterNode = FilterNodeFactory.from({ ...filterNode });
```
tmeyer2115: Was this just unused?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -209,7 +196,6 @@ export default class GeoLocationComponent extends Component {
       isFilterSearch: true,
       container: '.js-yxt-GeoLocationFilter-autocomplete',
       originalQuery: this.query,
-      originalFilter: this.filter,
```
tmeyer2115, line 207: This attribute is no longer supported by Autcomplete, right?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -278,9 +263,9 @@ export default class GeoLocationComponent extends Component {
    * @private
    */
   _saveDataToStorage (query, filter, displayValue, position) {
-    this.core.persistentStorage.set(`${this.name}.${StorageKeys.QUERY}`, query);
+    this.core.persistentStorage.set(`${StorageKeys.QUERY}.${this.name}`, query);
+    this.core.persistentStorage.set(`${StorageKeys.FILTER}.${this.name}`, filter);
```
tmeyer2115, line 267: Just wanna make sure I solidify my understanding. This component sets/modifies data in persistent and globalStorage. But, it never actually pulls data from either of them? Looking at the original GeoLocation component (before any of your changes), it looks like it would pull data from globalStorage to set this.filter. That seems unnecessary though since originalFilter isn't a thing for Autocomplete any more. So, the original component didn't actually need to pull from globalStorage at all, is that correct?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -128,7 +115,7 @@ export default class FilterSearchComponent extends Component {
   // TODO(oshi): SPR-1925 check that it is safe to remove this, it runs an extra search
   // For no obvious reasons
   onCreate () {
-    if (this.query && this.filter) {
+    if (this.query) {
```
tmeyer2115: For backwards compatibility, I think we need to add a new config option called "executeInitialSearch" or something like that. That option will have the following behavior:

this.executeInitialSearch = config.executeInitialSearch || config.filter

Not sure if "executeInitialSearch" is the best name. If we just check for this.query here, it could alter the behavior of existing sites. A client may have a config.query, but not config.filter, and is not expecting a network call to be issued when this component is created. We should update the README too.

src/ui/components/filters/geolocationcomponent.js
oshi97: Just checked again. It looks like this option was added in  #15.
When I set my local git back to #15 the only references to originalFilter are in geolocationcomponent.js, filterpicker.js, and filtersearchcomponent.js, where they are being put into the state, but not referenced anywhere in the .hbs . Going through the autocomplete.hbs history (https://github.com/yext/answers/commits/master/src/ui/templates/search/autocomplete.hbs.) and ctrl+f-ing 'originalFilter' I also can't find any references to it. 

Reading through the PR it doesn't look like there was any discussion over this option. My best guess is that these were put there in anticipation of somebody overriding the component and wanting to use the value? It may be safer to keep it, but I looked through the AEB and also did not find any reference to originalFilter. I think the only possible way removing this breaks something is for raw usages of the SDK.

src/ui/components/filters/geolocationcomponent.js
oshi97: Yeah, the only thing they needed to pull from globalStorage for was for setting `this.filter` in the constructor. Since we don't need that anymore it's safe to not pull from globalStorage. In the referenced code block we store the filter in the url to be backwards compatible and match previous behavior.

src/ui/components/search/filtersearchcomponent.js
oshi97: Sure that sgtm, updated!

src/ui/components/filters/geolocationcomponent.js
oshi97: yeah, I was trying to match the old behavior 1 for 1, but if originalFilter wasn't doing anything this wasn't doing anything

src/ui/components/search/filtersearchcomponent.js
oshi97: added it as searchOnLoad

package.json
oshi97: Ah this is only for jest tests I think we'd have to change how our grunting or gulping for that

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -85,6 +85,12 @@ export default class FilterSearchComponent extends Component {
      */
     this.redirectUrl = config.redirectUrl || null;
 
+    /**
+     * Whether to execute a search when the page is loaded
+     * @type {boolean}
+     */
+    this.searchOnLoad = config.searchOnLoad;
```
tmeyer2115: I think we need the defaulting behavior to be more complex. If they haven't set searchOnLoad explicitly, we will default to false unless they've set config.filter explicitly. In that case, searchOnLoad should be true. This will support the case where someone has config.query and config.filter set and is expecting a network call in onCreate.

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -188,15 +198,18 @@ export default class FilterSearchComponent extends Component {
...
+   * other filters to finish rendering before searching.
    */
   search () {
     if (this._storeOnChange) {
       return;
     }
-    this.core.verticalSearch(this._config.verticalKey, {
-      resetPagination: true,
-      useFacets: true
+    window.setTimeout(() => {
```
tmeyer2115, line 208: Why do we need setTimeout here? I'm not sure what the significance of letting other filters render first is

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -102,6 +96,21 @@ export default class FilterSearchComponent extends Component {
...
+     */
+    this.filter = config.filter || this.core.globalStorage.getState(`${StorageKeys.FILTER}.${this.name}`) || Filter.empty();
+    if (typeof this.filter === 'string') {
+      try {
+        this.filter = JSON.parse(this.filter);
+      } catch (e) {}
+    }
+
+    const filterNode = this._buildFilterNode(this.query, this.filter);
+    this.core.setStaticFilterNodes(this.name, filterNode);
```
tmeyer2115, line 112: FilterSearch didn't call this.core.setFilter in the constructor. Was that a bug? Someone could have specified a config.filter and it wouldn't have been added to globalStorage.

oshi97: Yes I believe this was a bug. core.setFilter was being called in the onCreate of the filter for a long time, but was removed in #196 . (see https://github.com/yext/answers/pull/196/commits/2df0026140e80f2ca4eacedada71597823bfe490) . This PR looks like it was done purely for refactoring purposes, and should not have changed functionality, but it looks like this was not the case. 

SIDE RANT: There is also reference to a this._saveQueryAndFilter that straight up does not exist...? (see https://github.com/yext/answers/commit/b39bbbc8779d1c33245788a92dfc02dce5f5a4d2#diff-af64d904dfb3640da19a4d461b50d322R204) Even going back as far as to when lines referencing this._saveQueryAndFilter were created the function _saveQueryAndFilter() does not exist so I am still quite confused as to what was happening here. Either way it probably doesn't concern this PR.

src/ui/components/search/filtersearchcomponent.js
oshi97: Even before this PR, FilterBox onMount would call its _saveFiltersToStorage(), which would call core.setFilter. I noticed during testing that the on page load search ran by FilterSearch was missing filters that the on page load search ran by the SearchBar had. I believe this was because the onCreate of FilterSearch was finishing before the onMount of FilterBox. If I move my FilterSearch's addComponent call anywhere after my FilterBox's addComponent call of the ANSWERS.init onReady() this issue no longer occurs.

oshi97: I think I remember you being the one who told me about this originally but https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful

tmeyer2115 (approved):

Issue #624, opened 2020-04-01
Answers SearchBar Borderless on Mobile


benmcginnis:
<img width="369" alt="Screen Shot 2020-03-31 at 11 19 15 PM" src="https://user-images.githubusercontent.com/1206953/78095871-17d7b780-73a6-11ea-9a9a-2655ea4d7d6a.png">

Caused by [this](https://github.com/yext/answers/blob/69a33d0628c560e13626d9ad9d7ad98bbeeae02b/src/ui/sass/modules/_SearchBar.scss#L42) line being in a media query

rosiegrant:
this is intentional styling from jeremy

benmcginnis:
I don’t think it’s fully thought through.  It doesn’t work on a white background. I think if we’re going to do more heavy customizations of the experience going forward, we should consider separating the default answers experience into a theme that’s separate from the base configuration of the components, which would be less tied to the styling of the default theme.

rosiegrant:
point taken ☺️ 

For whomever works on this, I think we can just remove the media query restriction here, 

https://github.com/yext/answers/blob/13ea6669234ed61d201f01f62326d01bb0aa6320/src/ui/sass/modules/_SearchBar.scss#L40

Here's what the HHs have been adding (for reference)
```
.yxt-SearchBar
{
  &-container
  {
    @include bplte(sm) {
      border-radius: .4rem;
      border: 1px solid $gray-3;
    }
  }
}
```

rosiegrant:
edit: Jeremy is going to look at this this week and will provide some recommendations for the default UX

benmcginnis:
We should definitely make sure he considers the search bar independently of
its context, in addition to the whole default layout.

rosiegrant:
@bryanreed let's move forward keeping the search border the same on desktop and mobile. thanks!

rosiegrant:
resolved as of 1.0.0

PR #686, opened 2020-04-15
Add border around the SearchBar component when viewing on Mobile.

tmeyer2115:
This PR ensures that the SearchBar component has a border when viewed on
Mobile. The border is the same as the one that already exists on the Desktop
view. Without a border, the SearchBar's input was impossible to find against
a white background on Mobile. This resolves issue #624 .

J=SPR-2001
TEST=manual

Verified that the border appears the same across Desktop and Mobile in both
IE11 and Chrome.


nbramblett (approved):
LGTM

PR #688, opened 2020-04-15
Pagination styling

nbramblett:
Updates pagination styling to match the mocks as best I can.

Also adds a vertical results page to the new acceptance tests, and has one little test for pagination.


src/ui/sass/modules/_Pagination.scss
```diff
@@ -1,12 +1,13 @@
 /** @define Pagination */
 .yxt-Pagination
 {
-  $pagination-text-color: $color-text-primary !default;
-  $pagination-text-color-hover: $color-text-secondary !default;
+  $pagination-text-color: #616161 !default;
```
alexisgrow: Do we have a variable for this color? If not, could you add one in the colors.scss file?

src/ui/sass/modules/_Pagination.scss
```diff
@@ -1,12 +1,13 @@
...
+  $pagination-text-color-hover: #FFFFFF;
```
alexisgrow: You can use $color-brand-white here. Sometimes, brands use a white that is slightly off-white, using the variable makes it easily configurable by implementers.

src/ui/sass/modules/_Pagination.scss
```diff
@@ -24,7 +25,10 @@
     display: flex;
     background: none;
     border: none;
-    padding: 0 4px;
+    color: $pagination-text-color;
+    padding: 6px 7px;
```
alexisgrow: Rather than roughly equal padding, using `justify-content: center;` is preferred!

src/ui/sass/modules/_Pagination.scss
```diff
@@ -42,20 +47,19 @@
   }
 
   &-page, &-link {
-    padding: 0;
-    margin: 0 20px;
+    padding: 2px 8px;
+    margin: 0 2px;
```
alexisgrow: It looks like `margin: 0 2px;` is being added to a bunch of these, is it possible to add it once on their container instead? This makes it easier for an implementer if they want to adjust this margin - they'd only have to do it in one place.

src/ui/sass/modules/_Pagination.scss
```diff
@@ -33,6 +37,7 @@
       &:hover
       {
         color: $pagination-text-color-hover;
+        background-color: $pagination-color-hover;
```
alexisgrow: nit: It would be nice if this was called `pagination-background-color-hover`; I'm not sure if the spec provides any input here but it could be a little ambiguous since this is not meant for the pagination's CSS `color` property

src/ui/sass/modules/_Pagination.scss
```diff
@@ -24,7 +25,10 @@
     display: flex;
     background: none;
     border: none;
-    padding: 0 4px;
+    color: $pagination-text-color;
+    padding: 6px 7px;
+    margin: 0 2px;
```
alexisgrow: If this is not meant to override/reset vertical margin to 0, can it be `margin-left: 2px, margin-right: 2px`?

src/ui/sass/modules/_Pagination.scss
```diff
@@ -67,6 +71,11 @@
   {
     margin-right: -4px;
   }
+
+  &--wide {
+    padding: 6px 3.5px;
```
alexisgrow: Not 100% sure without looking at the page, but it seems like this is centering content. If so, using something like `display: flex; justify-content: center; align-items: center` is preferred.

alexisgrow (commented):
looking good overall! Some CSS comments 🖌️🎨

tests/acceptance/acceptancesuite.js
```diff
@@ -37,3 +38,27 @@ test('Basic universal flow', async t => {
...
+  .page`http://localhost:9999/tests/acceptance/fixtures/html/vertical`;
+
+test('pagination flow', async t => {
+  const searchComponent = VerticalPage.getSearchComponent();
+  await searchComponent.enterQuery('Virginia');
+  await searchComponent.submitQuery();
+  const paginationComponent = VerticalPage.getPaginationComponent();
+  await paginationComponent.clickNextPage();
+  const pageNum = await paginationComponent.getCurrentPage();
+  // t.wait(30000);
```
tmeyer2115: Remove the comment.

tests/acceptance/acceptancesuite.js
```diff
@@ -37,3 +38,27 @@ test('Basic universal flow', async t => {
...
   await t.expect(faqsSectionTitle).eql('FAQS');
 });
+
+fixture`Vertical search page works as expected`
+  .before(async ctx => {
+    const server = http.createServer((request, response) => {
+      return handler(request, response);
+    });
+    server.listen(9999);
+    ctx.server = server;
```
tmeyer2115: Let's factor this out into a shared method that both fixtures can use.

tests/acceptance/blocks/paginationcomponent.js
```diff
@@ -0,0 +1,23 @@
+import { Selector, t } from 'testcafe';
+
+export default class PaginationComponentBlock {
```
tmeyer2115, line 6: JSDoc

tests/acceptance/blocks/paginationcomponent.js
```diff
@@ -0,0 +1,23 @@
...
+  constructor () {
+    this._currentPage = Selector('span.yxt-Pagination-page');
+    this._buttons = Selector('.yxt-Pagination-button');
+    this._links = Selector('.yxt-Pagination-link');
+  }
+
+  /**
+   * Find the next-page button and click it
+   */
+  async clickNextPage () {
```
tmeyer2115: I think it could be good to suffix the method names in here with OfResults. In the context of TestCafe, methods like getCurrentPage() or getNextPage() could be understood to be something else, potentially something that returns a different Page Object.

src/ui/sass/modules/_Pagination.scss
nbramblett: The variable names at the top use "color" for background color and "text color" for color, so I figured it would be best to follow that convention.

src/ui/sass/modules/_Pagination.scss
nbramblett: The annoying issue with resolving this is that the PMs want the page items to all have a nice square-shaped highlighting on hover (which is why the weird padding) but they also are basing the spacing between items based on their un-hovered appearance, which means I can't use a container span without screwing up the margins. CSS is not my favorite 😛but I'll do my best to clean up the classes here with some fixed dimensions.

tests/acceptance/fixtures/html/vertical.html
```diff
@@ -0,0 +1,46 @@
...
+            <div class="results-container"></div>
+            <div class="pagination-container"></div>
+        </div>
+    
+        <script>
+            ANSWERS.init({
+                apiKey: 'df4b24f4075800e5e9705090c54c6c13',
+                experienceKey: 'rosetest',
+                businessId: '2287528',
+                experienceVersion: 'PRODUCTION',
```
tmeyer2115, line 22: add templateBundle: TemplateBundle.default to this config as well. I forgot to add that for the universal page and that's what bit me. Having the <script> tag for the compiled HBS templates isn't enough. Importing that tag makes an object called TemplateBundle.default available in the window. But, we still need to tell the SDK to actually use that as its templateBundle.

tests/acceptance/pageobjects/verticalpage.js
```diff
@@ -0,0 +1,38 @@
+import SearchComponentBlock from '../blocks/searchcomponent';
+import VerticalResultsComponentBlock from '../blocks/verticalresultscomponent';
+import PaginationComponentBlock from '../blocks/paginationcomponent';
+
+/**
+ * A model of a universal search page, containing block representations
```
tmeyer2115: Update JSDoc

tests/acceptance/pageobjects/verticalpage.js
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+     * Returns the {@link VerticalResultsComponentBlock} on the page.
+     */
+  getUniversalResultsComponent () {
+    return this._verticalResultsComponent;
+  }
+
+  /**
+   * Returns the {@link} on the page.
```
tmeyer2115: Update the link here

src/ui/sass/modules/_Pagination.scss
nbramblett: Going to leave this as unresolved in case people have further input on the relevant changes.

alexisgrow: Ah gotcha, so this is to make the padding a specific way for background color! This is good then!

src/ui/sass/modules/_Pagination.scss
```diff
@@ -19,43 +20,36 @@
...
   }
 
   &-page, &-link {
-    padding: 0;
-    margin: 0 20px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 24px;
+    width: 24px;
```
alexisgrow: Does this overflow for really long page numbers (e.g. 10000)?

src/ui/sass/modules/_Pagination.scss
```diff
@@ -19,43 +20,36 @@
...
+    margin-right: 2px;
+    margin-left: 2px;
+    border-radius: 2px;
+  }
+
+  &-icon {
+    font-size: $font-size-sm;
+    span {
+      padding-top: 3px; // The icon SVGs are not aligned nicely.
```
alexisgrow: Use flex for this!

nbramblett: It's already flexed :( the problem is that the svg has some additional space at the bottom so the flex container thinks it's aligned but visually it's imbalanced. It might be worth updating the icon svg but idk if it's used elsewhere, since that would suddenly look bad if I change it for here.

EDIT: I misread the svg component, it wasn't the icon but the outer styling, which is easily flexable. My bad. Thanks!

src/ui/sass/modules/_Pagination.scss
nbramblett: I don't know if the PMs really considered that case lol. I'll set it to min width since I don't think there's any clean fix.

tests/acceptance/blocks/paginationcomponent.js
```diff
@@ -1,23 +1,25 @@
...
+    this._currentPage = Selector('#active-page');
     this._links = Selector('.yxt-Pagination-link');
   }
 
   /**
    * Find the next-page button and click it
    */
-  async clickNextPage () {
-    await t.click(this._buttons.filter('.js-yxt-Pagination-next'));
+  async clickNextPageButton () {
```
tmeyer2115: Maybe something like getNextPageOfResults()? It's definitely a minor nit, but I think it's good to avoid using DOM-specific terms (like click or button) in the names of methods exposed by a block or page object. A consumer of these methods doesn't need to have any idea about what the actual DOM interaction is to get the next set of results.

tests/acceptance/blocks/paginationcomponent.js
```diff
@@ -1,23 +1,25 @@
...
+    await t.click(this._links.filter('.js-yxt-Pagination-next'));
   }
 
   /**
    * Find the current page
    */
-  async getCurrentPage () {
+  async getCurrentPageOfResults () {
```
tmeyer2115: getPageNumberOfCurrentResults()? This makes it really clear that we're only returning the page number, not an actual list of results.

tests/acceptance/blocks/paginationcomponent.js
```diff
@@ -1,23 +1,25 @@
...
    * Find the next-page button and click it
    */
-  async clickNextPage () {
-    await t.click(this._buttons.filter('.js-yxt-Pagination-next'));
+  async clickNextPageButton () {
+    await t.click(this._links.filter('.js-yxt-Pagination-next'));
   }
 
   /**
    * Find the current page
```
tmeyer2115: Get the current page number?

tests/acceptance/server.js
```diff
@@ -0,0 +1,10 @@
+import http from 'http';
+import handler from 'serve-handler';
+
+export default async function setupServer (ctx) {
```
tmeyer2115: Thinking about things a bit more, we might want to move the logic to close the server in here too. Even though it's just one line now, we might eventually want more complex, shared logic for that. We should add some JSDoc for both methods too.

src/ui/templates/results/pagination.hbs
```diff
@@ -1,8 +1,7 @@
 {{#if showControls}}
   <div class="yxt-Pagination">
     {{#if firstPageButtonEnabled}}
-      <button class="yxt-Pagination-button yxt-Pagination--wide js-yxt-Pagination-first {{#unless showFirstPageButton}}yxt-Pagination--hidden{{/unless}}"
-        {{#unless showFirstPageButton}}disabled{{/unless}}
+      <a class="yxt-Pagination-link yxt-Pagination-icon js-yxt-Pagination-first {{#unless showFirstPageButton}}yxt-Pagination--hidden{{/unless}}"
```
tmeyer2115, line 4: So, the SDK has a much more strict concept of what a "breaking change" is. A breaking change includes altering the DOM structure or CSS class names in such a way that a client's existing, custom CSS could break. Here, for example, we might have a customer who has written custom CSS for the yxt-Pagination-button class. They may have even written styling for 

.yxt-Pagination: {
  button: {...}
}

When upgrading to 14, they'd have to update this styling. It seems minor, but we've fielded a number of complaints about this in the past. Now, we try and avoid such breaking changes unless there is an absolutely pressing need for it. 😞

tests/acceptance/blocks/paginationcomponent.js
nbramblett: Hm, I'm not sure I agree with that. I would expect a method prefixed with "get" to return the next page explicitly. I'm not a huge fan of abstraction in testing, I'd rather know immediately what actual actions were being carried out.

tests/acceptance/blocks/paginationcomponent.js
nbramblett: Tried to make it as explicit as possible.

src/ui/templates/results/pagination.hbs
nbramblett: This was explicitly requested by rose & max so I'd assume they're ok with it.

https://yext.slack.com/archives/GREFVUD7V/p1587039824121800

tmeyer2115: I posted in the Slack thread. I want to make sure they're aware that this has the potential to break a user's existing CSS. Similar requests have been made in the past, not fully realizing the potential impact, and we've had to revert.

tests/acceptance/blocks/paginationcomponent.js
tmeyer2115: Terms like "clickNextPage" can be a bit loaded in Selenium and TestCafe. Usually, a method named like that would model navigation to some other Page Object/Model. This isn't a big deal, so feel free to "ack"

nbramblett: I could change it to clickNextButton so that it doesn't imply navigation as strongly?

tmeyer2115: Sounds good! Once that's done, LGTM. I'd have Alexis take another pass though, she's our resident expert on CSS.

src/ui/templates/results/pagination.hbs
nbramblett: Going to resolve this since Rose seems to be ok with it.

alexisgrow (approved):
LGTM!

PR #690, opened 2020-04-16
VerticalResults: Hide applied filters when displaying no results

oshi97:
When displayAllResults is true, before a page would display
applied filters in the results header, like it does for normal
results.

This PR sets showAppliedFilters to false when displaying no results, which will hide any applied filters in the header (so nlp filters, static/facet filters, and sort options)

J=SPR-2003
TEST=manual
Checked that page was displaying applied filters for no results
Made change
checked that filters no longer display for no results
checked that filters still display for regular results


nbramblett:
Shouldn't this kind of behavior be configurable in the noResults object?

oshi97:
> Shouldn't this kind of behavior be configurable in the noResults object?

hmm idk asking pms

oshi97:
Rose said to just do this for now, in the future we can use allResultsForVertical.appliedQueryFilters

oshi97:
changed base to 13.3

nbramblett (approved):
Seems fine to me, probably want someone else to approve too though

tmeyer2115:
> When displayAllResults is true, before a page would display
> applied filters in the results header, like it does for normal
> results.
> 
> It's still the case that only nlp filters are displayed.
> 
> J=SPR-2003
> TEST=manual
> Checked that page was displaying applied filters for no results
> Made change
> checked that filters no longer display for no results
> checked that filters still display for regular results

@oshi97 are the NLP filters displayed in this no results case? If so, I don't think that's something we want shown either.

oshi97:
> > When displayAllResults is true, before a page would display
> > applied filters in the results header, like it does for normal
> > results.
> > It's still the case that only nlp filters are displayed.
> > J=SPR-2003
> > TEST=manual
> > Checked that page was displaying applied filters for no results
> > Made change
> > checked that filters no longer display for no results
> > checked that filters still display for regular results
> 
> @oshi97 are the NLP filters displayed in this no results case? If so, I don't think that's something we want shown either.

Sorry I reread my commit message and it was really unclear, this commit stops NLP filters from being displayed as well.

tmeyer2115 (approved):

PR #691, opened 2020-04-16
Map: add noResults configuration

nbramblett:
Adds a noResults configuration which alters the behavior of the map component when a vertical search returns no results.

J=SPR-2002
TEST=manual

Verified that inputting a no-results query while displayAllResults was set to false would cause the
map to be empty (and hidden unless otherwise set to display an empty map).


src/ui/components/map/mapcomponent.js
```diff
@@ -71,6 +77,10 @@ export default class MapComponent extends Component {
       return this;
     }
 
+    if (data.resultsContext !== ResultsContext.NORMAL && !this._noResults.displayAllResults) {
```
oshi97: I think this might break if this._noResults is undefined? Besides that lgtm!

nbramblett: Hm I assumed that the storage would have a defined fallback but maybe that's wrong. I'll add a catch for that.

oshi97 (approved):
lgtm!

Issue #693, opened 2020-04-16
Breaking Change - Link Targets


rosiegrant:
1. For title link targets in standard and legacy card, update so setting a target the `parent_ops` level works, since admins were relying on this functionality (here's what it looked like, https://github.com/yext/answers/blob/v0.12.0/src/ui/templates/results/resultsitem.hbs#L34). 

- https://github.com/yext/answers/blob/a641837a09380bcb0cd338258ae243f3e58d8be8/src/ui/templates/cards/legacy.hbs#L34

- https://github.com/yext/answers/blob/master/src/ui/templates/cards/standard.hbs#L38

We should first check if there's a target set at the card level, use it, if not, then check if there's a target at the parentOps level. 

2. In addition, any place we are setting a link target (title and CTAs), it should default to "_self". we are currently defaulting to "_blank" and this has caused a breaking change. Since "_self" is the default target, we can just remove these defaults and it should work. 

- https://github.com/yext/answers/blob/a641837a09380bcb0cd338258ae243f3e58d8be8/src/ui/templates/cards/legacy.hbs#L34
- https://github.com/yext/answers/blob/master/src/ui/templates/results/resultsaccordion.hbs#L38

(Places where we aren't defaulting target so we don't need to update:
- https://github.com/yext/answers/blob/master/src/ui/templates/ctas/cta.hbs#L3
- https://github.com/yext/answers/blob/master/src/ui/templates/cards/standard.hbs#L38)

rosiegrant:
@nbramblett @bryanreed I mixed these up, I'm sorry. 

Part 1 is correct. 

Part 2:

Any place we are setting a link target (title and CTAs), **it should default to "_blank"**. in certain places, we used to default to "_blank", and by removing it, it caused a breaking change. Using _self causes it to open in the current frame, which is not 

- [ ] https://github.com/yext/answers/blob/master/src/ui/templates/results/resultsaccordion.hbs#L38 can be left as-is, since we are already defaulting to "_blank"
- [ ] https://github.com/yext/answers/blob/master/src/ui/templates/cards/legacy.hbs#L34 can be left as-is, since we are already defaulting to "_blank"
- [ ] https://github.com/yext/answers/blob/master/src/ui/templates/ctas/cta.hbs#L3 should default to _blank
- [ ] https://github.com/yext/answers/blob/master/src/ui/templates/cards/standard.hbs#L38 should default to _blank. 

Again really sorry for the mix-up.

[...]

PR #694, opened 2020-04-16
Cards: Add parentOpts as a target, ensure _blank default

nbramblett:
_config._parentOpts.target is a valid fallback option for standard and
legacy cards now, and the places where the target attribute defaulted
to _self have been updated to default to _blank.

This is in response to issue #693 

J=SPR-2025
TEST=manual

Manually verified that setting target in the VerticalResults config
would be reflected in the card links. Also verified that not setting a
target would default to _blank.


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -203,6 +203,7 @@ export default class VerticalResultsComponent extends Component {
         isUniversal: this._config.isUniversal,
         template: this._config.itemTemplate,
         render: this._config.renderItem,
+        target: this._config.target,
```
alexisgrow: Can this be above `...this._config.card,` so that if there is a target set at the card level, it is used?

src/ui/templates/cards/standard.hbs
```diff
@@ -35,7 +35,7 @@
   {{#if _config.title}}
     {{#if _config.url}}
       <a class="yxt-StandardCard-titleLink js-yxt-StandardCard-title"
-        href="{{_config.url}}" target="{{_config.target}}"
+        href="{{_config.url}}" {{#if _config.target}}target="{{_config.target}}"{{else if _config._parentOpts.target}}target="{{_config._parentOpts.target}}"{{/if}}
```
alexisgrow: Does this have a fallback? It looks like the fallback was removed from the JS in `standardcardcomponent.js`

nbramblett: Leaving it blank defaults to _self. With the updates to the issue, this now has a fallback to _blank.

src/ui/components/results/verticalresultscomponent.js
nbramblett: I didn't know that's how it works. Done.

src/ui/templates/cards/standard.hbs
```diff
@@ -35,7 +35,7 @@
   {{#if _config.title}}
     {{#if _config.url}}
       <a class="yxt-StandardCard-titleLink js-yxt-StandardCard-title"
-        href="{{_config.url}}" target="{{_config.target}}"
+        href="{{_config.url}}" target="{{#if _config.target}}{{_config.target}}{{else if _config._parentOpts.target}}{{_config._parentOpts.target}}{{else}}_blank{{/if}}"
```
oshi97, line 38: I think parentOpts here might not do anything! The component hierarchy goes vertical results creates -> card components which each create one -> standard card/accordion card. The line you added in vertical results should do the trick!

src/ui/templates/results/resultsaccordion.hbs
```diff
@@ -35,7 +35,7 @@
                         {{#if this.url}}
                           <a class="yxt-AccordionResult-cta"
                              href="{{this.url}}"
-                             target="{{#if this.target}}{{this.target}}{{else}}_blank{{/if}}"
```
oshi97: Just wanted to check that we don't want any default value here? If you open a universal results page without setting a target I think this might not end up as _blank?

oshi97: I did some investigation into where {{#each callsToAction}} comes from and it looks like it's driven by the response (core/models/result.js) so I think we still need this!

src/ui/templates/cards/standard.hbs
nbramblett: Yeah what the admins actually want was a bit unclear but I believe this should also handle the case where they mess with the Card Component somehow.

src/ui/templates/results/resultsaccordion.hbs
nbramblett: resultsAccordion instantiates CTA components, which automatically fall back to _blank now if target is undefined. I tested this and it worked fine.

oshi97: These aren't cta components though

oshi97: these are just regular anchor tags (!)

src/ui/templates/cards/standard.hbs
oshi97: ok!

src/ui/templates/results/resultsaccordion.hbs
nbramblett: I believe does add CTA components as children actually, and then iterates through them on the #each instead of mounting them separately. I went back to double-check and it does default to blank as-is.

oshi97: Are you sure you're using the AccordionResults and not AccordionCard?

oshi97: It's very confusing but they are completely separate things

nbramblett: ...well, that would probably be the reason it works for me lol. My bad, I'll revert this one.

oshi97 (approved):
lgtm!

PR #702, opened 2020-04-20
Update deploy_version job to additionally also deploy $MAJOR_VERSION

oshi97:
When a commit is made, circle ci runs the following jobs

on non-master branches:
build
unit_test
deploy_branch

on master:
build
unit_test
hold
deploy_latest (deploys to https://assets.sitescdn.net/answers/)

on tags starting with "v":
build
unit_test
hold
deploy_version
    (deploys to https://assets.sitescdn.net/answers/$MINOR_VERSION, i.e. v0.13)

This commit changes the deploy_version job to also deploy to $MAJOR_VERSION, i.e. for v0.13.3
https://assets.sitescdn.net/answers/v0/

TEST=manual
set $CIRCLE_TAG in local bash and tested that 
`echo "export MAJOR_VERSION="$(echo "${CIRCLE_TAG}" | cut -d '.' -f 1)""`
printed out
`export MAJOR_VERSION=1`


nbramblett (approved):
Shell script is a pain to read/debug, but as far as I can tell this is correct.

nbramblett:
Could you quickly verify that v0.x is correctly tagged as well? (just to confirm that minor version is still ok)

oshi97:
> Could you quickly verify that v0.x is correctly tagged as well? (just to confirm that minor version is still ok)

minor version should still work, since it hasn't been touched and has been working (see https://assets.sitescdn.net/answers/v0.13/answers.min.js)

oshi97:
holding off on merging until we decide exactly what we want the cdn behavior to be

tmeyer2115 (commented):
I think we should confirm desired behavior with @bryanreed. To confirm, with this change we have folders for every major version and one for every ${MAJOR}.${MINOR} permutation? So we'll have branches like: /v0/ and /v0.13/? I think it would be more straightforward to have directories structured like:

${MAJOR}/${MINOR}

So we'd have /v0/, /v0/12, /v0/13, etc. If you pointed only to /v0/answers.js, you'd get the answers.js for the most recent minor version. Additionally, we'll need to consider patches/hot-fixes here. I think that would look like:

${MAJOR}/${MINOR}/${HOTFIX}

If you only specified ${MAJOR}/${MINOR} in your CDN URL, you'd get the resources from that minor version's most recent patch. Were we maintaining a flat directory structure (top-level folders for /v0/ and /v0.13/) for backwards compatibility reasons? Let's chat with Bryan and make sure we're all on the same page about what the CDN folders should look like.

tmeyer2115 (approved):

PR #704, opened 2020-04-20
Make Event Polyfill match previous non-polyfilled behavior

oshi97:
#668 changed dom/dom.js to always use our custom event polyfill, instead of first trying to use the event constructor if it exists.

However, this polyfill did not match the behavior of the event
constructor it was replacing, and had cancelable and bubbles set
to false by default instead of true. This issue has existed since before v0.13,
but was likely not discovered because it would only appear when using the polyfill
(which before was only ie11).

TEST=manual
Tested on ie11, firefox, safari, chrome.
Test that typing in search bar and submitting a search still works.
Check that submitting a near me query, then resubmitting another query
does not prompt for the geolocation a second time.


src/ui/dom/dom.js
```diff
@@ -173,9 +173,14 @@ export default class DOM {
...
-    var evt = document.createEvent('CustomEvent');
-    evt.initCustomEvent(event, settings.bubbles, settings.cancelable, settings.detail);
+    const _settings = {
+      bubbles: true,
+      cancelable: true,
+      detail: null,
+      ...settings
+    };
+    const evt = document.createEvent('Event');
+    evt.initEvent(event, _settings.bubbles, _settings.cancelable, _settings.detail);
```
tmeyer2115: I think we still want to continue using createEvent('CustomEvent') and evt.initCustomEvent:

https://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-event-constructor-doesnt-work

initEvent is deprecated.

oshi97: updated and retested! initCustomEvent is also deprecated but seems like it's a good deal newer

tmeyer2115 (commented):
Why was the DS_Store file committed? Other than that, LGTM! Can you quickly do a smoke test of some of the other component besides the SearchBar in Chrome and IE11? I want to make sure we didn't inadvertently break something by changing these defaults.

oshi97:
Tested on ie11 and chrome for multiple facets, resetting facets, navigation tabs on universal, clicking and selecting autocomplete options, clearing the searchbar

tmeyer2115 (approved):

PR #708, opened 2020-04-21
update universal results borders

oshi97:
margin on top and bottom of yxt-Results-filters
was not receiving border-left and border-right
(because margin does not receive border in css)
causing gaps to appear

TEST=manual
check that vertical results on vertical search,
vertical results on universal searchh, accordion results
do not have gaps in the border, and otherwise appearance is unchanged


oshi97:
I'm slightly worried about this breaking somebody's page layout, but the only other alternative I could see was removing all of the border-left and border-right for everything in vertical results when it is in a universal result and adding a border around the container. This might be safer though?

src/ui/sass/modules/_Results.scss
```diff
@@ -86,8 +85,11 @@
   &-filters
   {
     display: flex;
-    padding: $base-spacing/2 $base-spacing;
+    margin-top: 0;
+    margin-bottom: 0;
+    padding: $base-spacing * 1.5 $base-spacing;
     background-color: $results-filters-background;
+    border-top: $results-border;
```
alexisgrow, line 92: Does this need to be removed from titleBar and added here? If possible, it would be better to leave below the title I think

oshi97: Added a second commit showing the other possible solution for the double borders (which gives yxt-Results-filters border-bottom instead of giving it a border-top). 

Prior to this commit yxt-Results-filters never has a border-top or border-bottom, so if yxt-Results-filters is missing borders are going to double up. I can't think of something better than just giving it either border-bottom or border-top and moving other borders around?

tmeyer2115 (approved):

src/ui/sass/modules/_Results.scss
oshi97: I think I like the first commit more, if only because it keeps the 
```
  & .yxt-Accordion-list	
  {	
    border-top: $border-legacy;	
  }
```
they asked me to add ._. and it's slightly shorter

PR #709, opened 2020-04-21
Add global noResults configuration option.

tmeyer2115:
This new option can be set in the top-level Answers.init configuration. It can
be used by both VerticalResults and the MapComponent. Both of these components
have logic that will favor an individual noResults config to the one in
globalStorage. Additionally, both components will apply the same sane defaults
to any noResults attributes not specified. A change was also made to ensure
that the VerticalResults component passes along its noResults config to any
child MapComponent, to ensure consistency.

Note that Rose requested the displayAllResults default be false for the
MapComponent

TEST=manual

Ensured correct behavior in all the following situations:

VerticalResults with included MapComponent

- Specified noResults config for VerticalResults.
- Specified noResults config at global level.
- Specified noResults config at both levels.
- Specified noResults config at neither level.

VerticalResults and separate MapComponent

- Specified noResults config for both components.
- Specified noResults config at global level.
- Specified noResults config at global and component levels.
- Specified noResults config at neither levels.


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -163,7 +178,8 @@ export default class VerticalResultsComponent extends Component {
       placeholders: new Array(this._config.maxNumberOfColumns + 1),
       numColumns: Math.min(this._config.maxNumberOfColumns, this.results.length),
       showResultsHeader: showResultsHeader,
-      useLegacyNoResults: this._config.isUniversal || !this._config.noResults
+      useLegacyNoResults: this._config.isUniversal 
+        || !(this._noResultsConfig.template || this._noResultsConfig.displayAllResults)
```
alexisgrow: If a user just enters 
`noResults: {}` or `noResults: true`
in the config for noResults, they would not get the new noResults, is that correct? I think either of those examples should give the user the new noResults.

alexisgrow (approved):
LGTM!

PR #710, opened 2020-04-21
Change context to be positive equality rather than negative

nbramblett:
UniversalResults is dumb and this accommodates the dumbness.

TEST=manual

UniversalResults shows a map now. Same difference for Vertical Results.


oshi97 (approved):
lgtm!

PR #711, opened 2020-04-21
Revert padding change to yxt-Results-filters

oshi97:
This change caused an undesired breaking styling change to the
dignity funerals site. The margin-top: 0 and margin-bottom: 0
can be kept, since that was done to override user-agent styling.

TEST=manual
check that padding has same value as before


oshi97:
I talked with Alexis a bit and think it's safe to say that all consulting sites would override `<ul> `styling to have margin: 0? If that's not the case the margin-top:0 and margin-bottom:0 can also be reverted.

alexisgrow (approved):

PR #713, opened 2020-04-22
Searchable Facets: Do not set state on option click

alexisgrow:
This fixes a bug where the options were re-rendered on click. When
that happened, the showMoreLimit would be reset, so if a user had
all options expanded, then clicked an option, it would collapse all
options. The re-render also caused a slight flash on click.

Options will still be re-rendered on click when the storeOnChange boolean
is true.

TEST=manual

Test by using the Facets component on my local machine and selecting
options. Note that they don't re-render on click. Clicking Apply correctly
applies the selected options. Test that options can be initialized properly
through config. Test with the `expanded: true` and with `expanded: false`
since there is branching logic in the template. Also test with `showReset: true`
and `showReset: false`. Confirmed that the reset button still appears when
greater than zero options are selected and the expand/collapse behavior is 
as expected.

### DIFF
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -342,12 +342,8 @@ export default class FilterOptionsComponent extends Component {
    * @returns {number}
    * @private
    */
-  _toggleResetIfPresent () {
+  _toggleReset () {
     const resetEl = DOM.query(this._container, '.js-yxt-FilterOptions-reset');
-    if (!resetEl) {
-      return;
-    }
-
     const selectedCount = this._getSelectedCount();
     if (selectedCount > 0) {
       resetEl.classList.remove('js-hidden');
@@ -447,7 +443,9 @@ export default class FilterOptionsComponent extends Component {
   }
 
   _updateOption (index, selected) {
-    this._toggleResetIfPresent();
+    if (this.config.showReset) {
+      this._toggleReset();
+    }
 
     if (this.config.control === 'singleoption') {
       this.config.options = this.config.options.map(o => Object.assign({}, o, { selected: false }));
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -137,7 +137,7 @@ class FilterOptionsConfig {
     }));
   }
 
-  getSelectedCount () {
+  getInitialSelectedCount () {
     return this.options.reduce(
       (numSelected, option) => option.selected ? numSelected + 1 : numSelected,
       0);
@@ -177,7 +177,7 @@ export default class FilterOptionsComponent extends Component {
       ...config
     });
 
-    const selectedCount = this.config.getSelectedCount();
+    const selectedCount = this.config.getInitialSelectedCount();
 
     /**
      * True if the option list is expanded and visible
@@ -200,7 +200,7 @@ export default class FilterOptionsComponent extends Component {
   }
 
   setState (data) {
-    const selectedCount = this.config.getSelectedCount();
+    const selectedCount = this.config.getInitialSelectedCount();
     super.setState(Object.assign({}, data, {
       name: this.name.toLowerCase(),
       ...this.config,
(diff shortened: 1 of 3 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -420,7 +425,10 @@ export default class FilterOptionsComponent extends Component {
 
     this.config.options[index] = Object.assign({}, this.config.options[index], { selected });
     this.updateListeners();
-    this.setState();
```
tmeyer2115, line 423: Was the only reason setState was called here to update the selected count? Removing this call won't have any other side effects?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -212,16 +212,21 @@ export default class FilterOptionsComponent extends Component {
...
+    const selectedCount = selectedEls && selectedEls.length;
+
     DOM.delegate(
       DOM.query(this._container, `.yxt-FilterOptions-options`),
       this.config.optionSelector,
       'click',
       event => {
+        let selectedCountEl = DOM.query(this._container, '.js-yxt-FilterOptions-selectedCount');
+        if (selectedCountEl) {
+          selectedCountEl.innerText = selectedCount;
```
tmeyer2115: If I click a new option, does selectedCount need to be updated here?

alexisgrow: Ah good point, technically yes it should. In practice, clicking the collapse/expand icon triggers another onMount so it ends up updating in the selectedCount whenever it is made visible. I'll change it to update the selectedCount here as well, it is more future-proof and this way the count is always up to date!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Yes! Updated this so that setState is only called when using the `storeOnChange` option -- which triggers a new search with the selected options. 

When not using config.storeOnChange, a setState is not needed. The UI changes are now in the onMount (and the selected option is stored in the URL, but that is/was handled outside of the setState)

Edit: looking at this again and testing more, the reset button's behavior should also be updated to follow this pattern. Updated this in a PR fix commit

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -204,35 +204,33 @@ export default class FilterOptionsComponent extends Component {
...
       expanded: this.expanded,
       selectedCount,
       isSingleOption: this.config.control === 'singleoption'
     }));
   }
 
   onMount () {
-    const selectedEls = DOM.queryAll(this._container, '.js-yxt-FilterOptions-checkboxInput:checked');
-    const selectedCount = selectedEls && selectedEls.length;
+    let selectedCount = this._getSelectedCount();
```
tmeyer2115: Do we need this variable any more now that we call _getSelectedCount() in the click handler below?

src/ui/templates/controls/filteroptions.hbs
```diff
@@ -28,25 +28,26 @@
       {{/if}}
       {{#if showReset}}
```
tmeyer2115: Do we sill need this outer showReset here and below? We're not conditionally rendering the button any longer, we're just triggering display:none via displayReset

alexisgrow: Oh I'm not sure! There are two cases where the reset button doesn't appear: 

(1) `showReset: true` but no options are selected. This is where we toggle display: none 

(2) implementer specifies `showReset: false` in their config, in which case the reset button wouldn't appear in the DOM at all

Should we change case (2) so that the reset button does appear in the DOM but is hidden?

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Ah yes true we do not, removing

src/ui/templates/controls/filteroptions.hbs
tmeyer2115: Yeah, let's update case (2) so that the button is always there. I think it will simplify the code a bit, we won't have to track the difference between displayResult and showResult (which I obviously can't do, haha)

alexisgrow: Hahaha, ok will update!

tmeyer2115 (approved):

Issue #664, opened 2020-04-08
yxt-Results-placeholders adding extra margin to page


creotutar:
Hello, on v0.13.1, using dist/answers-umd.js

Noticing that the `yxt-Results-placeholder` has margin styling even though it does not show any content. This should not be the case as you can have inconsistent spacing on the universal results for instance, between verticals.

![image](https://user-images.githubusercontent.com/32468132/78791883-6ecd2600-797e-11ea-8dfb-639d8314395e.png)

Also noticing the placeholder has content in it, despite having `visibility: hidden`. Would recommend we remove the content, could have unintended effects.

alexisgrow:
This has been fixed in v1.0.0, closing!

PR #715, opened 2020-04-22
Results Placeholder: Remove placeholder content and adjust styling

alexisgrow:
Change so placeholders are not added when the maxNumberOfColumns is 1.
Update placeholder to be an empty element rather than a repeated result card.
The placeholder has the same margins and border as a card so it is the
same size; add number of placeholders needed +1 to account for the
requirement that column count changes on screen re-size.

Addresses this issue: https://github.com/yext/answers/issues/664

TEST=manual

On a Universal page, visually inspect on dev site locally and
client site locally. Test on Vertical page with 4, 3, 2, and 1 columns,
adjusting screen size and noting that the cards stay the same size.


oshi97 (commented):
I think we need to change yxt-Results-placeholder to not have the yxt-Card styling, and in the vertical results component also not add any placeholder elements if the maxNumberOfColumns <= 1. The placeholders have to be visibility: hidden instead of display: none to support columns. We should also change the placeholders to not have any internal content.

(Adding placeholders even when maxNumberOfColumns <= 1 was originally done to satisfy a QA item that was actually not the SDK's fault. Custom css was shrinking vertical results on ie11 if there was a small number of results, so placeholders were added to inflate things.)

EDIT: I reread what I wrote and it was really unclear, the placeholder divs probably shouldn't have the yxt-Card class, so that people can style the yxt-Card without unintentionally styling the placeholders. the placeholders still need to have the flex-grow/flex-basis behavior cards have for columns though.

src/ui/templates/results/verticalresults.hbs
```diff
@@ -39,7 +39,7 @@
         </div>
       {{/each}}
       {{#each placeholders}}
-        <div class="yxt-Results-placeholder yxt-Card" data-component="Card" data-opts='{"_index": 0 }' aria-hidden="true"></div>
+        <div class="yxt-Card-placeholder"></div>
```
oshi97: Oo you definitely know better than me but do we need aria-hidden for empty divs?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -186,13 +189,23 @@ export default class VerticalResultsComponent extends Component {
...
       showResultsHeader: showResultsHeader,
       useLegacyNoResults: this._useLegacyNoResults
     }), val);
   }
 
+  /**
+   * helper to return the number of placeholder cards to ensure correct column alignment
+   * @returns {number}
+   */
+  _getNumPlaceholders (numResults, maxNumberOfColumns) {
```
nbramblett: can you @param these?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -186,13 +189,23 @@ export default class VerticalResultsComponent extends Component {
...
+    return maxNumberOfColumns > 1
+      ? (maxNumberOfColumns - (numResults % maxNumberOfColumns)) + 1
```
oshi97: I think `maxNumberOfColumns - 1 `is slightly better than `(maxNumberOfColumns - (numResults % maxNumberOfColumns)) + 1` if we don't want to go the slightly more optimal but very ugly route of

if maxNumberOfColumns = 4, compare `maxNumberOfColumns - (numResults % 4)` and `maxNumberOfColumns - (numResults % 3` and choose the larger

if maxNumberOfColumns = 3, compare `maxNumberOfColumns - (numResults % 3)` and `maxNumberOfColumns - (numResults % 2)` and choose the larger

if maxNumberOfColumns = 2, `maxNumberOfColumns - (numResults % 2)`

oshi97 (commented):
This looks amazing! Just a one small thing and a question

src/ui/components/results/verticalresultscomponent.js
nbramblett: If I have 3 columns, and 2 results, wouldn't this return 2 placeholders? Maybe I'm misunderstanding how the placeholders are used, but shouldn't it only give back 1 so that it has 3 total? Seems like this would result in an uneven number.

nbramblett: oliver clarified why this is the way it is, nvm

src/ui/components/results/verticalresultscomponent.js
alexisgrow: on it!

src/ui/templates/results/verticalresults.hbs
alexisgrow: Lol eek yes, will add it back bc this div is purely visual.

src/ui/components/results/verticalresultscomponent.js
alexisgrow: Ahh gotcha, we'll go with optimal & ugly

alexisgrow: Actually, adjusted the styling so extra placeholders shouldn't affect the DOM. Going to go with `maxNumberOfColumns - 1 `.

src/ui/components/results/verticalresultscomponent.js
alexisgrow: Removed this helper/calculation

oshi97 (approved):
looks great!

oshi97 (approved):

PR #717, opened 2020-04-23
Add border-right border-left to yxt-Accordion-list

oshi97:
Since we added a border to yxt-Results-titleBar oobx,
and since the titleBar sits on top of yxt-Accordion-list,
yxt-Accordion-list probably needs a border-left and border-right
as well so the results line up with the titlebar.

TEST=manual
test that dignity funerals site has lined up AccordionResults
and yxt-Results-titleBar after updating


alexisgrow (approved):
LGTM!

PR #725, opened 2020-04-23
Update Legacy Card to support IE11

nbramblett:
Handles issues with vertical overflow and word wrapping on small IE11
screens.

TEST=manual

Applied this changes to the yext answers test site and verified that the
issues on IE11 were no longer present.


alexisgrow (approved):
lg lg! 🐐

PR #727, opened 2020-04-23
Let url be null on Legacy card.

nbramblett:
Let url be null-ed by the user, so that it can be forced to default to
a non-linking title.

TEST=manual

Verified that setting the config url to null would make the title not display a link for legacy cards.


oshi97 (commented):
looks good to me but would like someone else to look if possible too!

src/ui/components/cards/legacycardcomponent.js
```diff
@@ -52,7 +52,7 @@ class LegacyCardConfig {
      * Url when you click the title
      * @type {string}
      */
-    this.url = this.url || result.link || rawResult.website;
+    this.url = this.url === null ? '' : this.url || result.link || rawResult.website;
```
tmeyer2115: This expression is a bit difficult to parse. Could we reorganize a bit or add parentheses to improve readability?

tmeyer2115 (commented):
Would it be worthwhile to add this to the StandardCard as well? Eventually, folks might transition off the LegacyCard and onto the StandardCard. So, we might want a similar mechanism for them to have a non-linking title.

tmeyer2115 (approved):

PR #729, opened 2020-04-23
Cards: Pass modifier from VerticalResults through to card

alexisgrow:
Added for backwards compatibility. The legacy card is the only one
that uses this variable.

TEST=manual

Test locally on Universal and Vertical pages that the config is
passed through and shows in the DOM.


nbramblett (approved):
LGTM!

PR #730, opened 2020-04-23
VerticalPages config: Add iconUrl option

alexisgrow:
Allow an implementer to specify an iconURL in the verticalPages config. This allows them to use an icon that is not in the built-in set that ships with the template bundle. If both iconUrl and icon are passed in, they will both be passed through to the IconComponent which has logic to determine which takes priority.

TEST=manual

In dev site for a Vertical page, use in an iconUrl in the verticalPages config. View in browser to confirm custom icon displays. Also test to verify that built-in icons still work.


src/core/models/alternativevertical.js
```diff
@@ -37,6 +37,18 @@ export default class AlternativeVertical {
...
+     * URL of an icon
+     * @type {string}
+     */
+    this.iconUrl = config.iconUrl;
+
+    /**
+     * Whether the vertical has an icon
+     * @type {string}
+     */
+    this.hasIcon = this.iconName || this.iconUrl;
```
nbramblett, line 50: doesn't iconName default to star? Does this work? It also seems a bit odd to have hasIcon not explicitly a boolean, even if it works thanks to falseyness.

oshi97: It kind of defaults to star but all the way in the our custom handlebars 'icon' helper/iconcomponent which is a lot deeper than this

README.md
```diff
@@ -89,7 +89,8 @@ verticalPages: [
     url: './index.html',    // The link for the navigation element
     isFirst: true,          // optional, will always show this item first
     isActive: true,         // optional, will add a special class to the item
-    icon: 'star',           // optional, the icon to use in no results and universal results, defaults to star
+    icon: 'star',           // optional, the name of an icon to use in no results, defaults to star
```
oshi97, line 92: do we still use these in universal results?

tmeyer2115 (commented):
Is this something we were supposed to add initially, but didn't? The spec just references 'icon'. Did Rose ask us to add this in to 13.3?

src/core/models/alternativevertical.js
alexisgrow: Haha, yes, this works. +2 to what oshi said about defaults. For the boolean thing, are you suggesting that we change the @type to boolean, or we force this value into a boolean with something like `!!(this.iconName || this.iconUrl)`?

README.md
alexisgrow: We don't use it in Universal Results, and afaict we never did 🙈 I think this was just a mistake originally, it's totally possible that it was going to be used there at some point but the code change didn't happen!

alexisgrow (commented):
This wasn't part of spec originally. This request was from Liz/Bryan to get into v1.0.0 for the SPR-2065 item. I think the main issue is that this inconsistency makes it weird for the HH theme, because the icon (in verticalsToConfig) is used for both UniversalResults (which supports iconUrl) and No Results (which doesn't).

tmeyer2115 (approved):

src/core/models/alternativevertical.js
alexisgrow: Going to merge this. We can update the variable type in a later PR if needed!

PR #736, opened 2020-04-24
Make updates to @babel plugins suggested by Dependabot.

tmeyer2115:
Dependabot suggested we update the versions of the following @babel plugins:

- @babel/plugin-proposal-object-rest-spread
- @babel/plugin-transform-arrow-functions
- @babel/plugin-transform-runtime

Even though these plugins are devDependencies, they should be kept up-to-date.
We use these plugins to transform our source code when creating the various
asset bundles. It's possible, then, that security issues in these plugins could
make their way into the builds of the SDK. These were only minor version
upgrades, so the risk should be low.

TEST=manual

These @babel plugins are mainly used to help polyfill the SDK's source code and
make it IE11 friendly. To test, then, I ran a few smoke tests on a demo site in
IE11. I made sure search, navigation, spell check, autocomplete, and qa
submission functionality worked and no errors were present in the console. I
also made sure to exercise some areas of the code where we use arrow functions
or the spread operator, to ensure the polyfilling of those continued to work.


oshi97 (approved):

oshi97:
lgtm!

MattCBowman (approved):
🧑‍🚒

PR #737, opened 2020-04-24
StandardCard: styling to line up ctas when there are images and no im…

oshi97:
…ages

On chrome, this lines up ctas perfectly whether there are images or not
On ie11 they still don't line up properly, but I hope this can be a more short-term
fix?

TEST=manual
test with maxNumberOfColumns = 4 and maxNumberOfColumns = 1,
using element.getBoundingClientRect(); for cards with images and without images


src/ui/sass/modules/_StandardCard.scss
```diff
@@ -78,6 +78,7 @@
   &-img {
     height: $standard-card-base-spacing * 10;
     min-width: $standard-card-base-spacing * 10;
+    max-width: $standard-card-base-spacing * 15;
```
alexisgrow, line 81: Setting a max-width and height may mess up the aspect ratio of the image. Is there a way to get avoid it in this case?

oshi97: since it's background-size: cover the aspect ratio doesn't change! Setting the max-width was to avoid weird scenarios where the image was very long but short

In retrospect I think the way I did images (with background-image) tries too make it smart but doesn't do it well enough, it probably would have been better to just set the image to a fixed size.

oshi97: image

vs

image

alexisgrow: Ah nice, did not realize this was a background image! LGTM!

alexisgrow (approved):

PR #742, opened 2020-04-24
Fix too tall mobile cards

oshi97:
Alexis found that min-height: 1px apparently fixes the cards collapsing on ie11 for first financial, as well as not making the mobile cards super inflated!


alexisgrow (approved):
IE 11 😿

PR #743, opened 2020-04-24
v1.0.0

tmeyer2115:
# Version 1.0.0

## Changes

* Added TestCafe acceptance tests for the SDK.
* Added the ability for users to specify a custom Card type.
* Added a top-level noResults config object. This is used by both the MapComponent and VerticalResultsComponent.
* Updated some of the dependencies used to supply the SDK with polyfills.

## Fixes

* Assorted fixes for the LegacyCard component. These fixes will improve IE11 styling and allow for bigDate to be specified.
* Assorted fixes for the CTAComponent. These fixes will promote backwards compatibility and improve styling.
* Styling fixes for the legacy AccordionResults component.
* Assorted fixes related to the 'No Results' case. In this case, the VerticalResultsComponent will not display applied filters. Additionally, the MapComponent can be configured to display all results for the vertical or not.


oshi97 (approved):
Would be great if somebody else checked over too but lgtm!

alexisgrow (approved):
LGTM!

PR #750, opened 2020-05-04
Formless search

tmeyer2115:
Allow SearchBar to be created without form as submission context.

This PR allows the SearchBar to be configured to not use a form as the query
submission context. By default, a form is used, unless a client sets the new
'useForm' config option to false. This was done to support clients that wanted
to use a SearchBar on an ASP.NET page.

J=SPR-2175
TEST=manual

Tested the following scenarios for SearchBars set to useForm true and false:

- Entering text and hitting enter.
- Entering text and hitting the submit button.
- Entering text and clearing.
- Clicking autocomplete option.
- Using keys to navigate to and select an autocomplete option.
- Clear button appears in the correct situations.
- Search animation works as expected.
- Near me query works as expected.

In the above scenarios, made sure Analytics worked correctly, if applicable.


src/ui/templates/search/search.hbs
```diff
@@ -5,54 +5,63 @@
...
+          data-eventoptions="{{eventOptions}}"
+          data-component="IconComponent"
+          data-opts='{ "iconName": "close" }'
+          data-prop="icon"
+  >
+    <span class="yxt-SearchBar-clearButtonText sr-only">
+      {{clearText}}
+    </span>
+  </button>
+  <button {{#if useForm}}type="submit"{{/if}} class="js-yext-submit yxt-SearchBar-button">
```
alexisgrow: nit: for HTML validation, could this have the space inside the "{{if}}" on one side? (Some HTML validators don't like two spaces in a row)

alexisgrow (approved):
Overall this LGTM! Definitely out of scope of this item, but for the future - do we want to do the same thing for QA, Geolocation, and Direct Answer components?

alexisgrow (approved):
LGTM!

tmeyer2115:
> Overall this LGTM! Definitely out of scope of this item, but for the future - do we want to do the same thing for QA, Geolocation, and Direct Answer components?

Yep, I imagine we will. SearchBar is the most common problem child for now, but will probably be good to offer formless options for all.

PR #753, opened 2020-05-05
README: Add modifier option to the VerticalResults portion of the doc…

alexisgrow:
Add modifier option to the VerticalResults portion of the documentation.

TEST=none


README.md
```diff
@@ -494,6 +494,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
+  // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
+  modifier: true,
```
alexisgrow: Yikes this is a bad example. Updating.

nbramblett: Do we want this to default to true? That would modify the DOM technically, wouldn't it?

alexisgrow: Updated!

README.md
```diff
@@ -494,6 +494,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
+  // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
+  modifier: 'Standard',
```
nbramblett: Similar question about defaulting to Standard. Is this a purely additive classname change?

alexisgrow: This isn't a default, this is an example, I will update it to empty.

nbramblett (approved):
LGTM!

PR #761, opened 2020-05-05
v1.1.0

tmeyer2115:
# Version 1.1.0

## Changes

* Added a new configuration option, useForm, to the SearchBar. This allows the user to specify if the query submission context should use a form. For backwards compatibility, this parameter is true by default.
* Added the BSD-3 license file.

## Fixes

* Assorted corrections and updates to the SDK's README.
* Fixed the bug causing the LocationBias component's accuracy text to always be 'Device'.


alexisgrow (approved):
LGTM 👍

PR #763, opened 2020-05-06
change yxt-Results-titleBar border to match cards border

oshi97:
merging into v1.1.1 because I think this is arguably a bugfix
TEST=manual
open a universal page with accordion results and vertical results
check that border matches for vertical results but accordion results is unchanged


alexisgrow (approved):
i'm not sure what version this goes into but the change LGTM!

PR #765, opened 2020-05-06
Update border colors for applied filters

nbramblett:
Updates border color so that it's actually visible on the verticals.

### DIFF
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -98,8 +98,8 @@
     padding: $base-spacing / 2 $base-spacing;
     background-color: $results-filters-background;
     border-top: $results-border;
-    border-right: $border-legacy;
-    border-left: $border-legacy;
+    border-right: $border-default;
+    border-left: $border-default;
     border-bottom: none;
   }
 
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -163,8 +163,8 @@
     border: $border-legacy;
 
     .yxt-Results-filters {
-      border-left: none;
-      border-right: none;
+      border-left: $border-legacy;
+      border-right: $border-legacy;
     }
   }
 


src/ui/sass/modules/_Results.scss
```diff
@@ -92,8 +92,8 @@
     padding: $base-spacing / 2 $base-spacing;
     background-color: $results-filters-background;
     border-top: $results-border;
-    border-right: $border-legacy;
-    border-left: $border-legacy;
+    border-right: $border-default;
```
oshi97, line 101: lgtm! do we want to make this not affect AccordionResults (and only VerticalResults) just to be extra safe?

nbramblett: Not sure! It seems like AccordionResults might want to be changes as well, but idk. @tmeyer2115 do you have thoughts on this?

alexisgrow: hi all! (was added as a reviewer so I'll chime in, but Tom please feel free to correct!) 

AccordionResults currently has a more specific styling so I don't think this will apply to the filters on AccordionResults (line 159 of this file).
`      
&.yxt-Accordion .yxt-Results-filters {
  border-left: none;
  border-right: none;
}`

That is probably intended, because we have an open Github Issue to from Rose to change that AccordionResults styling (line 159 of this file) to:

`
&.yxt-Accordion .yxt-Results-filters {
  border-right: $border-legacy;
  border-left: $border-legacy;
}`

oshi97 (approved):
lgtm if alexis is okay too!

src/ui/sass/modules/_Results.scss
tmeyer2115: @nbramblett, since Rose and @alexisgrow have basically laid out what needs to be done for the AccordionResults styling, why don't we include that here too. Two birds with one stone.

src/ui/sass/modules/_Results.scss
```diff
@@ -163,8 +163,8 @@
     border: $border-legacy;
 
     .yxt-Results-filters {
-      border-left: none;
-      border-right: none;
+      border-left: $border-legacy;
```
oshi97, line 166: I know this is being super nitty but didn't we explicitly set these to border-left/right: none? Did something change?

oshi97: like a few weeks ago

tmeyer2115 (approved):

src/ui/sass/modules/_Results.scss
nbramblett: Not sure, but tom approved so it seems not

Issue #766, opened 2020-05-07
Pagination alignment


rosiegrant:
Better align the pagination buttons by default

 @alexisgrow had this suggested fix: 
```
.yxt-Pagination-chevron, .yxt-Pagination-doubleChevron {
  display: flex;
}
```
before:
![Screen Shot 2020-05-07 at 5 45 47 PM](https://user-images.githubusercontent.com/22160788/81347673-a36df380-908a-11ea-9b53-a47a70fdd272.png)

after:
![Screen Shot 2020-05-07 at 5 45 31 PM](https://user-images.githubusercontent.com/22160788/81347688-aa950180-908a-11ea-858b-b64c9ebfdd96.png)

alexisgrow:
This has been fixed in v1.1.1, closing!

PR #767, opened 2020-05-07
Pagination: Center chevron icons vertically

alexisgrow:
In response to this issue (https://github.com/yext/answers/issues/766).

TEST=manual,visual

Test with local answers site. Add pagination component and visually test that the icons were vertically centered.

### DIFF
--- a/src/ui/sass/modules/_Pagination.scss
+++ b/src/ui/sass/modules/_Pagination.scss
@@ -55,4 +55,10 @@
   {
     margin-right: -4px;
   }
+
+  &-chevron,
+  &-doubleChevron
+  {
+    display: flex;
+  }
 }
--- a/src/ui/sass/modules/_Pagination.scss
+++ b/src/ui/sass/modules/_Pagination.scss
@@ -57,7 +57,9 @@
   }
 
   &-chevron,
-  &-doubleChevron
+  &-doubleChevron,
+  &-chevron--left,
+  &-doubleChevron--left
   {
     display: flex;
   }


oshi97 (approved):
lgtm!

src/ui/sass/modules/_Pagination.scss
```diff
@@ -55,4 +55,10 @@
   {
     margin-right: -4px;
   }
+
+  &-chevron,
```
oshi97, line 59: oh just so I can learn css how come they weren't centered vertically without display: flex?

src/ui/sass/modules/_Pagination.scss
```diff
@@ -55,4 +55,10 @@
...
+  &-doubleChevron
+  {
+    display: flex;
```
tmeyer2115, line 64: Is vertical centering an issue for the left chevrons as well?

nbramblett: The left chevron just applies a reversal to the right chevron, iirc.

src/ui/sass/modules/_Pagination.scss
alexisgrow: This is an element wrapping the chevron icon, but it was reaching the height of its parent container! `display: flex` tells it to match its child element. So flex isn't necessarily centering here, it's just removing extra space inside the div that's making it off-center.

src/ui/sass/modules/_Pagination.scss
alexisgrow: Oof good catch, I did not realize that the left chevrons had different class names. Updated to add them here!

tmeyer2115 (approved):

Issue #509, opened 2020-02-24
facets searchOnChange Causing Console Errors (and Impacting SpellCheck)


rosiegrant:
**Current behavior:** When I set "searchOnChange" in the facets component to true, get a bunch of console errors. (In addition, the spell check component does not show up, but I'm guessing it has to do with some issues in setting state). When I remove the "searchOnChange" attribute, no console errors, and spellcheck works fine. 

**Expected behavior:** Setting the search on change attribute should work fine (and the spell check component should show up)

**Version:**
Verified this on dev/v0.13.0 (disregard the fact that I'm using the styling and templates) -- can work to verify on an earlier version (I'm guessing it's always been there since we didn't change anything having to do with facets in this version).

**Console Error:**

![image](https://user-images.githubusercontent.com/22160788/75183631-eae21600-5710-11ea-92a5-0d4ee5c89978.png)

`<!doctype html>
<html>

<head>
    <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,700&display=swap" rel="stylesheet">

    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=5">
    <meta name="format-detection" content="telephone=no">
    <script src='./sharedconfig.js'></script>
    <script src='./combiner.js'></script>
    <script src='./builderHelpers.js'></script>
    
    <script src="./templates.js"></script>

    <link rel="stylesheet" href="../css/all.css">
    <link rel="stylesheet" type="text/css" href="https://assets.sitescdn.net/answers/dev/vr/filters2/answers.css">

    <link rel="stylesheet" href="../css/healthcare.css">
    <style>
        ul, li {
            list-style: none;
            padding: 0;
            margin: 0;
        }
        .yxt-StandardCard-image {
            min-width: 25%;
        }

        .yxt-Results-titleBar { 
            display: none !important; 
        }

    </style>

    
    <title>Healthcare Demo</title>
    <script>
        function initAnswers() {
            const verticalKey = "providers";
            ANSWERS.init({
                apiKey: sharedConfig.apiKey,
                experienceKey: sharedConfig.experienceKey,
                debug: true,
                businessId: sharedConfig.businessId,
                templateUrl: "https://assets.sitescdn.net/answers/dev/vr/filters2/answerstemplates.compiled.min.js",
                locale: 'en',
[...]

PR #768, opened 2020-05-07
FilterBox: Add null check to filter event listener

alexisgrow:
When using the `searchOnChange` config option, the reset button is not present, and we get a 'cannot addEventListener of null' error. Since `this.config.resetFilters` is true by default, implementers will see this error unless they specify false here, which seems incorrect. Adding a guard so that the SDK will only try to bind the click listener if the element exists. Solves this issue: (https://github.com/yext/answers/issues/509).

TEST=manual

Add a FilterBox component with the `searchOnChange` option, notice error before change. After change, notice that the apply button is missing and clicking options triggers a search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,7 +848,7 @@ ANSWERS.addComponent('FilterBox', {
   showCount: true,
   // Optional, execute a new search whenever a filter selection changes
   searchOnChange: false,
-  // Optional, show a reset button per filter group
+  // Optional, show a reset button per filter group, this will only display if searchOnChange is false
   resetFilter: false,
   // Optional, the label to use for the reset button above
   resetFilterLabel: 'reset',
@@ -868,7 +868,7 @@ ANSWERS.addComponent('FilterBox', {
   expand: true,
   // Optional, show the number of applied filter when a group is collapsed
   showNumberApplied: true,
-  // Optional, the label to show on the apply button
+  // Optional, the label to show on the apply button, this will only display if searchOnChange is false
   applyLabel: 'apply',
   // Optional, whether or not this filterbox contains dynamic filters, default false
   isDynamic: true
--- a/README.md
+++ b/README.md
@@ -846,15 +846,15 @@ ANSWERS.addComponent('FilterBox', {
   title: 'Filters',
   // Optional, show number of results for each filter
   showCount: true,
-  // Optional, execute a new search whenever a filter selection changes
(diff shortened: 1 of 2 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -227,10 +227,11 @@ export default class FilterBoxComponent extends Component {
 
     // Initialize reset button
     if (this.config.resetFilters) {
-      DOM.on(
-        DOM.query(this._container, '.js-yxt-FilterBox-reset'),
-        'click',
-        this.resetFilters.bind(this));
+      let resetEl = DOM.query(this._container, '.js-yxt-FilterBox-reset');
```
oshi97: oh hmm if config.resetFilters is true I think the '.js-yxt-FilterBox-reset' should be present. do you mind double checking that there wasn't some typo or logical error from before?

alexisgrow: Yeah that was my thought initially too. The "reset" button only shows when "showApplyButton" (which is !config.searchOnChange) and the config.resetFilters are true. 

After looking into it from your comments, I just updated the PR to remove the config check entirely here so we don't duplicate the logic. Now, the setState is responsible for the logic for whether the reset filters element is on the page or not. If the element is present, this will initialize it. That way the logic is not in two places and it doesn't get out of sync

tmeyer2115 (commented):
LGTM! Let's update the README for Filterbox though, to make it clear that resetFilters isn't enough for the 'Reset' button to be displayed.

README.md
```diff
@@ -848,7 +848,7 @@ ANSWERS.addComponent('FilterBox', {
   showCount: true,
   // Optional, execute a new search whenever a filter selection changes
   searchOnChange: false,
-  // Optional, show a reset button per filter group
+  // Optional, show a reset button per filter group, this will only display if searchOnChange is false
   resetFilter: false,
```
tmeyer2115, line 852: We also want to make it clear that resetFilters behaves this way too.

alexisgrow: Oh true, I'm going to add it to all of the "reset*" config options to be clearer since it's possible that someone will only read the comment for the one they're trying to change.

alexisgrow: updated!

tmeyer2115 (approved):

PR #770, opened 2020-05-11
Pagination: cleanup

nbramblett:
Addressed Tom's comments on previous CRs.

TEST=manual,auto

Eye test and unit tests.

### DIFF
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -175,7 +175,7 @@ export default class PaginationComponent extends Component {
     const currentOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
     const currentPageNumber = (currentOffset / this._limit) + 1;
     const newPageNumber = (offset / this._limit) + 1;
-    const maxPageCount = Math.trunc((results.resultsCount - 1) / this._limit) + 1;
+    const maxPageCount = this._computeMaxPage(results.resultsCount);
     this._onPaginate(newPageNumber, currentPageNumber, maxPageCount);
     this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, offset);
     this.core.persistentStorage.set(StorageKeys.SEARCH_OFFSET, offset);
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -165,7 +165,7 @@ export default class PaginationComponent extends Component {
 
   updatePage (offset) {
     const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
-    let oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
+    const oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
     const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
     const oldPageNumber = (oldOffset / limit) + 1;
     const pageNumber = (offset / limit) + 1;
@@ -251,7 +251,7 @@ export default class PaginationComponent extends Component {
 
   setState (data) {
     const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
-    let offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
+    const offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
     const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
     const pageNumber = (offset / limit) + 1;
     const isMoreResults = results.resultsCount > offset + limit;
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -100,6 +100,13 @@ export default class PaginationComponent extends Component {
      */
     this._onPaginate = config.onPaginate || this.scrollToTop;
 
+    /**
+     * The maximum number of results per page
+     * @type {number}
+     * @private
+     */
+    this._limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
+
     const offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
     this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, Number(offset));
     this.core.globalStorage.on('update', StorageKeys.SEARCH_OFFSET, offset => {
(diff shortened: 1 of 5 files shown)


src/ui/components/results/paginationcomponent.js
```diff
@@ -164,7 +164,13 @@ export default class PaginationComponent extends Component {
   }
 
   updatePage (offset) {
-    this._onPaginate();
+    const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
+    let oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
+    const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
```
creotutar: are we guaranteed to get a search config?

src/ui/components/results/paginationcomponent.js
```diff
@@ -164,7 +164,13 @@ export default class PaginationComponent extends Component {
   }
 
   updatePage (offset) {
-    this._onPaginate();
+    const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
+    let oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
```
creotutar: very small nit: can this be a const?

src/ui/components/results/paginationcomponent.js
```diff
@@ -164,7 +164,13 @@ export default class PaginationComponent extends Component {
...
+    const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
+    const oldPageNumber = (oldOffset / limit) + 1;
```
creotutar: can limit be 0?

src/ui/components/results/paginationcomponent.js
nbramblett: Limit defaults to 15 if not set by the user, iirc. I don't think it's possible to be null unless someone actively tried to null out limit, in which case they shouldn't really be using pagination.

src/ui/components/results/paginationcomponent.js
nbramblett: Hm, this is probably one of those cases where it could technically happen but would probably not ever happen (since it would mean you get 0 results). I'm somewhat of the opinion that if you set limit to 0 and then attempt to paginate then you deserve whatever implosion you get. They could also attempt to set limit to a negative number and bork themselves that way as well.

@tmeyer2115  do you have thoughts on this?

src/ui/components/results/paginationcomponent.js
tmeyer2115: We are guaranteed to get a SearchConfig. A default one is created if not supplied to Answers.init. The limit in this default SearchConfig will be 20.

src/ui/components/results/paginationcomponent.js
tmeyer2115: There is validation in the SDK that the limit supplied to a SearchConfig must be between 1 and 50.

src/ui/components/results/paginationcomponent.js
```diff
@@ -164,7 +164,13 @@ export default class PaginationComponent extends Component {
   }
 
   updatePage (offset) {
-    this._onPaginate();
+    const results = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
+    const oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
+    const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
+    const oldPageNumber = (oldOffset / limit) + 1;
+    const pageNumber = (offset / limit) + 1;
+    const maxPageCount = Math.trunc((results.resultsCount - 1) / limit) + 1;
```
tmeyer2115: We recompute the limit many times in this class. But, this quantity is effectively static (once the page has loaded). To avoid inefficiency, maybe we should compute limit once, in the constructor. I think we should also make a helper to compute maxPageCount. That way, we can refer to the helper and not have this logic written out explicitly in multiple places.

src/ui/components/results/paginationcomponent.js
tmeyer2115: Minor nit, but currentPageNumber?

src/ui/components/results/paginationcomponent.js
```diff
@@ -239,18 +245,18 @@ export default class PaginationComponent extends Component {
         desktopBack: this._pinFirstAndLastPage && desktopBackLimit > 0,
```
tmeyer2115, line 259: Do we need to check that it's greater than 1 here? If the backLimit is 1, wouldn't we not want ellipses?

tests/ui/components/results/paginationcomponent.js
```diff
@@ -145,6 +145,6 @@ describe('rendering the page numbers', () => {
     const next = wrapper.find('.js-yxt-Pagination-next');
     expect(next).toHaveLength(1);
     next.first().simulate('click');
-    expect(paginate).toHaveBeenCalled();
```
tmeyer2115, line 148: Do we have test coverage for all the different edge cases that can occur in _allocate? That's one of the more complex methods in the Pagination component. We'll want to make sure we have coverage for the case in which _allocate will distribute pages evenly about the middle page and when it will be uneven since it's hit a boundary in some direction.

nbramblett: I've added some unit tests, it's a bit of a PITA to test comprehensively using enzyme, but I'll see about more thoroughly checking this.

src/ui/components/results/paginationcomponent.js
nbramblett: this is an interesting point, I suppose it would be a bit odd to see 1 ... 2 3 4 5 6 ... 10

I'll update this so that numbers have to actually be skipped for the ellipses to show up.

src/ui/components/results/paginationcomponent.js
```diff
@@ -165,12 +172,11 @@ export default class PaginationComponent extends Component {
...
-    const oldOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
-    const limit = this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).limit;
-    const oldPageNumber = (oldOffset / limit) + 1;
-    const pageNumber = (offset / limit) + 1;
-    const maxPageCount = Math.trunc((results.resultsCount - 1) / limit) + 1;
-    this._onPaginate(pageNumber, oldPageNumber, maxPageCount);
+    const currentOffset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET) || 0;
+    const currentPageNumber = (currentOffset / this._limit) + 1;
+    const newPageNumber = (offset / this._limit) + 1;
+    const maxPageCount = Math.trunc((results.resultsCount - 1) / this._limit) + 1;
```
tmeyer2115: Use _computeMaxPage here

src/ui/components/results/paginationcomponent.js
```diff
@@ -239,27 +253,32 @@ export default class PaginationComponent extends Component {
     }
 
     return {
-      ellipses: {
+      pinnedNumbers: {
```
tmeyer2115, line 256: This is kind of confusing. What's the difference between pinnedNumbers and ellipses?

src/ui/components/results/paginationcomponent.js
```diff
@@ -239,27 +253,32 @@ export default class PaginationComponent extends Component {
...
         mobileBack: this._pinFirstAndLastPage && mobileBackLimit > 0,
```
tmeyer2115, line 257: If the backLimit is 1, will this show both Page '1' twice?

src/ui/templates/results/pagination.hbs
```diff
@@ -38,11 +40,12 @@
     </a>
 
     <a
-      class="yxt-Pagination-link js-yxt-Pagination-link{{#unless ellipses.desktopBack}} desktop-hidden{{/unless}}{{#unless ellipses.mobileBack}} mobile-hidden{{/unless}}"
+      class="yxt-Pagination-link js-yxt-Pagination-link{{#unless pinnedNumbers.desktopBack}} desktop-hidden{{/unless}}{{#unless pinnedNumbers.mobileBack}} mobile-hidden{{/unless}}"
```
tmeyer2115, line 43: Would it be simpler to just conditionally render these tags instead of toggling hidden CSS classes on or off?

src/ui/components/results/paginationcomponent.js
nbramblett: ellipses will only show if there's a gap between the last page number and the pinned page number

src/ui/components/results/paginationcomponent.js
nbramblett: It will not.

src/ui/templates/results/pagination.hbs
nbramblett: the limits are different values for mobile and desktop, and there's no way to conditionally render them from the handlebars while knowing the size of the browser (at least not without some extremely hacky js)

tmeyer2115 (approved):

PR #771, opened 2020-05-12
SCSS: Move font-family top level instead of having it in text mixin

alexisgrow:
TEST=manual

Test on a Jambo site in SGS that overriding the font-family overrides for the SDK on a vertical-map page as well as universal. Test that without a font-family override, the SDK default is displayed.

### DIFF
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -236,6 +236,9 @@ export default class Component {
     });
 
     DOM.addClass(this._container, this._className);
+
+    // We use this class to scope some of our styling.
+    DOM.addClass(this._container, 'yxt-Container');
     return this;
   }
 
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -238,7 +238,7 @@ export default class Component {
     DOM.addClass(this._container, this._className);
 
     // We use this class to scope some of our styling.
-    DOM.addClass(this._container, 'yxt-Container');
+    DOM.addClass(this._container, 'yxt-ComponentContainer');
     return this;
   }
 
--- a/src/ui/sass/_mixins.scss
+++ b/src/ui/sass/_mixins.scss
@@ -5,7 +5,6 @@
   $style: normal,
   $color: $color-text-primary
 ) {
-  font-family: $font-family;
   font-size: $size;
   line-height: $line-height;
   font-weight: $weight;
--- a/src/ui/sass/_base.scss
+++ b/src/ui/sass/_base.scss
@@ -14,7 +14,7 @@ body
   -moz-osx-font-smoothing: grayscale;
 }
 
-.yxt-Container
+.yxt-ComponentContainer
 {
   font-family: $font-family;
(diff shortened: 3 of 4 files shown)


src/ui/components/component.js
```diff
@@ -236,6 +236,9 @@ export default class Component {
     });
 
     DOM.addClass(this._container, this._className);
+
+    // We use this class to scope some of our styling.
+    DOM.addClass(this._container, 'yxt-Container');
```
tmeyer2115: How about 'yxt-componentContainer',  'yxt-Component-Container', or something like that? A bit more descriptive about what the container div represents.

src/ui/sass/_mixins.scss
```diff
@@ -5,7 +5,6 @@
   $style: normal,
   $color: $color-text-primary
 ) {
-  font-family: $font-family;
```
tmeyer2115, line 8: Just want to confirm, but removing this won't break anybody's custom styling, right? If someone built their site with template-answers, they could conceivably have used this mixin in any custom styling. They'd be fine though, even with this removed, since whichever class they'd have styled should be underneath a 'yxt-Container' div.

src/ui/components/component.js
alexisgrow: Yes that sgtm! Updating!

src/ui/sass/_mixins.scss
alexisgrow: Yes exactly! Most implementers styling that would depend on this will still work - e.g. changing the SDK's font to their custom one - because this makes the styling less specific

This would break custom styling if 

1. This mixin was used outside of an answers component container. I don't know what our recommendation would be here, what do you think?
2.  If an implementer was applying a custom font-family to the answers component container CSS selector **and** their styling is imported before ours. Implementers would need to ensure their styling was imported after the SDK styling - this is currently how it works in client implementations, so no change required.
3.  If an implementer was applying styling to a more-specific selector than the answers component container but expecting it **not** to work and instead have the SDK's font-family applied. This seems like a bad practice, it may be OK to break this if it's happening, what do you think?

tmeyer2115: 1. I don't think we need to guarantee correctness if they use Answers CSS for non-Answers HTML. That's not something we'd recommend anyone do.
2. Almost always, if you're trying to override a library's styling, you want your styling to be loaded afterwards. That's certainly how it's always worked overriding the SDK's styling.
3. That does seem like a bad practice. If they added a font-family to a more specific selector, they may have wanted it to work. Now, it will. If they didn't actually want that font-family and left it, they left broken/irrelevant CSS in their style sheets. They were depending on what's arguably a bug in the SDK. I don't think we'd want to support that. It seems pretty unlikely.

alexisgrow: Amazing, that sounds right to me!

tmeyer2115 (approved):

PR #772, opened 2020-05-12
Add nice handlebars helpers we have from the theme

oshi97:
copy pasted from the HH theme


nbramblett:
Do neither of these already exist in some capacity?

oshi97:
it looks like we have `and` and `or` I'll close this but I think it would be nice in the future if everything was standardized

PR #776, opened 2020-05-12
Implement front-end killswitch for an Answers experience.

tmeyer2115:
This PR adds a killswitch to the SDK that can disable the front-end of any
Answers experience. Essentially, the onReady supplied to Answers.init will not
be invoked. A disabled page will contain none of the SDK's components. To check
if the front-end for an experience should be disabled, the corresponding
AnswersStatus page is consulted. The SDK errs on the side of enabling the
experience's front-end in that any AnswersStatus request failure is treated
as disabled: false.

J=SPR-1974
TEST=auto, manual

Added unit tests for the different scenarios. Manually tested an experience
that had been explicitly enabled and disabled. Confirmed the correct behavior
for these scenarios. Also tested an experience that did not yet have a status
page to ensure that the 404 was handled as disabled: false.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -165,7 +165,6 @@ class Answers {
 
     const masterSwitchApi = new MasterSwitchApi(globalStorage);
     masterSwitchApi.isDisabled(parsedConfig.apiKey, parsedConfig.experienceKey)
-      .then(isDisabled => { console.log(isDisabled); return isDisabled; })
       .then(isDisabled => !isDisabled && this._initInternal(parsedConfig, globalStorage, persistentStorage))
       .catch(() => this._initInternal(parsedConfig, globalStorage, persistentStorage));
   }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -28,6 +28,7 @@ import MockSearchService from './core/search/mocksearchservice';
 import ComponentManager from './ui/components/componentmanager';
 import VerticalPagesConfig from './core/models/verticalpagesconfig';
 import { SANDBOX, PRODUCTION } from './core/constants';
+import MasterSwitchApi from './core/utils/masterswitchapi';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
@@ -137,6 +138,12 @@ class Answers {
     return this.instance;
   }
 
+  /**
+   * Initializes the SDK with the provided configuration. Note that before onReady
+   * is ever called, a check to the relevant Answers Status page is made.
+   *
+   * @param {Object} config The Answers configuration.
+   */
   init (config) {
     const parsedConfig = this.parseConfig(config);
     this.validateConfig(parsedConfig);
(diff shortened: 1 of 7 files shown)


src/core/utils/masterswitchapi.js
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param {string} apiKey The apiKey of the experience.
+   * @param {string} experienceKey The identifier of the experience.
+   * @returns {Promise<boolean>} A Promise containing a boolean indicating if the front-end
+   *                             should be disabled.
+   */
+  isDisabled (apiKey, experienceKey) {
+    const requestConfig = {
+      apiKey,
+      baseUrl: 'https://answersstatus.pagescdn.com/',
```
MattCBowman: It's ok to have this as a default, but maybe we allow overriding in the constructor so that if we need to change it for a particular customer, we can do it?

src/core/utils/masterswitchapi.js
```diff
@@ -0,0 +1,39 @@
...
+      endpoint: `${apiKey}/${experienceKey}/status.json`
```
MattCBowman: Same comment here as above about letting the URL to check be injected instead of being created in the class and locked away

src/core/utils/masterswitchapi.js
tmeyer2115: We could add something to the constructor that would allow us to override the path to the status.json for a particular customer. I think we'd need to add a new attribute(s) to the SDK configuration as well if we wanted to support such an override on-the-fly, without any code changes to the SDK.

tmeyer2115:
@MattCBowman, I added the override you mentioned. Let me know if this is what you had in mind!

src/answers-umd.js
```diff
@@ -156,6 +165,24 @@ class Answers {
     globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
 
+    const masterSwitchApi = statusPage
+      ? new MasterSwitchApi({ apiKey: parsedConfig.apiKey, ...statusPage }, globalStorage)
```
bryanreed, line 169: do we want to be passing in the api key here still?  I feel like we should take the provided status page as gospel.  If for some wack reason they provide us a statuspage url with an api key param i wouldn't want to override it.

src/core/utils/masterswitchapi.js
```diff
@@ -0,0 +1,48 @@
...
+   * Creates a new {@link MasterSwitchApi} from the provided parameters.
+   *
+   * @param {string} apiKey The apiKey of the experience.
+   * @param {string} experienceKey The identifier of the experience.
+   * @param {GlobalStorage} globalStorage The {@link GlobalStorage} instance.
+   * @returns {MasterSwitchApi} The new {@link MasterSwitchApi} instance.
+   */
+  static from (apiKey, experienceKey, globalStorage) {
+    const requestConfig = {
+      apiKey,
```
bryanreed, line 42: is this accessible?

src/answers-umd.js
tmeyer2115: The apiKey doesn't really matter here. It's an idiosyncrasy of the ApiRequest class that we include it here. It is not used in determining the actual URL that is hit. We will still hit the one they provide in the statusPage arg.

src/core/utils/masterswitchapi.js
tmeyer2115: Not from outside the SDK, no. This method assumes the status page URL follows a particular format, so someone wanting to override wouldn't want to use this.

bryanreed: sorry, I just meant this looks weird to me since its a string amongst key value pairs here but maybe I am just a JS noob.

bryanreed (approved):

src/core/utils/masterswitchapi.js
tmeyer2115: Ohhh, that's syntactic sugar. It's short for "apiKey: apiKey"

PR #777, opened 2020-05-12
AppliedFilters: update styling based on new mocks/spec

oshi97:
TEST=manual
try to match mocks

### DIFF
--- a/src/ui/sass/modules/_ResultsHeader.scss
+++ b/src/ui/sass/modules/_ResultsHeader.scss
@@ -15,8 +15,6 @@ $results-header-filters-font-size: var(--yxt-font-size-md) !default;
 $results-header-filters-color: var(--yxt-color-text-secondary) !default;
 $results-header-filters-line-height: var(--yxt-line-height-md) !default;
 
-$results-header-horizontal-spacing: calc(var(--yxt-results-header-spacing) / 2) !default;
-
 :root {
   --yxt-results-header-font-size: #{$results-header-font-size};
   --yxt-results-header-spacing: #{$results-header-spacing};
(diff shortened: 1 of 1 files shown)


src/ui/sass/modules/_ResultsHeader.scss
```diff
@@ -51,7 +50,7 @@
   &-filterLabel
   {
     display: flex;
-    margin-right: $results-header-spacing * 3 / 8;
+    margin-right: $results-header-spacing / 2;
```
tmeyer2115: This seems like a fairly common quantity in this CSS file, let's make a variable for it

tmeyer2115 (commented):
Are these breaking CSS changes? Or is this styling for something that's net-new with your AppliedQueryFilters work?

src/ui/sass/modules/_ResultsHeader.scss
oshi97: I tried creating a $results-header-filters-spacing and then that didn't feel right because it wasn't only for applied filters so then I called it $results-header-horizontal-spacing, but then basically everything except for the padding-top/bottom value was replaced. I feel like this is not particularly more useful than $results-header-spacing but I could be wrong.

oshi97: updated to show what this might look like I'm fine either way

oshi97:
> Are these breaking CSS changes? Or is this styling for something that's net-new with your AppliedQueryFilters work?

Rose said this is fine as long as it's in a minor version and documented https://yext.slack.com/archives/GREFVUD7V/p1590094431498300

src/ui/sass/modules/_ResultsHeader.scss
```diff
@@ -82,7 +84,7 @@
   &-filterSeparator
   {
     color: $results-header-filters-color;
-    margin-right: $results-header-spacing * 3 / 8;
+    margin-right: $results-header-horizontal-spacing;
```
tmeyer2115: Isn't results-header-horizontal-spacing 0.5 * results-header-spacing? The previous styling was 3/8 * results-header-spacing. Was that incorrect?

tmeyer2115 (commented):
Ehh, you're right, not much gain by creating results-header-horizontal-spacing. We can just relate everything to results-header-spacing as before. Sorry about that.

oshi97:
rebased onto v1.4.0!

src/ui/sass/modules/_ResultsHeader.scss
oshi97: looking at the older mocks it looks like this was incorrect. I want to say I saw in the mocks that the spacing was 6px instead of the 8px it currently is (I'm sure I didn't make that weird fraction up on my own) but I can't find a mock showing this

oshi97:
in most recent commit also removed 

&-appliedFilters {
  align-items: center

just noticed it was misaligning things

tmeyer2115 (approved):

PR #779, opened 2020-05-13
css variables

creotutar:


nbramblett (approved):
LGTM

bryanreed (approved):

src/ui/sass/_colors.scss
```diff
@@ -7,8 +7,21 @@ $color-brand-white: #fff !default;
 $color-text-primary: #212121 !default;
 $color-text-secondary: #757575 !default;
 
-$color-link-primary: $color-brand-primary !default;
+$color-link-primary: var(--color-brand-primary) !default;
```
tmeyer2115: What are the mechanics of this kind of default? Does the SASS compiler replace all instances of "$color-link-primary" with "var(--color-brand-primary)"?

src/ui/sass/modules/_AlternativeVerticals.scss
```diff
@@ -13,7 +13,7 @@ $alternative-verticals-emphasized-font-weight: var(--yxt-font-weight-medium) !de
   background-color: var(--yxt-color-brand-white);
   border: var(--yxt-border-default);
 
-  @media (min-width: var(--yxt-breakpoint-mobile-max)) {
```
tmeyer2115: Are there other media queries that need to be similarly updated?

src/ui/sass/modules/_AlternativeVerticals.scss
```diff
@@ -13,7 +13,7 @@ $alternative-verticals-emphasized-font-weight: var(--yxt-font-weight-medium) !de
...
+  @media (min-width: $breakpoint-mobile-max) {
```
tmeyer2115, line 16: If someone updates the var(--yxt-breakpoint-mobile-max), those changes won't get reflected here, right? That seems like it could be an issue.

tmeyer2115: With the HH Theme, will someone need to add explicit styling to override a breakpoint? Sounds like we can't support altering breakpoints with just CSS vars/Ponyfill.

src/ui/sass/_colors.scss
creotutar: yup yup!

src/ui/sass/modules/_AlternativeVerticals.scss
creotutar: there are not-- i took a look at all other media queries

src/ui/sass/modules/_AlternativeVerticals.scss
creotutar: you are right! will remove the breakpoint css variables.

with the HH theme, we are dropping the ability to dynamically set media query boundaries due do this css restriction.

they can, however, still override styling inside of media queries by explicitly writing their own

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #780, opened 2020-05-13
MapboxMapProvider: use mapConfig.pin.svg for custom marker svg

oshi97:
This commit allows people to specify a custom svg for their
mapbox map marker. This behavior already existed for google,
but not mapbox.
For more customization see https://docs.mapbox.com/help/tutorials/custom-markers-gl-js/

TEST=manual, auto
changed hitchhiker theme to point to local answers
checked that new custom map pins in the HH theme worked

opened universal page with maps (custom raw html site, not theme)
checked that by default would still use original map marker
checked that could specify an svg in mapConfig.pin function

### DIFF
--- a/src/ui/components/map/providers/mapboxmapprovider.js
+++ b/src/ui/components/map/providers/mapboxmapprovider.js
@@ -169,7 +169,7 @@ export class MapBoxMarkerConfig {
           marker);
       }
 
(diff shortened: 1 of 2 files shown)


oshi97:
Sorry actually I need to double check a few things changing to draft for now

PR #784, opened 2020-05-14
Fix gridded yxt-Card styling for 2 columns

### LINKED ISSUE #783: maxNumberOfColumns does not support 2 columns width
https://github.com/yext/answers/blob/master/src/ui/sass/modules/_Card.scss#L63

The width for cards should be 50% when the maxNumberOfColumns is specified to be 2. This is working correctly for 3 and 4.

creotutar:
Fixes #783

We should be applying width to the Card for 2 columns just as we are doing for 3 and 4 columns.


creotutar:
Closing to start off of v0.14.0

PR #785, opened 2020-05-15
Support click analytics for RTF links.

tmeyer2115:
This PR adds click analytics for links in a formatted RTF value. Note that such
values occur only in Cards or the DirectAnswers component. Now, when the
formatRichText function is called, it places a wrapper div around the newly
formatted RTF value. A 'cta-type' data attribute is also added to all links
in the value. Click events from the links are bubbled up to this wrapper div
and handled in both the CardComponent and DirectAnswersComponent classes.

The adding of the 'cta-type' attribute is handled by the markdown-it-for-inline
plugin.

Right now, we only support one RTF formatted value per Card. The solution in
this PR can be easily extended to support multiple RTF values in the future.
Some room for future improvement would be to add classes to the links in
formatRichText. This would allow us to write click handlers for the links
directly, instead of having the container div around the entire formatted RTF
value. Unfortunately, it seems the markdown-it-for-inline plugin does not
support this functionality currently.

J=SPR-2232
TEST=auto, manual

Added unit tests for the new RichTextFormatter class. Manually tested the
following scenarios for DirectAnswers and Cards (Vertical and Universal):

- Clicking outside a link, in the formatted value, does nothing.
- Clicking a phone-type link in the formatted value works as expected and
  issues the correct AnalyticsEvent.
- Clicking an email-type link in the formatted value works as expected and
  issues the correct AnalyticsEvent.
- Clicking a normal link in the formatted value works as expected and issues
  the correct AnalyticsEvent.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -408,6 +408,18 @@ class Answers {
     this.core.setQuery(searchConfig.defaultInitialSearch);
   }
 
+  /**
+   * Sets the geolocation tag in global storage, overriding other inputs. Do not use in conjunction
+   * with other components that will set the geolocation internally.
+   * @param {number} lat
+   * @param {number} long
+   */
+  setGeolocation (lat, lng) {
+    this.core.globalStorage.set(StorageKeys.GEOLOCATION, {
+      lat, lng, radius: 0
+    });
+  }
+
   /*
    * Updates the css styles with new current variables. This is useful when the css
    * variables are updated dynamically (e.g. through js) or if the css variables are
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -1,7 +1,6 @@
 /** @module */
 
 import Core from './core/core';
-import RtfConverter from '@yext/rtf-converter';
 import cssVars from 'css-vars-ponyfill';
 
 import {
@@ -16,7 +15,7 @@ import ConsoleErrorReporter from './core/errors/consoleerrorreporter';
 import { AnalyticsReporter, NoopAnalyticsReporter } from './core';
 import PersistentStorage from './ui/storage/persistentstorage';
 import GlobalStorage from './core/storage/globalstorage';
-import { AnswersComponentError, AnswersCoreError } from './core/errors/errors';
+import { AnswersComponentError } from './core/errors/errors';
 import AnalyticsEvent from './core/analytics/analyticsevent';
 import StorageKeys from './core/storage/storagekeys';
 import SearchConfig from './core/models/searchconfig';
@@ -30,6 +29,7 @@ import ComponentManager from './ui/components/componentmanager';
 import VerticalPagesConfig from './core/models/verticalpagesconfig';
 import { SANDBOX, PRODUCTION } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
+import RichTextFormatter from './core/utils/richtextformatter';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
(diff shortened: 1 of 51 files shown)


src/ui/components/results/directanswercomponent.js
```diff
@@ -93,19 +93,55 @@ export default class DirectAnswerComponent extends Component {
...
       });
     });
 
     // Is this actually necessary? I guess it's only necessary if the
     // submit button is hidden.
     DOM.on(this._thumbsUpSelector, 'click', () => { DOM.trigger(this._formEl, 'submit'); });
     DOM.on(this._thumbsDownSelector, 'click', () => { DOM.trigger(this._formEl, 'submit'); });
+
+    const rtfElement = DOM.query(this._container, '.js-yxt-rtfValue');
+    rtfElement && this._handleRtfClickAnalytics(rtfElement);
```
oshi97: Could we move `DOM.on(rtfElement, 'click', eventHandler);` up here, and change the eventHandler method in _handleRtfClickAnalytics to just be the main method? 
Right now _handleRtfClickAnalytics isn't really an event handler it's an event registerer. We could also use the arrow function to get rid of const self = this like
`rtfElement && DOM.on(rtfElement, 'click', e => this._handleRtfClickAnalytics(e, rtfElement);`

*same for the click analytics in cardcomponent.js

oshi97 (commented):
I think DOM.query only takes the first matching element, so does this work if there are multiple instances of .js-yxt-rtfValue (and do we care...?)

src/ui/components/cards/cardcomponent.js
```diff
@@ -64,11 +67,52 @@ export default class CardComponent extends Component {
...
+   *
+   * @param {Node} rtfElement The Rich Text attribute.
+   */
+  _handleClickAnalytics (rtfElement) {
+    const self = this;
+    const eventHandler = event => {
+      if (!event.target.hasAttribute('data-cta-type')) {
+        return;
+      }
+      const ctaType = event.target.getAttribute('data-cta-type');
```
oshi97: could we do something like `const { ctaType, fieldName } = rtfElement.dataset` here?

src/ui/components/results/directanswercomponent.js
```diff
@@ -93,19 +93,55 @@ export default class DirectAnswerComponent extends Component {
...
+  /**
+   * A handler for the click analytics of a Rich Text Direct Answer.
+   *
+   * @param {Node} rtfElement The {@link Node} containing the Rich
+   *                          Text Direct Answer.
+   */
+  _handleRtfClickAnalytics (rtfElement) {
+    const self = this;
+    const eventHandler = event => {
+      if (!event.target.hasAttribute('data-cta-type')) {
```
oshi97: could this be const ctaType = rtfElement.dataset.ctaType instead?

tmeyer2115:
@oshi97, right now we're just supporting one RTF attribute in a Card: the card's details. In the future, we will support multiple! To do that, we can update what I have to be a querySelectorAll! I'd also like to figure out how we could inject custom class names into the anchor tags via a markdown-it plugin. Then we wouldn't need this top-level div.

oshi97:
Think this needs to be rebased on top of 1.2.0 but after that lgtm!

src/answers-umd.js
```diff
@@ -81,14 +81,8 @@ class Answers {
...
-    this.formatRichText = (markdown) => {
-      if (typeof markdown !== 'string') {
-        throw new AnswersCoreError(
-          `Rich text "${markdown}" needs to be a string. Currently is a ${typeof markdown}`
-        );
-      }
-      return RtfConverter.toHTML(markdown);
-    };
+    this.formatRichText = (fieldName, markdown) =>
+      RichTextFormatter.format(fieldName, markdown);
```
creotutar, line 85: is this meant to be `RichTextFormatter.format(markdown, fieldName);`?

based on https://github.com/yext/answers/pull/785/files#diff-be9a7814200300fa639a80e119e09cd8R24

PR #786, opened 2020-05-15
SearchDataTransformer: centralize allResultsForVertical parsing

oshi97:
This commit moves _formResponseFromAllResultsForVertical from models/verticalresults.js
up one level into searchdatatransformer, so that other models for the same set of request,
like DYNAMIC_FILTERS, do not have to redo this parsing.

The only functional change of this commit is that before verticalresults determined
it was using allResultsForVertical if response.results && response.results.length > 0;
was false, and dynamicfilters used allResultsForVertical if
response.resultsCount === 0 && response.allResultsForVertical was true. Now they both
use the former.

J=SPR-1923
TEST=manual
checked that can run searches and apply facets for both normal and no results.


oshi97:
HOW TO READ THIS PR:
dynamicfilters.js:
- changed response to facets because was only using that portion of response
- response is already transformed to use allResultsForVertical when needed
- logic for using allResultsForVertical is slightly different implementation-wise but end effect should be the same

verticalresults.js
- _formResponseFromAllResultsForVertical moved into searchdatatransformer
- response is already transformed to use allResultsForVertical when needed
- results context passed in from searchdatatransformer

searchdatatransformer.js
- transform data.response using same logic from verticalresults.js before
- also determine resultsContext here

-update tests

src/core/search/searchdatatransformer.js
```diff
@@ -31,16 +32,44 @@ export default class SearchDataTransformer {
   }
 
   static transformVertical (data, formatters, verticalKey) {
-    // TODO (tmeyer): SPR-1923, centralize the allResultsForVertical parsing here.
+    const resultsContext = SearchDataTransformer._determineResultsContext(data.response);
+    const response = SearchDataTransformer._formResponseWithContext(data.response, resultsContext);
```
alexisgrow: This `_formResponseWithContext` only works with vertical responses, do you think it should have `vertical` in the function signature?

alexisgrow (approved):
LGTM

src/core/search/searchdatatransformer.js
oshi97: sure sgtm!

src/core/search/searchdatatransformer.js
```diff
@@ -31,16 +32,44 @@ export default class SearchDataTransformer {
...
+    const hasResults = response.results && response.resultsCount > 0;
+    return hasResults ? ResultsContext.NORMAL : ResultsContext.NO_RESULTS;
+  }
+
+  /**
+   * Form response as if the results from `allResultsForVertical` were the actual
+   * results in `results`
+   * @param {Object} response The server response
+   */
+  static _formVerticalResponseWithContext (response, resultsContext) {
```
tmeyer2115: I think this name is kind of confusing. Maybe parseVerticalResponse? I think it could also simplify things a bit if we made resultsContext part of our VerticalResults model. So, this method would return an instance of VerticalResults that included a resultsContext property. Since resultsContext doesn't make sense without a VerticalResults and we need a resultsContext to interpret VerticalResults, I think it makes sense to add resultsContext to the VerticalResults model.

tmeyer2115 (commented):
One minor thing, other than that LGTM!

src/core/search/searchdatatransformer.js
oshi97: The other models don't currently take in the VerticalResults model, they take in things like response.facets and response.searchIntents. resultsContext is also already part of the VerticalResults model - are you saying we want something like a VerticalResponse model that also contains resultsContext?

tmeyer2115: Gotcha. Yeah, I think we should have the VerticalResponse (created here) have resultsContext. I think we should also rename this method to make its purpose clearer.

oshi97: ah okay that makes a lot more sense than making it a special property, even though it's the only one that property that doesn't come directly from the response, updated!

tmeyer2115 (approved):

PR #787, opened 2020-05-15
Add css-vars-ponyfill for css variables support on IE11

creotutar:
Invoke the ponyfill only for legacy browsers (ie11 being the relevant
browser for us). The ponyfill will be invoked on an ANSWERS.init and
should happen /after/ the answers css variables have been loaded.

Expose an ANSWERS.polyfillCssVariables for updating dynamic css
variables

TEST=manual

Tested on a JAMBO site and an AEB site on our supported browsers
and on Windows, iOS, and Android through browserstack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1620,6 +1620,9 @@ The Answers SDK supports styling specific elements with CSS variables for runtim
 
 Most browsers have native css variable support. In legacy browsers, 
 we use the [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill).
+* The SDK will do automatic resolution of CSS variables on initialization. Variables should be loaded
+before initialization. Overrided variables should be loaded after sdk variables are loaded.
+* You can opt-out with the `disableCssVariablesPonyfill` flag in ANSWERS.init.
 * If you opt-out of automatic resolution of variables, you should call ANSWERS.ponyfillCssVariables()
 after css variables are loaded and before components are added.
 * If you change a css variable value after initialization and wish to see the change in variable
@@ -1628,11 +1631,8 @@ value in a legacy browser, you should call ANSWERS.ponyfillCssVariables() after 
 
 ```js
 ANSWERS.ponyfillCssVariables({
-        onBeforeSend: function() {},
         onError: function() {},
-        onWarning: function() {},
         onSuccess: function() {},
-        onComplete: function() {},
         onFinally: function() {},
 });
 ```
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ function initAnswers() {
     onVerticalSearch: function() {},
     // Optional, analytics callback after a universal search, see onUniversalSearch Configuration for additional details
     onUniversalSearch: function() {},
+    // Optional, opt-out of automatic css variable resolution on init for legacy browsers
+    disableCssVariablesPolyfill: false,
   })
 }
 ```
@@ -1594,3 +1596,43 @@ ANSWERS.formatRichText(rtfFieldValue)
 ```
 
 For instance, this function can be used in the `dataMappings` of a Card to display an RTF attribute. When using this function, you must ensure that the relevant Handlebars template correctly unescapes the value's resultant HTML.
+
+# CSS Variable Styling
+
+The Answers SDK supports styling specific elements with CSS variables for runtime styling.
+* All sdk variables are exposed in the `:root` ruleset
+* All overridden variables must be included the `:root` ruleset
+* All overrides should be defined after Answers CSS is imported and before ANSWERS.init is called
+* To see available variables, see the scss modules in the SDK
+* For example, to override:
+
+```html
+<style>
+  :root {
+    --yxt-font-font-family: sans-serif;
+    --yxt-color-brand-primary: green;
+    --yxt-color-text-primary: #212121;
+    --yxt-searchbar-button-background-color-hover: red;
+    --yxt-nav-border-color: #e9e9e9;
+  }
+</style>
+```
+
+Most browsers have native css variable support. In legacy browsers, 
+we use the [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill).
+* If you opt-out of automatic resolution of variables, you should call ANSWERS.polyfillCssVariables()
+after css variables are loaded and before components are added.
+* If you change a css variable value after initialization and wish to see the change in variable
+value in a legacy browser, you should call ANSWERS.polyfillCssVariables() after the value is changed.
+* We support all callback functions, described [here](https://jhildenbiddle.github.io/css-vars-ponyfill/#/?id=onbeforesend)
+
+```js
+ANSWERS.polyfillCssVariables({
+        onBeforeSend: function() {},
+        onError: function() {},
+        onWarning: function() {},
+        onSuccess: function() {},
+        onComplete: function() {},
+        onFinally: function() {},
+});
+```
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ function initAnswers() {
     // Optional, analytics callback after a universal search, see onUniversalSearch Configuration for additional details
     onUniversalSearch: function() {},
     // Optional, opt-out of automatic css variable resolution on init for legacy browsers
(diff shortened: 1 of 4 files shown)


src/answers-umd.js
```diff
@@ -388,6 +389,19 @@ class Answers {
     this.core.globalStorage.set('queryTrigger', 'initialize');
     this.core.setQuery(searchConfig.defaultInitialSearch);
   }
+
+  /*
+   * Updates the css styles with new current variables. This is useful when the css
+   * variables are updated dynamically (e.g. through js).
+   * @param {Object} config Additional config to pass to the polyfill
+   */
+  polyfillCssVariables (config) {
```
tmeyer2115: Do we need to give users the ability to pass in config here? I think it would be better for us to prescribe how cssVars should be invoked when used with the SDK. A client would need to read the css-vars-ponyfill docs to know what to do here. If they made a mistake, it could cause some issues with their site. It doesn't seem like they'll really need to configure the call to cssVars, provided we set it up correctly.

tmeyer2115 (commented):
I think we should add some info to the README about how users can use CSS variables to affect SDK styling. 

Also, to minimize room for error, I think we should call the polyfillCSSVariables function, internally, as part of ANSWERS.init. I don't think we need to expose it to the user. Clients could easily forget to call it themselves in ANSWERS.init. I'm assuming this function is not too computationally expensive, so it'd be okay to always call it? We'll just need to make it clear in the README that any CSS needs to be loaded before ANSWERS.init is invoked.

tmeyer2115:
It looks like css-vars-ponyfill provides a number of callbacks:

`  // Callbacks
  onBeforeSend: function(xhr, elm, url) {
    // ...
  },
  onError: function(message, elm, xhr, url) {
    // ...
  },
  onWarning: function(message) {
    // ...
  },
  onSuccess: function(cssText, elm, url) {
    // ...
  },
  onComplete: function(cssText, styleElms, cssVariables, benchmark) {
    // ...
  },
  onFinally: function(hasChanged, hasNativeSupport, benchmark) {
    // ...
  }`

Do we need to make sure onReady is invoked in one (or more) of those callbacks? If we don't could we run into a race condition where components are rendered before the cssVars function is complete?

src/answers-umd.js
creotutar: Updated to only allow callbacks!

creotutar:
As discussed, changed to call polyfillCssVariables on ANSWERS.init with the option to opt-out if needed. This is mainly for client integration reasons, should they need to call the ponyfill function at some other time.

README.md
```diff
@@ -124,6 +124,8 @@ function initAnswers() {
     onVerticalSearch: function() {},
     // Optional, analytics callback after a universal search, see onUniversalSearch Configuration for additional details
     onUniversalSearch: function() {},
+    // Optional, opt-out of automatic css variable resolution on init for legacy browsers
+    disableCssVariablesPolyfill: false,
```
tmeyer2115: Ponyfill

src/answers-umd.js
```diff
@@ -172,6 +173,10 @@ class Answers {
     masterSwitchApi.isDisabled(parsedConfig.apiKey, parsedConfig.experienceKey)
       .then(isDisabled => !isDisabled && this._initInternal(parsedConfig, globalStorage, persistentStorage))
       .catch(() => this._initInternal(parsedConfig, globalStorage, persistentStorage));
+
+    if (!parsedConfig.disableCssVariablesPolyfill) {
+      this.polyfillCssVariables();
```
tmeyer2115: Do we need to define onError and onSuccess handlers here that invoke onReady? If we don't could there be a race condition?

src/answers-umd.js
```diff
@@ -241,20 +237,32 @@ class Answers {
...
+      }
+    };
+
     if (parsedConfig.useTemplates === false || parsedConfig.templateBundle) {
       if (parsedConfig.templateBundle) {
         this.renderer.init(parsedConfig.templateBundle);
       }
 
-      this._onReady();
+      callPonyfillCssVariables(this._onReady.bind(this));
```
creotutar: I am worried about having our onReady be guarded by a successful call to `onFinally` by a third part library. I think the tradeoff here is

(1) call onReady whenever
* css variables may not be resolved as components show on the page, showing unstyled components until they are resolved (this would only happen on legacy browsers; IE11 afaik)
* onReady guaranteed to happen

(2) call onReady only when ponyfill is finished
* more risk in relying on third party lib to always call onFinally in every case
* longer wait time for components to show on the page on IE11

I don't swing either way, just expressing my worries about guarding our onReady because it could stop an experience from loading altogether

src/answers-umd.js
```diff
@@ -236,20 +237,32 @@ class Answers {
 
     this._onReady = parsedConfig.onReady || function () {};
 
+    const callPonyfillCssVariables = (callback) => {
```
tmeyer2115: I think we should give this a different name and make it an actual function. I generally feel like anything over a couple of lines shouldn't be an inline function. Maybe we could name the function handlePonyfillCssVariables or something like that? Using the prefix 'call' makes it seem like we will always call the ponyfill. Depending on the value of disableCssVariablesPonyfill, we may or may not.

src/answers-umd.js
tmeyer2115: I'm not too concerned about relying on a 3rd party library to invoke onFinally. We've thoroughly tested this item and their repository is well documented and maintained. I think we've done our due diligence. I think something would need to go very wrong, structurally, with their library for onFinally to not be invoked and that would have been caught by our testing. The longer wait time is another issue. Have you noticed significant slowness in IE11 with this change?

src/answers-umd.js
```diff
@@ -388,6 +401,24 @@ class Answers {
...
     this.core.setQuery(searchConfig.defaultInitialSearch);
   }
+
+  /*
+   * Updates the css styles with new current variables. This is useful when the css
+   * variables are updated dynamically (e.g. through js) or if the css variables are
+   * added after the ANSWERS.init
+   * @param {Object} config Additional config to pass to the ponyfill
+   */
+  ponyfillCssVariables (config = {}) {
```
tmeyer2115, line 417: I think, for now, it's sufficient to just give people hooks into onFinally, onSuccess, and onError. The other callbacks seem more specialized. We can give people access to these if they request it later on.

README.md
```diff
@@ -1593,4 +1595,44 @@ ensure that a Rich Text Formatted value is shown properly on the page. To use th
...
+    --yxt-color-text-primary: #212121;
+    --yxt-searchbar-button-background-color-hover: red;
+    --yxt-nav-border-color: #e9e9e9;
+  }
+</style>
+```
+
+Most browsers have native css variable support. In legacy browsers, 
+we use the [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill).
+* If you opt-out of automatic resolution of variables, you should call ANSWERS.ponyfillCssVariables()
```
tmeyer2115, line 1626: Before this point, I would mention that we will automatically do the reconciliation for them. All they need to do is make sure things get loaded in the correct order.

tmeyer2115: I think we should reiterate here that they can set that boolean to disable this.

src/answers-umd.js
creotutar: good call!

src/answers-umd.js
creotutar: you bet

src/answers-umd.js
```diff
@@ -237,37 +237,37 @@ class Answers {
...
     this.templates = new DefaultTemplatesLoader(templates => {
       this.renderer.init(templates);
-      callPonyfillCssVariables(this._onReady.bind(this));
+      this._handlePonyfillCssVariables(parsedConfig, this._onReady.bind(this));
     });
 
     return this;
   }
 
+  _handlePonyfillCssVariables (parsedConfig, callback) {
```
tmeyer2115: Let's add some JSDoc for this function and it's params. Minor nit, but let's just pass in a ponyfillDisabled booelan. We don't need the entire parsedConfig passed in.

tmeyer2115 (approved):

PR #789, opened 2020-05-18
Pagination: add config.noResults and 'visible' option

oshi97:
This commit adds noResults config to the pagination component,
and allows pagination to accept a 'visible' option in this config.
Uses global noResults if none specified, and does not merge the two configs.

TEST=manual
for regular results, pagination still displays no matter what
for no results, check that will prioritize component level no results
over global no results without merging, and will use 'visible' property
if present otherwise will default to displayAllResults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -798,7 +798,13 @@ ANSWERS.addComponent('Pagination', {
   // Optional, display a double arrow allowing users to jump to the last page of results
   showLast: true,
   // Optional, label for a page of results
-  pageLabel: 'Page'
+  pageLabel: 'Page',
+  // Optional, configuration for the map's behavior when a query returns no results
+  noResults: {
+    // Optional, whether pagination should be visible when displaying no results.
+    // Defaults to false.
+    visible: false
+  }
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -799,7 +799,7 @@ ANSWERS.addComponent('Pagination', {
   showLast: true,
   // Optional, label for a page of results
   pageLabel: 'Page',
-  // Optional, configuration for the map's behavior when a query returns no results
+  // Optional, configuration for the pagination behavior when a query returns no results
   noResults: {
     // Optional, whether pagination should be visible when displaying no results.
     // Defaults to false.
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -3,7 +3,7 @@
 import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
 import DOM from '../../dom/dom';
-import { AnswersComponentError } from '../../../core/errors/errors';
+import { AnswersComponentError, AnswersConfigError } from '../../../core/errors/errors';
 import SearchStates from '../../../core/storage/searchstates';
 import ResultsContext from '../../../core/storage/resultscontext';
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -798,7 +798,13 @@ ANSWERS.addComponent('Pagination', {
   // Optional, display a double arrow allowing users to jump to the last page of results
   showLast: true,
   // Optional, label for a page of results
-  pageLabel: 'Page'
+  pageLabel: 'Page',
+  // Optional, configuration for the map's behavior when a query returns no results
```
tmeyer2115: Update comment

src/ui/components/results/paginationcomponent.js
```diff
@@ -67,6 +67,17 @@ export default class PaginationComponent extends Component {
         this.setState();
       }
     });
+
+    /**
+     * Configuration for the behavior when there are no vertical results.
+     */
+    this._noResults = config.noResults || this.core.globalStorage.getState(StorageKeys.NO_RESULTS_CONFIG);
```
tmeyer2115: Let's break this line up, it's pretty long.

tmeyer2115: Do we need to default to {} if it's not present in the component's config or global storage?

src/ui/components/results/paginationcomponent.js
```diff
@@ -67,6 +67,17 @@ export default class PaginationComponent extends Component {
...
+    if (typeof this._noResults !== 'object') {
```
alexisgrow: I'm not super familiar with typeof checks in JS, since noResults is optional config, does this need a truthy check to make sure noResults is present before the typeof?

src/ui/components/results/paginationcomponent.js
```diff
@@ -79,10 +90,11 @@ export default class PaginationComponent extends Component {
 
   shouldShowControls (results, limit) {
     const hasResults = results.searchState === 'search-complete' && results.resultsCount > limit;
-    const noResultsConfig = this.core.globalStorage.getState(StorageKeys.NO_RESULTS_CONFIG) || {};
-    const showControls = hasResults &&
-      (results.resultsContext === ResultsContext.NORMAL || noResultsConfig.displayAllResults);
-    return showControls;
+    const isNormalResults = results.resultsContext === ResultsContext.NORMAL;
+    const isVisibleForNoResults = 'visible' in this._noResults
```
alexisgrow, line 90: Does `this._noResults` fall back to an empty object? If so, could this be `this._noResults.visible || this._noResults. displayAllResults`?

src/ui/components/results/paginationcomponent.js
oshi97: Yeah you're right it does, I'll just remove this check I don't think it's particularly useful either way

src/ui/components/results/paginationcomponent.js
oshi97: If they set noResults.visible to be false I didn't want it to be true if displayAllResults happened to be true as well. I doubt anybody would ever set both but it feels a little more accurate to me this way

src/ui/components/results/paginationcomponent.js
oshi97: yeah missed this

nbramblett (approved):
Seems legit to me

tmeyer2115 (approved):

src/ui/components/results/paginationcomponent.js
nbramblett: typeof should just return 'null' or 'undefined' in that case

PR #790, opened 2020-05-18
Facets: hide facets for no results

oshi97:
When displaying no results, for now facets should not render.
TEST=manual
normal search: facets still render
no results search: no facets display on the page anymore, no container in the dom either

### DIFF
--- a/README.md
+++ b/README.md
@@ -879,7 +879,7 @@ ANSWERS.addComponent('FilterBox', {
 
 This component is only for Vertical pages.
 
-The Facets component displays filters relevant to the current search, configured on the server, automatically.
+The Facets component displays filters relevant to the current search, configured on the server, automatically. The Facets component will hide itself when itself when displaying no results.
 
 ```html
 <div class="facets-container"></div>
--- a/README.md
+++ b/README.md
@@ -879,7 +879,7 @@ ANSWERS.addComponent('FilterBox', {
 
 This component is only for Vertical pages.
 
-The Facets component displays filters relevant to the current search, configured on the server, automatically. The Facets component will hide itself when itself when displaying no results.
+The Facets component displays filters relevant to the current search, configured on the server, automatically. The Facets component will be hidden when a query returns no results.
 
 ```html
 <div class="facets-container"></div>
--- a/src/ui/components/filters/facetscomponent.js
+++ b/src/ui/components/filters/facetscomponent.js
@@ -2,6 +2,7 @@
 
 import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
+import ResultsContext from '../../../core/storage/resultscontext';
 
 class FacetsConfig {
   constructor (config) {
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -879,7 +879,7 @@ ANSWERS.addComponent('FilterBox', {
 
 This component is only for Vertical pages.
 
-The Facets component displays filters relevant to the current search, configured on the server, automatically.
+The Facets component displays filters relevant to the current search, configured on the server, automatically. The Facets component will hide itself when itself when displaying no results.
```
tmeyer2115: "The Facets component will be hidden when a query returns no results"?

src/ui/components/filters/facetscomponent.js
```diff
@@ -177,9 +185,9 @@ export default class FacetsComponent extends Component {
       this._filterbox.remove();
     }
 
-    let { filters } = this._state.get();
+    let { filters, resultsContext } = this._state.get();
 
-    if (!filters) {
+    if (!filters || resultsContext === ResultsContext.NO_RESULTS) {
```
tmeyer2115, line 190: For now, would it just be easier to have the SearchDataTransformer._parseVerticalResponse return a response with an empty list of facets if ResultsContext.NO_RESULTS? Would that work?

oshi97: The future plan is to have some specific behavior for facets w/no results though right?

README.md

src/ui/components/filters/facetscomponent.js
oshi97: I think bryan said they wanted something weird like when displaying no results, display all facets, and if any are applied reset the search to blank string and apply the facets

tmeyer2115: We'd talked about how they might want this to work in the future. The current thinking is to have it behave like you said. That hasn’t been finalized in a spec though and a lot has been subject to change with noResults. I'd rather not leak the resultsContext implementation detail into Facets for now, if we can avoid it. It adds to the complexity of the component and serves a use-case we don't yet have final confirmation on. If we do get the go-ahead for the more complex behavior, we can always add resultsContext then and undo the small change to _parseVerticalResponse. We're not painting ourselves into a corner by leaving it out now and make the Facets component simpler by doing so. If you feel strongly, though, we can keep it in! 😃

oshi97: Oh there was one more thing when I was testing this on friday, even if you have facets as [] the component container and things Facets title will still render. Otherwise I think I would definitely agree with you and just return an empty list of facets!

tmeyer2115: Yeah, I didn't see this was in onMount. That could cause issues with the approach I was proposing too. The component might get mounted due to a valid query. If someone makes a subsequent query that returns noResults (and thus an empty array of facets), we'd need this component to disappear. But, I don't think onMount would get called again in that case, right? So, we definitely need to add this to setState as well.

tmeyer2115 (approved):

PR #791, opened 2020-05-18
SortOptions: hide when displaying no results

oshi97:
This commit hides the sort options component when the
resultsContext is NO_RESULTS

TEST=manual
saw the sort options disappear for no results and reappear for normal results

### DIFF
--- a/README.md
+++ b/README.md
@@ -1273,6 +1273,7 @@ ANSWERS.addComponent('LocationBias', {
 ## Sort Options Component
 
 The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search.
+The Facets component will hide itself when itself when displaying no results.
 Currently, there may be only one sort options component per page.
 
 ```html
--- a/README.md
+++ b/README.md
@@ -1272,8 +1272,7 @@ ANSWERS.addComponent('LocationBias', {
 
 ## Sort Options Component
 
-The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search.
-The sort options component will hide itself when itself when displaying no results.
+The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search. When a query returns no results, the component will not be rendered on the page.
 Currently, there may be only one sort options component per page.
 
 ```html
--- a/README.md
+++ b/README.md
@@ -1273,7 +1273,7 @@ ANSWERS.addComponent('LocationBias', {
 ## Sort Options Component
 
 The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search.
-The Facets component will hide itself when itself when displaying no results.
+The sort options component will hide itself when itself when displaying no results.
 Currently, there may be only one sort options component per page.
 
 ```html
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1273,6 +1273,7 @@ ANSWERS.addComponent('LocationBias', {
 ## Sort Options Component
 
 The sort options component displays a list of radio buttons that allows users to sort the results of a vertical search.
+The Facets component will hide itself when itself when displaying no results.
```
tmeyer2115: This is confusingly worded.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -21,9 +22,16 @@ export default class SortOptionsComponent extends Component {
     this.options[this.selectedOptionIndex].isSelected = true;
     this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex < this._config.showMoreLimit;
     this.showReset = this._config.showReset && this.selectedOptionIndex !== 0;
+
+    /**
+     * This component listens to updates to vertical results, and sets its state to it when
+     * an update occurs.
+     * @type {string}
+     */
+    this.moduleId = StorageKeys.VERTICAL_RESULTS;
```
tmeyer2115, line 31: This makes me a bit nervous. Are we certain this won't have unintended consequences? This component will get re-rendered every time there is an update to the VerticalResults.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -21,9 +22,16 @@ export default class SortOptionsComponent extends Component {
...
   }
 
-  setState (data) {
+  setState (data = {}) {
```
tmeyer2115, line 34: Do we need to update the HBS template for this component? If resultsContext is NO_RESULTS, could we just return here instead of calling super.setState? That'd save us having to clutter the HBS with more conditional logic.

src/ui/components/filters/sortoptionscomponent.js
oshi97: Double checked this by 
1. click show more
2. choose non default sort option within show more
3. ran normal search, sort option did not change
4. ran no results search, sort option disappears
5. ran another normal search, sort option came back with same show/more and selected option state

I think this makes sense since everything in the sort options component is tied pretty closely to a variable stored on the component like this.selectedOptionIndex. I should have checked this before though.;

src/ui/components/filters/sortoptionscomponent.js
oshi97: Tried it, I don't think things work like that because the component's `render()` method is called during `mount()`, and `mount()` is called when `this._state.on('update', () =>`, which needs the setState call. Even then the template still needs to have some conditional logic otherwise everything will render since changing setState only changes what's passed into the template not the actual template itself. Basically fake react strikes again.

README.md
oshi97: updated! is this better?

tmeyer2115 (approved):

PR #792, opened 2020-05-18
SortOptions: add null checking to onMount

oshi97:
The SortOptions component was throwing console errors because it was trying
to register listeners for undefined elements, since when no results are
returned from a search new behavior was added to hide the sort options component.

TEST=manual
no more console errors when loading a page with a search that returns no results

### DIFF
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -48,29 +48,42 @@ export default class SortOptionsComponent extends Component {
   onMount () {
     // Handle radio button selections
     const containerEl = DOM.query(this._container, '.yxt-SortOptions-fieldSet');
-    containerEl && DOM.on(containerEl, 'change', evt =>
-      this.handleOptionSelection(parseInt(evt.target.value))
+    containerEl && DOM.on(
+      containerEl,
+      'change',
+      evt => this.handleOptionSelection(parseInt(evt.target.value))
     );
 
     // Register more/less button
     if (this._config.showMore) {
       const toggleEl = DOM.query(this._container, '.yxt-SortOptions-showToggle');
-      toggleEl && DOM.on(toggleEl, 'click', () => {
-        this.hideExcessOptions = !this.hideExcessOptions;
-        this.setState();
-      });
+      toggleEl && DOM.on(
+        toggleEl,
+        'click', () => {
+          this.hideExcessOptions = !this.hideExcessOptions;
+          this.setState();
+        }
+      );
     }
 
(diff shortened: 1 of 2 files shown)


src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -46,6 +46,9 @@ export default class SortOptionsComponent extends Component {
   }
 
   onMount () {
+    if (this.getState('resultsContext') === ResultsContext.NO_RESULTS) {
+      return;
+    }
```
alexisgrow: Ooh what do you think about adding null checks for the selectors before setting the on-click listeners?

oshi97: oo at first I was like "but I kind of like how there's no safety net and I will get console errors if I make a mistake". Then I realized I could just add a unit test

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -47,40 +47,30 @@ export default class SortOptionsComponent extends Component {
 
   onMount () {
     // Handle radio button selections
-    DOM.on(
-      DOM.query(this._container, '.yxt-SortOptions-fieldSet'),
-      'change',
-      evt => this.handleOptionSelection(parseInt(evt.target.value))
+    const containerEl = DOM.query(this._container, '.yxt-SortOptions-fieldSet');
+    containerEl && DOM.on(containerEl, 'change', evt =>
+      this.handleOptionSelection(parseInt(evt.target.value))
```
tmeyer2115: Every argument to DOM.on should be on its own line since not all the params can fit on one line

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -47,40 +47,30 @@ export default class SortOptionsComponent extends Component {
...
-      DOM.on(
-        DOM.query(this._container, '.yxt-SortOptions-showToggle'),
-        'click',
-        () => {
-          this.hideExcessOptions = !this.hideExcessOptions;
-          this.setState();
-        }
-      );
+      const toggleEl = DOM.query(this._container, '.yxt-SortOptions-showToggle');
+      toggleEl && DOM.on(toggleEl, 'click', () => {
```
tmeyer2115: Same thing about breaking up the lines per param

src/ui/components/filters/sortoptionscomponent.js
oshi97: updated!

tmeyer2115 (approved):

PR #797, opened 2020-05-18
MapComponent: add config.noResults.visible flag

oshi97:
This commit adds the visible option to the map's noResults config.
This will take priority over showEmptyMap when no results are returned.

TEST=manual
tested show empty map works still
tested that noResults.visible takes priority over showEmptyMap for no results

### DIFF
--- a/README.md
+++ b/README.md
@@ -1376,8 +1376,11 @@ ANSWERS.addComponent('Map', {
   onLoaded: function () {},
   // Optional, configuration for the map's behavior when a query returns no results
   noResults: {
-    // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
-    displayAllResults: true
+    // Optional, whether the map should be visible when displaying no results.
+    // Defaults to false.
+    visible: false,
+    // DEPRECATED, please use 'visible' instead.
+    displayAllResults: false
   },
   // Optional, the custom configuration override to use for the map markers, function
   pin: function () {
--- a/README.md
+++ b/README.md
@@ -1376,11 +1376,10 @@ ANSWERS.addComponent('Map', {
   onLoaded: function () {},
   // Optional, configuration for the map's behavior when a query returns no results
   noResults: {
-    // Optional, whether the map should be visible when displaying no results.
-    // Defaults to false.
-    visible: false,
-    // DEPRECATED, please use 'visible' instead.
-    displayAllResults: false
+    // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
+    displayAllResults: false,
+    // Optional, whether the map should be visible when no results are returned. This takes priority over showEmptyMap. Defaults to false
+    visible: false
   },
   // Optional, the custom configuration override to use for the map markers, function
   pin: function () {
--- a/README.md
+++ b/README.md
@@ -1376,9 +1376,9 @@ ANSWERS.addComponent('Map', {
   onLoaded: function () {},
   // Optional, configuration for the map's behavior when a query returns no results
   noResults: {
-    // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
+    // Optional, whether to display map pins for all possible results when no results are found. Defaults to false.
     displayAllResults: false,
-    // Optional, whether the map should be visible when no results are returned. This takes priority over showEmptyMap. Defaults to false
+    // Optional, whether to display the map when no results are found, taking priority over showEmptyMap. If displayAllResults is false, this will be an empty map. Defaults to false.
     visible: false
   },
   // Optional, the custom configuration override to use for the map markers, function
--- a/src/ui/components/map/mapcomponent.js
+++ b/src/ui/components/map/mapcomponent.js
@@ -79,8 +79,11 @@ export default class MapComponent extends Component {
     if (Object.keys(data).length === 0) {
       return this;
     }
-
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1376,8 +1376,11 @@ ANSWERS.addComponent('Map', {
   onLoaded: function () {},
   // Optional, configuration for the map's behavior when a query returns no results
   noResults: {
-    // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
-    displayAllResults: true
+    // Optional, whether the map should be visible when displaying no results.
```
nbramblett: I think this is pretty misleading given that it actually still determines whether to display all results. The above description would fit more for showEmptyMap.

oshi97: sorry one sec I misread something

README.md
```diff
@@ -1377,7 +1377,9 @@ ANSWERS.addComponent('Map', {
   // Optional, configuration for the map's behavior when a query returns no results
   noResults: {
     // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
```
tmeyer2115: Is this documentation correct? It references a no results card. That won't be shown by the Map component.

src/ui/components/map/mapcomponent.js
```diff
@@ -81,7 +81,9 @@ export default class MapComponent extends Component {
     }
 
     if (data.resultsContext === ResultsContext.NO_RESULTS && !this._noResults.displayAllResults) {
```
tmeyer2115, line 83: Does this conditional need to be updated to check _noResults.visible as well?

README.md
tmeyer2115: Do we also need to update the "this._noResults" block in the constructor?

src/ui/components/map/providers/mapprovider.js
```diff
@@ -88,6 +95,14 @@ export default class MapProvider {
     };
   }
 
+  shouldHideMap (mapData, resultsContext) {
```
tmeyer2115, line 98: What if someone had not set noResults.visible but had _showEmptyMap as true? Is that handled correctly here? In that case, if there's a query that doesn't return actual results, wouldn't they expect to see a Map?

tmeyer2115 (commented):
Disclaimer: This is me venting. I know these code changes are unavoidable and I appreciate you tackling them!

The interface for this component (and others) has become very confusing. These "backwards compatible" changes are, I feel, very problematic. They will confuse users and already confuse us. It adds significant complexity to our code and prevents us from having a firm grasp of how our components work. If we struggle to understand how all these new and "backwards compatible" options work together, imagine how a client will feel. I think we need to make a V2 of the library sooner rather than later where we abandon these "backward compatible" options. 

In the mean time, I think we should take full advantage of Component config classes. In general, I think these classes have been overused. Here, though, they serve an actual purpose. They translate the config provided by the user into a configuration object that can be understood by the Component. I think that the Components should be written to work with only the most recent configuration options. A Component should not know about any legacy, deprecated configuration. Instead, it is the job of the config class to translate deprecated config into modern config.

src/ui/components/map/providers/mapprovider.js
oshi97: I think in that case we want to hide the map. The spec says if noResults is not specified, "don't show" for no results.

oshi97: i.e. we don't need to fall back to past showEmptyMap behavior

tmeyer2115: I'd check with Rose about the behavior in this case. I'm not sure if the spec meant "Default if noResults not specified" or "Default if noResults not specified and showEmptyMap not specified"

src/ui/components/map/mapcomponent.js
oshi97: asking Rose, I interpreted this as noResults.visible only controls whether an empty map will show up, and will not affect whether map pins will show up or not (which is what displayAllResults does). I think I likely misunderstood things.

src/ui/components/map/providers/mapprovider.js
oshi97: ~~yup just posted in slack~~

edit: 
sorry I misread your comment, if both showEmptyMap and noResults are not specified we should almost definitely hide the map. If showEmptyMap is unspecified it will default to false, and if noResults.visible is unspecified OR noResults is unspecified it should also default to false

oshi97: Ok apparently I didn't misread things, 
| so noResults.visible is true, show the map on no results. if noResults.displayAllResults is also true, show the pins on the map

src/ui/components/map/mapcomponent.js
oshi97: > so noResults.visible is true, show the map on no results. if
> noResults.displayAllResults is also true, show the pins on the map

Ok apparently I didn't misread things

oshi97: In other words, no this conditional should not be updated.

README.md
oshi97: updated the readme and the this._noResults block in mapcomponent.js's constructor. I wasn't sure if it made sense to be explicit about 'visible' in that file, since it only gets used in mapprovider.js not mapcomponent.js, but on second thought I don't know where else it would go.

src/ui/components/map/providers/mapprovider.js
oshi97: if only noResults.visible is not specified it should still hide the map for no results (based on the spec at least)

tmeyer2115 (approved):

PR #801, opened 2020-05-19
FilterOptions: appliedOnLoad setting for options

oshi97:
This commit adds the appliedOnLoad flag for options.
Options with appliedOnLoad: true will be automatically
checked on page load, if no prior options were stored in persistent storage.
If the filteroptions mode is single-option, then only the first option
marked as appliedOnLoad will be selected.

TEST=manual
tested that for singleoption, only the first option marked as appliedOnLoad
   is selected
tested that for multioption, all options are selected
tested using both filterbox and filteroptions
tested that persistent storage will take priority by unchecking boxes and
   then refreshing the page, but refreshing the page AND erasing url params
   will result in the original appliedOnLoad options

### DIFF
--- a/README.md
+++ b/README.md
@@ -1072,6 +1072,9 @@ Filter components can be used in a FilterBox or on their own to affect a search.
 ### FilterOptions
 
 FilterOptions displays a set of filters with either checkboxes or radio buttons.
+As a user interacts with FilterOptions, information on which options are selected
+is stored in the url. Returning to that same url will load the page with those saved
+options already selected.
 
 ```html
 <div class="filter-container"></div>
@@ -1093,12 +1096,16 @@ ANSWERS.addComponent('FilterOptions', {
       // The api field to filter on, configured on the Yext platform
       field: 'c_openNow',
       // The value for the above field to filter by
-      value: true
+      value: true,
+      // Whether this option will be selected on page load. Selected options stored in the url
+      // will take priority over appliedOnLoad. Defaults to false.
+      appliedOnLoad: false
     },
     {
       label: 'Dog Friendly',
       field: 'c_dogFriendly',
-      value: true
+      value: true,
+      appliedOnLoad: true
     },
     {
       label: 'Megastores',
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -135,7 +135,7 @@ class FilterOptionsConfig {
   /**
    * If no previous options were stored in persistentStorage, default to options marked
    * as appliedOnLoad.
-   * @param {*} options 
+   * @param {*} options
    */
   setAppliedOnLoad (options) {
     if (this.control === 'singleoption') {
@@ -144,7 +144,7 @@ class FilterOptionsConfig {
       firstAppliedOption.selected = true;
       return _options;
     }
-    return options.map(o =>  ({
+    return options.map(o => ({
       ...o,
       selected: o.appliedOnLoad
     }));
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -110,36 +110,44 @@ class FilterOptionsConfig {
         config.previousOptions = [];
       }
     }
-    const previousOptions = config.previousOptions || [];
-    this.options = this.setDefaultSelectedValues(this.options, previousOptions);
+    // previousOptions will be null if there were no previousOptions in persistentStorage
+    const previousOptions = config.previousOptions;
+    if (previousOptions) {
+      this.options = this.setPreviousOptions(this.options, previousOptions || []);
+    } else {
+      this.options = this.setAppliedOnLoad(this.options);
+    }
   }
 
   /**
-   * Sets selected options on load based on options stored in persistent storage,
-   * and if persistent storage is empty then the appliedOnLoad flag.
+   * Sets selected options on load based on options stored in persistent storage.
    * @param {Array<Object>} options
    * @param {Array<string>} previousOptions
    * @returns {Array<Object>}
    */
-  setDefaultSelectedValues (options, previousOptions) {
-    if (previousOptions.length) {
-      return options.map(o => ({
-        ...o,
-        selected: previousOptions.length
-          ? previousOptions.includes(o.label)
-          : o.selected
-      }));
+  setPreviousOptions (options, previousOptions) {
+    return options.map(o => ({
+      ...o,
+      selected: previousOptions.includes(o.label)
+    }));
+  }
+
+  /**
+   * If no previous options were stored in persistentStorage, default to options marked
+   * as appliedOnLoad.
+   * @param {*} options 
+   */
+  setAppliedOnLoad (options) {
+    if (this.control === 'singleoption') {
+      const _options = options.map(o => ({ ...o }));
+      const firstAppliedOption = _options.find(o => o.appliedOnLoad);
+      firstAppliedOption.selected = true;
+      return _options;
     }
-    for (let o of options) {
-      if (!o.appliedOnLoad) {
-        continue;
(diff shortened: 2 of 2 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,17 +110,36 @@ class FilterOptionsConfig {
...
-  setDefaultSelectedValues (options, selectedOptions) {
-    return options.map(o => ({
-      ...o,
-      selected: selectedOptions.length
-        ? selectedOptions.includes(o.label)
-        : o.selected
-    }));
+  /**
+   * Sets selected options on load based on options stored in persistent storage,
+   * and if persisten storage is empty then the appliedOnLoad flag.
```
nbramblett: *persistent

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,17 +110,36 @@ class FilterOptionsConfig {
...
+  setDefaultSelectedValues (options, previousOptions) {
+    if (previousOptions.length) {
+      return options.map(o => ({
+        ...o,
+        selected: previousOptions.length
+          ? previousOptions.includes(o.label)
+          : o.selected
+      }));
+    }
+    for (let o of options) {
```
nbramblett: I believe this loop could be replaced with 

```
const filterCond = o => !o.appliedOnLoad && o.selected;
return this.control === 'singleoption' ? options.find(filterCond) : options.filter(filterCond);
```

src/ui/components/filters/filteroptionscomponent.js
oshi97: *facepalm*

src/ui/components/filters/filteroptionscomponent.js
oshi97: naw we need to return all options

nbramblett: Ah whoops I misunderstood what this was doing.

nbramblett (approved):

README.md
```diff
@@ -1093,12 +1096,16 @@ ANSWERS.addComponent('FilterOptions', {
       // The api field to filter on, configured on the Yext platform
       field: 'c_openNow',
       // The value for the above field to filter by
-      value: true
+      value: true,
+      // Whether this option will be selected on page load. Selected options stored in the url
+      // will take priority over appliedOnLoad. Defaults to false.
+      appliedOnLoad: false
```
tmeyer2115, line 1102: To confirm, there's no intelligent merging logic here, right? We don't try and reconcile the selected options in the url params with appliedOnLoad'd options, right? If there are options in the url params, that's all we show.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,17 +110,36 @@ class FilterOptionsConfig {
         config.previousOptions = [];
       }
     }
-    let selectedOptions = config.previousOptions || [];
-    this.options = this.setDefaultSelectedValues(this.options, selectedOptions);
+    const previousOptions = config.previousOptions || [];
```
tmeyer2115: For my own edification what are previousOptions? Is this what's stored in the URL? If so, it's confusing that this data is coming from the config. Why not query persistent storage or global storage directly?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,17 +110,36 @@ class FilterOptionsConfig {
...
+   * and if persistent storage is empty then the appliedOnLoad flag.
+   * @param {Array<Object>} options
+   * @param {Array<string>} previousOptions
+   * @returns {Array<Object>}
+   */
+  setDefaultSelectedValues (options, previousOptions) {
+    if (previousOptions.length) {
+      return options.map(o => ({
+        ...o,
+        selected: previousOptions.length
```
tmeyer2115: This check seems redundant.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,17 +110,36 @@ class FilterOptionsConfig {
...
+          ? previousOptions.includes(o.label)
+          : o.selected
+      }));
+    }
+    for (let o of options) {
```
tmeyer2115: We should be consistent in this method about how we're iterating through options and returning a result. Above, we were using Array.map, but here we're using a for loop. I'm partial to the .map way of doing things since it doesn't modify the options param itself. I think a .forEach could work here as well. That would let you break early and we could still use the more functional style of iteration. With a forEach though, I think we'd want to be operating on a copy of options.

src/ui/components/filters/filteroptionscomponent.js
oshi97: We have this very special block in the constructor that passes "previousOptions" into the config class

```
    let previousOptions = this.core.globalStorage.getState(this.name);
    this.core.globalStorage.delete(this.name);

    /**
     * The component config
     * @type {FilterOptionsConfig}
     */
    this.config = new FilterOptionsConfig({
      previousOptions,
      ...config
    });
```

oshi97: so even though calling it config.previousOptions might make you think it's from the config, it's not! pretty fun

oshi97: Would it make more sense to move this to a separate parameter?

src/ui/components/filters/filteroptionscomponent.js
oshi97: This check is here so that if there are any options in persistentStorage, we use those. Otherwise we use appliedOnLoad

src/ui/components/filters/filteroptionscomponent.js
tmeyer2115: That's pretty weird. Don't worry about it for now, this was before your PR! Let's just make an item to address.

oshi97: but how can I scope creep the entire codebase into this then

src/ui/components/filters/filteroptionscomponent.js
oshi97: Ah okay you're right sorry about that. This isn't really an excuse but Github made it look like you were talking about the first check
image

src/ui/components/filters/filteroptionscomponent.js
oshi97: Refactored!

oshi97:
I realized I missed an edge case in my original commit. When persistentStorage contains an empty array, we should treat that as no options selected on page load, rather than defaulting to appliedOnLoad. Updated and tested this behavior.

README.md
oshi97: yeah, was very careful about this.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -110,16 +110,43 @@ class FilterOptionsConfig {
...
+      const firstAppliedOption = _options.find(o => o.appliedOnLoad);
+      firstAppliedOption.selected = true;
+      return _options;
+    }
     return options.map(o => ({
       ...o,
-      selected: selectedOptions.length
-        ? selectedOptions.includes(o.label)
-        : o.selected
+      selected: o.appliedOnLoad
```
alexisgrow: how does this work with the legacy "selected" config option?

oshi97: Ohh shoot I did some more digging and that was being used by facets, updated and rested things! (only tested FilterBox and FilterOptions the first time)

alexisgrow (approved):
LGTM!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Oh nice! sweet this lg!

PR #804, opened 2020-05-20
DynamicFilters: null handling for response.facets

oshi97:
We aren't guaranteed to receive an allResultsForVertical.facets
or allResultsForVertical property in the vertical search response.

TEST=manual
copied rose's qa site with this issue
saw that no results query, with response missing allResultsForVertical,
renders no results correctly

### DIFF
--- a/src/core/models/dynamicfilters.js
+++ b/src/core/models/dynamicfilters.js
@@ -25,7 +25,7 @@ export default class DynamicFilters {
    * @returns {DynamicFilters}
    */
   static from (response) {
-    const { facets, resultsContext } = response;
+    const facets = response.facets || [];
     const dynamicFilters = facets.map(f => ({
       label: f['displayName'],
       fieldId: f['fieldId'],
@@ -39,7 +39,7 @@ export default class DynamicFilters {
 
     return new DynamicFilters({
       filters: dynamicFilters,
-      resultsContext: resultsContext
+      resultsContext: response.resultsContext
     });
   }
 }


src/core/models/dynamicfilters.js
```diff
@@ -39,7 +39,7 @@ export default class DynamicFilters {
 
     return new DynamicFilters({
       filters: dynamicFilters,
-      resultsContext: resultsContext
+      resultsContext: response.resultsContext
```
creotutar, line 42: can you access an array with a dot modifier?

oshi97: oh response isn't an array! it's an object

oshi97: response.facets is the array

creotutar: im silly

creotutar: does it matter if resultsContext is empty?

creotutar (approved):
🚀 🚀 🚀 LGTM

src/core/models/dynamicfilters.js
oshi97: is fine in this case, but it should never be empty! the _parseVerticalResponse method in searchdatatransformer.js always assigns a resultsContext. I wanted to avoid adding defaults for results context everywhere it's used if possible, and just trusting that at the very top we guarantee it always has a value.

PR #809, opened 2020-05-21
FilterBox: allow custom apply logic for children filters

oshi97:
FilterBox was deciding on the behavior of children behavior when the
were applied. This causes difficulty for the RADIUS_FILTER, since
it doesn't want to use this behavior.

It's possible this change is only needed in FilterOptions, and not in
DateRangeFilter and RangeFilter. It IS possible to add both a DateRangeFilter
and a RangeFilter to a FilterBox, however you have to add a dummy options
config option, like options: [{}], otherwise FilterBox will throw an error.
This leads me to believe FilterBox is only meant to work with FilterOptions.
I added this change to those files but would not mind removing changes to those files. The other
filters do not have a getFilter() method and cannot be hacked into a FilterBox, so were not touched.

Functionally speaking, this change does not change anything. Before,
FilterBox was AND-ing together all of its children filters, then adding this
AND filter to the array of filters in globalStorage. In the search request,
this globalStorage array of filters would again be AND-ed together. Now,
FilterBox's children filters are added individually to the globalStorage
array of filters, and then all of these filters are AND-ed together.
Since ANDing is associative this causes no logical change.

TEST=manual
create a FilterBox with FilterOptions, RangeFilter, and DateRangeFilter.
inspect networks tab in dev tools and check that search request creates expected filters

### DIFF
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -122,11 +122,6 @@ export default class DateRangeFilterComponent extends Component {
     return this._buildFilter();
   }
 
-  apply () {
-    const filter = this._buildFilter();
-    this.core.setFilter(this.name, filter);
-  }
-
   /**
    * Updates the current state of the date range
    * @param {string} key The key for the date value
@@ -139,7 +134,7 @@ export default class DateRangeFilterComponent extends Component {
 
     const filter = this._buildFilter();
     if (this._storeOnChange) {
-      this.apply();
+      this.core.setFilter(this.name, filter);
     }
     this.core.persistentStorage.set(`${this.name}.min`, this._date.min);
     this.core.persistentStorage.set(`${this.name}.max`, this._date.max);
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -122,6 +122,11 @@ export default class DateRangeFilterComponent extends Component {
     return this._buildFilter();
   }
 
+  apply () {
+    const filter = this._buildFilter();
+    this.core.setFilter(this.name, filter);
+  }
+
   /**
    * Updates the current state of the date range
    * @param {string} key The key for the date value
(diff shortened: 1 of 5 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -102,7 +101,7 @@ class FilterBoxConfig {
      * The list of filters to display and control, ignoring empty sections
      * @type {object[]}
      */
-    this.filterConfigs = config.filters.filter(f => f.options.length !== 0);
```
oshi97, line 105: This line makes me think FilterBox was only ever meant to support FilterOptions

oshi97: You can hack either DateRangeFilter or RangeFilter into FilterBox by specifying options: [{}] in the config, though, and they will work as expected (or should I say as unexpected?)

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -122,6 +122,11 @@ export default class DateRangeFilterComponent extends Component {
     return this._buildFilter();
   }
 
+  apply () {
```
tmeyer2115: A hack is required to add the range filters to FilterBox, right? Let's first confirm with Rose that these filters are not really meant to be used with FilterBox. If that's the case, then I'd say let's not modify them to work with FilterBox. If they should be supported in FilterBox, I think the first step would be to retool FilterBox so that a hack is not needed to support the range filters.

tmeyer2115 (commented):
To confirm, the FilterBox was AND'ing all of its child filters together. This AND'd filter was then made part of another AND (to accommodate filters outside the FilterBox) before the query was issued. Now we just have that one, final round of AND'ing before a query is submitted? 

I think we could potentially use this child "apply" pattern to further simplify FilterBox. Right now, FilterBox has some conditional logic in _saveFiltersToStorage to handle Facets vs. other Filters. If we added an "apply" method to Facets, I think we could get rid of that conditional logic in FilterBox. It would always just call the apply of its children and trust them to handle storage properly. I think that's a better design for FilterBox in general. Delegate storage handing to the child, constituent filters. Could we add a to-do to Facets for this?

src/ui/components/filters/daterangefiltercomponent.js
oshi97: asked on slack! I also highly doubt anybody adds a range filter to FilterBox. https://yext.slack.com/archives/GREFVUD7V/p1590157084068000

oshi97: actually, RangeFilter and DateRangeFilter are missing the clearOptions method FIlterBox uses to clear FIlterOptions, so I think it's very safe to say we can ignore them.

alexisgrow (approved):
This LGTM! 🔥

PR #810, opened 2020-05-22
Geolocation: add geolocationOptions and geolocationTimeoutAlert config

oshi97:
This commit adds the geolocationOptions, which directly set options
used by the geolocation api, and geolocationTimeoutAlert, which
sets whether a window.alert should appear when geolocation fails and
what message it contains.

Specaroni:
https://docs.google.com/document/d/1Dw69VKPn-tsrMy1hQNgFBZ5i4n1WN7UCkSSo0Dgs128/edit#

TEST=manual
Check that setting timeout 0, will cause geolocation to fail, and
that when this happens the window.alert is disabled by default, but can
be turned on, and the message will default correctly but can also be set.
Did not test maximumAge or enableHighAccuracy (not sure how?) but double
checked my spelling.
tested the above for all 3 components

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,17 +398,17 @@ ANSWERS.addComponent('SearchBar', {
   useForm: 'true',
   // Optional, the input element used for searching and wires up the keyboard interaction
   inputEl: '.js-yext-query',  
-  // Optional, options to pass to the geolocation api.
+  // Optional, options to pass to the geolocation api, which is used to fetch the user's current location.
   // https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions
   geolocationOptions: {
     // Optional, whether to improve accuracy at the cost of response time and/or power consumption, defaults to false.
     enableHighAccuracy: false,
-    // Optional, the maximum amount of time (in ms) a geolocation call is allowed to take before defaulting, defaults to 1000ms.
+    // Optional, the maximum amount of time (in ms) a geolocation call is allowed to take before defaulting, defaults to 1 second.
     timeout: 1000,
-    // Optional, the maximum amount of time (in ms) to cache a geolocation call, defaults to 300000ms (5 minutes).
+    // Optional, the maximum amount of time (in ms) to cache a geolocation call, defaults to 5 minutes.
     maximumAge: 300000,
   },
-  // Optional, options for the alert given when geolocation defaults.
+  // Optional, options for an alert when the geolocation call fails.
   geolocationTimeoutAlert: {
     // Optional, whether to display a window.alert() on the page, defaults to false.
     enabled: false,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -397,7 +397,24 @@ ANSWERS.addComponent('SearchBar', {
   // Note that WCAG compliance is not guaranteed if a form is not used as the context.
   useForm: 'true',
   // Optional, the input element used for searching and wires up the keyboard interaction
-  inputEl: '.js-yext-query'
+  inputEl: '.js-yext-query',  
+  // Optional, options to pass to the geolocation api.
+  // https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions
```
tmeyer2115, line 402: I think we should add a quick blurb mentioning that we use the geolocation api to fetch a user's current location.

README.md
```diff
@@ -397,7 +397,24 @@ ANSWERS.addComponent('SearchBar', {
...
+  geolocationOptions: {
+    // Optional, whether to improve accuracy at the cost of response time and/or power consumption, defaults to false.
+    enableHighAccuracy: false,
+    // Optional, the maximum amount of time (in ms) a geolocation call is allowed to take before defaulting, defaults to 1000ms.
+    timeout: 1000,
+    // Optional, the maximum amount of time (in ms) to cache a geolocation call, defaults to 300000ms (5 minutes).
+    maximumAge: 300000,
+  },
+  // Optional, options for the alert given when geolocation defaults.
```
tmeyer2115: options for an alert when the geolocation call fails?

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -217,11 +238,20 @@ export default class GeoLocationComponent extends Component {
...
         },
-        () => this.setState({ geoError: true })
+        () => this._handleGeolocationError(),
+        this._geolocationOptions
       );
     }
   }
 
+  _handleGeolocationError () {
+    this.setState({ geoError: true });
```
tmeyer2115, line 248: What behavior is triggered by geoError true? There's not already some kind of UI notification is there?

src/ui/components/search/locationbiascomponent.js
```diff
@@ -70,16 +91,23 @@ export default class LocationBiasComponent extends Component {
             radius: position.coords.accuracy
           });
           this._doSearch();
-        }, (err) => {
-          if (err.code === 1) {
-            this._disableLocationUpdate();
-          }
-        });
+        }, (err) => this._handleGeolocationError(err),
```
tmeyer2115: Move to new line

src/ui/components/search/searchcomponent.js
```diff
@@ -494,7 +515,13 @@ export default class SearchComponent extends Component {
...
                   resolve(this.search(query));
                 },
-                () => resolve(this.search(query)))
+                () => {
+                  resolve(this.search(query));
+                  const { enabled, message } = this._geolocationTimeoutAlert;
+                  if (enabled) {
+                    window.alert(message);
+                  }
+                }, this._geolocationOptions)
```
tmeyer2115: this._geolocationOptions on a newline

README.md
```diff
@@ -397,7 +397,24 @@ ANSWERS.addComponent('SearchBar', {
...
   useForm: 'true',
   // Optional, the input element used for searching and wires up the keyboard interaction
-  inputEl: '.js-yext-query'
+  inputEl: '.js-yext-query',  
+  // Optional, options to pass to the geolocation api.
+  // https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions
+  geolocationOptions: {
+    // Optional, whether to improve accuracy at the cost of response time and/or power consumption, defaults to false.
+    enableHighAccuracy: false,
+    // Optional, the maximum amount of time (in ms) a geolocation call is allowed to take before defaulting, defaults to 1000ms.
```
tmeyer2115: Here and below, let's mention the defaults in seconds (or minutes). Small nit, but that will save the reader some mental math.

README.md
```diff
@@ -397,7 +397,24 @@ ANSWERS.addComponent('SearchBar', {
...
+    timeout: 1000,
```
tmeyer2115, line 407: I know the spec suggests this timeout, is it reasonable? Experimentally, does this seem like a good default? Most geolocation calls take less than a second?

src/ui/components/filters/geolocationcomponent.js
oshi97: yup there is the GeolocationFilter will display whatever it's error text config is set to, defaulting to 'Unable To Use Location'

README.md
oshi97: in my experience most take more than one second. Apparently there are a number of other factors that affect speed like whether your device has a gps in it (i.e. is a smartphone)

README.md
oshi97: I wanted to avoid using the word "fail" but this might be ptsd from 6th grade

README.md
oshi97: good point math is for nerds

tmeyer2115 (approved):

PR #812, opened 2020-05-26
V1.3.0

oshi97:
# Version 1.3.0

## Changes

* Added geolocationOptions and geolocationTimeoutAlert config to the SearchBar, GeolocationFilter, and LocationBias components. These options let the user configure the HTML5 Geolocation API used to determine a user's location, as well as whether to display an alert when this fails or timeouts (by default no alert is shown).


alexisgrow (approved):
🎊😁

PR #813, opened 2020-05-26
FilterOptions: filter by locationRadius w/ optionType RADIUS_FILTER

oshi97:
This commit adds a new config option for FilterOptions called optionType, which
can be either RADIUS_FILTER or STATIC_FILTER, defaulting to STATIC_FILTER which
is the original behavior. Filtering by radius is done by appending a new
locationRadius param to a query.

TEST=manual
Test Facets, FilterBox, FilterOptions
Facets: smoke test
FilterBox:
- check that regular STATIC_FILTER FilterBox still works, and can filter by distance
- check that reseting FilterBox then applying also clears the locationRadius
- check that works with searchOnChange (after fixing reset button init bug with #768 )

FilterOptions:
- check that defaults to STATIC_FILTER, throws helpful errors
when trying to use unrecognized optionType or trying to use multioption
with RADIUS_FILTER
- check that setting a value of 0 for RADIUS_FILTER clears locationRadius
- check that selected filter persists over page refresh
- check that can filter by locationRadius by searching "new york" under vertical key "people"
  locationRadius = 1 gives no results, locationRadius = 8046.72 = 1 result
- check that clicking the reset button in FilterOptions then searching again clears the 
  locationRadius

### DIFF
--- a/README.md
+++ b/README.md
@@ -1057,6 +1057,12 @@ ANSWERS.addComponent('FilterOptions', {
   container: '.filter-container',
   // Required, control type: 'singleoption' or 'multioption'
   control: 'singleoption',
+  // The type of options to filter by, either 'STATIC_FILTER' or 'RADIUS_FILTER'.
+  // Defaults to 'STATIC_FILTER'.
+  optionType: 'STATIC_FILTER',
+  // If true, the filter value is saved on change and sent with the next search.
+  // Defaults to false.
+  storeOnChange: true,
   // Required, list of options
   options: [
     /** Depends on the above optionType, either 'STATIC_FILTER' or 'RADIUS_FILTER', see below. **/
--- a/README.md
+++ b/README.md
@@ -1059,17 +1059,7 @@ ANSWERS.addComponent('FilterOptions', {
   control: 'singleoption',
   // Required, list of options
   options: [
-    {
-      // Required, label to show next to the filter option
-      label: 'Open Now',
-      // Required, the field's API name to filter on, configured in the Yext platform
-      field: 'c_openNow',
-      // Required, the value for the above field to filter by
-      value: true,
-      // Optional, whether the option is selected by default
-      selected: true,
-    },
-    ...
+    /** Depends on the above optionType, either 'STATIC_FILTER' or 'RADIUS_FILTER', see below. **/
   ],
   // Optional, if true, the filter value is saved on change and sent with the next search. Defaults to false.
   storeOnChange: false,
@@ -1098,6 +1088,74 @@ ANSWERS.addComponent('FilterOptions', {
   // Optional, the label to be used in the legend, defaults to 'Filters'
   label: 'Filters'
 });
+```
+
+The options config varies depending on whether the optionType is 'STATIC_FILTER' or 'RADIUS_FILTER'.
+
+##### STATIC_FILTER
+
+```js
+{
+  options: [
+    {
+      // Label to show next to the filter option
+      label: 'Open Now',
+      // The api field to filter on, configured on the Yext platform
+      field: 'c_openNow',
+      // The value for the above field to filter by
+      value: true,
+      // Whether this option will be selected on page load. Selected options stored in the url
+      // take priority over appliedOnLoad. Defaults to false.
+      appliedOnLoad: false
+    },
+    {
+      label: 'Dog Friendly',
+      field: 'c_dogFriendly',
+      value: true,
+      appliedOnLoad: true
+    },
+    {
+      label: 'Megastores',
+      field: 'c_storeType',
+      value: 'Megastore'
+    }
+  ]
+}
+```
+
+##### RADIUS_FILTER
+
+```js
+{    
+  options: [
+    {
+      // Label to show next to the filter option
+      label: '5 miles', 
+      // The value of the radius to apply (in meters). If this value is 0, will not filter by radius.
+      value: 8046.72,
+      // Whether this option will be selected on page load. Selected options stored in the url
+      // will take priority over appliedOnLoad. Defaults to false.
+      appliedOnLoad: false
+    },
+    {
+      label: '10 miles',
+      value: 16093.4,
+      appliedOnLoad: true
+    },
+    {
+      label: '25 miles',
+      value: 40233.6
+    },
+    { 
+      label: '50 miles',
+      value: 80467.2
+    },
+    {
+      label: "Do not filter by radius",
+      value: 0
+    }
+  ],
+}
 ```
 
 ### RangeFilter
--- a/README.md
+++ b/README.md
@@ -1111,14 +1111,14 @@ The options config varies depending on whether the optionType is 'STATIC_FILTER'
       // Optional, the label to show next to the filter option.
       label: 'Open Now',
       // Optional, whether this option will be selected on page load. Selected options stored in the url
-      // take priority over appliedOnLoad. Defaults to false.
-      appliedOnLoad: false
+      // take priority over this. Defaults to false.
+      selected: false
     },
     {
       field: 'c_dogFriendly',
       value: true,
       label: 'Dog Friendly',
-      appliedOnLoad: true
+      selected: true
     },
     {
       field: 'c_storeType',
@@ -1140,13 +1140,13 @@ The options config varies depending on whether the optionType is 'STATIC_FILTER'
       // Optional, the label to show next to the filter option.
       label: '5 miles',
       // Optional, whether this option will be selected on page load. Selected options stored in the url
-      // take priority over appliedOnLoad. Defaults to false.
-      appliedOnLoad: false
+      // take priority over this. Defaults to false.
+      selected: false
     },
     {
       value: 16093.4,
       label: '10 miles',
-      appliedOnLoad: true
+      selected: true
     },
     {
       value: 40233.6,
--- a/README.md
+++ b/README.md
@@ -1098,26 +1098,26 @@ The options config varies depending on whether the optionType is 'STATIC_FILTER'
 {
   options: [
     {
-      // Label to show next to the filter option
-      label: 'Open Now',
-      // The api field to filter on, configured on the Yext platform
+      // Required, the api field to filter on, configured on the Yext platform.
       field: 'c_openNow',
-      // The value for the above field to filter by
+      // Required, the value for the above field to filter by.
       value: true,
-      // Whether this option will be selected on page load. Selected options stored in the url
+      // Optional, the label to show next to the filter option.
+      label: 'Open Now',
+      // Optional, whether this option will be selected on page load. Selected options stored in the url
       // take priority over appliedOnLoad. Defaults to false.
       appliedOnLoad: false
     },
     {
-      label: 'Dog Friendly',
       field: 'c_dogFriendly',
       value: true,
+      label: 'Dog Friendly',
       appliedOnLoad: true
     },
     {
-      label: 'Megastores',
       field: 'c_storeType',
-      value: 'Megastore'
+      value: 'Megastore',
+      label: 'Megastores'
     }
   ]
 }
@@ -1129,30 +1129,30 @@ The options config varies depending on whether the optionType is 'STATIC_FILTER'
 {    
   options: [
     {
-      // Label to show next to the filter option
-      label: '5 miles', 
-      // The value of the radius to apply (in meters). If this value is 0, will not filter by radius.
+      // Required, the value of the radius to apply (in meters). If this value is 0, will not filter by radius.
       value: 8046.72,
-      // Whether this option will be selected on page load. Selected options stored in the url
-      // will take priority over appliedOnLoad. Defaults to false.
+      // Optional, the label to show next to the filter option.
+      label: '5 miles',
+      // Optional, whether this option will be selected on page load. Selected options stored in the url
+      // take priority over appliedOnLoad. Defaults to false.
       appliedOnLoad: false
     },
     {
-      label: '10 miles',
       value: 16093.4,
+      label: '10 miles',
       appliedOnLoad: true
     },
     {
-      label: '25 miles',
-      value: 40233.6
+      value: 40233.6,
+      label: '25 miles'
     },
     { 
-      label: '50 miles',
-      value: 80467.2
+      value: 80467.2,
+      label: '50 miles'
     },
     {
-      label: "Do not filter by radius",
-      value: 0
+      value: 0,
+      label: "Do not filter by radius"
     }
   ],
 }
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -153,7 +153,8 @@ export default class Core {
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
-        sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS)
+        sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS),
+        locationRadius: this.globalStorage.getState(StorageKeys.LOCATION_RADIUS)
       })
       .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))
       .then(data => {
@@ -358,6 +359,21 @@ export default class Core {
 
   setStaticFilterNodes (namespace, filterNode) {
     this.filterRegistry.setStaticFilterNodes(namespace, filterNode);
+  }
+
+  /**
+   * Sets the locationRadius param for vertical searches.
+   * @param {number} locationRadius
+   */
+  setLocationRadius (locationRadius) {
+    this.globalStorage.set(StorageKeys.LOCATION_RADIUS, locationRadius);
+  }
+
+  /**
+   * Clears the locationRadius param for vertical searches.
+   */
+  clearLocationRadius () {
+    this.globalStorage.delete(StorageKeys.LOCATION_RADIUS);
   }
 
   enableDynamicFilters () {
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -1140,6 +1122,74 @@ ANSWERS.addComponent('FilterOptions', {
...
 
+The options config varies depending on whether the optionType is 'STATIC_FILTER' or 'RADIUS_FILTER'.
+
+##### STATIC_FILTER
+
+```js
+{
+  options: [
+    {
+      // Label to show next to the filter option
```
alexisgrow: Can these comments have a "Required" or "Optional" prefix?

oshi97: updated! Reordered things to have required options first

README.md
```diff
@@ -1085,33 +1085,15 @@ ANSWERS.addComponent('FilterOptions', {
   container: '.filter-container',
   // Control type, singleoption or multioption
   control: 'singleoption',
+  // The type of filtering to apply to the options, either 'STATIC_FILTER' or 'RADIUS_FILTER'.
+  // Defaults to 'STATIC_FILTER'.
```
tmeyer2115, line 1061: I think it might be more accurate to say that optionType determines the kind of the options, not so much the filtering applied to them. STATIC_FILTER type options refer to values for a profile field, right? RADIUS_FILTER options are distances.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -141,7 +155,9 @@ class FilterOptionsConfig {
     if (this.control === 'singleoption') {
       const _options = options.map(o => ({ ...o }));
       const firstAppliedOption = _options.find(o => o.appliedOnLoad);
-      firstAppliedOption.selected = true;
+      if (firstAppliedOption) {
+        firstAppliedOption.selected = true;
+      }
       return _options;
```
tmeyer2115: I think this method can be made more efficient. We're essentially traversing the options array twice. Once with the .map and again with the .find. We should be able to consolidate the two traversals by making the mapping function a bit more complex.

oshi97:
rebased applied-filters onto v0.14.0, distance-filter onto applied-filters, then this branch onto distance-filter

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -322,8 +351,22 @@ export default class FilterOptionsComponent extends Component {
   }
 
   apply () {
-    const filter = this._buildFilter();
-    this.core.setFilter(this.name, filter);
+    switch (this.config.optionType) {
+      case OptionTypes.RADIUS_FILTER:
```
tmeyer2115, line 355: Minor nit, but could consolidate things a bit here. 
const selectedOption = this.config.options.find(o => o.selected && o.value !== 0);
selectedOption && this.core.setLocationRadius(selectedOption.value);

src/ui/components/filters/filteroptionscomponent.js
oshi97: I was worrying about not modifying the original array but it's pretty idiotic to worry about that when the array is just being set to the return value anyways. I don't really want to make the function more complex just to force the code to use .map

src/ui/components/filters/filteroptionscomponent.js
oshi97: I feel like this is not logically equivalent. If the selected value is === 0 I still want to call core.clearLocationRadius. I could move the "if the value is 0 clear the locationRadius" logic into core.setLocationRadius and delete core.clearLocationRadius though, if you like that better? The backend will throw an error if you set locationRadius to 0 so I think that makes sense, but maybe we don't want to tie our behavior with theirs.

README.md
oshi97: I wanted to avoid documentation that basically takes a config option like chocolateChipCookie and tells you "it is a cookie made with chocolate chips" but I think what you said makes sense here, updated!

src/ui/components/filters/filteroptionscomponent.js
oshi97: removed .map() call!

oshi97:
The spec was changed from `appliedOnLoad` to `selected` so this PR was changed to match. This was originally going to go in a separate PR but it looks like it would make things simpler for everyone to just put it here.

oshi97:
FilterOptions: rename 'appliedOnLoad' config option to 'selected'

THe 'selected' config option already exists (as a by product of implementation
details) but the spec has been changed to use 'selected' instead of 'appliedOnLoad'
and it doesn't look like there are any issues with this.

TEST=manual
Facets: smoke test
FilterOptions: check that works with multioption/singleoption, only selecting
the first selected option for singleoption, and using persistentStorage instead
if any stored data exists

README.md
tmeyer2115: I think it's okay to say that optionType represents the type or kind of the options if we also provide some quick details about where 'STATIC_FILTER' options (profile field values) and 'RADIUS_FILTER' options (distances) come from. If we provide the additional details about what the different kinds of options represent, you're providing the reader with additional info apart from the 'optionType is type of options' tautology. You don't need much detail here since you expand on them below.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -134,20 +148,27 @@ class FilterOptionsConfig {
...
   /**
    * If no previous options were stored in persistentStorage, default to options marked
-   * as appliedOnLoad.
+   * as selected. If multiple options are marked as selected for 'singleoption', only the
+   * first should be selected.
    * @param {*} options
+   * @returns {Array<Object>}
    */
-  setAppliedOnLoad (options) {
+  setSelected (options) {
```
tmeyer2115: Can we consolidate setPreviousOptions and setSelected into one, single setSelected method? I think it could help reduce a bit of complexity if we had a single method whose responsibility was setting the selected attribute for an option(s). As it stands, we essentially have two different methods for doing this.

oshi97: updated!

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -126,35 +126,25 @@ class FilterOptionsConfig {
...
    * first should be selected.
-   * @param {*} options
+   * @param {Array<Object>} options
+   * @param {Array<string>} previousOptions
    * @returns {Array<Object>}
    */
-  setSelected (options) {
-    if (this.control === 'singleoption') {
+  setSelectedOptions (options, previousOptions) {
+    if (previousOptions) {
```
tmeyer2115: I think there's a weird edge case here. Someone could first configure FilterOptions to be 'multioption', then save multiple selected options in persistent storage. If they then decide to change FilterOptions to be 'singleoption', but use the URL from before, could they wind up with multiple options selected here, even though it's 'singleoption'?

oshi97: Ah I didn't consider that. What happens is that multiple radio buttons are marked as checked in the html. In Chrome and Firefox this is interpreted as visually only the last radio button being selected, however if you inspect the dom multiple radio buttons are given the 'checked' attribute. If you apply the filters, since the component doesn't care about whether it's multioption/singleoption when it builds the filter it includes all the selected options.

Is this something we need to handle? I think no matter what we do the url won't be valid anymore.

tmeyer2115: So they could generate a filter with multiple options selected even though it's type 'singleoption'? I think we should handle this, even though it's an extreme edge case. For consistency's sake, I think we should just pick the first of the previousOptions in this case. What do you think?

oshi97: yeah sounds good to me!

oshi97: updated to add this handling, also added unit tests for RADIUS_FILTER and options.selected logic

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -25,11 +33,17 @@ class FilterOptionsConfig {
...
+     * @type {string}
+     */
+    this.optionType = config.optionType || OptionTypes.STATIC_FILTER;
+
     /**
      * The list of filter options to display with checked status
      * @type {object[]}
      */
-    this.options = config.options;
+    this.options = config.options.map(o => ({ ...o }));
```
tmeyer2115, line 46: Minor nit, you can use slice() to make a copy

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -233,3 +229,267 @@ describe('filter options component', () => {
     expect(setStaticFilterNodes.mock.calls).toHaveLength(6);
   });
 });
+
+describe('filter options component selected options', () => {
```
tmeyer2115: Can you provide a more descriptive name for this suite of tests? The name doesn't make it very clear what's being tested.

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -233,3 +229,267 @@ describe('filter options component', () => {
...
+      ]
+    };
+
+    const component = COMPONENT_MANAGER.create('FilterOptions', config);
+    const options = component.config.options;
+    expect(options).toHaveLength(3);
+    expect(options.filter(o => o.selected)).toHaveLength(0);
+  });
+
+  it('can choose selected options for multioption', () => {
```
tmeyer2115: How about 'properly sets selected options for multioption'? This kinda makes it sound like we're testing that a user can click multiple options.

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -233,3 +229,267 @@ describe('filter options component', () => {
...
+    const component = COMPONENT_MANAGER.create('FilterOptions', config);
+    const options = component.config.options;
+    expect(options).toHaveLength(4);
+    const selectedOptions = options.filter(o => o.selected);
+    expect(selectedOptions).toHaveLength(2);
+    expect(selectedOptions[0].label).toEqual('label3');
+    expect(selectedOptions[1].label).toEqual('label4');
+  });
+
+  it('prioritizes previous options over selected config for multioption', () => {
```
tmeyer2115: 'prioritizes previously selected options over selected options in config for multioption'

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -233,3 +229,267 @@ describe('filter options component', () => {
...
+    const component = COMPONENT_MANAGER.create('FilterOptions', config);
+    const options = component.config.options;
+    expect(options).toHaveLength(4);
+    const selectedOptions = options.filter(o => o.selected);
+    expect(selectedOptions).toHaveLength(2);
+    expect(selectedOptions[0].label).toEqual('label1');
+    expect(selectedOptions[1].label).toEqual('label2');
+  });
+
+  it('can choose selected option for singleoption', () => {
```
tmeyer2115: 'properly sets selected option for singleoption'

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -233,3 +229,267 @@ describe('filter options component', () => {
...
+
+    const component = COMPONENT_MANAGER.create('FilterOptions', config);
+    const options = component.config.options;
+    expect(options).toHaveLength(4);
+    const selectedOptions = options.filter(o => o.selected);
+    expect(selectedOptions).toHaveLength(1);
+    expect(selectedOptions[0].label).toEqual('label3');
+  });
+
+  it('prioritizes previous options over selected config for singleoption', () => {
```
tmeyer2115: 'prioritizes previously selected option over selected option in config for singleoption'

src/ui/components/filters/filteroptionscomponent.js
oshi97: for the record I did NOT know about this tip, but after some testing it looks like slice doesn't make a deep copy of the objects. What I did was
```
let a = [{ a: 123}, {a: 345, b: true}]
let b = a.slice()
b[1].a = 123
console.log(a[1].a === b[1].a)
```

and in this case array `a` was also modified

tests/ui/components/filters/filteroptionscomponent.js
oshi97: updated!

tests/ui/components/filters/filteroptionscomponent.js
oshi97: thank you for your brainpower

src/core/filters/filterregistry.js
```diff
@@ -76,6 +76,16 @@ export default class FilterRegistry {
     return Facet.fromFilters(this.availableFieldIds, ...filters);
   }
 
+  /**
+   * Gets the locationRadius to send in a search query.
+   * @returns {number}
+   */
+  getLocationRadiusPayload () {
```
tmeyer2115: Are all these payload methods used only in core.js? If so, I think we should move them to that class. Arguably, I think it's beyond the responsibility of this class to have payload methods. The role of this class is to maintain a collection of filters. It should have setters for adding filter nodes and getters as well. It deals in the currency of FilterNodes. Consumers of this class can then translate filter nodes into whatever structure they need. 

I'd suggest we make an item to remove the existing payload methods. For this particular method, what if we instead add a getFilterNodeByKey method to the registry? We have getters for all static filters, facets, etc. Might be nice to have a general purpose method for pulling exactly one filter, by key. In core.js, we can use this method to get the location radius FilterNode, then get the enclosed value.

src/core/models/filter.js
```diff
@@ -192,6 +192,16 @@ export default class Filter {
     return Filter._fromMatcher('builtin.location', '$near', { lat, lng, radius });
   }
 
+  /**
+   * Create a new RADIUS_FILTER.
+   * @param {number} radius
+   */
+  static locationRadius (radius) {
```
tmeyer2115: This method doesn't do much besides call the Filter constructor. Is it really necessary? I'm hesitant to clutter the Filter class with more of these static method. There are already a lot.

src/core/filters/filterregistry.js
```diff
@@ -76,6 +76,16 @@ export default class FilterRegistry {
...
+    const filterNode =
+      this.globalStorage.getState(`${this.filterRegistryId}.${StorageKeys.LOCATION_RADIUS}`);
```
tmeyer2115: As we talked about last night, I think we should remove filterRegistryId as it's no longer being used. For the sake of separation of concerns, I think we should do that in a separate PR.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -313,6 +386,28 @@ export default class FilterOptionsComponent extends Component {
     });
   }
 
+  getLocationRadiusFilterNode () {
+    const selectedOption = this.config.options.find(o => o.selected);
+    if (!selectedOption) {
+      return FilterNodeFactory.from();
```
tmeyer2115, line 392: This is a minor nit, but do we want to return you anything if no option was selected?

src/ui/sass/modules/_ResultsHeader.scss
```diff
@@ -28,11 +28,11 @@ $results-header-filters-line-height: var(--yxt-line-height-md) !default;
   --yxt-results-header-filters-color: #{$results-header-filters-color};
   --yxt-results-header-filters-line-height: #{$results-header-filters-line-height};
 }
-
 .yxt-ResultsHeader {
   padding: calc(var(--yxt-results-header-spacing) / 2) var(--yxt-results-header-spacing);
   display: flex;
   align-items: baseline;
+  height: 20px;
```
tmeyer2115: Are these changes part of another PR?

src/core/filters/filterregistry.js
oshi97: got it

src/ui/components/filters/filteroptionscomponent.js
oshi97: I thought maybe it was simpler to just return an empty filternode? alternatively I could add a clearLocationRadiusFilterNode() method

src/ui/sass/modules/_ResultsHeader.scss
oshi97: rebased!

src/core/models/filter.js
oshi97: removed

src/core/filters/filterregistry.js
oshi97: added a todo

tmeyer2115: Sorry, I should have clarified. I think we should add the getFilterNodeByKey in this PR and not include this locationRadiusPayload method. We can remove the other payload methods later, but I don't think we should add another in this PR.

src/core/models/filter.js
```diff
@@ -192,6 +192,13 @@ export default class Filter {
...
```
tmeyer2115: Can we remove this static method?

oshi97: sorry I must have mouse slipped or something

src/core/filters/filterregistry.js
oshi97: I see, removed locationRadiusPayload method and added getFilterNodeByKey

src/core/core.js
```diff
@@ -154,7 +156,7 @@ export default class Core {
         queryTrigger: this.globalStorage.getState('queryTrigger'),
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
         sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS),
-        locationRadius: this.filterRegistry.getLocationRadiusPayload()
+        locationRadius: locationRadiusFilterNode ? locationRadiusFilterNode.getFilter().value : null
```
tmeyer2115, line 159: Just to confirm, is locationRadius: null the same thing as leaving out locationRadius on the back-end?

tmeyer2115 (approved):
LGTM, just had the one question!

src/core/core.js
oshi97: yup, we have a data cleaning step that removes all null and undefined params

PR #814, opened 2020-05-26
SearchApi: remove radius param

oshi97:
Apparently this is not being used by the backend anywhere.
Also causes confusion with the new locationRadius param.

TEST=manual
use location bias component to set location
test that query doesn't error even without radius param


alexisgrow (approved):
🤔 glad we're removing it if it's unused

PR #816, opened 2020-05-26
AnalyticsReporter: only send SCROLL_TO_BOTTOM_OF_PAGE when queryId ex…

oshi97:
…ists

If queryId does not exist in the global event options,
SCROLL_TO_BOTTOM_OF_PAGE will 500.

T=320737
J=SPR-2193
TEST=manual

- w/ show allowEmptySearch: false for searchBar
  tested that on page with no results, which causes there to be no
  queryId (checked this too), before would fire the event and 500 but
  no longer does this

- after adding a query, check that event still fires without 500ing


alexisgrow (approved):
lg lg!

PR #818, opened 2020-05-28
Fix malformed calc in Card.scss

creotutar:
Division of the css variables performed outside of the calc in Card.scss
for the margin. This caused the FAQ to not have spacing in between
cards.

J=None
TEST=manual

Tested on a new Jambo site referencing a local sdk dist (served through npm
serve package). Verified that the FAQ cards did not have spacing before
and have margin between them now.


nbramblett (approved):

oshi97:
Do you know of any good linter we can use to find these issues for us? I messed up a few times while rebasing and a little worried about my eyes missing some places

oshi97 (approved):

oshi97 (approved):

PR #821, opened 2020-05-29
Applied filters display all

oshi97:


src/ui/components/results/resultsheadercomponent.js
```diff
@@ -9,7 +9,11 @@ const DEFAULT_CONFIG = {
   showFieldNames: false,
   resultsCountSeparator: '|',
   verticalURL: undefined,
-  showChangeFilters: false
+  showChangeFilters: false,
+  hiddenFields: ['builtin.location'],
+  removable: false, // TODO implement
+  delimiter: '|',
```
tmeyer2115, line 14: ResultsHeaderComponent is not something that people will directly instantiate, correct? I think it is only every created through another component (like VerticalResults). If it's the case that nobody would ever addComponent this directly, I don't think we need to worry about synonyms here. We can just update resultsCountSepeartor to be called delimiter.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -9,7 +9,11 @@ const DEFAULT_CONFIG = {
   showFieldNames: false,
   resultsCountSeparator: '|',
   verticalURL: undefined,
-  showChangeFilters: false
+  showChangeFilters: false,
+  hiddenFields: ['builtin.location'],
```
tmeyer2115: Do we need to include hiddenFields here? If a field is hidden can we not just exclude it's corresponding filter(s) from data.appliedQueryFilters?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -56,15 +97,21 @@ export default class ResultsHeaderComponent extends Component {
 
   setState (data) {
     const offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET);
-    const shouldShowFilters = this.appliedQueryFilters.length > 0 && this._config.showAppliedFilters;
+    const simpleFilterNodes = this._getSimpleFilterNodes([
+      ...this.core.getStaticFilterNodes(),
+      ...this.core.getFacetFilterNodes(),
+      this.core.getLocationRadiusFilterNode()
+    ]);
+    const appliedFiltersArray = this.getAppliedFiltersArray(this.appliedQueryFilters, simpleFilterNodes);
```
tmeyer2115: This is confusing. I'm not understanding the difference between this.appliedQueryFilters and simpleFilterNodes. It seems like simpleFilterNodes includes every single root FilterNode. Why do we need appliedQueryFilters then?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -152,11 +163,21 @@ export default class VerticalResultsComponent extends Component {
...
     this.numColumns = this._config.maxNumberOfColumns;
 
+    const appliedFiltersOpts = this._config.appliedFilters || {};
+
     /**
      * Config options used in the {@link ResultsHeaderComponent}
      */
-    const { showFieldNames, resultsCountSeparator, showResultCount, showAppliedFilters, showChangeFilters } = this._config;
-    this.resultsHeaderOpts = { showFieldNames, resultsCountSeparator, showResultCount, showAppliedFilters, showChangeFilters };
+    this.resultsHeaderOpts = {
```
tmeyer2115, line 180: In this component and any others, I want us to avoid referring to deprecated configuration options. It adds a lot of complexity to a component. We should use the VerticalResultsConfig as a translation mechanism. It takes in user-supplied configuration (potentially with deprecated options) and converts it to an object that the VerticalResultsComponent can understand. In this case, the VerticalResults component should now only refer to the appliedFilters object. It should not refer to things like _config.showFieldNames any more. 

The VerticalResultsConfig class has the responsibility of converting 'showFieldNames' into 'appliedFilters.showFieldNames', etc. For those configuration options that were removed and renamed, we should have their default values be the values of the legacy attributes. So, appliedFilters.showFieldNames would default to showFieldNames, which defaults to false. VerticalResultsConfig should produce an up-to-date, correct configuration for VerticalResults. That means a fully populated appliedFilters object with any defaults. 

I'd like to avoid using the "parseConfig" method as well. I think such a class could definitely be useful, but it warrants a bit more discussion.

oshi97: updated to only use new config in vertical results (outside of the config class).

src/ui/components/results/resultsheadercomponent.js
oshi97: updated vertical results to turn appliedQueryFilters (nlpFilters) into FilterNodes, and sending ONLY filter nodes to the results header

src/ui/components/results/resultsheadercomponent.js
oshi97: updated to do all hidden fields filtering (as well as empty filternode filtering) inside of vertical results, reducing the complexity required by the results header as we discussed

src/ui/components/results/resultsheadercomponent.js
oshi97: removed any mention of DEPRECATED and now only use new config names for vertical results. for results header config names were not changed because this is a component designed purely for internal use

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -31,22 +34,31 @@ export default class ResultsHeaderComponent extends Component {
...
     return true;
   }
 
-  getAppliedFiltersArray () {
+  /**
+   * Combine all of the applied filters into a format the handlebars
+   * template can work with.
+   * @param {Array<AppliedQueryFilter>} appliedQueryFilters
+   * @param {Array<SimpleFilterNode>} simpleFilterNodes
+   * @returns {Array<Object>}
```
tmeyer2115, line 51: Update the JSDoc

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -155,8 +163,15 @@ export default class VerticalResultsComponent extends Component {
...
+      showFieldNames: this._config.appliedFilters.showFieldNames,
+      resultsCountSeparator: this._config.appliedFilters.resultsCountSeparator,
+      showAppliedFilters: this._config.appliedFilters.show,
+      showChangeFilters: this._config.showChangeFilters,
+      showResultCount: this._config.showResultCount,
+      removable: this._config.appliedFilters.removable, // TODO implement
+      delimiter: this._config.appliedFilters.delimiter || '|'
```
tmeyer2115: We should include this default in the VerticalResultsConfig class as well. It should provide a fully populated appliedFilters config (using any defaults necessary to do this). The component should not need any additional defaulting behavior for _config.appliedFilters.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -191,14 +206,14 @@ export default class VerticalResultsComponent extends Component {
...
-    this.appliedQueryFilters = (data.appliedQueryFilters || [])
+    this.nlpFilters = (data.appliedQueryFilters || [])
       .filter(f => !this._config.hiddenFields.includes(f.fieldId));
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
     const displayResultsIfExist = this._config.isUniversal ||
       this._displayAllResults ||
       data.resultsContext === ResultsContext.NORMAL;
     const showResultsHeader = this.resultsHeaderOpts.showResultCount ||
-      (this.resultsHeaderOpts.showAppliedFilters && this.appliedQueryFilters.length > 0);
+      (this.resultsHeaderOpts.showAppliedFilters && this.nlpFilters.length > 0);
```
tmeyer2115: Does this logic need to change? We're only going to show applied filters if nlpFilters are present. What if there are no nlpFilters but there are static filters and facets? Is that possible?

src/core/models/section.js
```diff
@@ -81,10 +83,16 @@ export default class Section {
...
   constructor (appliedQueryFilter) {
     this.key = appliedQueryFilter.key || appliedQueryFilter.displayKey;
     this.value = appliedQueryFilter.value || appliedQueryFilter.displayValue;
-    this.fieldId = Object.keys(appliedQueryFilter.filter)[0];
+    this.filter = Filter.from(appliedQueryFilter.filter);
+    this.fieldId = this.filter.getFilterKey();
+    this.metadata = new FilterMetadata({
+      fieldName: this.key,
+      displayValue: this.value
+    });
```
tmeyer2115: Minor nit, but I'd suggest we not mess with this class until we do SPR-2394. We can have VerticalResults do the full conversion for now.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -243,6 +258,53 @@ export default class VerticalResultsComponent extends Component {
     return 'results/verticalresults';
   }
 
+  /**
+   * Given an array of nlp filters from the backend (see class AppliedQueryFilter within section.js),
+   * turn them into an array of SimpleFilterNodes
+   * @param {Array<Object>} nlpFilters
```
tmeyer2115: Array< AppliedQueryFilter >, then you can get rid of the "(see class ...)"

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -243,6 +258,53 @@ export default class VerticalResultsComponent extends Component {
...
+    }));
+  }
+
+  /**
+   * Recursively get all of the root FilterNodes of the given array,
+   * which will all be SimpleFilterNodes
+   * @param {Array<FilterNode>} filterNodes
+   * @returns {Array<SimpleFilterNode>}
+   */
+  _getSimpleFilterNodes (filterNodes) {
```
tmeyer2115: I'm wondering if there's a better place for this method. Conceivably, we could use it somewhere else. It feels more like an instance method we should add to the FilterNode class. What if we add a getSimpleAncestors() to the FilterNode class? It returns all SimpleFilterNode ancestors (at any level) of the FilterNode.

src/core/utils/configutils.js
```diff
@@ -0,0 +1,32 @@
...
+ * will first look for config['first']['second']['option'].
+ *
+ * This is mostly needed for boolean config values, since boolean operators,
+ * which we commonly use for defaulting config options, do not work properly
+ * in those cases.
+ * @param {Object} topLevelConfig
+ * @param  {Array<string>}
+ * @param {any} defaultValue
+ */
+export function parseConfig (topLevelConfig, synonyms, defaultValue) {
```
tmeyer2115: Let's give this a more descriptive name. Maybe "defaultConfigOption" or something like that. The name "parseConfig" indicates this parses the entire config object and returns some parsed version of it. I think we should rename topLevelConfig to be just "config". Theoretically, someone could use at any level of config.

oshi97: sure I definitely had a lot of trouble coming up with accurate and descriptive names for this, updated! there's also already a parseConfig in answers-umd so its good not to share names

src/ui/components/results/verticalresultscomponent.js
oshi97: Aw man I remember trying to push for this with FilterView but I guess it makes a lot more sense on a node class, added!

src/ui/components/results/verticalresultscomponent.js
oshi97: updated!

src/core/models/section.js
oshi97: sure, I thought this was pretty safe and would make things 1% easier later but updated

src/ui/components/results/verticalresultscomponent.js
oshi97: yup updated!

src/ui/components/results/verticalresultscomponent.js
oshi97: sorry brainfartedI also missed delimiter and removable in the config class

src/ui/components/results/resultsheadercomponent.js
oshi97: updated

src/core/core.js
```diff
@@ -140,7 +140,11 @@ export default class Core {
       this.globalStorage.delete(StorageKeys.SEARCH_OFFSET);
     }
 
-    const locationRadiusFilterNode = this.filterRegistry.getFilterNodeByKey(StorageKeys.LOCATION_RADIUS);
+    if (!useFacets) {
+      this.filterRegistry.setFacetFilterNodes([], []);
```
tmeyer2115, line 144: For my own edification, why do we need to do this?

src/core/filters/filternode.js
```diff
@@ -19,11 +19,17 @@ export default class FilterNode {
...
    * Returns the children of this node.
    * @returns {Array<FilterNode>}
    */
   getChildren () {}
+
+  /**
+   * Recursively get all of the leaf SimpleFilterNodes.
+   * @returns {Array<SimpleFilterNode>}
+   */
+  getSimpleAncesors () {}
```
tmeyer2115: Beautiful

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -191,16 +220,15 @@ export default class VerticalResultsComponent extends Component {
...
     this.verticalKey = data.verticalConfigId;
     this.resultsContext = data.resultsContext;
-    this.appliedQueryFilters = (data.appliedQueryFilters || [])
-      .filter(f => !this._config.hiddenFields.includes(f.fieldId));
+    this.nlpFilters = data.appliedQueryFilters || [];
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
     const displayResultsIfExist = this._config.isUniversal ||
       this._displayAllResults ||
       data.resultsContext === ResultsContext.NORMAL;
+    this.appliedFilterNodes = this._getAppliedFilterNodes(this._convertNlpFiltersToFilterNodes(this.nlpFilters));
```
tmeyer2115: Line break after '='

tmeyer2115: Minor nit, but how about we have getAppliedFilterNodes be responsible for calling the nlpFilter conversion too? I think that could consolidate things nicely.

src/core/filters/filternode.js
oshi97: oof

src/core/core.js
oshi97: When you trigger a search with the searchbar, all of the facets are cleared, as opposed to when you trigger a search with facets or any static filters. What was happening was that I would click some facets, apply them, and see them pop up in the applied filters bar. sweet. then I would search with the search bar, and the applied filters bar would pretend those same facets were applied to the search, when in fact they were NOT because useFacets was false (since the search was triggered with the search bar). This was because the facets were still stored in the filter registry, after the search, even though they were not used.

src/ui/components/results/verticalresultscomponent.js
oshi97: Sure, updated!

src/core/filters/filternode.js
oshi97: fixed!

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -290,12 +290,12 @@ export default class VerticalResultsComponent extends Component {
    * Returns an array of all filter nodes currently being applied to the search.
    * Filters out filterNodes without fieldName or displayValue, or that have a
    * fieldId listed in this._config.hiddenFields.
-   * @param {AppliedQueryFilter} nlpFilters
+   * @param {Array<FilterNode>} filterNodes
```
tmeyer2115: Is this an Array of AppliedQueryFilter?

tmeyer2115 (approved):

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #825, opened 2020-06-01
ResultsHeader: Removable Applied Filters

oshi97:
This commit implements the removable config option of applied filters.
This was done by adding a callback to the filternode each filtercomponent
creates, which is called when the applied filter removable pill is clicked.
This allows for external component logic to be executed from within resultsheader.

TEST=manual
tested that geolocationfilter, filtersearch, filteroptions, filterbox, facets, range,
daterange, can be removed, and work in combination with each other
test that REMOVED_FILTER event fires correctly for each component
test univeral results uses its own default vertical config defaults correctly
(specifically checking for showAppliedFilters, hiddenFields, delimiter, resultsCountSeparator, showResultCount, showFieldNames, and that removable does nothing since universal is only nlp filters, which are nonremovable)


oshi97:
There is one outstanding question on the REMOVED_FILTER event fired when removing a radius filter, https://yext.slack.com/archives/GREFVUD7V/p1591021851002000, but I can't imagine this causing more than just radius filter specific logic when firing the analytics event.

src/core/filters/simplefilternode.js
```diff
@@ -60,4 +66,11 @@ export default class SimpleFilterNode extends FilterNode {
   getSimpleAncestors () {
     return this;
   }
+
+  /**
+   * Removes this filter node from the FilterRegistry.
+   */
+  remove () {
+    this._remove();
```
tmeyer2115, line 74: It seems like at least a couple of the remove methods don't actually remove the filter from persistent or globalStorage. Instead they set the filter to some empty value in storage. Can we always ensure we're deleting the filter and its remnants?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -247,6 +257,16 @@ export default class FilterBoxComponent extends Component {
...
   }
 
+  /**
+   * Handle filter node removal in a child FilterOptions component.
+   * @param {number} index
+   * @param {FilterNode} filterNode
+   */
+  onFilterNodeRemoval (index, filterNode) {
+    this._filterNodes[index] = filterNode;
+    this._saveFilterNodesToStorage();
```
tmeyer2115: I'm a bit confused, how does this method work? I would've thought this method first created a version of _filterNodes without this particular filterNode, called the remove method for that filterNode, then called saveFilterNodesToStorage. If I remove a filterNode here, how is the corresponding child filter getting re-rendered?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -10,9 +13,10 @@ const DEFAULT_CONFIG = {
   resultsCountSeparator: '|',
   verticalURL: undefined,
   showChangeFilters: false,
-  removable: false, // TODO implement
+  removable: false,
   delimiter: '|',
-  isUniversal: false
+  isUniversal: false,
+  verticalKey: undefined
```
tmeyer2115: Do we need this default? Aren't we always guaranteed to have a verticalKey?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +38,91 @@ export default class ResultsHeaderComponent extends Component {
...
+    const removableFilterTags = DOM.queryAll(this._container, '.js-yxt-ResultsHeader-removableFilterTag');
+    removableFilterTags.forEach(tag => {
+      DOM.on(tag, 'click', () => this._removeFilterTag(tag));
+    });
+  }
+
+  _removeFilterTag (tag) {
+    const { filterId } = tag.dataset;
+    const simpleFilterNode = this.removableFilterNodes[filterId];
+    simpleFilterNode.remove();
```
tmeyer2115: This feels kinda weird to me. I think the FilterNodeRegistry should have some kind of remove method that we can call. A registry suggests you an add or delete FilterNodes.

src/ui/components/filters/filterboxcomponent.js
oshi97: _saveFilterNodesToStorage does that, I guess the name isn't terribly accurate

oshi97: This is also only used for facets, so I should probably rename it

src/ui/components/results/resultsheadercomponent.js
oshi97: can get rid of it wasn't sure if we wanted to list out all possible config options or not

src/ui/components/results/resultsheadercomponent.js
oshi97: maybe a better flow might be to give every filterNode (not just the top level ones) a filterId like we talked about, and call a this.core.removeFilterNode(filterId)? I didn't really want to do this right now but if you think it's necessary I can

src/ui/components/filters/filterboxcomponent.js
oshi97: ok this method name is garbage deleting it like we talked about

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -200,9 +207,12 @@ export default class FilterBoxComponent extends Component {
           showReset: this.config.resetFilter,
           resetLabel: this.config.resetFilterLabel,
           showExpand: this.config.expand,
+          isDynamic: this.config.isDynamic,
+          originComponent: this.config.originComponent,
           onChange: filterNode => {
             this.onFilterNodeChange(i, filterNode);
-          }
+          },
+          onFilterNodeRemoval: filterNode => this.onFilterNodeRemoval(i, filterNode)
```
tmeyer2115: Talked offline about consolidating onChange and onFilterNodeRemoval

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -236,6 +255,9 @@ export default class FilterOptionsComponent extends Component {
      * @type {boolean}
      */
     this.allShown = false;
+    if (this.config.storeOnChange) {
+      this.apply();
+    }
```
tmeyer2115, line 264: For my own edification, why do we need to call this.apply() in the constructor?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -391,23 +414,31 @@ export default class FilterOptionsComponent extends Component {
     if (!selectedOption) {
```
tmeyer2115, line 605: Could this be !selectedOption || selectedOption.valye === 0?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -391,23 +414,31 @@ export default class FilterOptionsComponent extends Component {
...
     } else {
       return FilterNodeFactory.from({
-        metadata: metadata,
+        ...filterNode,
         filter: Filter.empty()
       });
     }
   }
 
+  _removeFilterNode (option) {
```
tmeyer2115: Could we rename this method? It's not that we're removing the filter node right, we're just unselecting an option?

src/ui/components/results/resultsheadercomponent.js
oshi97: will resolve in next pr

src/ui/components/results/resultsheadercomponent.js
tmeyer2115: I don't think we want to include it here since it's not something we can actually default (gracefully).

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +38,91 @@ export default class ResultsHeaderComponent extends Component {
...
+      displayValue: filterNode.getMetadata().displayValue,
+      dataFilterId: index,
+      removable: false
     });
-    // Has to be parsed into an array because our handlebars can only loop through arrays, not objects.
+    const irremovableGrouped = groupArray(this.irremovableFilterNodes, keyFunc, irremovableValueFunc);
+    const removableValueFunc = (filterNode, index) => ({
+      displayValue: filterNode.getMetadata().displayValue,
+      dataFilterId: index,
+      removable: true
```
tmeyer2115: Does this need to be configurable?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +38,91 @@ export default class ResultsHeaderComponent extends Component {
...
-      const { fieldName, displayValue } = fn.getMetadata();
-      if (!groupedFilters[fieldName]) {
-        groupedFilters[fieldName] = [];
-      }
-      groupedFilters[fieldName].push(displayValue);
+  _groupAppliedFilters () {
+    const keyFunc = filterNode => filterNode.getMetadata().fieldName;
+    const irremovableValueFunc = (filterNode, index) => ({
+      displayValue: filterNode.getMetadata().displayValue,
+      dataFilterId: index,
```
tmeyer2115: Do these need a dataFIlterId?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -286,25 +288,28 @@ export default class VerticalResultsComponent extends Component {
     }));
   }
 
+  _getAppliedFilterNodes () {
```
tmeyer2115, line 296: Add some JSDoc

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -358,7 +363,8 @@ export default class VerticalResultsComponent extends Component {
       const resultsHeaderData = {
         resultsLength: this.results.length,
         resultsCount: this.resultsCount,
-        appliedFilterNodes: this.appliedFilterNodes,
+        removableFilterNodes: this.appliedFilterNodes,
+        irremovableFilterNodes: this.nlpFilterNodes,
```
tmeyer2115: Minor nit, but I think we shouldn't leak the concept of removable or irremovable into VerticalResults. That's internal to the ResultsSectionHeader itself. As far as VerticalResults is concerned, it passes nlpFilters and appliedFilters into the ResultsHeader component. I think using hiddenFields in VerticalResults was a bit different because that would have caused VerticalResults to pass the ResultsSectionHeader unnecessary data.

src/ui/components/filters/filteroptionscomponent.js
oshi97: it was to apply the filters on page load, I felt like even though it wasn't specifically asked for in this item since FilterBox was already successfully applying filters on page load, FilterOptions should also do so. I did not add this to the other filters (except for FilterSearch in a past PR because we felt it was necessary for proper backwards compatibility)

src/ui/components/filters/filteroptionscomponent.js
oshi97: I don't think so we still want the FilterMetadata when selectedOption.value is 0, tried to think of a simpler way but didn't want to spend too long

src/ui/components/filters/filteroptionscomponent.js
oshi97: yup! I renamed it to _clearSingleOption and it's only 3 lines after our discussion

src/ui/components/results/resultsheadercomponent.js
oshi97: yes sorry missed this

src/ui/components/results/resultsheadercomponent.js
oshi97: nope!

src/ui/components/results/verticalresultscomponent.js
oshi97: this is necessary because some FilterNodes passed to the ResultsHeader component (namely nlpFilters but could be anything) are not able to be removable.

oshi97: We could rename FilterMetadata.originComponent to just origin, change that to "nlpFilter" for nlpFilters, but I think it actually makes sense to leak this since the logic for removable/irremovable is inside of ResultsHeader (also it's just simpler in the current moment lol)

src/core/filters/simplefilternode.js
oshi97: updated!

src/ui/components/filters/filterboxcomponent.js
oshi97: updated!

src/core/core.js
```diff
@@ -387,6 +387,14 @@ export default class Core {
     this.filterRegistry.setLocationRadiusFilterNode(filterNode);
   }
 
+  clearStaticFilterNode (namespace) {
```
tmeyer2115, line 429: Add some JSDoc for these new methods.

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -118,6 +118,16 @@ export default class DateRangeFilterComponent extends Component {
     this._updateRange('max', date);
   }
 
+  _removeFilterNode () {
+    this._date = Object.assign({}, this._date, { 'min': null });
+    this._date = Object.assign({}, this._date, { 'max': null });
```
tmeyer2115: Can we combine these two lines?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -238,11 +248,14 @@ export default class FilterBoxComponent extends Component {
    * Handle changes to child filter components
    * @param {number} index The index of the changed filter
    * @param {FilterNode} filterNode The new filter node
+   * @param {boolean} options
```
tmeyer2115: Update JSDoc

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -382,34 +392,44 @@ export default class FilterOptionsComponent extends Component {
...
         filter: Filter.empty()
       });
+    } else {
+      return FilterNodeFactory.from(filterNode);
     }
   }
 
+  _clearSingleOption (option) {
+    option.selected = false;
+    this.updateListeners(true, true);
```
tmeyer2115, line 625: Is this the only place updateListeners is called? If so, I think we can simplify things a bit. For one, updateListeners wouldn't need to take in any params. blockSearchOnChange would always be true, as would saveFilterNodes. In turn, if the only place config.onChange is called is updateListeners, we can really simplify the onChange passed in from FilterBox.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +37,93 @@ export default class ResultsHeaderComponent extends Component {
...
+     */
+    this.irremovableFilterNodes = data.irremovableFilterNodes || [];
   }
 
   static areDuplicateNamesAllowed () {
     return true;
   }
 
+  onMount () {
+    const removableFilterTags = DOM.queryAll(this._container, '.js-yxt-ResultsHeader-removableFilterTag');
```
tmeyer2115: new line

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +37,93 @@ export default class ResultsHeaderComponent extends Component {
...
+    const removedFilter = optionType === 'STATIC_FILTER'
+      ? JSON.stringify(simpleFilterNode.getFilter())
+      : simpleFilterNode.getFilter().value || 0;
+    analyticsEvent.addOptions({
+      verticalKey: this._config.verticalKey,
+      removedFromComponent: originComponent,
+      optionType: optionType,
+      removedFilter: removedFilter
+    });
+    this.analyticsReporter.report(analyticsEvent);
```
tmeyer2115: Just for the sake of factoring things a bit, I'd suggest making a method that creates the analyticsEvent for this.

src/ui/components/results/verticalresultscomponent.js
tmeyer2115: Oh, I was thinking we'd just have VerticalResults pass in two data attributes: nlpFilterNodes and appliedFilterNodes. The ResultsSectionHeader would have the logic for knowing nlp filters aren't removable.

src/core/filters/filtermetadata.js
```diff
@@ -20,6 +20,19 @@ export default class FilterMetadata {
...
+     * DateRangeFilter.
+     * @type {string}
+     */
+    this.originComponent = originComponent;
+
+    /**
+     * The type of filter this is, either 'RADIUS_FILTER' or 'STATIC_FILTER'.
+     * @type {string}
+     */
+    this.optionType = optionType || 'STATIC_FILTER';
```
tmeyer2115: Neither optionType nor originComponent is expressive enough on it's own. It seems like what we really want is some higher-level construct, "filterType". This type can be: "Facet", "Distance", "Combined", "NLP", "Geolocation", "Range", etc. When computing the analytics event for removing a filter, we can maintain a mapping of filterType to component type. I think having such a granular "filterType" could come in handy. What do you think? I noticed the spec said one example of "removed_from_component" was "VerticalResults". When would we see that?

oshi97: hmm I missed that, no idea what they mean

oshi97: VerticalResults definitely doesn't make any filters

src/ui/components/filters/filteroptionscomponent.js
oshi97: it's not the only place it's called

oshi97: clearSingleOption() is basically the preexisting clearOptions() method but for a single option

src/ui/components/results/resultsheadercomponent.js

src/ui/components/results/resultsheadercomponent.js

src/ui/components/filters/daterangefiltercomponent.js
oshi97: updated!

src/ui/components/results/verticalresultscomponent.js
oshi97: ah okay, updated! that does feel kinda nice

src/core/core.js
```diff
@@ -359,22 +359,49 @@ export default class Core {
...
+   * @param {Array<FilterNode>} filterNodes
+   */
   setFacetFilterNodes (availableFieldids = [], filterNodes = []) {
     this.filterRegistry.setFacetFilterNodes(availableFieldids, filterNodes);
   }
 
+  /**
+   * Sets the specified {@link FilterNode} under the given key.
+   * Will replace a preexisting node if there is one.
+   * @param {string} key
```
tmeyer2115: namespace

src/core/core.js
```diff
@@ -387,6 +414,21 @@ export default class Core {
     this.filterRegistry.setLocationRadiusFilterNode(filterNode);
   }
 
+  /**
+   * Remove the static FilterNode with this namespace.
+   * @param {string} key
```
tmeyer2115: namespace

src/core/filters/filtermetadata.js
tmeyer2115: We discussed last night no longer needing originComponent. I've asked Max a question about optionType. I'm not sure what the value of that attribute is supposed to be for Facets. Is it unset, is it STATIC_FILTER? I'm not sure.

I think filterType would still be a useful thing to add because it gives us more information than just optionType. optionType just allows us to differentiate between two kinds of static filters. But, we can have dynamic filters, nlp filters, etc. There are also more than just two kinds of static filter. filterType would allow us to encapsulate all that in one value (instead of originComponent + optionType). 

We could also add neither filterType nor optionType right now. By looking at the structure of the filter, you can differentiate between normal static and distance filter options, right? A distance filter is just a number, correct? Other filters are objects. I'd be fine if we didn't add either optionType or filterType for now, with the plan being adding filterType later on.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -238,11 +248,15 @@ export default class FilterBoxComponent extends Component {
    * Handle changes to child filter components
    * @param {number} index The index of the changed filter
    * @param {FilterNode} filterNode The new filter node
+   * @param {boolean} saveFilterNodes Whether to save filternodes to storage
+   * @param {boolean} searchOnChange Whether to conduct a search
    */
-  onFilterNodeChange (index, filterNode) {
+  onFilterNodeChange (index, filterNode, saveFilterNodes, searchOnChange) {
```
tmeyer2115, line 254: It's possible that someone could have saveFilterNodes as false and searchOnChange as true. Is that a problematic state? They'd issue a query without actually saving the updated nodes in storage first. It seems like they'd always want to saveFilterNodes if searchOnChange is true.

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -200,8 +207,11 @@ export default class FilterBoxComponent extends Component {
           showReset: this.config.resetFilter,
           resetLabel: this.config.resetFilterLabel,
           showExpand: this.config.expand,
-          onChange: filterNode => {
-            this.onFilterNodeChange(i, filterNode);
+          originComponent: this.config.originComponent,
+          onChange: (filterNode, saveFilterNodes, blockSearchOnChange) => {
```
tmeyer2115: One thing that's bugging me about this lambda function a bit is the "saveFilterNodes" param above. It's a little bit misleading. Based on the name, if I set this param to false, I'd expect no filter nodes to be saved/updated. But, that's not necessarily the case, depending on the value of config.searchOnChange.

tmeyer2115 (commented):
Will a later PR add REMOVE_ANALYTICS events when filters are reset in FilterOptions/FilterBox? It seems like we no longer need originComponent, is that right? It's VerticalResults when REMOVE_ANALYTICS is fired from ResultsSectionHeader.

src/ui/components/filters/filterboxcomponent.js
oshi97: *sure I think it makes sense to always saveFilterNodes if searchOnChange is true

src/ui/components/filters/filterboxcomponent.js
oshi97: updated to alwaysSaveFilterNodes!

oshi97:
PR has been updated with the new REMOVED_FILTER analytics spec

src/core/filters/filtermetadata.js
oshi97: added filtertype which is either STATIC, FACET, NLP, or RADIUS!

src/core/utils/eventutils.js
```diff
@@ -0,0 +1,40 @@
+import FilterType from '../filters/filtertype';
```
tmeyer2115: I think we should call this file analyticseventutils.js or something like that. Event is a very loaded term in Yext's codebase, haha.

src/core/utils/eventutils.js
```diff
@@ -0,0 +1,40 @@
...
+import AnalyticsEvent from '../analytics/analyticsevent';
+import { AnswersAnalyticsError } from '../errors/errors';
+
+/**
+ * Create the 'REMOVED_FILTER' analytics event, using the given {@link FilterNode}.
```
tmeyer2115: given Filter?

src/core/utils/eventutils.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param {Filter} removedFilter The {@link Filter} that was removed.
+ * @param {string} optionType The optionType to send in the request, either STATIC_FILTER, FACET_FILTER, or RADIUS_FILTER.
```
tmeyer2115: Minor nit, but I think we should take in filterType here and do the conversion internally. I feel the SDK should operate off filterType internally and only reference optionType right at the boundaries (i.e. when shipping something to the Analytics backend).

src/core/utils/eventutils.js
```diff
@@ -0,0 +1,40 @@
...
+    verticalKey: verticalKey,
+    removedFromComponent: removedFromComponent,
+    optionType: optionType,
+    removedFilter: removedFilter
+  });
+  return analyticsEvent;
+}
+
+/**
+ * Parse the 'optionType' analytics event param.
```
tmeyer2115: We're parsing the filterType, aren't we?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -234,18 +240,35 @@ export default class FilterBoxComponent extends Component {
   }
 
   resetFilters () {
+    const validFilterNodes = this._filterNodes.filter(fn => fn.getFilter().getFilterKey());
```
tmeyer2115: Do we do this calculation elsewhere in the component? If so, I think we should make a getValidFilterNodes method.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -487,19 +507,38 @@ export default class FilterOptionsComponent extends Component {
     return (option && filter) ? option.indexOf(filter) : -1;
   }
 
+  /**
+   * Clears all selected options. If this component is not part of a FilterBox,
+   * also send the REMOVED_FILTER analytics event. Otherwise, let FilterBox do it.
+   */
   clearOptions () {
+    if (!this.config.isDynamic) {
```
tmeyer2115: Are we relying on config.isDynamic to indicate if this came from FilterBox? If so, that's a bit of a code smell. That kind of logic can be dangerous as isDynamic was not meant to represent whether or not FilterBox was the parent. If FilterBox is not being used with Facets, but with a collection of normal FilterOptions, will we see duplicate analytics events getting fired?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -487,19 +507,38 @@ export default class FilterOptionsComponent extends Component {
...
+      const removedFilterParam = this.config.optionType === 'RADIUS_FILTER'
+        ? this.getLocationRadiusFilterNode().getFilter().value
+        : JSON.stringify(this.getFilterNode().getFilter());
```
tmeyer2115: Could we move the serialization of the removed filter into createRemovedFilterEvent as well? Then we wouldn't need to duplicate that code. I think it also fits within the responsibility of that method. It takes in SDK quantities (a Filter, fitlerType, etc. ) and serializes them so they can be understood by the Analytics backend.

src/core/utils/eventutils.js
oshi97: The issue I was running into is that if a CombinedFilterNode is removed, then there is no FilterMetadata, which means there is no filterType since filterType lives on the FilterMetadata.

src/ui/components/filters/filteroptionscomponent.js
oshi97: Yup there'll be dupes, clearOptions is called both when FilterBox resets all of its FilterOptions as well as when FilterOptions resets itself

src/ui/components/filters/filterboxcomponent.js
oshi97: yup done in two places, added the helper method

src/core/utils/eventutils.js
oshi97: yeah somehow what I wrote made sense at the time

src/core/utils/eventutils.js
oshi97: fixed

src/core/utils/eventutils.js
oshi97: sure, renamed!

src/ui/templates/results/resultsheader.hbs
```diff
@@ -30,8 +30,16 @@
...
-          <div class="yxt-ResultsHeader-filterValue">{{this}}{{#unless @last}}, {{/unless}}</div>
+        {{#each filterDataArray}}
+          {{#if removable}}
+            <div class="yxt-ResultsHeader-removableFilterTag js-yxt-ResultsHeader-removableFilterTag"
+              data-filter-id={{dataFilterId}}>
+              <span class="yxt-ResultsHeader-removableFilterValue">{{displayValue}}</span>
+              <span class="yxt-ResultsHeader-removableFilterX">&times;</span>
+            </div>
+          {{else}}
+            <div class="yxt-ResultsHeader-filterValue">{{displayValue}}{{#unless @last}}, {{/unless}}</div>
```
tmeyer2115: Do we want spans in this case as well? Could be good to add now in case we want to target that text with styling at some point.

src/ui/components/filters/filteroptionscomponent.js
oshi97: removing analytics event from this PR so this is no longer an issue, otherwise would have added a new 'isChildOfFilterBox' config like we discussed

src/core/utils/eventutils.js
oshi97: removing analytics event so no longer an issue

src/ui/components/filters/filteroptionscomponent.js
oshi97: will do this when we eventually implement the analytics event

src/ui/templates/results/resultsheader.hbs
oshi97: sure! I love making consulting happy no cap

oshi97:
@tmeyer2115 The analytics event is gone now, but should we keep filterType around in this PR? FilterType was being used mostly for the analytics event. Removing it would not be hard, but if that's the correct long term direction then maybe we can keep it?

src/core/utils/arrayutils.js
```diff
@@ -5,14 +5,17 @@
  * @param {params} Object
  * @returns {string}
```
tmeyer2115: Update JSDoc

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +36,87 @@ export default class ResultsHeaderComponent extends Component {
...
+   * removable filter tag.
+   * @param {HTMLElement} tag
+   */
+  _removeFilterTag (tag) {
+    const { filterId } = tag.dataset;
+    const filterNode = this.appliedFilterNodes[filterId];
+    filterNode.remove();
+    this.core.verticalSearch(this._config.verticalKey, {
+      resetPagination: true,
+      useFacets: true
```
tmeyer2115, line 76: Will useFacets always be true here?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -138,6 +138,9 @@ export default class FilterSearchComponent extends Component {
   }
 
   onMount () {
+    if (this.autoCompleteComponent) {
+      this.autoCompleteComponent.remove();
+    }
```
tmeyer2115, line 143: For my own edification, was this added to fix a bug?

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -34,40 +36,87 @@ export default class ResultsHeaderComponent extends Component {
...
-      label: label, displayValues: groupedFilters[label]
+      label: label,
+      filterDataArray: groupedFilters[label]
     }));
   }
 
   setState (data) {
     const offset = this.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET);
-    const appliedFiltersArray = this.getAppliedFiltersArray(this.appliedFilterNodes);
+    const groupedAppliedFilters = this._groupAppliedFilters();
```
tmeyer2115: Do we need both appliedFiltersArray and groupedAppliedFilters in setState?

tests/ui/components/filters/daterangecomponent.js
```diff
@@ -87,31 +87,27 @@ describe('date range filter component', () => {
...
-    };
+    });
     expect(component._buildFilter()).toEqual(filter);
     expect(component._buildFilterMetadata()).toEqual(metadata);
-    expect(component.getFilterNode()).toEqual(FilterNodeFactory.from({
-      filter: filter,
-      metadata: metadata
-    }));
+    expect(component.getFilterNode().getFilter()).toEqual(filter);
+    expect(component.getFilterNode().getMetadata()).toEqual(metadata);
```
tmeyer2115, line 95: Does getFilterNode essentially just invoke _buildFilter and _buildFilterMetadata? If so, I think these last two assertions are sufficient. We want full test coverage, but want to avoid redundant tests.

tests/core/filters/filterregistry.js
```diff
@@ -170,6 +170,19 @@ describe('FilterRegistry', () => {
     expect(locationRadiusFilterNode.getFilter().value).toEqual(1234);
     registry.setLocationRadiusFilterNode(FilterNodeFactory.from());
```
tmeyer2115, line 178: I think you should split this into another test for "can clear location radius filter node" (or something like that).

tests/ui/components/filters/daterangecomponent.js
```diff
@@ -87,31 +87,27 @@ describe('date range filter component', () => {
...
```
tmeyer2115, line 106: Same question here and elsewhere in this file.

tests/ui/components/filters/filterboxcomponent.js
```diff
@@ -148,31 +149,48 @@ describe('filter box component', () => {
...
     );
-    expect(setStaticFilterNodes.mock.calls[8][1]).toEqual(child0.getFilterNode());
-    expect(setStaticFilterNodes.mock.calls[9][1]).toEqual(child1.getFilterNode());
+    expect(child1.getFilterNode().getMetadata()).toEqual(
+      FilterNodeFactory.and(nodes1[0], FilterNodeFactory.or(nodes1[3], nodes1[4])).getMetadata()
+    );
+    expect(setStaticFilterNodes.mock.calls[8][1].getFilter()).toEqual(child0.getFilterNode().getFilter());
+    expect(setStaticFilterNodes.mock.calls[8][1].getMetadata()).toEqual(child0.getFilterNode().getMetadata());
+    expect(setStaticFilterNodes.mock.calls[9][1].getFilter()).toEqual(child1.getFilterNode().getFilter());
+    expect(setStaticFilterNodes.mock.calls[9][1].getMetadata()).toEqual(child1.getFilterNode().getMetadata());
```
tmeyer2115: This test is really long, with a ton of assertions. That seems like a bit of a code smell to me. It suggests a couple possibilities:
- This one test is used to verify too many things and should be split apart.
- Some of these assertions are redundant or overlap. 

With our Jest tests, there's a bit of a balance. We want full test coverage, but we want to avoid any overlap between tests and assertions as much as possible. The more tests there are and the larger they are, the slower our unit test suite is to run. So, we want to be judicious about the new assertions we add or the new tests we add. We want to ensure that by adding something, we are expanding our coverage in a meaningful way.

src/ui/components/results/resultsheadercomponent.js
oshi97: Pretty sure! I think the idea is we only don't use the facets (which resets them since the backend will send new facets) when the query is altered

src/ui/components/search/filtersearchcomponent.js
oshi97: Kind of! It was more to support removing an applied filter tag created by FilterSearch. Without this line, even if this.setState() is called in FilterSearch, the autocomplete text will not update. This same line was also added to the GeolocationFilter a year ago by Julian, since the components act very similarly I think it should be a safe addition.

oshi97: see https://github.com/yext/answers/blob/master/src/ui/components/filters/geolocationcomponent.js#L183

src/ui/components/results/resultsheadercomponent.js
oshi97: nope, made a one line helper for this!

tests/ui/components/filters/daterangecomponent.js
oshi97: updated!

tests/core/filters/filterregistry.js
oshi97: separated

tests/ui/components/filters/daterangecomponent.js
oshi97: also updated the rangefiltercomponent test

tests/ui/components/filters/filterboxcomponent.js
oshi97: Yeah looking back when I have something like
`let actualNode = blah; let expectedNode = blah`
and then be continually changing what the expected and actual values are equal to, it can probably be easily separated into multiple tests. I think my tests are still a little too complicated but things should be much better now

src/core/utils/arrayutils.js
```diff
@@ -1,9 +1,13 @@
 
 /**
- * Groups an array by a given key function.
- * @param {string} url
- * @param {params} Object
- * @returns {string}
+ * Groups an array into an object using a given key and value function, and an initial object
+ * to add to. By default the key and value functions will perform not perform any transformations
```
tmeyer2115: key and value functions will not perform

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -113,10 +113,18 @@ export default class ResultsHeaderComponent extends Component {
     }));
   }
 
+  /**
+   * Returns an array of object the handlebars can understand and render
+   * the applied filters bar from.
+   * @returns {Array<Object>}
+   */
+  _getAppliedFiltersArray () {
+    return this._convertGroupedFiltersToArray(this._groupAppliedFilters());
```
tmeyer2115: Is this the only place convertGroupdFiltersToArray is called? If so, I think we should just consolidate that method into this one. This method's not doing too much currently.

oshi97: sure that's simpler

src/core/utils/arrayutils.js
oshi97: ugh sorry

tmeyer2115 (approved):

PR #826, opened 2020-06-01
add pin hover event functions for map component

creotutar:
* add pin hover event functions for map component

Add a pin hover function option for the map component that will be
called on a mouseon event.  This function will be called with the
selected marker as a parameter (including the item's result data).

* rename function for pin mouse over and add mouse out for map component

Commonly, if a user wants to use hover events, they also want offhover
events (for example, to show and hide a modal). This allows the user to
provide mouse over and mouse out event handlers for pins on Google and
Mapbox for the Map component on verticals

J=SPR-2229
TEST=manual

Tested on a Jambo site sourcing a locally built SDK dist
Tested on a vertical-map vertical using Map component on
        - Google provider
        - Mapbox provider
Made sure hovered item was available to the on pin hover function.
Tested on Chrome,IE11

### DIFF
--- a/README.md
+++ b/README.md
@@ -1428,7 +1428,9 @@ ANSWERS.addComponent('Map', {
   // Optional, callback to invoke when a pin is clicked. The clicked item(s) are passed to the callback
   onPinClick: null,
   // Optional, callback to invoke when a pin is hovered. The clicked item(s) are passed to the callback
-  onPinHover: null,
+  onPinMouseOver: null,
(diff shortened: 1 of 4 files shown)


oshi97 (approved):
~~lgtm! could we just remove the onPinHover commit?  it looks like the onPinHover commit was completely overridden by onMouseOver/Out~~
edit: ignore what I said didn't realize we didn't rebase yet!

oshi97:
ignore what I said didn't realize we didn't rebase yet!

oshi97 (approved):

PR #828, opened 2020-06-01
Facets merge v1.4.0

alexisgrow:
Add Searchable Facets

Merges the following Searchable Facets PRs into the version branch, additional information below.

* Make facets searchable, update show more/less logic (#673)
* Add searchable facets highlight on match and overrides (#680)
* Implement levenshtein distance matching (#681)
* Searchable Facets: Add styling (#699)
* Facets/FilterOptions: Show selected facets first by default (#666)
* Searchable Facets: Do not set state on option click (#713)

Make facets searchable, update show more/less logic, highlight on match and overrides (#673) (#680):
 * Highlight filter option items that match the search terms. Allow field-specific overrides for placeholderText, searchable, and all of the config for FilterOptions.
* TEST=manual
* Test overrides manually on local test site, run and visually test changes in Chrome. Test highlighted matches locally on both Chrome and IE11.

Implement levenshtein distance matching (#681):
* Allow searchable facets to support a one-character typo if the search string is longer than three characters, using the levenshtein distance.
* TEST=manual
* Tested on Chrome and IE11 using our demo account. The current minimum query length for typos is 3, so tested with a 4 character string, with and without a 1 character typo. Verify that results come back correctly.

Searchable Facets: Add styling (#699):
* Update search input styling, add custom checkbox. 
* TEST=manual
* Visually inspect in Chrome and IE11.

Facets/FilterOptions: Show selected facets first by default (#666):
* Floats selected checkboxes/radio buttons for FilterOptions within a Facets component. Facets rebirth a new FilterBox (which births new FilterOptions) every time a new search is returned, so applying this all that is needed. Adding it as a listener to clicking the apply button actually creates a weird shuffling effect which is undesired, since what would happen is 1. user clicks apply button 2. selected options float to the top 3. new search is returned with new options
* J=SPR-1970
* TEST=manual
* Tested that Facets are still applied on searches, and after clicking the apply button (but not the search button or changing any of the options) the filter options will be sorted.

Searchable Facets: Do not set state on option click (#713):
* This fixes a bug where the options were re-rendered on click. When
that happened, the showMoreLimit would be reset, so if a user had
all options expanded, then clicked an option, it would collapse all
options. The re-render also caused a slight flash on click.
* TEST=manual
* Test by using the Facets component on my local machine and selecting
options. Note that they don't re-render on click. Clicking Apply correctly
applies the selected options. Test that options can be initialized properly
through config.


alexisgrow (commented):
These have already been approved in other PRs, going to merge.

(Opened this PR because GH would not let me view and keep refreshing the massive diff when I was trying to ensure that merge conflicts didn't overwrite anything.)

PR #829, opened 2020-06-02
Support click analytics for RTF links.

tmeyer2115:
This PR adds click analytics for links in a formatted RTF value. Note that such
values occur only in Cards or the DirectAnswers component. Now, when the
formatRichText function is called, it places a wrapper div around the newly
formatted RTF value. A 'cta-type' data attribute is also added to all links
in the value. Click events from the links are bubbled up to this wrapper div
and handled in both the CardComponent and DirectAnswersComponent classes.

The adding of the 'cta-type' attribute is handled by the markdown-it-for-inline
plugin.

Right now, we only support one RTF formatted value per Card. The solution in
this PR can be easily extended to support multiple RTF values in the future.
Some room for future improvement would be to add classes to the links in
formatRichText. This would allow us to write click handlers for the links
directly, instead of having the container div around the entire formatted RTF
value. Unfortunately, it seems the markdown-it-for-inline plugin does not
support this functionality currently.

J=SPR-2232
TEST=auto, manual

Added unit tests for the new RichTextFormatter class. Manually tested the
following scenarios for DirectAnswers and Cards (Vertical and Universal):

- Clicking outside a link, in the formatted value, does nothing.
- Clicking a phone-type link in the formatted value works as expected and
  issues the correct AnalyticsEvent.
- Clicking an email-type link in the formatted value works as expected and
  issues the correct AnalyticsEvent.
- Clicking a normal link in the formatted value works as expected and issues
  the correct AnalyticsEvent.
- For cards, verified that if no eventOptionsFieldName is specified,
  clicking a link would produce an Analytics event without fieldName. A warning
  would also be displayed in the console.

### DIFF
--- a/src/ui/components/cards/cardcomponent.js
+++ b/src/ui/components/cards/cardcomponent.js
@@ -1,7 +1,9 @@
 /** @module CardComponent */
 
 import Component from '../component';
+import DOM from '../../dom/dom';
 import { cardTypes } from './consts';
+import AnalyticsEvent from '../../../core/analytics/analyticsevent';
 
 class CardConfig {
   constructor (config = {}) {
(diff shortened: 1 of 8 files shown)


src/ui/components/results/directanswercomponent.js
```diff
@@ -106,6 +106,38 @@ export default class DirectAnswerComponent extends Component {
...
+
+    const rtfElement = DOM.query(this._container, '.js-yxt-rtfValue');
+    rtfElement && DOM.on(rtfElement, 'click', e => this._handleRtfClickAnalytics(e));
+  }
+
+  /**
+   * A click handler for links in a Rich Text Direct Answer. When such a link
+   * is clicked, an {@link AnalyticsEvent} needs to be fired.
+   *
+   * @param {Event} event The click event.
```
oshi97: most unimportant of nits but this could be MouseEvent

src/ui/components/cards/cardcomponent.js
```diff
@@ -64,11 +66,53 @@ export default class CardComponent extends Component {
     this.verticalKey = data.verticalKey;
   }
 
+  onMount () {
+    const rtfElement = DOM.query(this._container, '.js-yxt-rtfValue');
+    if (rtfElement) {
+      const fieldName = rtfElement.dataset.fieldName;
```
oshi97, line 72: could fieldName go inside _handleRtfClickAnalytics, or is this for a performance improvement?

src/ui/components/cards/cardcomponent.js
```diff
@@ -64,11 +66,53 @@ export default class CardComponent extends Component {
...
+   *
+   * @param {Event} event The click event.
+   * @param {string} fieldName The name of the Rich Text field used in the
+   *                           attriubte.
+   */
+  _handleRtfClickAnalytics (event, fieldName) {
+    if (!event.target.dataset.ctaType) {
+      return;
+    }
+    const ctaType = event.target.dataset.ctaType;
```
oshi97: nit but could could this go before the if check, and then we just check `if(!ctaType)` instead of  `if (!event.target.dataset.ctaType) {`

oshi97 (commented):
just have some debatable nits!

src/ui/components/results/directanswercomponent.js
tmeyer2115: Not unimportant! Makes for better docs.

src/ui/components/cards/cardcomponent.js
tmeyer2115: It's a small performance optimization to compute it once in the onMount as opposed to recomputing it each time the handler is called.

oshi97 (approved):
lgtm!

PR #831, opened 2020-06-02
inject library version from git describe --tags

creotutar:
This is done to remove a version release step where we must make a
commit to change the library version in the SDK constants.js.

Library version is set to the output of `git describe --tags`. This will give us how far (in commit #) from a prior release tag we are as well as the current hash, if we are not currently on a tag.

We also want to have canary builds be identified by the commit hash (and persist for a long
time). We want to use dashes in the library versions to make it easier
to navigate the CDN url.

Canary builds will only be pushed to s3 for the develop branch.
As a result, /dev/branch CDN links will not be made for the develop branch.

These library version changes will affect three places we use the library:
        1. In a release tag, used by the CDN
        2. In a dist-* release tag
        3. In the NPM registry

For 1 and 2, we rely on the gulp build process, which has been changed to
inject this build version from the current git repository.

For 3, we do not currently have the sdk in the npm registry. When we do,
this build can be included in a preinstall and inject correctly into the
files before pushing to the registry.

J=SPR-2346
TEST=manual

For 1: Tested to see that the circleci build for the PR has the git described tag in the answers.js (and propogated in the apirequest).

For 2: Tested locally on a dist tag created locally and deleted to
see a v# lib version number.

For 3: Tested the git commands for a test npm package on the registry on the
pre-install and it successfully got the tag value when the test package
was referenced in another repository.


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -19,6 +20,23 @@ const uglify = require('gulp-uglify-es').default;
 
 const NAMESPACE = 'ANSWERS';
 
+function getLibVersion () {
+  const hash = require('child_process')
+    .execSync('git rev-parse HEAD')
+    .toString().trim();
+
+  const version = require('child_process')
+    .execSync('git describe --match "v[0-9]*" --abbrev=0 HEAD --always')
```
oshi97: I tried playing around with this locally for fun, and this command hasn't been working for me :( not sure what I'm doing wrong
image

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -19,6 +20,23 @@ const uglify = require('gulp-uglify-es').default;
...
+    .toString().trim();
+
+  let libVersion = `canary-${hash}`;
```
oshi97: So is the idea here that any dev branches will have the libVersion set to canary-${commit sha}?

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -19,6 +20,23 @@ const uglify = require('gulp-uglify-es').default;
...
+  if (hash !== version) {
+    libVersion = `${version}`;
```
oshi97: nit but could we just return here if version !== hash and at the bottom, and get rid of the libVersion variable? ${version} can also just be version

oshi97 (commented):
This looks really cool! Just a few questions and minor comments

conf/gulp-tasks/library.gulpfile.js
oshi97: I think my issue is I'm not using an annotated git tag, do we need to support lightweight tags as well?

src/core/constants.js
```diff
@@ -1,7 +1,7 @@
 /** @module */
 
 /** The current lib version, reported with errors and analytics */
-export const LIB_VERSION = 'v1.3.1';
+export const LIB_VERSION = '@@LIB_VERSION';
```
tmeyer2115, line 4: Let's add a comment(s) indicating that this value is now injected. That could help minimize confusion for any new person navigating our code.

conf/gulp-tasks/library.gulpfile.js
tmeyer2115: I'm wondering if we should hold off on shipping this PR until we figure out what canary will look like in the CDN. 

@bryanreed have you and Christian discussed some ideas about canary and how it will appear in the CDN? I don't think we'd want to have assets in the CDN for each commit in the develop/canary branch. That would mean a pretty big increase in the number of assets we store in S3. I was envisioning we'd have a single canary target in the CDN that would get updated every time a new commit was merged into develop/canary.

conf/gulp-tasks/library.gulpfile.js
creotutar: This is surprising. I would expect all of our release tags to be annotated tags instead of lightweight tags but that is not the case. It seems the Github UI does not allow for annotated tags. (All of our dist tags are annotated tags). @bryanreed What do you suggest here?

https://stackoverflow.com/questions/11514075/what-is-the-difference-between-an-annotated-and-unannotated-tag#:~:text=Annotated%20tags%20are%20meant%20for,private%20or%20temporary%20object%20labels.&text=annotated%20tags%20can%20contain%20a,without%20making%20a%20release%20commit.

creotutar: Using lightweight tags.

conf/gulp-tasks/library.gulpfile.js
creotutar: Hiya, as discussed, canary branches will only be added to the develop branch, but we will keep the hash so users can pin to them.

conf/gulp-tasks/library.gulpfile.js
creotutar: done

src/core/constants.js
creotutar: done in the comment above the const

creotutar:
building on a branch without tags will produce fatals below (but are caught)
image

I'm beginning to feel like we need a separate build for version releases vs. branches. Thus we only have to `git describe` for tag builds and not see the above.

oshi97:
would something like `git tag --points-at HEAD` work? from https://stackoverflow.com/questions/2324195/how-to-get-tags-on-current-commit

creotutar:
Hey @oshi97. That is actually just what I needed lol. Though, I talked to @bryanreed yesterday, we are going to move over to have all versions be just a plain `git describe --tags` as that gives us handy information like "how many commits away from the last tag you are" as well as the commit hash.

creotutar:
Lib version for this branch currently is `	var LIB_VERSION = 'v1.3.1-12-gf0817fe';`
https://assets.sitescdn.net/answers/dev/auto-version/answers.js

.circleci/config.yml
```diff
@@ -82,8 +82,23 @@ jobs:
       - image: circleci/python:2.7
     working_directory: ~/answers
     steps:
+      - run:
+          name: "Setup branch formatting, replace / with -"
+          command: echo 'export FORMATTED_CIRCLE_BRANCH=$(echo ${CIRCLE_BRANCH} | sed "s/\//-/g")' >> $BASH_ENV
```
oshi97, line 87: could you explain to me what sed does here exactly? my only experience with sed is a reddit link on somebody implementing snake with it (l o l), reading up on some docs right now

oshi97: is this replacing all forward slashes with dashes?

creotutar: sure thing, sed is editing whatever is input to it
`s` is the replace command, delimits params with the `/`
so it's searching for `\/` which is an escaped `/` and replacing it with `-`
the `g` flag makes it global, looking everywhere in the stream instead of just one place

that's from memory, correct me if im wrong anywhere

oshi97 (approved):
lgtm! if Tom could take a quick look on top that would be great

bryanreed (approved):

oshi97 (approved):
*tweet tweet*

tmeyer2115 (approved):

PR #832, opened 2020-06-02
change circleci resource class to large for unit tests

creotutar:
This is necessary for unit tests to run correctly for pull requests.
Otherwise we get errors

    Call retries were exceeded
      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)

J=None
TEST=manual

Copying this from the branch where it is working, v1.4.0


oshi97 (approved):
do it!!!!

PR #833, opened 2020-06-03
Accordion caret

sawong99:
Updated accordion-card caret dimensions in the SDK.

J=SPR-2343

TEST=manual

Tested on desktop with local test site.


.idea/workspace.xml
```diff
@@ -0,0 +1,280 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
oshi97: we probably shouldn't commit this file, it looks like an intellij specific thing?

package.json
```diff
@@ -4,52 +4,52 @@
   "description": "Javascript Answers Programming Interface",
   "main": "gulpfile.js",
   "dependencies": {
-    "@yext/rtf-converter": "^1.0.0",
+    "@yext/rtf-converter": "^1.1.1",
```
oshi97: I think updating our npm deps would be great, but it should be a separate PR. I believe our stance is we need to be careful to test that nothing breaks when we update a dependency (@tmeyer2115 correct me if I'm wrong). probably stuff used only in testing like testcafe and jest can be updated whenever,  but things that affect the actual bundle like handlebars have a bit of risk

oshi97 (commented):
mostly lgtm! Don't think we need to update package.json or check in the intellij workspace

oshi97 (approved):
lgtm! reminder to choose squash and merge it's easy to misclick (I wish github would make them different shades of green or something)

PR #834, opened 2020-06-03
FilterOptions: fix showMoreLimit for single-option

oshi97:
The template was not giving single-options the hidden
css class if they were over the showMoreLimit. It was
also giving the hidden class even if config.showMore was
false. Also added 2 js classes when the control is singleoption
to match multioption.

QA=v1.4.0 no. 5
TEST=manual,auto
Test that showMoreLimit works for both multioption and singleoption
inside a FilterBox
Test that works when showMore is true, false, and unset


alexisgrow (approved):
🎉

PR #835, opened 2020-06-03
Pagination: add missing --yxt-pagination-color-hover variable

oshi97:
This css var was being referenced but not initialized. This was probably
missed during a rebase. Now there are 7 sass variables and 7 matching scss variables.

TEST=manual
test that background color for pagination on hover is not white anymore


alexisgrow (approved):
LGTM!

PR #837, opened 2020-06-04
FilterOptions: Add clear button to searchable options

alexisgrow:
Add a clear button that clears the search input and otherwise has the same affect as an empty search filter: removes the search styling.

TEST=manual

Test that clear button is not there on page load. Start typing in the search input and see clear button appear. Click button and see (1) search input clear,  (2) clear button disappear, and (3) search styling removed.

### DIFF
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -342,30 +342,41 @@ export default class FilterOptionsComponent extends Component {
       const filterOptionEls = DOM.queryAll(this._container, '.js-yxt-FilterOptions-option');
       const filterContainerEl = DOM.query(this._container, '.js-yxt-FilterOptions-container');
 
-      // On clearSearchEl click, clear search
+      // On clearSearchEl click, clear search input
       if (clearSearchEl && searchInputEl) {
         DOM.on(clearSearchEl, 'click', event => {
           searchInputEl.value = '';
+          searchInputEl.dispatchEvent(new Event('input', {
+            'bubbles': true,
+            'cancelable': true
+          }));
           searchInputEl.focus();
-          this._clearFilterStyling(filterOptionEls, filterContainerEl, clearSearchEl);
         });
       }
 
       DOM.on(
         searchInputEl,
-        'keyup',
+        'input',
         event => {
           const filter = event.target.value;
 
           if (!filter) {
-            this._clearFilterStyling(filterOptionEls, filterContainerEl, clearSearchEl);
+            filterContainerEl.classList.remove('yxt-FilterOptions-container--searching');
+            clearSearchEl.classList.add('js-hidden');
           } else {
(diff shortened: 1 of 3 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -401,6 +411,33 @@ export default class FilterOptionsComponent extends Component {
...
+   * Returns label of a given filter option
+   * @returns {string}
+   * @private
+   */
+  _clearFilterStyling (filterOptionEls, filterContainerEl, clearSearchEl) {
+    filterContainerEl.classList.remove('yxt-FilterOptions-container--searching');
+
+    for (let option of filterOptionEls) {
+      const labelEl = DOM.query(option, '.js-yxt-FilterOptions-optionLabel--name');
+      clearSearchEl.classList.add('js-hidden');
```
tmeyer2115: Do we want this outside the loop?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -401,6 +411,33 @@ export default class FilterOptionsComponent extends Component {
     }
   }
 
+  /**
+   * Returns label of a given filter option
```
tmeyer2115: Update JSDoc

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -337,25 +337,35 @@ export default class FilterOptionsComponent extends Component {
...
+          searchInputEl.value = '';
+          searchInputEl.focus();
+          this._clearFilterStyling(filterOptionEls, filterContainerEl, clearSearchEl);
+        });
+      }
+
       DOM.on(
-        DOM.query(this._container, '.js-yxt-FilterOptions-filter'),
+        searchInputEl,
         'keyup',
```
tmeyer2115: Minor thing, but could we listen for an 'input' event here instead? That could allow us to consolidate a bit. In the handler for clicking the clearSearchEl, we wouldn't need to call this._clearFilterStyling (I think). We'd rely on this block for that.

alexisgrow: Ah, yes, that is much better. Updating!

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: 🙈

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: Yes definitely, updating

src/ui/components/filters/filteroptionscomponent.js
alexisgrow: I was testing this, requires firing an "input" event manually and is much cleaner! Was able to remove the majority of the JS changes previously in the PR 🎉

tmeyer2115 (approved):

PR #838, opened 2020-06-04
UniversalResults: add appliedFilters.showChangeFilters config option

oshi97:
This commit moves the showChangeFilters config option under the appliedFilters
object. This config option is only used on universal results, and displays
a link to the verticalURL for a specific vertical. Also updates the readme
for universal results to use the recently added appliedFilters config object.

QA=v1.4.0 no. 13
TEST=manual
test that on universal, can specify showChangeFilters in either the
root component config or in the appliedFilters config object, and that
the appliedFilters config takes precedence.
test that changeFilters do not show up on vertical results no matter
what

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,9 +486,9 @@ ANSWERS.addComponent('UniversalResults', {
       appliedFilters: {
         // If true, show any applied filters that were applied to the universal search. Defaults to false
         show: true,
-        // If showAppliedFilters is true, whether to display the field name of an applied filter, e.g. "Location: Virginia" vs just "Virginia". Defaults to false.
+        // If appliedFilters.show is true, whether to display the field name of an applied filter, e.g. "Location: Virginia" vs just "Virginia". Defaults to false.
         showFieldNames: false,
-        // If showAppliedFilters is true, this is list of filters that should not be displayed.
+        // If appliedFilters.show is true, this is list of filters that should not be displayed.
         // By default, builtin.entityType will be hidden
         hiddenFields: ['builtin.entityType'],
         // The character that separates the count of results (e.g. “1-6”) from the applied filter bar. Defaults to '|'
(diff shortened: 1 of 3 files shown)


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -104,20 +104,20 @@ class VerticalResultsConfig {
...
        * @type {boolean}
        */
-      removable: defaultConfigOption(config, ['appliedFilters.removable'], false)
+      removable: defaultConfigOption(config, ['appliedFilters.removable'], false),
+
+      /**
+       * Whether to show the change filters link on universal results.
+       * @type {boolean}
+       **/
+      showChangeFilters: defaultConfigOption(config, ['appliedFilters.showChangeFilters', 'showChangeFilters'], false)
```
tmeyer2115, line 113: If appliedFilters.show is false, should showChangeFilters always be false as well? Or is this option not dependent on appliedFilters.show at all?

README.md
```diff
@@ -482,18 +482,23 @@ ANSWERS.addComponent('UniversalResults', {
...
+        show: true,
+        // If appliedFilters.show is true, whether to display the field name of an applied filter, e.g. "Location: Virginia" vs just "Virginia". Defaults to false.
+        showFieldNames: false,
+        // If appliedFilters.show is true, this is list of filters that should not be displayed.
+        // By default, builtin.entityType will be hidden
+        hiddenFields: ['builtin.entityType'],
+        // The character that separates the count of results (e.g. “1-6”) from the applied filter bar. Defaults to '|'
+        resultsCountSeparator: '|',
+        // Whether to display the change filters link in universal results. Defaults to true.
+        showChangeFilters: true,
```
tmeyer2115: Is the spec incorrect? It lists a default of false

src/ui/components/results/verticalresultscomponent.js
oshi97: The current behavior is: if there are no applied filters OR appliedFilters.show is false, the change filters link won't render even if it's true

README.md
oshi97: ugh sorry I assumed the old default was correct but it was not

tmeyer2115 (approved):

PR #839, opened 2020-06-04
FilterBox: fix resetFilters and resetFiltersLabel config options

oshi97:
These config options probably have not worked for a very long time.
Now, resetFilters will control whether or not FilterBox displays a reset all button.
If resetFilters is not specified, it will default to its previous behavior, which was
displaying a reset button if searchOnChange was false. The text on the button will
now correctly try to use config.resetFiltersLabel first before using the default text.

QA=v1.4.0 no. 7
TEST=manual, auto
Test that if resetFilters is not specified, will display a reset button and apply button if searchOnChange
is true, otherwise neither will be displayed
Test that if resetFilters IS false, whether searchOnChange is true or false no reset button appears
Test that if resetFilters is true, whether searchOnChange is true or false a reset button will appear
Test that facets still work properly
Sent codepen with dev branch to Rose to verify styling is ok

### DIFF
--- a/README.md
+++ b/README.md
@@ -908,9 +908,9 @@ ANSWERS.addComponent('FilterBox', {
   resetFilter: false,
   // Optional, the label to use for the reset button above, this will only display if searchOnChange is false
   resetFilterLabel: 'reset',
-  // Optional, show a reset-all button for the filter control, this will only display if searchOnChange is false
+  // Optional, show a reset-all button for the filter control. Defaults to displaying a reset button if searchOnChange is false.
   resetFilters: true,
-  // Optional, the label to use for the reset-all button above, this will only display if searchOnChange is false
+  // Optional, the label to use for the reset-all button above, this will only display if resetFilters is true.
   resetFiltersLabel: 'reset-all',
   // Optional, allow collapsing excess filter options after a limit
   showMore: true,
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -40,13 +40,13 @@ class FilterBoxConfig {
      * If true, show a "reset all" button to reset all facets
      * @type {boolean}
      */
-    this.resetFilters = config.resetFacets === undefined ? true : config.resetFacets;
+    this.resetFilters = config.resetFilters === undefined ? !config.searchOnChange : config.resetFilters;
 
     /**
      * The label to show for the "reset all" button
      * @type {string}
      */
-    this.resetFiltersLabel = config.resetFacetsLabel || 'reset all';
+    this.resetFiltersLabel = config.resetFiltersLabel || 'reset all';
 
     /**
      * The max number of facets to show before displaying "show more"/"show less"
--- a/src/ui/sass/modules/_FilterBox.scss
+++ b/src/ui/sass/modules/_FilterBox.scss
@@ -60,5 +60,9 @@
 
     margin-left: 10px;
     letter-spacing: 0.5px;
+
+    &--noApplyButton {
+      margin-left: 0;
+    }
   }
 }
--- a/src/ui/templates/filters/filterbox.hbs
+++ b/src/ui/templates/filters/filterbox.hbs
@@ -12,14 +12,16 @@
     <div class="js-yext-filterbox-filter{{@index}} yxt-FilterBox-filter">
     </div>
   {{/each}}
-  {{#if showApplyButton}}
+  {{#if (or showApplyButton showReset)}}
     <div class="yxt-FilterBox-controls">
-      <button type="button"
-              class="js-yext-filterbox-apply yxt-FilterBox-apply">
-        {{applyLabel}}
-      </button>
+      {{#if showApplyButton}}
+        <button type="button"
+                class="js-yext-filterbox-apply yxt-FilterBox-apply">
+          {{applyLabel}}
+        </button>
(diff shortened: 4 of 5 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -40,13 +40,13 @@ class FilterBoxConfig {
      * If true, show a "reset all" button to reset all facets
      * @type {boolean}
      */
-    this.resetFilters = config.resetFacets === undefined ? true : config.resetFacets;
+    this.resetFilters = config.resetFilters === undefined ? !config.searchOnChange : config.resetFilters;
```
oshi97, line 43: Note!: This change does not break facets, before facets were passing the value of config.resetFacets both as config.resetFilters AND as config.resetFacets, see https://github.com/yext/answers/blob/master/src/ui/components/filters/facetscomponent.js#L209

README.md
```diff
@@ -908,9 +908,9 @@ ANSWERS.addComponent('FilterBox', {
   resetFilter: false,
   // Optional, the label to use for the reset button above, this will only display if searchOnChange is false
   resetFilterLabel: 'reset',
-  // Optional, show a reset-all button for the filter control, this will only display if searchOnChange is false
+  // Optional, show a reset-all button for the filter control. Defaults to displaying a reset button if searchOnChange is false.
```
tmeyer2115, line 911: The commit message says the default behavior was to display a reset button if searchOnChange was true. Is that incorrect?

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -40,13 +40,13 @@ class FilterBoxConfig {
...
 
     /**
      * The label to show for the "reset all" button
      * @type {string}
      */
-    this.resetFiltersLabel = config.resetFacetsLabel || 'reset all';
+    this.resetFiltersLabel = config.resetFiltersLabel || 'reset all';
```
tmeyer2115, line 49: Was the Facets component also setting resetFacetsLabel and resetFiltersLabel to be the same?

src/ui/components/filters/filterboxcomponent.js
tmeyer2115: resetFilter is also listening to resetFacet. Can we correct that too?

tmeyer2115: Where is the default of !config.searchOnChange coming from? Is that maintaining backwards compatibility? It looks like the default was true before.

oshi97: The default WAS true before. But in the template, the reset button would only render if the showApplyButton is true, and showApply is set to !config.searchOnChange

oshi97: I was going to have fixing resetFilter as a separate PR, should I add it here?

src/ui/components/filters/filterboxcomponent.js
oshi97: yup! what it's currently doing is passing resetFiltersLabel: config.resetFacetsLabel, as well as passing the whole facets config in the object assign https://github.com/yext/answers/blob/v1.3.0/src/ui/components/filters/facetscomponent.js#L203

README.md
oshi97: Yeah this readme was wrong (for once I believe I am correct mwahaha!). 
1: in the config class config.resetFilters was being set to config.resetFacets, and defaulting to true. 
2: since the user should never be setting config.resetFacets, we can assume config.resetFilters was always true.
3: HOWEVER, in the template, the reset button would only render if showApplyButton is true.
4: showApplyButton is always set to !config.searchOnChange
5: Therefore the reset button only renders if !config.searchOnChange
6: Therefore the reset button would only render if searchOnChange is false

src/ui/components/filters/filterboxcomponent.js
tmeyer2115: Separate PR is fine! Thanks for taking care of it.

README.md
tmeyer2115: Makes sense to me! I think we want to update the commit message for this item then. It says: "If resetFilters is not specified, it will default to its previous behavior, which was
displaying a reset button if searchOnChange was true"

tmeyer2115 (commented):
Before, it sounds like the "Reset filters" button was completely controlled by searchOnChange. Now, if someone sets searchOnChange to true, they can still have a "Reset filters" button. Is that desired?

oshi97:
> Before, it sounds like the "Reset filters" button was completely controlled by searchOnChange. Now, if someone sets searchOnChange to true, they can still have a "Reset filters" button. Is that desired?

Yup asked Rose while working on this and she said that's what we want! Link to google sheets discussion https://docs.google.com/spreadsheets/d/1aOqBeX8rrVq7i26EiUhuWFTnZ55CjPycNVoK8tP0VtE/edit?disco=AAAAJocwiCI

>+oshi@yext.com codepen LGTM -- i think it's a little weird to show the link without the apply but dealers choice (and someone can add styling on top if they neeed to). searchOnChange still controls whether the apply button shows up, correct?

README.md
oshi97: Oh shoot that's my bad

oshi97: updated!

tmeyer2115 (approved):

PR #841, opened 2020-06-04
ResultsHeader: make removable filters a tabbable button

oshi97:
This commit updates the removable filter tags to be a tabbable button

v1.4.0 QA: 15, 16
TEST=manual
Test I can tab onto the filter tag
when it is tabbed onto the styling will update to the hover/focus styling
tabbing to it doesn't change my cursor to a pointer, but it does change if I hover over it
can press enter to remove the filter


tmeyer2115 (approved):
We group the non-removable filters first, right? So tab navigation will skip over that whole block to the first removable filter?

oshi97:
> We group the non-removable filters first, right? So tab navigation will skip over that whole block to the first removable filter?

Yup!

PR #844, opened 2020-06-04
Update accordion card to check for empty answer details

sawong99:
Updated the accordion card so that if answer details are empty/blank, the accordion card only displays its header and doesn't expand.

J=SPR-2375

TEST=manual

Tested on local test site. Changed accordion card dataMappings to test when details is null/"" or not.


src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -143,13 +143,13 @@ export default class AccordionCardComponent extends Component {
...
 
   onMount () {
-    const toggleEl = DOM.query(this._container, '.js-yxt-AccordionCard-toggle');
-    const accordionBodyEl = DOM.query(this._container, '.js-yxt-AccordionCard-body');
     const accordionEl = DOM.query(this._container, '.js-yxt-AccordionCard');
-
-    accordionBodyEl.style.height = `${this.isExpanded ? accordionBodyEl.scrollHeight : 0}px`;
-
-    DOM.on(toggleEl, 'click', () => this.handleClick(toggleEl, accordionBodyEl, accordionEl));
+    if (!(this._config.details == null || this._config.details === '')) {
```
oshi97: I think we can just do `if (this._config.details)` here!
@alexisgrow might remember better, but I think the weird `this.details = this.details === null ? null : (this.details || result.details || rawResult.details || '');` was written with regards to some odd qa item where they wanted the details to default to result data if it was undefined, but not default if it was null

alexisgrow: Yes! Can this be `if (this._config.details)` or `if (toggleEl)`?

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -52,7 +52,7 @@ class AccordionCardConfig {
     /**
      * @type {string}
      */
-    this.details = this.details === null ? null : (this.details || result.details || rawResult.description || '');
+    this.details = this.details === null ? null : (this.details || result.details || rawResult.details || '');
```
alexisgrow: Is the rawResult.**description** -> rawResult.**details** change part of this item?

src/ui/templates/cards/accordion.hbs
```diff
@@ -1,26 +1,41 @@
...
+          <div class="yxt-AccordionCard-ctas" data-component="CTACollection" data-prop="result"></div>
+        {{/if}}
+      </div>
     </div>
-  </div>
+  {{/if}}
 </div>
+    
+
+
```
alexisgrow: minor nit: Can these extra blank lines be deleted?

package-lock.json
```diff
@@ -7284,7 +7284,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
```
alexisgrow: Since there aren't package.json changes in this PR, let's change the package-lock in a separate PR. If the package-lock.json is not in-sync with the package.json, we should definitely update it, but it would be good to separate that change from these accordion card changes.

src/ui/sass/modules/_AccordionCard.scss
```diff
@@ -61,6 +61,18 @@ $accordion-card-spacing--expanded: var(--yxt-base-spacing) !default;
     border-top: 0;
   }
   
+  &-empty-header
```
alexisgrow: I don't see this selector being used in the .hbs, so if it is unused, could it be deleted? hahaha if not: this looks really similar to the `&-toggle` styling below. Can you add the "&-empty-header" selector to that rule-set and then use specificity or ordering to change the properties that are different?

alexisgrow (commented):
looking good so far 😸 left a few comments!

src/ui/components/cards/accordioncardcomponent.js
sawong99: Changed it to if (this._config.details)!

src/ui/components/cards/accordioncardcomponent.js
sawong99: Ah, I made the wrong change! I will change it back.

src/ui/components/cards/accordioncardcomponent.js
```diff
@@ -143,13 +143,13 @@ export default class AccordionCardComponent extends Component {
   }
 
   onMount () {
-    const toggleEl = DOM.query(this._container, '.js-yxt-AccordionCard-toggle');
-    const accordionBodyEl = DOM.query(this._container, '.js-yxt-AccordionCard-body');
     const accordionEl = DOM.query(this._container, '.js-yxt-AccordionCard');
```
oshi97: ah really minor nit but can we move `const accordionEl =` inside the if statement? it's not being used outside of it

src/ui/templates/cards/accordion.hbs
```diff
@@ -1,26 +1,38 @@
...
+        {{{_config.title}}}
+      </h3>
+      <div class="yxt-AccordionCard-icon js-yxt-AccordionCard-icon{{#if isExpanded}} yxt-AccordionCard-icon--expanded{{/if}}"
+        data-component="IconComponent"
+        data-opts='{"iconName": "chevron"}'
+        data-prop="icon">
+      </div>
+    </button>
+  {{else}}
+    <div class="yxt-AccordionCard-toggle">
```
alexisgrow: minor nit: This class has `cursor: pointer` in its styling. Can this element have a normal cursor styling since clicking it does not do anything?

oshi97:
lgtm! @alexisgrow does this look okay to you?

.circleci/config.yml
```diff
@@ -62,6 +62,7 @@ jobs:
   unit_test:
     docker:
       - image: circleci/node:7.10
+    resource_class: large
```
alexisgrow: It looks like this change has already been made in v1.3.2 here (https://github.com/yext/answers/commit/83843468000227ba9e273f345c6dbbb81dddbc59). Can you merge v1.3.2 into this branch so this PR only shows changes for the accordion card? Otherwise LGTM!

alexisgrow (approved):
🔥🔥🔥 LGTM!

PR #845, opened 2020-06-04
FilterOptions: pass additional needed state to the template

oshi97:
The show more/less mechanism FilterOptions was using was changing
DOM element classes but not propagating these changes into setState,
meaning if the component calls setState sometimes unexpected show more/less
behavior would occur.

Also update this._getSelectedCount to use this.config.options instead of
doing a DOM query. This should improve performance slightly, especially
for facets with 100 options.

Also remove unused variable this.allShown and unused method this.clear().

TEST=manual, auto
Added a unit test to make sure showMore renders for single option.
Tested that with storeOnChange both true and false, and multioption
and singleoption, can use the reset button and show more toggle without weird behavior

### DIFF
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -263,10 +263,12 @@ export default class FilterOptionsComponent extends Component {
     this.expanded = this.config.showExpand ? selectedCount > 0 : true;
 
     /**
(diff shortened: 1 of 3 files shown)


alexisgrow (approved):
🔥 🔥 🔥 lgtm!

alexisgrow (approved):
🥳 🥳

PR #846, opened 2020-06-04
FilterBox: fix resetFilter and resetFilterLabel

oshi97:
FilterBox was using the wrong config names.

TEST=manual
Test that can set resetFilterLabel, and also that toggling resetFilter
will cause reset buttons to appear for each FilterOptions if it is true


alexisgrow (approved):
😆 LG!

PR #850, opened 2020-06-04
FilterOptions: have cursor: pointer for singleoption radio button labels

oshi97:
Have the cursor be a pointer when it hovers over both the input and
the label (just like for multioption). Fix the for attribute attached
the label I goofed up earlier so clicking it selects the associated button.

TEST=manual
test that single options have cursor:pointer for both label and input, and
clicking either will select the option

### DIFF
--- a/src/ui/sass/modules/_FilterOptions.scss
+++ b/src/ui/sass/modules/_FilterOptions.scss
@@ -114,6 +114,10 @@
     }
   }
 
+  &-radioButtonLabel {
+    cursor: pointer;
+  }
+
   &-optionLabel {
     position: relative;
     font-size: $font-size-md;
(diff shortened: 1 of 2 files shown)


src/ui/sass/modules/_FilterOptions.scss
```diff
@@ -217,6 +221,10 @@
     margin-left: 0;
   }
 
+  &-radioButtonInput {
+    cursor: pointer;
+  }
+
```
alexisgrow, line 227: Would it make sense for this styling to go on the parent of `&-radioButtonInput` and `&-radioButtonLabel`? Thinking that maybe later we'll add margin between these elements and if the area between them is clickable then it may need a cursor: pointer too.

alexisgrow (approved):
left a q but lgtm as-is!

src/ui/sass/modules/_FilterOptions.scss
oshi97: just tried adding that, but it seemed a little weird because clicking that array wouldn't actually toggle the option

PR #851, opened 2020-06-04
ResultsHeader: shrink margin-right for nonremovable applied filters

oshi97:
Rose thought the spacing was too big (I like the new spacing too).

TEST=manual
check margin-right on universal and vertical


alexisgrow (approved):
😍

PR #852, opened 2020-06-04
FilterBox: filters config options overrides FilterBox top-level config

alexisgrow:
The FilterOptions config, which is more specific, takes priority over the FilterBox config.

TEST=manual

Set showMoreLimit in filterConfigs for a field, also set it in the FilterBox config. Note that the one set in filterConfigs takes affect.


oshi97 (approved):
amazing!

PR #854, opened 2020-06-05
Add ability to configure 'target' behavior of rich text content.

tmeyer2115:
This PR adds a 'target' parameter to the SDK's formatRichText function. This
parameter, when specified, sets the 'target' attribute of all the formatted RTF
links. It is optional.

J=SPR-2134
TEST=auto, manual

Added unit tests for the updated formatRichText function. Manually verified:
- Passing a 'target' of '_blank' to formatRichText added the correct 'target'
  attribute to all of the link anchor tags.
- When I did not provide a 'target' to formatRichText, the link anchor tags had
  no 'target' attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1860,14 +1860,22 @@ The Answers SDK exposes a `formatRichText` function which translates CommonMark 
 ensure that a Rich Text Formatted value is shown properly on the page. To use this function, call it like so:
 
 ```js
-ANSWERS.formatRichText(rtfFieldValue, target, eventOptionsFieldName)
+ANSWERS.formatRichText(rtfFieldValue, targetConfig, eventOptionsFieldName)
 ```
 
 For instance, this function can be used in the `dataMappings` of a Card to display an RTF attribute. 
 
 When clicking any link in the resultant HTML, an `AnalyticsEvent` will be fired. If the `eventOptionsFieldName` has been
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1860,13 +1860,16 @@ The Answers SDK exposes a `formatRichText` function which translates CommonMark
 ensure that a Rich Text Formatted value is shown properly on the page. To use this function, call it like so:
 
 ```js
-ANSWERS.formatRichText(rtfFieldValue, eventOptionsFieldName)
+ANSWERS.formatRichText(rtfFieldValue, target, eventOptionsFieldName)
```
oshi97: would we want to generalize this to an attributes param object, that lets you stick arbitrary attributes onto the html, or is that overengineered?

tmeyer2115: We currently don't have requirements for that. The HHs are really only interested in changing the 'target' attribute of a link.

oshi97 (approved):

oshi97 (approved):
lgtm! I guess whenever we want to use a specific plugin we have to do this add and disable pattern?

tmeyer2115:
> lgtm! I guess whenever we want to use a specific plugin we have to do this add and disable pattern?

Yep! If you're okay with the plugin you added being used during subsequent RichTextFormatter#format calls, no need to disable!

PR #855, opened 2020-06-05
FilterBox: fix overriden onChange

oshi97:
FilterBox uses FilterOption's onChange method as part of its implementation.
When the user sets a custom onChange method for a specific FilterBox, this
would override the onChange given by FilterBox, which was necessary for functionality.

TEST=manual
Test that can specify onChange for a child FilterOptions within FilterBox and
searchOnChange will still work
Test that works when custom onChange is not specified

### DIFF
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -204,7 +204,7 @@ export default class FilterBoxComponent extends Component {
           const _saveFilterNodes = this.config.searchOnChange || alwaysSaveFilterNodes;
           const _searchOnChange = this.config.searchOnChange && !blockSearchOnChange;
           this.onFilterNodeChange(i, filterNode, _saveFilterNodes, _searchOnChange, config.onChange);
-        },
+        }
       });
       if (this.config.isDynamic && typeof component.floatSelected === 'function') {
         component.floatSelected();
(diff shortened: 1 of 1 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -189,25 +189,23 @@ export default class FilterBoxComponent extends Component {
...
+        container: `.js-yext-filterbox-filter${i}`,
+        showReset: this.config.resetFilter,
+        resetLabel: this.config.resetFilterLabel,
+        showExpand: this.config.expand,
+        isDynamic: this.config.isDynamic,
+        ...config,
+        onChange: (filterNode, alwaysSaveFilterNodes, blockSearchOnChange) => {
+          const _saveFilterNodes = this.config.searchOnChange || alwaysSaveFilterNodes;
+          const _searchOnChange = this.config.searchOnChange && !blockSearchOnChange;
+          this.onFilterNodeChange(i, filterNode, _saveFilterNodes, _searchOnChange, config.onChange);
```
tmeyer2115: Could we just invoke config.onChange() here instead of needing to pass it into onFilterNodeChange? It's a bit strange having a method called onFilterNodeChange that takes in a function called onChange.

oshi97: yeah definitely, updated!

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -259,7 +260,6 @@ export default class FilterBoxComponent extends Component {
     if (searchOnChange) {
       this._search();
     }
-    onChange && onChange();
```
tmeyer2115: I think you need to update the onFilterNodeChange method signature now that this is removed.

oshi97: my bad, updated

tmeyer2115 (approved):

PR #856, opened 2020-06-05
ResultsHeader: align elements

oshi97:
Spacing was slightly misaligned. I think we have to put margin-bottom
on the elements within the header instead of on the container, otherwise
the margin-bottom on the removable filter tags needs to be removed,
which causes them to be stacked without spacing when they
overflow into multiple rows.

TEST=manual
check that all elements in the results header have the same bottom spacing
check that on universal top and bottom spacing of the bar is even

### DIFF
--- a/src/ui/sass/modules/_ResultsHeader.scss
+++ b/src/ui/sass/modules/_ResultsHeader.scss
@@ -137,6 +137,7 @@ $results-header-filters-line-height: var(--yxt-line-height-md) !default;
     padding-left: 5px;
     padding-right: 4px;
     margin-right: calc(var(--yxt-results-header-spacing) / 2);
+    white-space: nowrap;
     @include Text(
       $size: var(--yxt-font-size-sm),
       $color: var(--yxt-color-text-neutral),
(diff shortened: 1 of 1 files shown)


oshi97:
QA description: "Current: yxt-ResultsHeader-filterValue, and .yxt-ResultsHeader-changeFilters has a margin-bottom of 4px, which makes the top/bottom look uneven on universal search.

Expected: Up to you guys on where we put this, the margin should be .25rem top and bottom for all the applied filters components (on both universal and vertical). "

tmeyer2115 (approved):
LGTM! How does the spacing/styling look if we have two rows of non-removable filters? Or a row of non-removable and a row of removable?

oshi97:
> LGTM! How does the spacing/styling look if we have two rows of non-removable filters? Or a row of non-removable and a row of removable?

Think it looks alright! I think I just need to add whitespace: nowrap so the X button doesn't wrap to the next line in certain cases

image

oshi97:
image

PR #860, opened 2020-06-08
ResultsHeader: different filter spacing when removable

oshi97:
item below
"Current: Margin-right for .yxt-ResultsHeader-resultsCount, yxt-ResultsHeader-filterLabel, yxt-ResultsHeader-filterValue  AND .yxt-ResultsHeader-filterSeparator, .yxt-ResultsHeader-resultsCountSeparator is calc(var(--yxt-results-header-spacing)/2)
Expected: When removable: false and always on universal, margin-right for these classes is calc(var(--yxt-results-header-spacing)/4)

I know this differs from the mocks there's just a lot of space right now that feels unnecessary. We need them for removability because there's supposed to be 1rem of spacing between touch targets.  "

TEST=manual
check that on universal, or vertical with appliedFilters.removable = false, has spacing of 4px
check that with appliedFilters.removable = true, has spacing of 8px

### DIFF
--- a/src/ui/sass/modules/_ResultsHeader.scss
+++ b/src/ui/sass/modules/_ResultsHeader.scss
@@ -42,7 +42,7 @@ $results-header-filters-line-height: var(--yxt-line-height-md) !default;
 
   &-resultsCount
   {
-    margin-right: calc(var(--yxt-results-header-spacing) / 2);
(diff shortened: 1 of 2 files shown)


alexisgrow (approved):
LGTM

PR #861, opened 2020-06-08
Filters: Allow field level overrides for showCount, pass through expanded

alexisgrow:
v1.4.0 QA: showCount and expanded should be passed through from the "fields" config in Facets.

TEST=manual

On local test experience, set "expanded: false" at a field-level in config with "expanded: true" at the component-level config. See field's filter un-expanded and the rest of the fields in the component expanded. Set "showCount: false" at a field-level config. See the count is hidden for the field, but it still displays for all the other fields.

### DIFF
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -154,16 +154,10 @@ export default class FilterBoxComponent extends Component {
      */
     this._filterNodes = [];
 
-    if (!this.config.showCount) {
-      this.config.filterConfigs.forEach(config => {
-        config.options.forEach(option => {
-          option.countLabel = null;
-        });
-      });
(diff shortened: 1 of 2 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -161,6 +161,14 @@ export default class FilterBoxComponent extends Component {
         });
       });
     }
+
```
oshi97: would it be possible to do this in the above loop?

alexisgrow: Ah yes, I can refactor that loop so that it works for both!

oshi97 (approved):
lgtm!

tmeyer2115 (approved):

PR #862, opened 2020-06-08
FilterOptions: Add focus state for radio button input

alexisgrow:
TEST=manual

Locally run experience, click radio buttons and see the focus state apply focused input.


oshi97 (approved):
nice!

PR #863, opened 2020-06-08
DirectAnswer: allow for custom component

oshi97:
This commit allows users the use a custom component for DirectAnswer.
The main difference between this and directly adding a custom component
with ANSWERS.addComponent is that you don't have to import StorageKeys
from the SDK.

I tried to find a way to directly call this.addChild() within the onMount()
of the direct answer component, to reduce the number of unnecessary containers
in the DOM. I found that when doing this the SDK will create an additional copy
of the original container, and decided not to mess with it after a while.

TEST=manual
tested with below component

class CustomDA1 extends ANSWERS.Component {
  constructor(config = {}, systemConfig = {}) {
    super(config, systemConfig);
    this.setTemplate('{{log this}}<div>{{answer.value}}</div>')
  }

  setState(data) {
    return super.setState(data);
  }

  static get type() {
    return 'CustomDA1';
  }
}
ANSWERS.registerComponentType(CustomDA1);

saw that direct answer data was being passed through both in the {{log this}} and
that it can be used in the template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,8 @@ The Direct Answer Component will render the BEST result, if found, based on the 
 ANSWERS.addComponent('DirectAnswer', {
   // Required, the selector for the container element where the component will be injected
   container: '.direct-answer-container',
+  // Optional, a custom card component to use.
+  defaultCard: 'MyCustomCard',
   // Optional, the selector for the form used for submitting the feedback
   formEl: '.js-directAnswer-feedback-form',
   // Optional, the selector to bind ui interaction to for reporting
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -86,8 +86,10 @@ export default class DirectAnswerComponent extends Component {
    * we want to wire up the behavior for interacting with the quality feedback reporting (thumbsup/down)
    */
   onMount () {
-    // Avoid bindings if the feedback has previously been submitted
-    if (this.getState('feedbackSubmitted') === true) {
+    const isUsingCustomCard = this._config.defaultCard;
+    const feedbackSubmitted = this.getState('feedbackSubmitted') === true;
+    // Avoid bindings if the feedback has previously been submitted or is using a custom card.
+    if (isUsingCustomCard || feedbackSubmitted) {
       return this;
     }
 
@@ -180,5 +182,12 @@ export default class DirectAnswerComponent extends Component {
       });
 
     this.analyticsReporter.report(event);
+  }
+
+  addChild (data, type, opts) {
+    if (type === this._config.defaultCard) {
+      return super.addChild(this.getState(), type, opts);
+    }
+    return super.addChild(data, type, opts);
   }
 }
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -86,14 +86,15 @@ export default class DirectAnswerComponent extends Component {
    * we want to wire up the behavior for interacting with the quality feedback reporting (thumbsup/down)
    */
   onMount () {
-    // Avoid bindings if the feedback has previously been submitted
-    if (this.getState('feedbackSubmitted') === true) {
+    const isUsingCustomCard = this._config.defaultCard;
+    const feedbackSubmitted = this.getState('feedbackSubmitted') === true;
+    // Avoid bindings if the feedback has previously been submitted or is using a custom card.
+    if (isUsingCustomCard || feedbackSubmitted) {
       return this;
     }
 
     // For WCAG compliance, the feedback should be a submittable form
-    const formEl = DOM.query(this._container, this._formEl);
-    formEl && DOM.on(this._formEl, 'submit', (e) => {
+    DOM.on(this._formEl, 'submit', (e) => {
       const formEl = e.target;
       const checkedValue = DOM.query(formEl, 'input:checked').value === 'true';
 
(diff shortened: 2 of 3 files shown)


src/ui/templates/results/directanswer.hbs
```diff
@@ -1,3 +1,6 @@
+{{#if _config.defaultCard}}
```
creotutar, line 1: nit: do we want to be consistent with the hbs tabbing and tab everything inside the {{if}}{{else}} blocks?

src/ui/templates/results/directanswer.hbs
```diff
@@ -1,3 +1,6 @@
...
+<div class="yxt-DirectAnswer" data-component="{{_config.defaultCard}}"></div>
```
creotutar: do we want to add a modifier class here to let people conditionally style this class depending on which type of DirectAnswer is used? like `yxt-DirectAnswer yxt-DirectAnswer--something`?

README.md
```diff
@@ -432,6 +432,8 @@ The Direct Answer Component will render the BEST result, if found, based on the
 ANSWERS.addComponent('DirectAnswer', {
   // Required, the selector for the container element where the component will be injected
   container: '.direct-answer-container',
+  // Optional, a custom card component to use.
+  defaultCard: 'MyCustomCard',
```
creotutar, line 436: this being called defaultCard is kind of confusing to me. why not card or customCard?

oshi97: I'll post in the spec I was wondering as well

src/ui/templates/results/directanswer.hbs
oshi97: I've been starting to prefer not indenting after a handlebars helper (I think this is what we've been doing in the theme? also reduces the number of lines changed when you move handlebars logic around ) but you're right, should be consistent here

src/ui/templates/results/directanswer.hbs
oshi97: sure

oshi97: I think the sdk will automatically add a 'component' css class in this case, but what you said feels more contractual

src/ui/components/results/directanswercomponent.js
```diff
@@ -181,4 +183,11 @@ export default class DirectAnswerComponent extends Component {
 
     this.analyticsReporter.report(event);
   }
+
+  addChild (data, type, opts) {
+    if (type === this._config.defaultCard) {
+      return super.addChild(this.getState(), type, opts);
```
alexisgrow: This is putting this component's (the parent's) config in the child component's `config.data._config`. I don't think we want that, since the child's `container` is different, and other config could be as well. A child could get the parent's config through `config._parentOpts`, and/or we could make the interface here super explicit, what do you think?

README.md
tmeyer2115: Yeah, defaultCard is definitely not the best name. Maybe answerCardType or something like that? I'm hesitant to call it cardType since we don't want people thinking they can plug results cards into here.

src/ui/components/results/directanswercomponent.js
```diff
@@ -86,8 +86,10 @@ export default class DirectAnswerComponent extends Component {
    * we want to wire up the behavior for interacting with the quality feedback reporting (thumbsup/down)
    */
   onMount () {
-    // Avoid bindings if the feedback has previously been submitted
-    if (this.getState('feedbackSubmitted') === true) {
+    const isUsingCustomCard = this._config.defaultCard;
+    const feedbackSubmitted = this.getState('feedbackSubmitted') === true;
+    // Avoid bindings if the feedback has previously been submitted or is using a custom card.
```
tmeyer2115, line 97: Have you and @alexisgrow talked about where these bindings should live for a custom card (here vs the Theme)? We are going to want most, if not all, of these bindings for a custom direct answers card as well. These bindings are handle RTF click analytics, form submission, and submission of feedback. I think any default direct answers card in the Theme will need these bindings. We could have this component handle the bindings. We'd just check for the existence of the related DOM element before doing the binding.

oshi97: hmm, I thought it might be clearer to do this in the theme, but I guess we pass down all of the selectors through the config

src/ui/components/results/directanswercomponent.js
oshi97: Sure we can be a bit cleaner about what gets passed through

README.md
oshi97: pinged rose still waiting, will ping again in an hour if no response

src/ui/components/results/directanswercomponent.js
tmeyer2115: We can discuss as a group if you'd like! I just think it's important that all parties are on the same page about where this will live.

README.md
oshi97: https://yext.slack.com/archives/GREFVUD7V/p1591712960014200

tldr: we want to call it defaultCard because in the future there are plans to add cardOverrides

src/ui/components/results/directanswercomponent.js
oshi97: okay, I think we agreed on letting the base theme component do the bindings!

src/ui/components/results/directanswercomponent.js
tmeyer2115: @alexisgrow and @oshi97, do we want to include some of the _parentOps in opts? For instance, would we want to include the various selector config options from the parent as top-level config for the child?

alexisgrow: I think it makes sense to include all of these from the README, what do you think @oshi97?

  // Optional, the selector for the form used for submitting the feedback
  formEl: '.js-directAnswer-feedback-form',
  // Optional, the selector to bind ui interaction to for reporting
  thumbsUpSelector: '.js-directAnswer-thumbUp',
  // Optional, the selector to bind ui interaction to for reporting
  thumbsDownSelector: '.js-directAnswer-thumbDown',
  // Optional, the display text for the View Details click to action link
  viewDetailsText: 'View Details',
  // Optional, the screen reader text for positive feedback on the answer
   positiveFeedbackSrText: 'This answered my question',
  // Optional, the screen reader text for negative feedback on the answer
   negativeFeedbackSrText: 'This did not answer my question',
  // Optional, the footer text to display on submission of feedback
   footerTextOnSubmission: 'Thank you for your feedback!'

alexisgrow (approved):
this LGTM!

PR #865, opened 2020-06-09
DirectAnswer: add cardOverrides

oshi97:
This commit adds cardOverrides to direct answers, which let you specify an
array of rules for specifying a specific custom component based on the
direct answer received. Currently supports matching by fieldName, fieldType,
and entityType.

defaultConfigOption() in configutils.js was renamed to getOrDefault() so that it could be used within this context

TEST=manual
tested with cardOverrides: [
{
  cardType: 'CustomDA1',
  fieldName: 'Phone Number'
},
{
  cardType: 'bogus component',
  fieldName: 'Phone Number'
}
]

checked that would use custom card for a phone number direct answer, but not for
others. checked that the first match would be used. checked that page did not
error when cardOverrides were commented out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,8 +432,8 @@ The Direct Answer Component will render the BEST result, if found, based on the 
 ANSWERS.addComponent('DirectAnswer', {
   // Required, the selector for the container element where the component will be injected
   container: '.direct-answer-container',
-  // Optional, a custom card component to use.
-  defaultCard: 'MyCustomCard',
+  // Optional, a custom direct answer card to use, which is the default when there are no matching card overrides.
+  defaultCard: 'MyDefaultDirectAnswer',
   // Optional, the selector for the form used for submitting the feedback
   formEl: '.js-directAnswer-feedback-form',
   // Optional, the selector to bind ui interaction to for reporting
@@ -447,7 +447,24 @@ ANSWERS.addComponent('DirectAnswer', {
   // Optional, the screen reader text for negative feedback on the answer
    negativeFeedbackSrText: 'This did not answer my question',
   // Optional, the footer text to display on submission of feedback
-   footerTextOnSubmission: 'Thank you for your feedback!'
+  footerTextOnSubmission: 'Thank you for your feedback!',
+  // Optional, card overrides that allow you to specify a specific direct answers card depending on the direct answer. The first matching card will be used, otherwise defaultCard will be used.
+  cardOverrides: [
+    {
+      cardType: 'MenuItemDescriptionDirectAnswer',
+      fieldName: 'description',
+      entityType: 'ce_menuItem',
+      fieldType: 'rich_text'
+    },
+    {
+      cardType: 'DeliveryHoursDirectAnswer',
+      fieldName: 'c_deliveryHours'
+    },
+    {
+      cardType: 'PhoneDirectAnswer',
+      fieldType: 'phone'
+    }
+  ]
 })
 ```
 
--- a/README.md
+++ b/README.md
@@ -448,7 +448,7 @@ ANSWERS.addComponent('DirectAnswer', {
    negativeFeedbackSrText: 'This did not answer my question',
   // Optional, the footer text to display on submission of feedback
   footerTextOnSubmission: 'Thank you for your feedback!',
-  // Optional, card overrides that allow you to specify a specific direct answers card depending on the direct answer. The first matching card will be used, otherwise defaultCard will be used.
+  // Optional, card overrides that allow you to specify a specific direct answers card depending on the fieldName, entityType, and fieldType of the direct answer. The first matching card will be used, otherwise defaultCard will be used.
   cardOverrides: [
     {
       cardType: 'MenuItemDescriptionDirectAnswer',
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -4,7 +4,6 @@ import Component from '../component';
 import AnalyticsEvent from '../../../core/analytics/analyticsevent';
 import StorageKeys from '../../../core/storage/storagekeys';
 import DOM from '../../dom/dom';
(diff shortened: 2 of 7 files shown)


src/ui/components/results/directanswercomponent.js
```diff
@@ -87,15 +99,32 @@ export default class DirectAnswerComponent extends Component {
...
+   * Returns the custom card that should be used for the given direct answer.
+   * First, checks user given cardOverrides for a matching override, if there are none
+   * then returns the default card.
+   * @returns {string}
+   */
+  _getCustomCard (directAnswer) {
+    const answer = directAnswer.answer || {};
+    const matchingOverrides = this._cardOverrides.filter(override => {
+      const { fieldType, fieldName, entityType } = override;
+      return directAnswer.fieldType === fieldType &&
```
alexisgrow: Should these be ||s rather than &&s? The spec was saying that the first rule in the array that matches the direct answer would be applied, even if a more specific rule exists

oshi97: shoot yeah I goofed here. fieldType is not yet implemented, and I assumed entityType was in the same directAnswer.answer object as fieldName. It's not, so when I was testing only fieldName would be checked. Fixed now!

src/ui/components/results/directanswercomponent.js
```diff
@@ -87,15 +100,50 @@ export default class DirectAnswerComponent extends Component {
...
 
+  /**
+   * Check whether a given cardOverride matches a given directAnswer.
+   * @param {Object} directAnswer
+   * @param {Object} override
+   */
+  _overrideMatchesAnswer (directAnswer, override) {
+    const directAnswerPropeties = {
+      entityType: getOrDefault(directAnswer, ['relatedItem.data.type']),
+      fieldName: getOrDefault(directAnswer, ['answer.fieldName']),
```
tmeyer2115: Why are we using getOrDefault here? Why not just access these properties directly?

src/ui/components/results/directanswercomponent.js
```diff
@@ -87,15 +100,50 @@ export default class DirectAnswerComponent extends Component {
...
+  }
+
+  /**
+   * Returns the custom card that should be used for the given direct answer.
+   * First, checks user given cardOverrides for a matching override, if there are none
+   * then returns the default card.
+   * @returns {string}
+   */
+  _getCustomCard (directAnswer) {
+    const matchingOverrides = this._cardOverrides.filter(override => {
```
tmeyer2115: We could just use array.find here to find the first override that works.

src/ui/components/results/directanswercomponent.js
```diff
@@ -87,15 +100,50 @@ export default class DirectAnswerComponent extends Component {
...
+   * @param {Object} directAnswer
+   * @param {Object} override
+   */
+  _overrideMatchesAnswer (directAnswer, override) {
+    const directAnswerPropeties = {
+      entityType: getOrDefault(directAnswer, ['relatedItem.data.type']),
+      fieldName: getOrDefault(directAnswer, ['answer.fieldName']),
+      fieldType: directAnswer.fieldType
+    };
+    for (let [property, overrideValue] of Object.entries(override)) {
```
tmeyer2115: propertyToMatch and propertyValue?

src/ui/components/results/directanswercomponent.js
oshi97: It felt like cleaner code to me to use this helper method instead of having something like
```
const relatedItem = directAnswer.relatedItem || {};
const data = relatedItem.data || {};
```
alternatively maybe we could solidify the direct answer model?

oshi97: When there is no directAnswer direct answer will just be {}, it's probably better just to check for this case

src/ui/components/results/directanswercomponent.js
oshi97: updated!

src/ui/components/results/directanswercomponent.js
oshi97: updated!

src/ui/components/results/directanswercomponent.js
```diff
@@ -87,15 +99,52 @@ export default class DirectAnswerComponent extends Component {
...
+  /**
+   * Returns the custom card that should be used for the given direct answer.
+   * First, checks user given cardOverrides for a matching override, if there are none
+   * then returns the default card.
+   * @returns {string}
+   */
+  _getCustomCard (directAnswer) {
+    return this._cardOverrides.find(override => {
+      return this._overrideMatchesAnswer(directAnswer, override);
+    }).cardType || this._defaultCard;
```
tmeyer2115: Minor style nit: this is kinda difficult to read. This is easier to parse:

const cardOverride = ...;
return cardOverride ? cardOverride.cardType : this._defaultCard.

I think what's there now could also blow up if no matching override is found. We'd be trying to access cardType of undefined.

oshi97: fixed! sorry I should have caught this

README.md
```diff
@@ -447,7 +447,24 @@ ANSWERS.addComponent('DirectAnswer', {
   // Optional, the screen reader text for negative feedback on the answer
    negativeFeedbackSrText: 'This did not answer my question',
   // Optional, the footer text to display on submission of feedback
-   footerTextOnSubmission: 'Thank you for your feedback!'
+  footerTextOnSubmission: 'Thank you for your feedback!',
+  // Optional, card overrides that allow you to specify a specific direct answers card depending on the direct answer. The first matching card will be used, otherwise defaultCard will be used.
```
tmeyer2115: Add a little more detail here. We want them to know what features of the direct answer they can use when creating an override.

tmeyer2115 (approved):

oshi97:
The plan is to merge this into the develop branch when our git flow prep is ready. Bryan said this should go into v1.5.0

PR #867, opened 2020-06-10
Pagination: useFacets when running search

oshi97:
Pagination was not applying facets when running a new search.

TEST=manual
set search limit to 3, choose facets with more than 3 results
check that going to the second page still uses the applied facets


alexisgrow (approved):
LG!

PR #868, opened 2020-06-10
ResultsHeader: default background color to white on universal

oshi97:
"Current: On universal, the filters no longer has a background color
Expected: On universal, applied filters should have background-color: var(--yxt-results-filters-background) as they did before "

TEST=manual
check that I can set the results-header-background variable to --yxt-color-brand-primary,
to make the background blue, then change it to --yxt-color-brand-white to make it white
set background color of a vertical and universal test page to alicia blue, and checked the results header was only white for the vertical page

### DIFF
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -115,7 +115,7 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     margin-top: 0;
     margin-bottom: 0;
     padding: calc(var(--yxt-base-spacing) / 2) var(--yxt-base-spacing);
-    background-color: $results-filters-background;
+    background-color: var(--yxt-results-filters-background);
     border-right: var(--yxt-results-border);
     border-left: var(--yxt-results-border);
     border-bottom: var(--yxt-results-border);
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -1,7 +1,7 @@
 /** @define Results */
 
 $results-title-bar-background: var(--yxt-color-background-highlight) !default;
-$results-filters-background: white !default;
+$results-filters-background: var(--yxt-color-brand-white) !default;
 
 $results-title-bar-text-color: var(--yxt-color-text-primary) !default;
 $results-title-bar-text-font-size: var(--yxt-font-size-md-lg) !default;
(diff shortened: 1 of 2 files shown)


src/ui/sass/modules/_ResultsHeader.scss
```diff
@@ -15,6 +15,8 @@ $results-header-filters-font-size: var(--yxt-font-size-md) !default;
 $results-header-filters-color: var(--yxt-color-text-secondary) !default;
 $results-header-filters-line-height: var(--yxt-line-height-md) !default;
 
+$results-header-universal-background: var(--yxt-color-brand-white) !default;
```
creotutar, line 18: net new variables technically do not need a sass variable counterpart. though it is nice for any implementations that will use the sass directly to have.

src/ui/sass/modules/_Results.scss
```diff
@@ -115,7 +115,7 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     margin-top: 0;
     margin-bottom: 0;
     padding: calc(var(--yxt-base-spacing) / 2) var(--yxt-base-spacing);
-    background-color: $results-filters-background;
+    background-color: var(--yxt-results-filters-background);
```
creotutar, line 118: this is `white` whereas the filters background color is yxt-color-brand-white which will not necessarily always be `white`. did we want them to match by default?

src/ui/sass/modules/_ResultsHeader.scss
oshi97: I am lowkey hoping we add something someday that transforms our sass variables automatically into css variables

src/ui/sass/modules/_Results.scss
oshi97: sure we can make them match

creotutar: i think we want them to match via some variable. that way if one changes the other changes. can we make them both source from `var(--yxt-color-brand-white)` or something similar?

oshi97: I was worried about that being more "breaking" but I guess there's so many more breaking changes and using the variable is definitely better

oshi97: ok we keeping it white since it's semi deprecated and don't wanna break stuff

creotutar (approved):
🥇 🥇 🥇 LGTM

Good call on the breaking change. Given that this ResultsHeader is not going to be used going forward, I think it makes sense to not change the previous variable.

PR #869, opened 2020-06-10
Pagination: fix show first/last page button defaults

oshi97:
Expected: would default to showFirstAndLastButton
Actual: would default to showFirstAndLastButton if it was user specified, otherwise would be false

The issue was that if showFirstAndLastButton defaulted to true, the component would not use the defaulted value

Also add nik's lost in rebase readme update

TEST=manual
test that by default the last/first buttons are displayed

### DIFF
--- a/README.md
+++ b/README.md
@@ -888,10 +888,10 @@ ANSWERS.addComponent('Pagination', {
   },
   // Function invoked when a user clicks to change pages. By default, scrolls the user to the top of the page.
   onPaginate: (newPageNumber, oldPageNumber, totalPages) => {},
-  // DEPRECATED with showFirstAndLastButton.
+  // DEPRECATED, please use showFirstAndLastButton instead.
   // Display a double arrow allowing users to jump to the first page of results. Defaults to showFirstAndLastButton.
   showFirst: true,
-  // DEPRECATED with showFirstAndLastButton.
+  // DEPRECATED, please use showFirstAndLastButton instead.
   // Display a double arrow allowing users to jump to the last page of results. Defaults to showFirstAndLastButton.
   showLast: true,
 });
--- a/README.md
+++ b/README.md
@@ -870,15 +870,15 @@ ANSWERS.addComponent('Pagination', {
   container: '.pagination-component',
   // Required*, the vertical for pagination, *if omitted, will fall back to the search base config
   verticalKey: 'verticalKey',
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -870,10 +870,14 @@ ANSWERS.addComponent('Pagination', {
   container: '.pagination-component',
   // Required*, the vertical for pagination, *if omitted, will fall back to the search base config
   verticalKey: 'verticalKey',
-  // Optional, display a double arrow allowing users to jump to the first page of results
```
creotutar, line 873: When we deprecate an option but it is still used in our code, do we want to mark it as deprecated instead of removing it entirely until it is phased out? I can see this being confusing when an HH is using it, upgrades, and finds that it is no longer in the README but it is still working. Unsure if this is something we want to support

oshi97: asking Tom about his thoughts!

oshi97: Tom said that what we've been doing is just removing any deprecated options from the readme. After bringing up your points I think the verdict is we should keep these around and mark them as deprecated
> Hmm, the Theme could reference older, deprecated SDK stuff. We'd want to keep the Theme up-to-date as much as possible, but that's no guarantee. Maybe we should still refer to it in the README (but note that it's deprecated and they should use [the new option])

oshi97: updated!

README.md
```diff
@@ -884,6 +888,12 @@ ANSWERS.addComponent('Pagination', {
   },
   // Function invoked when a user clicks to change pages. By default, scrolls the user to the top of the page.
   onPaginate: (newPageNumber, oldPageNumber, totalPages) => {},
+  // DEPRECATED with showFirstAndLastButton.
+  // Display a double arrow allowing users to jump to the first page of results. Defaults to showFirstAndLastButton.
+  showFirst: true,
+  // DEPRECATED with showFirstAndLastButton.
```
creotutar: I read this as "both showFirstAndLastButton and showLast are deprecated". Could we try something like `DEPRECATED, use x instead`

oshi97: yeah that's a lot better

creotutar (approved):
🎸 🎸 🎸 LGTM

creotutar:
can we re-run the unit test?

oshi97:
> can we re-run the unit test?

works locally!

PR #871, opened 2020-06-10
UniversalResults: add display: flex to the icon component

oshi97:
This was done in the old resultssectionheader.hbs, and fixes
vertical alignment issues. Adding the original class="yxt-Results-titleIconWrapper"
adds this.

TEST=manual
test that if I set &-titleIconWrapper to an arbitrary height the
icon will be vertically aligned with display: flex and will not be
aligned without it


src/ui/sass/modules/_Results.scss
```diff
@@ -161,11 +161,6 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     background-color:  var(--yxt-color-background-highlight);
   }
 
-  &-title .Icon
-  {
-    margin-right: calc(var(--yxt-base-spacing) / 2);
```
creotutar, line 166: is this margin change intentional? if so, can we update the commit message and/or desc? doesn't seem related to the vertical alignment (unless I'm missing something)

oshi97: this margin is already added in yxt-Results-titleIconWrapper so is redundant

creotutar (approved):
🌻 🌻 🌻 LGTM

PR #872, opened 2020-06-11
bust cache for xhr requests in css vars ponyfill for ie11

creotutar:
This change is necessary for the following situation:

Link tags make requests to servers without CORS headers. Because v1.3.1
(and all patch version CDN links) have a cache-control with a non-zero
max-age, this hints to a browser (especially IE11) that the requests
should be cached. Thus IE11, in the XMLHttpRequests made by the
css-vars-ponyfill, tries to access the request, is told it should used
the cache request, and finds that the cached request does not have CORS
headers and fails the XHRs.

We cannot make the CORS headers show up for the link tag without
removing the cache-control max-age or adding crossorigin attribute
(which, lucky for us, is not available in IE11). So, we must try to bust
the cache in IE11 for the XHRs made by css vars ponyfill.

This is done by adding a random string (current time) as the query
parameter to the links before they're sent. This _ query parameter is
used by jQuery to do similar cache busting.

Note: If the asset is not a valid URL (for the [URL constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)), then we do not add the cache-busting URL param. This affects two URLs of note: relative URLs (e.g. `href="bundle.css"`) and relative protocol URLs (e.g. `href="//www.yext.com/some.css"`). We also do not want to block JS if we cannot process the URL.

T=330545
TEST=manual

Tested that all link tags in the affected Virtusa site (and a local
Jambo site) are fetched a second time with the cache busting query
param.

Tested a script with and without a query param already in the url.

With the link tag before the ANSWERS.init call, tested in IE11 (the
affected browser) and Chrome. Verified that the styling with css
variables are showing on IE11 on page load.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -413,6 +413,10 @@ class Answers {
    * Updates the css styles with new current variables. This is useful when the css
    * variables are updated dynamically (e.g. through js) or if the css variables are
    * added after the ANSWERS.init
+   *
+   * To solve issues with non-zero max-age cache controls for link/script assets in IE11,
+   * we add a cache busting parameter so that XMLHttpRequests succeed.
+   *
    * @param {Object} config Additional config to pass to the ponyfill
    */
   ponyfillCssVariables (config = {}) {
@@ -422,11 +426,11 @@ class Answers {
       onSuccess: config.onSuccess || function () {},
       onFinally: config.onFinally || function () {},
       onBeforeSend: (xhr, node, url) => {
-        const uri = new URL(url);
-        const params = new SearchParams(uri.search);
+        const uriWithCacheBust = new URL(url);
+        const params = new SearchParams(uriWithCacheBust.search);
         params.set('_', new Date().getTime());
-        uri.search = params.toString();
-        xhr.open('GET', uri.toString());
+        uriWithCacheBust.search = params.toString();
+        xhr.open('GET', uriWithCacheBust.toString());
       }
     });
   }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -3,12 +3,12 @@
 import Core from './core/core';
 import RtfConverter from '@yext/rtf-converter';
 import cssVars from 'css-vars-ponyfill';
-import URI from 'urijs';
 
 import {
   DefaultTemplatesLoader,
   Renderers,
-  DOM
+  DOM,
+  SearchParams
 } from './ui/index';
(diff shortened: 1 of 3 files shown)


src/answers-umd.js
```diff
@@ -419,7 +420,11 @@ class Answers {
       onlyLegacy: true,
       onError: config.onError || function () {},
       onSuccess: config.onSuccess || function () {},
-      onFinally: config.onFinally || function () {}
+      onFinally: config.onFinally || function () {},
+      onBeforeSend: (xhr, node, url) => {
+        const uri = new URI(url).addQuery('_', new Date().getTime());
```
oshi97: could we name this variable something that hints at what it does, and maybe add a blurb to the jsdoc? I think it's hard to understand what this is for without reading your commit message

creotutar:
💀 This will need to be addressed.
```
  dist/answers.min.js
 
  Package size limit has exceeded by 4.55 KB
 
  Package size: 154.55 KB
 
  Size limit:   150 KB
 

 
  dist/answers-modern.min.js
 
  Package size limit has exceeded by 1.43 KB
 
  Package size: 101.43 KB
 
  Size limit:   100 KB
```

src/answers-umd.js
```diff
@@ -412,14 +413,29 @@ class Answers {
...
+      onFinally: config.onFinally || function () {},
+      onBeforeSend: (xhr, node, url) => {
+        try {
+          const uriWithCacheBust = new URL(url);
+          const params = new SearchParams(uriWithCacheBust.search);
+          params.set('_', new Date().getTime());
+          uriWithCacheBust.search = params.toString();
+          xhr.open('GET', uriWithCacheBust.toString());
+        } catch (e) {
+          // Catch the error and continue if the URL provided in the asset is not a valid URL
```
oshi97, line 436: what kinds of errors might we get here, and is it okay to just ignore them? also love that we don't import any new dependencies

creotutar: Ideally, this cache busting will not be blocking JS on the page if a query param cannot be added, thus the continue.

This affects two URLs of note: relative URLs (e.g. href="bundle.css") and relative protocol URLs (e.g. href="//www.yext.com/some.css").

If we try to use those kinds of URLs as parameters for the URL constructor, we get a JS error, not fitting the schema.

oshi97 (approved):
lgtm!

PR #873, opened 2020-06-11
VerticalResults: add back removed styling from #569

oshi97:
This commit reverts unnecessary styling changes in #569. Namely the removal of css classes. Also makes changes to ResultsHeader.scss to be more backwards compatible with preexisting yxt-Results styling

TEST=manual
no visible changes on my local testing site

### DIFF
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -72,6 +72,72 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     font-size: var(--yxt-results-title-bar-icon-size);
   }
 
+  &-viewAllLink
+  {
+    display: flex;
+    align-items: center;
+
+    @include Text(
+      var(--yxt-results-title-bar-link-font-size),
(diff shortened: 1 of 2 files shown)


creotutar:
I believe this suggestion was to fix the applied query filters business on the Yext Answers page (pictured below). The center-align does help for sure, but I think we would want to address the extreme height with the filter.

image

src/ui/sass/modules/_Results.scss
```diff
@@ -72,6 +72,72 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     font-size: var(--yxt-results-title-bar-icon-size);
   }
 
+  &-viewAllLink
```
creotutar, line 75: this doesn't exist yet right? another PR?

oshi97: it's being used in resultssectionheader.hbs, I think I deleted resultssectionheader.hbs in #569 but then we brought it back at some point

creotutar (approved):
🍭 🍭 🍭 LGTM

PR #875, opened 2020-06-11
VerticalResults: only update when search complete

oshi97:
> Current: Removing an applied filter hides the entire component and causes a flash
Expected: No flash, the result count edit: and vertical results should remain
I think this is similar to the problem we have with pagination.
can we make it such that we wait to update the vertical results until the new set returns (instead of hiding it)?

This commit changes VerticalResultsComponent to listen to changes to StorageKeys.VERTICAL_RESULTS the same way pagination does, only updating when the search is complete. Previously, VerticalResultsComponent listened to changes to
StorageKeys.VERTICAL_RESULTS by settings it's moduleId, which downstream is
hooked up to call setState when that storage key is updated.

TEST=manual
check that search results still update when removing a filter, applying a filter, and
making a new search

### DIFF
--- a/src/ui/components/results/verticalresultscomponent.js
+++ b/src/ui/components/results/verticalresultscomponent.js
@@ -171,7 +171,12 @@ export default class VerticalResultsComponent extends Component {
      */
     this._noResultsTemplate = this._noResultsConfig.template;
 
-    this.moduleId = StorageKeys.VERTICAL_RESULTS;
+    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
+      if (results.searchState === SearchStates.SEARCH_COMPLETE) {
+        this.setState(results);
+      }
+    });
+
     /**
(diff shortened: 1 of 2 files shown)


creotutar (approved):
🎃 🎃 🎃 LGTM

I think we want to take a look in the future about the moduleId in componentmanager. It feels weird to me that update handlers are added to components /sometimes/ and only if moduleId is non-undefined/null. It was not clear to me that removing this moduleId was removing that update (or if it might be doing other things). Just something to think about for the future

oshi97:
> 🎃 🎃 🎃 LGTM
> 
> I think we want to take a look in the future about the moduleId in componentmanager. It feels weird to me that update handlers are added to components /sometimes/ and only if moduleId is non-undefined/null. It was not clear to me that removing this moduleId was removing that update (or if it might be doing other things). Just something to think about for the future

yes for sure moduleId is pretty confusing and also doesn't really do very much

PR #876, opened 2020-06-11
 use xlarge unit test circleci resource class

creotutar:
We are bumping the resource class to the next tier so that v1.4.0 PRs
will run unit test tests correctly without out of memory errors.

J/T=None
TEST=manual

circleci config validate
Verify the PR unit tests run without out of memory errors

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
   unit_test:
     docker:
       - image: circleci/node:7.10
-    resource_class: large
+    resource_class: xlarge
     working_directory: ~/answers
     steps:
       - setup-workspace


oshi97 (approved):
lgtm! I think other things we can do in the future are
1. run jest tests in parallel (see https://dev.to/drazisil/test-splitting-with-jest-on-circleci-1i0c)
2. upgrade our dependencies (we're using some relatively outdated stuff)
3. refactor jest tests

creotutar:
I've been told the tests about that are failing are checks that happen when the base branch is master. That is not the case here, the branch was initially master and then I switched it to v1.4.0.

PR #877, opened 2020-06-11
VerticalResults: use <h2> for yxt-Results-title

oshi97:
To match resultssectionheader.hbs

TEST=manual
check that answers.yext.com's yxt-Results-title has margin: 0 from the css reset.
We can expect all consulting sites to apply this css reset, so adding margin: 0 is not a breaking change.
check that visually looks the same out of the box whether it is h2 or div

### DIFF
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -213,6 +213,7 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
 
   &-title
   {
+    margin: 0;
     text-transform: uppercase;
     @include Text(
       $color: var(--yxt-color-text-primary),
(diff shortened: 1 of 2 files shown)


src/ui/sass/modules/_Results.scss
```diff
@@ -213,6 +213,7 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
 
   &-title
   {
+    margin: 0;
```
creotutar, line 216: was this necessary to maintain the styling after changing to h2 from div?

oshi97: yup, otherwise user agent styling gives it a huge margin

oshi97: (which doesn't match the mocks)

oshi97: image

creotutar (approved):
💯 💯 💯 LGTM

PR #878, opened 2020-06-11
VerticalResults: rename yxt-Results-viewMore -> yxt-Results-viewAll

oshi97:
In order to be more backwards compatible, we should use the older
viewAllLink css class. Styling for AccordionResults was not changed
with this commit, even though the yxt-Results-viewAllLink class is
now defined in two different places, because the preexisting AccordionResults
specific styling (luckily) completely overrides the new styling.

TEST=manual
checked in chrome inspector that the viewAllLink in AccordionResults
does not carry over any of the new styling. checked that for both on
hover and not, that all css properties from yxt-Results-viewAllLink
are overriden by the more specific .yxt-Accordion .yxt-Results-viewAllLink.

### DIFF
--- a/src/ui/templates/results/verticalresults.hbs
+++ b/src/ui/templates/results/verticalresults.hbs
@@ -8,7 +8,7 @@
       {{> resultsHeader }}
       {{> map}}
       {{> results}}
-      {{> viewMore}}
+      {{> viewAll}}
     </section>
   {{/if}}
 {{/if}}
@@ -69,11 +69,11 @@
   {{/if}}
 {{/inline}}
 
-{{#*inline "viewMore"}}
+{{#*inline "viewAll"}}
   {{#if (and _config.isUniversal _config.viewMore)}}
-    <div class="yxt-Results-viewMore">
-      <a class="yxt-Results-viewMoreLink" href="{{ verticalURL }}">
-        <div class="yxt-Results-viewMoreLabel">{{_config.viewMoreLabel}}</div>
+    <div class="yxt-Results-viewAll">
+      <a class="yxt-Results-viewAllLink" href="{{ verticalURL }}">
+        <div class="yxt-Results-viewAllLabel">{{_config.viewMoreLabel}}</div>
         <div data-component="IconComponent" data-opts='{ "iconName": "chevron" }'></div>
       </a>
     </div>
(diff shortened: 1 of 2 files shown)


src/ui/templates/results/verticalresults.hbs
```diff
@@ -69,11 +69,11 @@
...
 
-{{#*inline "viewMore"}}
+{{#*inline "viewAll"}}
   {{#if _config.isUniversal}}
-    <div class="yxt-Results-viewMore">
-      <a class="yxt-Results-viewMoreLink" href="{{ verticalURL }}">
-        <div class="yxt-Results-viewMoreLabel">{{_config.viewMoreLabel}}</div>
+    <div class="yxt-Results-viewAll">
+      <a class="yxt-Results-viewAllLink" href="{{ verticalURL }}">
+        <div class="yxt-Results-viewAllLabel">{{_config.viewMoreLabel}}</div>
```
oshi97, line 76: We still want to use {{_config.viewMoreLabel}} instead of the older {{_config.viewAllText}}, since we want to use the updated config names where possible. Setting only config.viewAllText will still work here, because in the universal results component we translate deprecated config into their updated names.

src/ui/templates/results/verticalresults.hbs
```diff
@@ -69,11 +69,11 @@
   {{/if}}
 {{/inline}}
 
-{{#*inline "viewMore"}}
+{{#*inline "viewAll"}}
   {{#if _config.isUniversal}}
-    <div class="yxt-Results-viewMore">
-      <a class="yxt-Results-viewMoreLink" href="{{ verticalURL }}">
-        <div class="yxt-Results-viewMoreLabel">{{_config.viewMoreLabel}}</div>
+    <div class="yxt-Results-viewAll">
```
creotutar, line 74: This is interesting.

Anyone hoping to hide the View All link in older versions would have to change their target to viewAll if they wanted this styling to go away. Does this solve the initial problem?

creotutar: The initial problem being, Yext Answers was hiding viewAllLink to make sure it does not show on its promo card.

creotutar: eg. It will still look like this
image

creotutar: What I said was false. Yext Answers is not hiding viewAllLink, it was hiding the titleBar. Regardless of this change, if they wanted to hide the viewAllLink, they would need to add net new styling.

oshi97: hopefully partially addressed in #879

creotutar (approved):
👁️ 👁️ 👁️ LGTM

PR #879, opened 2020-06-11
UniversalResults: make viewMore config option work

oshi97:
This option is supposed to prevent the viewMore div from
rendering if false. It was being ignored.

TEST=manual
check that I can set viewMore to false (and that the default is true))
and that the viewMore box will not render in this case. check that
setting it to true will render the box


creotutar (approved):
🍐 🍐 🍐 LGTM

As discussed offline, universalcomponents.js default has an easy to miss `...config` that makes the config option work with the default. We should take a look at this in the future.

PR #880, opened 2020-06-11
VerticalResults: use ::after for viewAllLink icon

oshi97:
The AEB currently adds it's own icon to yxt-Results-viewAllLink::after. Since the 
oobx mock for universal results also wants an icon there, we end up with a
very annoying double icon.

This commit changes the viewAllLink icon to use ::after styling instead of
an icon component, setting the desired chevron icon as a base64 encoded
background-image.

Generally, it is preferable to use an IconComponent when an icon is desired.
This mixin is for cases when we want to add an icon in a place where
SDK implementors commonly would add an icon with ::after. Using an IconComponent
in those cases would result in a double icon, while adding the desired icon with ::after
will only result in the new SDK icon being overriden.

TEST=manual
check that we still get a chevron icon with the ::after styling

### DIFF
--- a/src/ui/sass/_mixins.scss
+++ b/src/ui/sass/_mixins.scss
@@ -100,3 +100,21 @@
     $hover-decoration: none
   );
 }
+
+// Generally, it is preferable to use an IconComponent when an icon is desired.
+// This mixin is for cases when we want to add an icon in a place where 
+// SDK implementors commonly would add an icon with ::after. Using an IconComponent
+// in those cases would result in a double icon, while adding the desired icon with ::after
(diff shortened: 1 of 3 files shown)


oshi97:
screenshot of the double chevron

image

If we hate this change I'm totally fine with either changing the css class of yxt-Results-viewAllLink for VerticalResults and telling people they're old viewAllLink styling isn't going to be carried over, or asking Rose if we can just not do this viewMore change and keep it at the top right of universal results.

oshi97:
Closing for now, think we prefer to just add this to the update docs since Christian said most sites don't do this

PR #881, opened 2020-06-12
Use MasterSwitchApi to block only the invocation of onReady.

tmeyer2115:
Previously, the MasterSwitchApi call blocked much of the SDK initialization, as well as onReady, until it was determined that the page was enabled. This caused problems, however. Some instance methods do not work properly unless the SDK initialization is complete. Prior to the introduction of the Status checking, initialization was guaranteed to be complete immediately after any call to Answers.init. It was not necessary to wait until onReady had fired. With the Status check blocking most of the initialization, this synchronous behavior was not guaranteed.

The AEB relied on this old behavior. It called setConversionsOptIn right after invoking Answers.init. However, this setter needs eligibleForAnalytics to be initialized. Otherwise, it will not work correctly. This was the cause of the Conversion Tracking Enablement TechOps.

With this change, the MasterSwitchApi blocks only onReady. Now, the call to Answers.init will, again, synchronously initialize the SDK. Also included in this PR is a timeout for the Status check. The default timeout is 100ms. If nothing is returned in this time, the experience is assumed to be enabled.

T=331657
TEST=manual

Tested the following in IE11 and Chrome:

- onReady is fired correctly when the site is enabled.
- onReady is not fired when the site is disabled.
- onReady is fired when the Status request times out.
- I pointed one of the AEB sites from the TechOps to my local copy of Answers. The onReady fired as expected and the conversion tracking was properly enabled.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -166,24 +166,10 @@ class Answers {
     globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
 
-    const masterSwitchApi = statusPage
+    this._masterSwitchApi = statusPage
       ? new MasterSwitchApi({ apiKey: parsedConfig.apiKey, ...statusPage }, globalStorage)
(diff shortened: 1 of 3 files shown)


oshi97 (approved):
lgtm! I think this might also let people do their jank double inits again

PR #882, opened 2020-06-12
v1.2.2

tmeyer2115:
# Version 1.2.2

## Fixes

* SDK initialization is no longer blocked by a Status page check. Instead, it will happen synchronously. This will ensure other instance methods can be invoked immediately after Answers.init.


oshi97 (approved):
lgtm!

PR #883, opened 2020-06-12
VerticalResults: reduce icon color specificity

oshi97:
The default color of the title icon was changed to color-brand-primary,
where before it was set to inherit. This change makes it easier to
override the sdk default color.

TEST=manual
checked that icon color still defaults to color-brand-primary
check that on accordionresults color defaults to text color (default black)


creotutar (approved):
🚀 🚀 🚀 LGTM

PR #884, opened 2020-06-12
UniversalResults: hide viewAllText if set to blank value

oshi97:
If somebody manually sets their viewAllText to blank, it should be hidden

TEST=manual
test that setting it to blank will hide it
test that if unset it will still default to 'view more'
test that can use either viewAllText or viewMoreLabel, and that viewMoreLabel takes priority

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,8 @@ ANSWERS.addComponent('UniversalResults', {
       renderItem: function(data) {},
       // Optional, override the handlebars template for each item in this vertical
       itemTemplate: `my item {{name}}`,
+      // DEPCREATED, please use viewMoreLabel instead. The text for the view more link. Has the same effect as viewMoreLabel, and is out-prioritized by viewMoreLabel. Defaults to 'View More'.
+      viewAllText: 'View All Results For Vertical'
     }
   },
   // Optional, override the render function for each item in the result list
--- a/README.md
+++ b/README.md
@@ -521,7 +521,7 @@ ANSWERS.addComponent('UniversalResults', {
       renderItem: function(data) {},
       // Optional, override the handlebars template for each item in this vertical
       itemTemplate: `my item {{name}}`,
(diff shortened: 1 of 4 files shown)


src/ui/components/results/universalresultscomponent.js
```diff
@@ -67,7 +67,7 @@ export default class UniversalResultsComponent extends Component {
       // Show a view more link by default, which also links to verticalURL.
       viewMore: true,
       // By default, the view more link has a label of 'View More'.
-      viewMoreLabel: config.viewAllText || 'View More',
+      viewMoreLabel: defaultConfigOption(config, ['viewMoreLabel', 'viewAllText'], 'View More'),
```
creotutar, line 70: i dont see viewAllText in the README. is this something we want to add back (as deprecated)?

oshi97: sigh yeah we should add it all the deprecated config is really upsetting though

README.md
```diff
@@ -521,6 +521,8 @@ ANSWERS.addComponent('UniversalResults', {
       renderItem: function(data) {},
       // Optional, override the handlebars template for each item in this vertical
       itemTemplate: `my item {{name}}`,
+      // DEPCREATED, please use viewMoreLabel instead. The text for the view more link. Has the same effect as viewMoreLabel, and is out-prioritized by viewMoreLabel. Defaults to 'View More'.
```
creotutar: typo `DEPCREATED`

could we word somewhere along the lines of the value of `viewMoreLabel` takes precedence over `viewAllText`

oshi97: Does 
>Has the same effect as viewMoreLabel, and is out-prioritized by viewMoreLabel.

work?

oshi97: updated to the wording we agreed on

creotutar (approved):
🚀 🚀 🚀 LGTM

PR #885, opened 2020-06-12
Facets: patch expand field override

oshi97:
First thing was expand always being set to true if the top level was set to
true even if the fieldOverride expand was set to false.

Second thing was that going from FilterBox to FilterOptions the config name
changes from expand to showExpand, so relying on Object.assign
to correctly override config doesn't work in this case.

It's super dumb that showExpand and expand are the same config option
but have different names.

TEST=manual
check that after the changes I can set
            fields: {
              c_employeeDepartment: {
                control: 'singleoption',
                expand: false
              }
            }
in my facets config, and the the Employee Department FilterOptions will
not be expandable

### DIFF
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -199,9 +199,9 @@ export default class FilterBoxComponent extends Component {
         container: `.js-yext-filterbox-filter${i}`,
         showReset: this.config.resetFilter,
         resetLabel: this.config.resetFilterLabel,
-        showExpand: config.showExpand === undefined ? this.config.expand : config.showExpand,
         isDynamic: this.config.isDynamic,
         ...config,
+        showExpand: config.showExpand === undefined ? this.config.expand : config.showExpand,
         onChange: (filterNode, alwaysSaveFilterNodes, blockSearchOnChange) => {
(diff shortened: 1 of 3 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -199,7 +199,7 @@ export default class FilterBoxComponent extends Component {
         container: `.js-yext-filterbox-filter${i}`,
         showReset: this.config.resetFilter,
         resetLabel: this.config.resetFilterLabel,
-        showExpand: this.config.expand,
+        showExpand: config.expand === undefined ? this.config.expand : config.expand,
```
tmeyer2115: Could we not keep it showExpand in the FacetsComponent and just refer to it properly here? 

showExpand: config.showExpand === undefined ? ....

I'd rather the FacetsComponent use the correct terminology when specifying config for FilterOptions.

oshi97: The correct config option name for FilterBox is `expand` though, not `showExpand`

oshi97: whereas the correct config name for FilterOptions is `showExpand`

oshi97: I can't read

tmeyer2115 (approved):

PR #887, opened 2020-06-12
 Add a hardcoded default for js lib version when not in a git repository

creotutar:
This is for the case when we are building the repository but without git
tracking. This is happening in the case of the netlify tests in our PRs
to master. Add this hardcoded default when we are not in a git working
tree.

J=None
TEST=manual

Check that after a gulp build, the three following files have (1) the
git describe --tags output as the js lib version as normal and (2) TEST
as the js lib version when the .git folder is removed.

* dist/answers.js
* dist/answers-umd.js
* dist/answers-modern.js

Verify the PR tests (starting with master to start the netlify tests,
then changing to correct base branch).

### DIFF
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -21,8 +21,8 @@ const uglify = require('gulp-uglify-es').default;
 const NAMESPACE = 'ANSWERS';
 
 function getLibVersion () {
-  try { 
-    let insideWorkTree  = require('child_process')
+  try {
+    let insideWorkTree = require('child_process')
       .execSync('git rev-parse --is-inside-work-tree 2>/dev/null')
       .toString().trim();
     if (insideWorkTree) {
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -21,9 +21,21 @@ const uglify = require('gulp-uglify-es').default;
 const NAMESPACE = 'ANSWERS';
 
 function getLibVersion () {
-  return require('child_process')
-    .execSync('git describe --tags')
-    .toString().trim();
+  try { 
+    let insideWorkTree  = require('child_process')
+      .execSync('git rev-parse --is-inside-work-tree 2>/dev/null')
+      .toString().trim();
+    if (insideWorkTree) {
+      return require('child_process')
+        .execSync('git describe --tags')
+        .toString().trim();
+    }
+  } catch (e) {
+    // if above command fails, catch error and continue, as we are not in a git repository
+  }
+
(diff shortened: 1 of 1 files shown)


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -21,9 +21,21 @@ const uglify = require('gulp-uglify-es').default;
 const NAMESPACE = 'ANSWERS';
 
 function getLibVersion () {
-  return require('child_process')
-    .execSync('git describe --tags')
-    .toString().trim();
+  try {
+    let insideWorkTree = require('child_process')
```
tmeyer2115: Minor nit, make this const

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -21,9 +21,21 @@ const uglify = require('gulp-uglify-es').default;
...
+      .execSync('git rev-parse --is-inside-work-tree 2>/dev/null')
+      .toString().trim();
+    if (insideWorkTree) {
```
tmeyer2115: Is insideWorkTree a string? Do we need to convert it to a boolean since "false" is truthy?

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -21,9 +21,21 @@ const uglify = require('gulp-uglify-es').default;
...
+      return require('child_process')
+        .execSync('git describe --tags')
+        .toString().trim();
+    }
+  } catch (e) {
```
tmeyer2115, line 33: Which of the commands above can cause the exception?

conf/gulp-tasks/library.gulpfile.js
creotutar: My worry was that this would be too restrictive for getting a valid version. I have only personally seen the value `true` and then if we were not in a git repository, we did not get a value back and instead just got a fatal error (explaining the catch). I didn't have any luck finding all the possible values of the function, but I can keep digging around here

https://github.com/git/git/blob/3bab5d56259722843359702bc27111475437ad2a/setup.c#L403

conf/gulp-tasks/library.gulpfile.js
creotutar: Both of the commands can cause an exception. The latter, I have only seen an exception when we are not in a git repository, so that case should be caught by the first command.

conf/gulp-tasks/library.gulpfile.js
creotutar: You betcha

conf/gulp-tasks/library.gulpfile.js
creotutar: Underlying functions return booleans and fatals, will check specifically for true here

tmeyer2115 (approved):

PR #890, opened 2020-06-15
VerticalResults: fix viewMore styling

oshi97:
per slack post w/Jeremy + Rose, the whole showMore div should be clickable,
and have background color #fafafa instead of white

TEST=manual
test that background color is #fafafa and whole div is clickable
and on hover text is underlined

### DIFF
--- a/src/ui/components/results/universalresultscomponent.js
+++ b/src/ui/components/results/universalresultscomponent.js
@@ -63,7 +63,7 @@ export default class UniversalResultsComponent extends Component {
       // Icon in the titlebar
       icon: config.sectionTitleIconName || config.sectionTitleIconUrl || 'star',
       // Url that links to the vertical search for this vertical.
-      verticalURL: config.url || verticalKey + '.html',
+      verticalURL: 'url' in config ? config.url : verticalKey + '.html',
       // Show a view more link by default, which also links to verticalURL.
       viewMore: true,
       // By default, the view more link has a label of 'View More'.
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -168,7 +168,8 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     text-decoration: none;
     display: flex;
     align-items: center;
-    &:hover, &:focus 
+    &:hover .yxt-Results-viewAllLabel,
+    &:focus .yxt-Results-viewAllLabel
     {
       color: $color-brand-hover;
       text-decoration: underline;
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -2,6 +2,7 @@
 
 $results-title-bar-background: var(--yxt-color-background-highlight) !default;
 $results-filters-background: white !default;
+$results-view-more-background: var(--yxt-color-background-highlight);
 
 $results-title-bar-text-color: var(--yxt-color-text-primary) !default;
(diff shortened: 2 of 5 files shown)


src/ui/sass/modules/_Results.scss
```diff
@@ -2,6 +2,7 @@
 
 $results-title-bar-background: var(--yxt-color-background-highlight) !default;
 $results-filters-background: white !default;
+$results-view-more-background: var(--yxt-color-background-highlight);
```
creotutar: Can we add !default if an AEB user wanted to override them

oshi97: sorry can't believe I forgot this

src/ui/sass/modules/_Results.scss
```diff
@@ -165,7 +167,8 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     text-decoration: none;
     display: flex;
     align-items: center;
-    &:hover, &:focus 
+    &:hover .yxt-Results-viewAllLabel,
+    &:focus .yxt-Results-viewAllLabel
     {
       color: $color-brand-hover;
```
creotutar: this seems to change the text color too, is that what we want?

creotutar: by removing this, are we affecting legacy sites relying on this `&:hover` and `&:focus` (without viewAllLabel)?

oshi97: there are no legacy sites on this

oshi97: no legacy sites with this styling, this was newly added

oshi97: hmmmm the color is a good point though, it's not in the mock and it wasn't mentioned in the slack post
image
so I think we can just remove it and life will be much simpler

src/ui/templates/results/verticalresults.hbs
```diff
@@ -71,11 +71,9 @@
 
 {{#*inline "viewAll"}}
   {{#if (and _config.isUniversal _config.viewMore _config.viewMoreLabel)}}
-    <div class="yxt-Results-viewAll">
-      <a class="yxt-Results-viewAllLink" href="{{ verticalURL }}">
-        <div class="yxt-Results-viewAllLabel">{{_config.viewMoreLabel}}</div>
-        <div data-component="IconComponent" data-opts='{ "iconName": "chevron" }'></div>
-      </a>
-    </div>
+    <a class="yxt-Results-viewAll yxt-Results-viewAllLink" href="{{ verticalURL }}">
```
creotutar, line 74: do we want to add check to see if there is a verticalURL before showing an empty `<a>` tag?

oshi97: updated!

creotutar (approved):
🥤 🥤 🥤 LGTM

src/ui/components/results/universalresultscomponent.js
```diff
@@ -63,7 +63,7 @@ export default class UniversalResultsComponent extends Component {
       // Icon in the titlebar
       icon: config.sectionTitleIconName || config.sectionTitleIconUrl || 'star',
       // Url that links to the vertical search for this vertical.
-      verticalURL: config.url || verticalKey + '.html',
+      verticalURL: 'url' in config ? config.url : verticalKey + '.html',
```
creotutar, line 66: can we add this to README?

oshi97: cleaned up readme section!

PR #892, opened 2020-06-15
DateRangeFilter: update display value

oshi97:
> Date range and number ranges follow the format "Between XXX and YYY". Spec says it should just be "XXX - YYY". Can we get rid of the words "between" and "and"?

TEST=manual
check that uses updated display value in applied filters bar


creotutar (approved):
⛵ ⛵ ⛵ LGTM

PR #894, opened 2020-06-15
FilterSearch + GeolocationFilter: remove quotes from filter metadata

oshi97:
>"Certain filter values appear in quotes, which they shouldn't. For example, any filter produced by the FilterSearch component will appear in quotes
Same problem for values from GeolocationFilter (e.g. `"New York, New York, United States"`).
Can we get rid of the quotes?"

TEST=manual
tested that geolocation and filtersearch filters dont have quotes in the applied filters bar


creotutar (approved):
🖌️ 🖌️ 🖌️ LGTM

PR #897, opened 2020-06-15
FilterBox: 0 padding-left on reset button when no apply button

oshi97:
A new feature was added where you are able to render only a reset-all button or only render an
apply button for FilterBox. Before, these two were always rendered together.

This commit updates the spacing around the reset-all button when there is no apply button, so that it lines up better with the rest of the filter box. This was not possible before, so it is not a breaking css change

TEST=manual
check that with an apply button AND reset button, styling is the same as before
with only a reset button, margin-left and padding-left are set to 0


creotutar (approved):
🎲 🎲 🎲 LGTM

PR #898, opened 2020-06-15
v1.3.2

tmeyer2115:
# Version 1.3.2

## Changes

* The SDK's version is now computed automatically, instead of being hard-coded in constants.js.

## Fixes

* The behavior of ```Answers.init``` changed when the ```Answers Status``` check was added to it. Specifically, the SDK's initialization became asynchronous. Prior to this change, invoking ```init``` would immediately initialize the internals of the library. The AEB is dependent on this synchronous behavior. By fixing the status check to block only ```onReady```, this patch ensures the desired, synchronous behavior. 
* Fixed incorrect margin styling for SDK cards.
* Corrected minor styling for the AccordionCard.
* Added a needed null check in the AccordionCard's Handlebars template.

### DIFF
--- a/src/ui/templates/cards/accordion.hbs
+++ b/src/ui/templates/cards/accordion.hbs
@@ -1,26 +1,36 @@
 <div class="yxt-AccordionCard js-yxt-AccordionCard{{#if isExpanded}} yxt-AccordionCard--expanded{{/if}}"
-  id="{{id}}">
-  <button class="yxt-AccordionCard-toggle js-yxt-AccordionCard-toggle"
-    aria-expanded="{{#if isExpanded}}true{{else}}false{{/if}}"
-    aria-controls="{{id}}"
-    tabindex="0"
-  >
+id="{{id}}">
+
+  {{#if _config.details}}
+    <button class="yxt-AccordionCard-toggle js-yxt-AccordionCard-toggle"
+      aria-expanded="{{#if isExpanded}}true{{else}}false{{/if}}"
+      aria-controls="{{id}}"
+      tabindex="0"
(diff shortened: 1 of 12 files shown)


src/ui/templates/cards/accordion.hbs
```diff
@@ -1,26 +1,36 @@
...
-    <div class="yxt-AccordionCard-icon js-yxt-AccordionCard-icon{{#if isExpanded}} yxt-AccordionCard-icon--expanded{{/if}}"
-      data-component="IconComponent"
-      data-opts='{"iconName": "chevron"}'
-      data-prop="icon">
+  {{/if}}
+
+  {{#if _config.details}}
+    <div class="yxt-AccordionCard-body js-yxt-AccordionCard-body" aria-hidden="{{#if isExpanded}}false{{else}}true{{/if}}">
+      <div class="yxt-AccordionCard-content">
+        <div class="yxt-AccordionCard-subtitle">{{{_config.subtitle}}}</div>
```
creotutar, line 28: I realize that what I am asking is a breaking change. Maybe we want an item for this for later? config.subtitle and config.title, are they guaranteed to be available? Do we want to check for their existence before showing the html (which have margin/padding styling)?

tmeyer2115: It's been unguarded like that for a while. Made an item to address in a subsequent release: https://yexttest.atlassian.net/browse/SPR-2445

alexisgrow (approved):
LGTM

creotutar (approved):
🚀 🚀 🚀 LGTM

PR #901, opened 2020-06-15
VerticalResults: add query to view all link

oshi97:
If the url was passed in from the config, the `query` url param was not being set. This is needed when going from universal to vertical search so that the same query is run on page load.

TEST=manual
test that locally, query will be passed through
test setting url to unset, null, undefined, '', defaults to <verticalKey>.html

### DIFF
--- a/src/ui/components/results/verticalresultscomponent.js
+++ b/src/ui/components/results/verticalresultscomponent.js
@@ -227,7 +227,7 @@ export default class VerticalResultsComponent extends Component {
 
   getVerticalURL (data = {}) {
     if ('verticalURL' in this._config) {
-      return this._config.verticalURL;
+      return addParamsToUrl(this._config.verticalURL, { query: this.query });
     }
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
     const verticalURL = verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
--- a/src/ui/components/results/verticalresultscomponent.js
+++ b/src/ui/components/results/verticalresultscomponent.js
@@ -226,11 +226,8 @@ export default class VerticalResultsComponent extends Component {
   }
 
   getVerticalURL (data = {}) {
-    if ('verticalURL' in this._config) {
-      return addParamsToUrl(this._config.verticalURL || '', { query: this.query });
-    }
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
-    const verticalURL = verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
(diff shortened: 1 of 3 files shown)


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -227,7 +227,7 @@ export default class VerticalResultsComponent extends Component {
 
   getVerticalURL (data = {}) {
     if ('verticalURL' in this._config) {
-      return this._config.verticalURL;
+      return addParamsToUrl(this._config.verticalURL || '', { query: this.query });
```
creotutar: Iif someone were to give `url` as undefined, null, or `''` in the universalresults, this would save them and give them an empty (ie index.html) url.

Is this the behavior we want?

oshi97: I'll ask Rose exactly what we want the behavior to be when verticalURL is set vs unset vs '' vs null vs undefined, and how it should default. I probably shouldn't have tried to come up with conclusions on my own on how this config option should work.

oshi97: Updated it to not try to be cute with the `in` check. Will wait for Rose's response

oshi97: sorry dodging the original question, no I don't think we want that defaulting behavior

oshi97: Rose said for the defaulting behavior just be consistent with the current code

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -252,7 +249,7 @@ export default class VerticalResultsComponent extends Component {
     const hasAppliedFilters = this.appliedFilterNodes.length || this.nlpFilterNodes.length;
     const showResultsHeader = this.resultsHeaderOpts.showResultCount ||
       (this.resultsHeaderOpts.showAppliedFilters && hasAppliedFilters);
-    this.query = this.core.globalStorage.getState(StorageKeys.QUERY);
+    this.query = this.core.globalStorage.getState(StorageKeys.QUERY) || '';
```
creotutar: Is this necessary for this to work?

oshi97: no we can remove this line, I was just worried because the default return value of globalStorage.getState() is null. I think right now it's really hard if not impossible to get this to happen, but if this.query does somehow become null we end up appending `?query=null` to the verticalURL.

creotutar: In the spirit of keeping changes atomic, could we add this as a separate change? I think we want to keep this strictly as QA changes for v1.4.0

I'm not sure about the downstream of affects of handling a `query=null` as a `query=` instead

oshi97: sure, will take it out

creotutar (approved):
🐭 🐭 🐭 LGTM

PR #903, opened 2020-06-15
ResultsHeader: add yxt-Results-filters css class for compatibility

oshi97:
This commit adds the yxt-Results-filters css class to yxt-ResultsHeader,
so that if yxt-Results-filters was being display: none'd, then the ResultsHeader
will also be display: none'd. This is only done on universal. On universal, since
v0.13.0 the newer yxt-ResultsHeader has already been in use, and adding this class
on vertical pages is more likely to confuse us than help us in the future.

TEST=manual
test that on local verizon site with display: none, nlp filters no longer show
up on universal

### DIFF
--- a/src/ui/sass/modules/_ResultsHeader.scss
+++ b/src/ui/sass/modules/_ResultsHeader.scss
@@ -32,13 +32,20 @@ $results-header-universal-background: var(--yxt-color-brand-white) !default;
   --yxt-results-header-universal-background: #{$results-header-universal-background};
 }
 
-.yxt-ResultsHeader {
+.yxt-ResultsHeader
+{
+  // reset styling from yxt-Results-filters, a css class added for backwards compatibility
+  margin: inherit;
+  background-color: inherit;
+  border: inherit;
+
   padding: calc(var(--yxt-results-header-spacing) / 4) var(--yxt-results-header-spacing);
   padding-bottom: 0;
   display: flex;
   align-items: baseline;
 
-  &-wrapper {
+  &-wrapper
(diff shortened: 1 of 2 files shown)


creotutar:
nits
> _then_ the ResultsHeader
will also be display: none'd.

> and adding this class
~to when~ on vertical _pages?_ is more likely to confuse us than help us in the future.

src/ui/sass/modules/_ResultsHeader.scss
```diff
@@ -32,13 +32,20 @@ $results-header-universal-background: var(--yxt-color-brand-white) !default;
   --yxt-results-header-universal-background: #{$results-header-universal-background};
 }
 
-.yxt-ResultsHeader {
+.yxt-ResultsHeader
+{
+  // reset styling from yxt-Results-filters, a css class added for backwards compatibility
```
creotutar: I'm worried about keeping reset styling in the SDK.

So the problem is we want to keep a class but we also want new styling. Is it possible to add this reset styling to a selector `yxt-Results-filters.yxt-ResultsHeader--universal` instead of for any use of `yxt-ResultsHeader`? That way it is easier to know when we can remove this styling.

oshi97: sure I like this a lot

oshi97: oh hmm we actually don't need a reset

creotutar (approved):
👍 👍 👍 LGTM

PR #904, opened 2020-06-16
DirectAnswers: color: inherit for icon

oshi97:
.yxt-Results-titleIconWrapper was previously changed to have color: $color-brand-primary.
This css class was also being used in DirectAnswers, which caused an undesired color change.
Double checked, and yxt-Results-titleIconWrapper is only being used in DirectAnswers and
VerticalResults.

This commit adds the currently unused yxt-DirectAnswer-titleIconWrapper class to the
`<div>`. It has the exact same styling as yxt-Results-titleIconWrapper, except for color: inherit.

TEST=manual
test that by default the direct answers icon color is white

### DIFF
--- a/src/ui/sass/modules/_DirectAnswer.scss
+++ b/src/ui/sass/modules/_DirectAnswer.scss
@@ -81,7 +81,7 @@ $direct-answer-footer-height: var(--yxt-module-footer-height) !default;
   {
     display: flex;
     margin-right: calc(var(--yxt-base-spacing) / 2);
-    height: 18px;
+    color: inherit;
   }
 
   &-entityName
--- a/src/ui/sass/modules/_DirectAnswer.scss
+++ b/src/ui/sass/modules/_DirectAnswer.scss
@@ -81,7 +81,11 @@ $direct-answer-footer-height: var(--yxt-module-footer-height) !default;
   {
     display: flex;
     margin-right: calc(var(--yxt-base-spacing) / 2);
-    color: inherit;
+
+    &.yxt-Results-titleIconWrapper
+    {
+      color: inherit;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -81,7 +81,7 @@ $direct-answer-footer-height: var(--yxt-module-footer-height) !default;
   {
     display: flex;
     margin-right: calc(var(--yxt-base-spacing) / 2);
-    height: 18px;
```
creotutar, line 84: It looks like this is styling that has been there for 7 months. Is this a breaking change we want to happen? (It looks like this class was used in v1.3.0)

oshi97: we're actually using yxt-Results-titleIconWrapper in 1.3.0 right now, instead of this yxt-DirectAnswer-titleIconWrapper, so shouldn't be breaking!

src/ui/sass/modules/_DirectAnswer.scss
```diff
@@ -81,7 +81,7 @@ $direct-answer-footer-height: var(--yxt-module-footer-height) !default;
...
+    color: inherit;
```
creotutar: Could we make it so this styling is only on `yxt-DirectAnswer-titleIconWrapper.yxt-Results-titleIconWrapper`?

oshi97: sure, updated!

creotutar (approved):
🚋 🚋 🚋 LGTM

creotutar (approved):
🎏 🎏 🎏 LGTM

PR #905, opened 2020-06-16
VerticalResults/ResultsHeader: add missing event options

oshi97:
This commit adds missing analytics events to the updated vertical results/
results header for the view all link and the change filters link, that were present in the older resultssectionheader.hbs.

These use the same event options as the original analytics events.

TEST=manual
check that on click these events fire with a 200 response


creotutar (approved):
🎃 🎃 🎃 LGTM

PR #910, opened 2020-06-17
ResultsHeader: display facets from response, not request

oshi97:
This commit moves applied filter display logic into the ResultsHeader,
instead of VerticalResults, and also sets up the component to listen
to changes to DYNAMIC_FILTERS. Now, the applied filters bar will display the facets calculated from the response, instead of the facets sent in the request.

Originally, I wanted to just make VerticalResults listen to changes to
DYNAMIC_FILTERS. However, VerticalResults already listens to changes
to VERTICAL_RESULTS, and the sdk does not play nice when a component
tries to listen to multiple StorageKeys. In order to do this the filter calculation
logic needed to be moved into the ResultsHeader.

Due to some complications below, it was necessary to create a bespoke storageKey that mirrors DYNAMIC_FILTERS.

Just registering a listener to DYNAMIC_FILTERS inside the ResultsHeader doesn't work, though
because every time VerticalResults (the parent component) has its setState called, a new ResultsHeader component will be created from scratch, and that new component will register a new, duplicate event listener, without cleaning up the old one. Cleaning up the old event listener to DYNAMIC_FILTERS breaks Facets, which also listen to DYNAMIC_FILTERS, because the current implementation of the off() method erases all event listeners for a moduleId, and cannot be targeted at a specific listener. This problem was worked around by adding a unique ResultsHeader StorageKey, and SPR-2455 was created to investigate this spaghetti code.

A future commit will add collision handling when an NLP filter and facet
share the same display value + fieldId.

TEST=manual
test that results header will not render if showResultCount and appliedFilters.show are both false
test that facets/filters can still be removed from the applied filters bar
test that checking a facet, applying it, then removing it and applying again does not display the removed facet
  in the applied filters bar (i.e. we display facets received from the response, not sent in the request)
test that when I search "consulting", both an nlp filter tag and a facet removable filter tag will be rendered, and that
  clicking the removable tag will remove the selected facet and trigger a new search

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -187,6 +187,7 @@ export default class Core {
 
         if (data[StorageKeys.DYNAMIC_FILTERS]) {
           this.globalStorage.set(StorageKeys.DYNAMIC_FILTERS, data[StorageKeys.DYNAMIC_FILTERS]);
+          this.globalStorage.set(StorageKeys.RESULTS_HEADER, data[StorageKeys.DYNAMIC_FILTERS]);
         }
         if (data[StorageKeys.SPELL_CHECK]) {
           this.globalStorage.set(StorageKeys.SPELL_CHECK, data[StorageKeys.SPELL_CHECK]);
--- a/src/core/filters/filterregistry.js
+++ b/src/core/filters/filterregistry.js
@@ -26,6 +26,22 @@ export default class FilterRegistry {
   }
 
   /**
+   * Returns an array containing all of the filternodes stored in global storage.
+   * @returns {Array<FilterNode>}
+   */
+  getAppliedFilterNodes () {
+    const globalStorageFilterNodes = [
+      ...this.getStaticFilterNodes(),
+      ...this.getFacetFilterNodes()
+    ];
+    const locationRadiusFilterNode = this.getFilterNodeByKey(StorageKeys.LOCATION_RADIUS);
+    if (locationRadiusFilterNode) {
+      globalStorageFilterNodes.push(locationRadiusFilterNode);
+    }
+    return globalStorageFilterNodes;
+  }
+
+  /**
    * Get all of the {@link FilterNode}s for static filters.
    * @returns {Array<FilterNode>}
    */
--- a/src/core/filters/filterregistry.js
+++ b/src/core/filters/filterregistry.js
@@ -29,7 +29,7 @@ export default class FilterRegistry {
    * Returns an array containing all of the filternodes stored in global storage.
    * @returns {Array<FilterNode>}
    */
-  getAppliedFilterNodes () {
+  getAllFilterNodes () {
     const globalStorageFilterNodes = [
       ...this.getStaticFilterNodes(),
       ...this.getFacetFilterNodes()
--- a/src/core/utils/filternodeutils.js
+++ b/src/core/utils/filternodeutils.js
@@ -20,7 +20,7 @@ export function convertNlpFiltersToFilterNodes (nlpFilters) {
 /**
  * Flattens an array of {@link FilterNode}s into an array
  * of their constituent leaf {@link SimpleFilterNode}s.
- * @param {Array<FilterNode>} filterNodes 
+ * @param {Array<FilterNode>} filterNodes
  * @returns {Array<SimpleFilterNode>}
  */
 export function flattenFilterNodes (filterNodes) {
--- a/src/core/utils/filternodeutils.js
+++ b/src/core/utils/filternodeutils.js
@@ -1,0 +1,41 @@
+import FilterNodeFactory from '../filters/filternodefactory';
+import Filter from '../models/filter';
+import FilterMetadata from '../filters/filtermetadata';
+
+/**
+ * Converts an array of {@link AppliedQueryFilter}s into equivalent {@link SimpleFilterNode}s.
+ * @param {Array<AppliedQueryFilter>} nlpFilters
+ * @returns {Array<SimpleFilterNode>}
+ */
+export function convertNlpFiltersToFilterNodes (nlpFilters) {
+  return nlpFilters.map(nlpFilter => FilterNodeFactory.from({
+    filter: Filter.from(nlpFilter.filter),
+    metadata: new FilterMetadata({
+      fieldName: nlpFilter.key,
+      displayValue: nlpFilter.value
+    })
+  }));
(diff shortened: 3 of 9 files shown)


oshi97:
looks like nobody is reviewing this right now, and it can probably be broken up into 2 PRs

src/core/core.js
```diff
@@ -187,6 +187,7 @@ export default class Core {
 
         if (data[StorageKeys.DYNAMIC_FILTERS]) {
           this.globalStorage.set(StorageKeys.DYNAMIC_FILTERS, data[StorageKeys.DYNAMIC_FILTERS]);
+          this.globalStorage.set(StorageKeys.RESULTS_HEADER, data[StorageKeys.DYNAMIC_FILTERS]);
```
tmeyer2115, line 190: I'm not sure I follow why we can't just configure ResultsSectionHeader to listen to DYNAMIC_FILTERS. Can you not turn off a specific callback listener with globalStorage.off? It looks like it takes a callback to turn off.

src/core/filters/filterregistry.js
```diff
@@ -25,6 +25,22 @@ export default class FilterRegistry {
     this.availableFieldIds = availableFieldIds;
   }
 
+  /**
+   * Returns an array containing all of the filternodes stored in global storage.
+   * @returns {Array<FilterNode>}
+   */
+  getAppliedFilterNodes () {
```
tmeyer2115: Can we rename this getAllFIlterNodes()? "Applied" filters are at a different level of abstraction that what the registry operates on. It's just a repository for filters.

src/core/utils/filternodeutils.js
```diff
@@ -0,0 +1,41 @@
...
+  return nlpFilters.map(nlpFilter => FilterNodeFactory.from({
+    filter: Filter.from(nlpFilter.filter),
+    metadata: new FilterMetadata({
+      fieldName: nlpFilter.key,
+      displayValue: nlpFilter.value
+    })
+  }));
+}
+
+export function flattenIntoSimpleFilterNodes (filterNodes) {
```
tmeyer2115: I think just calling this "flatten" and providing JSDoc would be enough. This name is kind of long.

src/core/utils/filternodeutils.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * Returns the given array of {@link FilterNode}s,
+ * removing FilterNodes that are empty or have a field id listed as a hidden.
+ * @param {Array<FilterNode>} filterNodes
+ * @param {Array<string>} hiddenFields
+ * @returns {Array<FilterNode>}
+ */
+export function purifyFilterNodes (filterNodes, hiddenFields) {
```
tmeyer2115: How about call thing "prune" or something like that? The term "purify" doesn't make it too clear what's happening.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -108,17 +115,29 @@ export default class ResultsHeaderComponent extends Component {
    * not objects, so we need to reformat the grouped applied filters.
    * @returns {Array<Object>}
    */
-  _getAppliedFiltersArray () {
+  _parseAppliedFiltersToArray () {
```
tmeyer2115: createAppliedFiltersArray?

src/core/core.js
oshi97: It looks like it does, but the callback is completely ignored. what `globalStorage.off()` does is call `this._moduleDataContainer[moduleId].off(evt, cb);`, which is an instance of `ModuleData`. `ModuleData` uses  EventEmitter's (its parent's) `off()`, which only takes a single argument, `off (evt) {`

src/core/utils/filternodeutils.js
oshi97: how about `flattenFilterNodes()` ? I feel like just calling it `flatten()` could imply it's flattening a generic array

src/ui/components/results/resultsheadercomponent.js
oshi97: (y)

src/core/utils/filternodeutils.js
oshi97: sure, updated!

src/core/filters/filterregistry.js
oshi97: sure sgtm

src/core/core.js
oshi97: this is the source code for EventEmitter's off()

eventemitter.js
```js
  off (evt) {
    delete this._listeners[evt];
    return this;
  }
```
each moduleId corresponds to one instance of `ModuleData`, so this ends up clearing every listener for a given moduleId

oshi97: moduleId listeners are automatically cleaned up by componentmanager's remove method

componentmanager.js
```js
  remove (component) {
    this._core.globalStorage.off('update', component.moduleId);

    const index = this._activeComponents.findIndex(c => c.name === component.name);
    this._activeComponents.splice(index, 1);
  }
```

oshi97: during a given component's `unMount()` all of the component's children have their remove() called.

component.js
```js
  unMount () {
    if (!this._container) {
      return this;
    }

    this._children.forEach(child => {
      child.unMount();
    });

    DOM.empty(this._container);
    this._children.forEach(c => c.remove());
    this._children = [];
    this.onUnMount();
  }
```

this `remove()` first recursively calls `remove()` on all of its children, and then removes itself using componentmanager's remove()

component.js
```js
  remove () {
    this._children.forEach(c => c.remove());
    this.componentManager.remove(this);
  }
```

src/core/utils/filternodeutils.js
```diff
@@ -0,0 +1,41 @@
...
+  return nlpFilters.map(nlpFilter => FilterNodeFactory.from({
+    filter: Filter.from(nlpFilter.filter),
+    metadata: new FilterMetadata({
+      fieldName: nlpFilter.key,
+      displayValue: nlpFilter.value
+    })
+  }));
+}
+
+export function flattenFilterNodes (filterNodes) {
```
tmeyer2115, line 26: Add some JSDoc

oshi97: added!

tmeyer2115 (approved):

PR #912, opened 2020-06-17
Merge v1.3.2 to develop

oshi97:
# Version 1.3.2
## Changes
* The SDK's version is now computed automatically, instead of being hard-coded in constants.js.
## Fixes
* The behavior of ```Answers.init``` changed when the ```Answers Status``` check was added to it. Specifically, the SDK's initialization became asynchronous. Prior to this change, invoking ```init``` would immediately initialize the internals of the library. The AEB is dependent on this synchronous behavior. By fixing the status check to block only ```onReady```, this patch ensures the desired, synchronous behavior. 
* Fixed incorrect margin styling for SDK cards.
* Corrected minor styling for the AccordionCard.
* Added a needed null check in the AccordionCard's Handlebars template.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
   unit_test:
     docker:
       - image: circleci/node:7.10
-    resource_class: large
+    resource_class: xlarge
     working_directory: ~/answers
     steps:
       - setup-workspace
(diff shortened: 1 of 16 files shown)


creotutar (approved):
☎️ ☎️ ☎️ LGTM

PR #914, opened 2020-06-17
Merge v1.4.0 QA into develop

oshi97:
## QA Updates

* Readme/jsdoc fixes
* DateRangeFilter: applied filters bar display value should always use `${min} - ${max}`, instead of having different displayValues for inclusive vs exclusive ranges
* Facets: fix showExpand fieldOverride
* AppliedFilters: Remove quotes from display value for GeoLocationFilter and FilterSearch
* UniversalResults: fix `url` config option being ignored by VerticalResults
* Add backwards compatibility to styling changes to .yxt-Results-viewAll`, `.yxt-Results-filters`, by using a combination of old + new css class for updated styling. This isn't 100% backwards compatible and has no guarantee to work with layout related css, but it will work for things like color, and display: none. 
* Css for the FilterBox reset-all button when there is only the reset-all button and no apply button has been updated for better vertical alignment. The ability to only have a reset-all button, with no apply button, is net new, and should not be breaking.
* Background for .yxt-Results-viewAll has been changed from white to #fafafa following a comment from jeremy. This css class is net new, and is applied along with the older '.yxt-Results-viewAllLabel' css class.
* UniversalResult's "View More" and "Change Filters" links were missing analytics events previously present in resultssectionheader.hbs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1045,7 +1045,7 @@ ANSWERS.addComponent('Facets', {
       // Optional, if true, display the filter option search input
       searchable: false,
       // Optional, control type, singleoption or multioption
-      control: false,
+      control: 'singleoption',
     }
   },
   // Optional, the label to show on the apply button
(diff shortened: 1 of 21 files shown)


creotutar (approved):
🛸 🛸 🛸 LGTM

PR #916, opened 2020-06-18
Add context as URL parameter for Search API, add ANSWERS.setContext

creotutar:
We add the URL parameter for query rules in the Answers backend. A user
can set the context with the function and expect the context to persist
for subsequent searches in the same page (on both Universal and
Vertical).

Example:
ANSWERS.setContext({
  state: 'tx',
});

Implementation note: Because some of our URL generation functions calculate URLs ad-hoc
based on the URL at time of function call, and because
persistentStorage.set is behind a timeout, the URL generation functions may be called
before the URL reflects the updated context. Thus, we must pass in the
context from globalStorage instead of from the URL. In the future, it
would be good to have knowledge of when the URL was updated from the
persistentStorage.set (a callback system?) or to remove the gathering of
URL params from the current URL and instead rely on storage of these
params.

Implementation note: We use globalStorage instead of persistentStorage with the context that globalStorage will be superseding persistentStorage in the future.

J=SPR-2406
TEST=manual

Tested on a newly created local environment (using Jambo), testing a
local dist build of answers.

Make sure the context is not added by default by looking at the Network
tab in Chrome dev tools.

After calling ANSWERS.setContext (In both JavaScript and from Chrome console),
In both Universal and Vertical Results pages
Test query API requests of subsequent searches include context as a URL param
Test query API requests of application of a facet includes context as a URL parameter.
Test query API requests of pagination includes context as a URL parameter.

Test navigation tab links maintain the context in the URL
Test alternative verticals links maintain the context in the URL
Test "change filters" applied links maintain the context in the URL
Test "view more" links maintain the context in the URL

Test landing on a page with the context in the URL query params shows context in subsequent API requests

Test validation of context through both setContext and loading the page with a URL query param
Add tests for isValidContext

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -428,6 +428,22 @@ class Answers {
       onFinally: config.onFinally || function () {}
     });
   }
+
+  /*
+   * Adds context as a parameter for the query API calls.
+   * @param {Object} context The context object passed in the API calls
+   */
+  setContext (context = {}) {
+    this.core.persistentStorage.set(
+      StorageKeys.API_CONTEXT,
+      JSON.stringify(Object.assign({}, context)),
+      true
+    );
+    this.core.globalStorage.set(
+      StorageKeys.API_CONTEXT,
+      JSON.stringify(Object.assign({}, context))
+    );
+  }
 }
 
 /**
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -448,15 +448,8 @@ class Answers {
       return;
     }
 
-    this.core.persistentStorage.set(
-      StorageKeys.API_CONTEXT,
-      contextString,
-      true
-    );
-    this.core.globalStorage.set(
-      StorageKeys.API_CONTEXT,
-      contextString
-    );
+    this.core.persistentStorage.set(StorageKeys.API_CONTEXT, contextString, true);
+    this.core.globalStorage.set(StorageKeys.API_CONTEXT, contextString);
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -441,19 +441,21 @@ class Answers {
    * Adds context as a parameter for the query API calls.
    * @param {Object} context The context object passed in the API calls
    */
-  setContext (context = {}) {
-    if (context && !isValidContext(JSON.stringify(context))) {
+  setContext (context) {
+    const contextString = JSON.stringify(context);
+    if (!isValidContext(contextString)) {
+      console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
       return;
     }
 
     this.core.persistentStorage.set(
       StorageKeys.API_CONTEXT,
-      JSON.stringify(Object.assign({}, context)),
+      contextString,
       true
     );
     this.core.globalStorage.set(
       StorageKeys.API_CONTEXT,
-      JSON.stringify(Object.assign({}, context))
+      contextString
     );
   }
 }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -30,6 +30,7 @@ import VerticalPagesConfig from './core/models/verticalpagesconfig';
 import { SANDBOX, PRODUCTION } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
 import RichTextFormatter from './core/utils/richtextformatter';
+import { isValidContext } from './core/utils/apicontext';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
@@ -159,6 +160,13 @@ class Answers {
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
     globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
+
+    const context = globalStorage.getState(StorageKeys.API_CONTEXT);
+    if (context && !isValidContext(context)) {
+      persistentStorage.delete(StorageKeys.API_CONTEXT, true);
+      globalStorage.delete(StorageKeys.API_CONTEXT);
+      console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
+    }
 
     this._masterSwitchApi = statusPage
       ? new MasterSwitchApi({ apiKey: parsedConfig.apiKey, ...statusPage }, globalStorage)
@@ -434,6 +442,10 @@ class Answers {
    * @param {Object} context The context object passed in the API calls
    */
   setContext (context = {}) {
+    if (context && !isValidContext(JSON.stringify(context))) {
+      return;
+    }
+
     this.core.persistentStorage.set(
       StorageKeys.API_CONTEXT,
       JSON.stringify(Object.assign({}, context)),
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -169,7 +169,7 @@ export default class Core {
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
         sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS),
         locationRadius: locationRadiusFilterNode ? locationRadiusFilterNode.getFilter().value : null,
-        context: this.persistentStorage.get(StorageKeys.API_CONTEXT)
+        context: this.globalStorage.getState(StorageKeys.API_CONTEXT)
       })
       .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))
       .then(data => {
@@ -236,7 +236,7 @@ export default class Core {
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
-        context: this.persistentStorage.get(StorageKeys.API_CONTEXT)
+        context: this.globalStorage.getState(StorageKeys.API_CONTEXT)
       })
       .then(response => SearchDataTransformer.transform(response, urls, this._fieldFormatters))
       .then(data => {
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -168,7 +168,8 @@ export default class Core {
         queryTrigger: this.globalStorage.getState('queryTrigger'),
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
         sortBys: this.globalStorage.getState(StorageKeys.SORT_BYS),
-        locationRadius: locationRadiusFilterNode ? locationRadiusFilterNode.getFilter().value : null
+        locationRadius: locationRadiusFilterNode ? locationRadiusFilterNode.getFilter().value : null,
+        context: this.persistentStorage.get(StorageKeys.API_CONTEXT)
       })
       .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))
       .then(data => {
@@ -234,7 +235,8 @@ export default class Core {
         geolocation: this.globalStorage.getState(StorageKeys.GEOLOCATION),
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
-        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN)
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
+        context: this.persistentStorage.get(StorageKeys.API_CONTEXT)
       })
       .then(response => SearchDataTransformer.transform(response, urls, this._fieldFormatters))
       .then(data => {
--- a/src/core/utils/apicontext.js
+++ b/src/core/utils/apicontext.js
@@ -1,0 +1,21 @@
+export function isValidContext (context) {
+  // should be both valid JSON and a map
+  let parsed;
+  try {
+    parsed = JSON.parse(context);
+  } catch (e) {
+    return false;
+  }
+
+  if (Array.isArray(parsed)) {
+    return false;
+  }
+  if (typeof parsed === 'string') {
+    return false;
+  }
+  if (!parsed) {
+    return false;
+  }
+
+  return true;
+}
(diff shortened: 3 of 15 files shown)


src/core/core.js
```diff
@@ -234,7 +235,8 @@ export default class Core {
         geolocation: this.globalStorage.getState(StorageKeys.GEOLOCATION),
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: this.globalStorage.getState('queryTrigger'),
-        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN)
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN),
+        context: this.persistentStorage.get(StorageKeys.API_CONTEXT)
```
oshi97: currently we try to use globalStorage instead of persistentStorage where possible. Think this helps separate responsibility a bit, and eventually we want to turn persistentStorage (or however we persist session data) into an implementation detail of globalStorage.

src/ui/components/search/searchcomponent.js
```diff
@@ -141,6 +141,10 @@ export default class SearchComponent extends Component {
       this.debouncedSearch(q);
     });
 
+    this.core.globalStorage.on('update', StorageKeys.API_CONTEXT, () => {
+      this.debouncedSearch(this.query);
+    });
```
tmeyer2115: Was this functionality referenced in the spec somewhere?

src/answers-umd.js
```diff
@@ -428,6 +428,22 @@ class Answers {
       onFinally: config.onFinally || function () {}
     });
   }
+
+  /*
+   * Adds context as a parameter for the query API calls.
+   * @param {Object} context The context object passed in the API calls
+   */
+  setContext (context = {}) {
```
tmeyer2115: I saw this thread in the spec: https://docs.google.com/document/d/1YQTGReyXbrET9FV6_UUZ8xZfYcYLcNQ9W3BtXjbFJpg/edit?disco=AAAAJsIxVwM. Was the decision that the SDK will always have this method and the ability to parse a context from the URL (even if an iframe is not being used)?

src/core/core.js
creotutar: Sure. So the situation here is that we definitely need the context in the persistentStorage. We want people to be able to copy & paste a link with the same context and get the same experience. I can change this to use globalStorage.getState instead of persistentStorage.get, I just thought it was weird that we're /relying/ on persistentStorage being added to globalStorage in ANSWERS.init instead of getting the value from the source (persistentStorage directly). AFAIK nothing should be changing the globalStorage version of this separate from the persistentStorage version of this, and I only added it to globalStorage  in `setContext` because it was already added to globalStorage by the `globalStorage.setAll(persistentStorage.getAll())`

If the context is that we will eventually get rid of persistentStorage and move toward one whole globalStorage, I am fine with removing this getter.

oshi97: yeah we wanna avoid persistent storage like its the plague. in general we want one way of doing things

src/ui/components/search/searchcomponent.js
creotutar: This was not referenced in the spec. Max gave feedback that we would like to have the context persisted as we navigate /within/ the answers experience. That is, it should be reflected in the URLs. This search was the solution I landed on for updating the Nav tab links and the View more links with the updated context params. Do you have any thoughts?

src/answers-umd.js
creotutar: Yup, the decision is that regardless of its iframed or not, we will parse the `context` query on ANSWERS.init (allowing for people to share context links to people and receive the same experience), and we will also let people `setContext` to dynamically add context through JavaScript.

tmeyer2115: Sounds good!

src/ui/components/search/searchcomponent.js
tmeyer2115: I think we should edit those bits of code making the nav tab urls and view more links directly. We can just pull whatever is in globalStorage for the context there. Doing things in an indirect way, like this, may have potential unwanted side-effects.

creotutar: Updated!

src/core/core.js
creotutar: Updated!

src/answers-umd.js
```diff
@@ -428,6 +436,26 @@ class Answers {
...
+    if (context && !isValidContext(JSON.stringify(context))) {
+      return;
```
tmeyer2115, line 448: Do we want to console.error here as well?

src/answers-umd.js
```diff
@@ -428,6 +436,26 @@ class Answers {
...
+    }
+
+    this.core.persistentStorage.set(
+      StorageKeys.API_CONTEXT,
+      JSON.stringify(Object.assign({}, context)),
```
tmeyer2115: Let's factor the JSON.stringify out so that we only need to do it once.

src/answers-umd.js
```diff
@@ -160,6 +161,13 @@ class Answers {
     globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
 
+    const context = globalStorage.getState(StorageKeys.API_CONTEXT);
+    if (context && !isValidContext(context)) {
+      persistentStorage.delete(StorageKeys.API_CONTEXT, true);
+      globalStorage.delete(StorageKeys.API_CONTEXT);
+      console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
```
tmeyer2115, line 168: Sounds like you Bryan, Max, and yourself discussed validation! Was the consensus that in the failure case, we remove the invalid context from any subsequent requests and show a console error?

src/core/utils/apicontext.js
```diff
@@ -0,0 +1,21 @@
+export function isValidContext (context) {
+  // should be both valid JSON and a map
+  let parsed;
+  try {
+    parsed = JSON.parse(context);
+  } catch (e) {
+    return false;
+  }
+
+  if (Array.isArray(parsed)) {
```
tmeyer2115: Is it sufficient to check:
a) JSON.parse works
b) return parsed && typeof parsed === 'Object'?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -180,6 +180,10 @@ export default class NavigationComponent extends Component {
 
     this.checkOutsideClick = this.checkOutsideClick.bind(this);
     this.checkMobileOverflowBehavior = this.checkMobileOverflowBehavior.bind(this);
+
+    this.core.globalStorage.on('update', StorageKeys.API_CONTEXT, () => {
+      this.setState(this.getState());
```
tmeyer2115: I'm not sure this will work. When we investigated yesterday, we found that for this component, setState is passed in globalStorage[StorageKeys.NAVIGATION], which is structured differently than the component's state.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -53,6 +55,15 @@ export default class AlternativeVerticalsComponent extends Component {
...
      */
     this._isShowingResults = opts.isShowingResults || false;
+
+    this.core.globalStorage.on('update', StorageKeys.API_CONTEXT, () => {
+      this.verticalSuggestions = AlternativeVerticalsComponent._buildVerticalSuggestions(
+        this._alternativeVerticals,
+        this._verticalsConfig,
+        this.core.globalStorage.getState(StorageKeys.API_CONTEXT)
+      );
+      this.setState(this.getState());
```
tmeyer2115: Same thing here about what's being passed into setState. I think we need to pass in what's stored in global storage for ALTERNATIVE_VERTICALS. I think this will go for UniversalResults and VerticalResults as well.

src/core/utils/apicontext.js
oshi97: typeof [] is 'object' because javascript. this can definitely still be simplified though, I think ideally the whole method is within the try catch?

src/answers-umd.js
```diff
@@ -428,6 +436,26 @@ class Answers {
...
+    }
+
+    this.core.persistentStorage.set(
+      StorageKeys.API_CONTEXT,
+      JSON.stringify(Object.assign({}, context)),
+      true
+    );
+    this.core.globalStorage.set(
+      StorageKeys.API_CONTEXT,
+      JSON.stringify(Object.assign({}, context))
```
oshi97: nit but could we use spread here? also is this even needed if the isValidContext already guarantees that this is an object?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -96,9 +107,14 @@ export default class AlternativeVerticalsComponent extends Component {
    * @param {object} alternativeVerticals alternativeVerticals server response
    * @param {object} verticalsConfig the configuration to use
    */
-  static _buildVerticalSuggestions (alternativeVerticals, verticalsConfig) {
+  static _buildVerticalSuggestions (alternativeVerticals, verticalsConfig, context) {
     let verticals = [];
-    let queryParams = window.location.search;
+
```
oshi97, line 112: what do you think about refactoring this and other usages of SearchParams to use addParamsToUrl() from urlutils.js?

oshi97: If context is being added to persistent storage, I think it also shouldn't be necessary to explicitly add it to SearchParams. Correct me if I'm wrong, but the param should already be included in window.location.search.substring(1) right? That way we don't have to add this if check every time we create a url in the SDK

src/answers-umd.js
creotutar: That is right!

>if we validate and it's incorrect it should just not pass it to the backend (not like error out or anything)

src/answers-umd.js
creotutar: Good catch! Added

src/answers-umd.js
creotutar: Done, for all three stringify's

src/core/utils/apicontext.js
creotutar: Because of short circuiting, would need to check `parsed` alone. But like the idea of the allowlist and shortening the return!

src/ui/components/navigation/navigationcomponent.js
creotutar: Ah forgot! I also definitely had a misunderstanding about the difference between globalStorage and getState. Changed all to use globalStorage (as that is the data from the /search/) so that we do not pass in extraneous data. However, could you expand on why you don't think this would work. I understand wanting to keep the input data format the same (thus changed), but it looks like the variables from getState would not override anything in the setState call. (Also tested and it was working locally)

src/ui/components/results/alternativeverticalscomponent.js
creotutar: Changed! Thanks

src/ui/components/results/alternativeverticalscomponent.js
oshi97: edit: okay I read your commit message again and this is because these values are sometimes calculate before context is set. would building something like verticalSuggestions within setState instead of in the constructor fix some of these issues?

src/answers-umd.js
creotutar: Ah didn't see this, but yes removed! Thank you!

src/ui/components/results/alternativeverticalscomponent.js
creotutar: * re SearchParams here: sure! what a handy function
* re SearchParams elsewhere: that feels dangerous to me and a bit out of scope of this particular item, would be down to refactor as part of another item
* re verticalSuggestions: I'm not sure I understand. The problem is that the URL is not updated at the time they are called. These functions are taking the URL from window.location.search. Building it in setState would still have the same timing issues, no?

oshi97: when setState is called the component's html updates so it should update the url

creotutar: Ah, as discussed offline, setContext is not necessarily called before all `addComponent`s, would need to check before every update:(

src/answers-umd.js
tmeyer2115: 👍

src/answers-umd.js
```diff
@@ -428,6 +436,28 @@ class Answers {
...
+    const contextString = JSON.stringify(context);
+    if (!isValidContext(contextString)) {
+      console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
+      return;
+    }
+
+    this.core.persistentStorage.set(
+      StorageKeys.API_CONTEXT,
+      contextString,
+      true
```
tmeyer2115: Minor style thing, but seems like these should be able to fit on the same line. Same goes for below.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -457,7 +465,11 @@ export default class NavigationComponent extends Component {
     return tabOrder;
   }
 
-  generateTabUrl (baseUrl, params = new URLSearchParams()) {
+  generateTabUrl (baseUrl, params = new URLSearchParams(), context) {
```
tmeyer2115, line 468: Is the defaulting of params to new URLSearchParams needed any more? It seems like we call this method in one place and we pass in a non-null value there.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -96,9 +107,13 @@ export default class AlternativeVerticalsComponent extends Component {
...
+    const params = {};
+    if (context) {
+      params[StorageKeys.API_CONTEXT] = context;
+    }
```
tmeyer2115, line 116: What about the other params in window.location.search? Seems like we were including those in the generated url before. Does add paramsToUrl take care of that for us?

creotutar: Yup!
```
export function addParamsToUrl (url, params = {}) {
  const urlParams = new SearchParams(window.location.search.substring(1));
```

src/answers-umd.js
creotutar: Sounds good! Do we know if there is anything that will catch these kinds of things before it gets to PR?

src/ui/components/navigation/navigationcomponent.js
creotutar: For sure! As discussed offline, will make another item for this

src/answers-umd.js
```diff
@@ -428,6 +436,21 @@ class Answers {
       onFinally: config.onFinally || function () {}
     });
   }
+
+  /*
+   * Adds context as a parameter for the query API calls.
+   * @param {Object} context The context object passed in the API calls
+   */
+  setContext (context) {
+    const contextString = JSON.stringify(context);
```
tmeyer2115, line 445: Does this blow up if context is null or undefined?

creotutar: Nope!

image

tmeyer2115 (approved):

PR #917, opened 2020-06-18
VerticalResults: move filternode calculation logic outside of component

oshi97:
The _convertNlpFiltersToFilterNodes, _getAppliedFilterNodes, and _processFilterNodes methods inside VerticalResults component were moved into common locations.

A future commit will move these calls inside the ResultsHeader, so that
the ResultsHeader can hook up a globalStorage listener that updates
after DYNAMIC_FILTERS is updated.

TEST=manual, auto
updated unit tests
checked that I can still select a facet, it will be removable, and clicking
  the removable tag will remove the facet and trigger a search

### DIFF
--- a/src/core/filters/filterregistry.js
+++ b/src/core/filters/filterregistry.js
@@ -26,6 +26,22 @@ export default class FilterRegistry {
   }
 
   /**
+   * Returns an array containing all of the filternodes stored in global storage.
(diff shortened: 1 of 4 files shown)


oshi97:
@tmeyer2115 said he prefers the larger PR, since there's basically 0 complexity in this PR, closing this!

PR #919, opened 2020-06-18
ResultsHeader: hide nlp filters that duplicate another filter

oshi97:
This commit changes the ResultsHeader component to ignore nlpFilterNodes
that are duplicates of a different applied FilterNode. This way, if
somebody were to search for a nlpFilter like "technology", that is
configured on the back-end to also return a selected facet in the response,
only the facet filter tag will be displayed.

To do this a filter.equals instance method was added.

TEST=manual, auto
check that I can search consulting or technology, the facet will be checked
off, and only the removable facet filter tag will be rendered.


src/ui/components/results/resultsheadercomponent.js
```diff
@@ -86,6 +86,21 @@ export default class ResultsHeaderComponent extends Component {
...
+   * Returns the currently applied nlp filter nodes, with nlp filter nodes that
+   * are duplicates of other static filter nodes removed.
+   * @returns {Array<FilterNode>}
+   */
+  _pruneDuplicateNlpFilterNodes () {
+    return this.nlpFilterNodes.filter(filterNode => {
+      const isDuplicate = this.appliedFilterNodes.find(node =>
+        filterNode.getFilter().getFilterKey() === node.getFilter().getFilterKey() &&
+        filterNode.getMetadata().displayValue === node.getMetadata().displayValue
+      );
```
tmeyer2115, line 98: Is this the only way to check that two nodes are the "same"? It makes me a bit nervous that we're doing a comparison based on something like displayValue. That's not always the best indicator of two objects sharing the same value.

oshi97: you're right, filters should probably have some kind of equals method

oshi97: Tried to add some lodash helpers for this, which increased the bundle size by a few kb.

```

  dist/answers.min.js
  Package size limit has exceeded by 4.42 KB
  Package size: 154.42 KB
  Size limit:   150 KB

  dist/answers-modern.min.js
  Package size limit has exceeded by 2.93 KB
  Package size: 102.93 KB
  Size limit:   100 KB

  dist/answerstemplates.compiled.min.js
  Package size: 58.83 KB
  Size limit:   100 KB
```

Before this change, the bundle size was

```
  dist/answers.min.js
  Package size: 147.46 KB
  Size limit:   150 KB

  dist/answers-modern.min.js
  Package size: 96.08 KB
  Size limit:   100 KB

  dist/answerstemplates.compiled.min.js
  Package size: 58.87 KB
  Size limit:   100 KB

  With all dependencies, minified and gzipped
```

edit: re ran build on my machine instead of copy pasting the circle ci one. The circle ci one is slightly larger.

src/core/utils/filternodeutils.js
```diff
@@ -45,3 +45,35 @@ export function pruneFilterNodes (filterNodes, hiddenFields) {
...
 }
+
+/**
+ * Check whether two non-combined (i.e. do not contain an array of child Filters) filters
+ * are equal to each other.
+ * @param {Filter} filterNode1
+ * @param {Filter} filterNode2
+ * @returns {boolean}
+ */
+export function haveEqualSimpleFilters (filterNode1, filterNode2) {
```
tmeyer2115: Does this method determine if two SimpleFilterNodes are equal? If so, I think this is better suited as an "equal" method for the SimpleFilterNode class. The "equal" method takes in another SimpleFilterNode and compares it to the current one.

oshi97: it does not, it only compares their filters

oshi97: actually, it can probably be turned into that, though. the filternodes created from the nlp filters also have a valid display value

src/core/models/filter.js
```diff
@@ -23,6 +23,34 @@ export default class Filter {
     }
   }
 
+  /**
+   * Check whether this simple filter is equal to another simple filter.
+   * @param {Filter} filter
+   * @returns {boolean}
+   */
+  isEqualToSimpleFilter (filter) {
+    if (this.getFilterKey() !== filter.getFilterKey()) {
```
tmeyer2115: Since we're only checking one-level deep, I think we can simplify this method a bit. I think instead of your for loop, we could do something like fieldIds1.any(id -> !fieldIdsToValues2.hasOwnProperty(id) || fieldIdsToValues2[id] !== fieldIdsToValues1[id]).

src/core/models/filter.js
```diff
@@ -23,6 +23,34 @@ export default class Filter {
...
+      return false;
+    }
+    const fieldIdsToValues1 = this[this.getFilterKey()];
+    const fieldIdsToValues2 = filter[filter.getFilterKey()];
+    const fieldIds1 = Object.keys(fieldIdsToValues1).sort();
+    const fieldIds2 = Object.keys(fieldIdsToValues2).sort();
+    if (fieldIds1.length !== fieldIds2.length) {
```
tmeyer2115: I think we should give these things better names. The 1 and 2 suffix aren't particularly descriptive. Are these objects maps of fieldId to value? I thought the filterKey itself was a fieldId.

src/core/filters/simplefilternode.js
```diff
@@ -73,4 +73,15 @@ export default class SimpleFilterNode extends FilterNode {
...
+
+  /**
+   * Returns whether this SimpleFilterNode's filter is equal to another SimpleFilterNode's
+   * @param {SimpleFilterNode} node
+   * @returns {boolean}
+   */
+  hasSameFilterAs (node) {
+    const filter1 = this.getFilter();
+    const filter2 = node.getFilter();
+    return filter1.isEqualToSimpleFilter(filter2);
```
tmeyer2115: Even though isEqualToSimpleFilter is operating on Filter instances, I think it may be better to fold that method into this one here. The Filter class has no concept of "simple". Right now, that concept only makes sense in the context of a FilterNode. If we were writing a generic Filter equality checker, that would definitely live in the Filter class. But, the checker we've added to that class depends on assumptions that are at a different level of abstraction.

src/core/models/filter.js
oshi97: Yes you're right. I was worried about duplicate entries, so if you had something like
```js
{
  a: 1,
  a: 1,
  b: 2
},
{
  a: 1,
  b: 2,
  b: 2,
}
```
using the for loop implementation would return that they are unequal, while using the simpler version would return that they are equal.
these are definitely equal Filters, though, even if the objects themselves are not equal. I can't think of any possible matcher that this would ever be an issue.

src/core/filters/simplefilternode.js
oshi97: I was thinking that, eventually we might want to only have "simple" filters and have filter nodes take care of combination. Then, in that world, we could rename this filter method to just equals(). that's not the case right now, though, probably shouldn't be part of this item, and I shouldn't write the SDK into a corner.

oshi97:
ready for review again!

src/core/models/filter.js
oshi97: updated! not sure how much better my new names are, but this is kind of what lodash names their params in their readme

src/core/filters/simplefilternode.js
```diff
@@ -73,4 +73,29 @@ export default class SimpleFilterNode extends FilterNode {
...
+    }
+    const thisMatchersToValues = thisFilter[thisFieldId];
+    const otherMatchersToValues = otherFilter[otherFieldId];
+    const thisMatchers = Object.keys(thisMatchersToValues);
+    const otherMatchers = Object.keys(otherMatchersToValues);
+    if (thisMatchers.length !== otherMatchers.length) {
+      return false;
+    }
+    return thisMatchers.every(m => {
+      return otherMatchersToValues.hasOwnProperty(m) && otherMatchersToValues[m] === thisMatchersToValues[m];
```
tmeyer2115: Minor nit, I think you can remove the return here.

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -86,6 +86,20 @@ export default class ResultsHeaderComponent extends Component {
     });
   }
 
+  /**
+   * Returns the currently applied nlp filter nodes, with nlp filter nodes that
+   * are duplicates of other static filter nodes removed.
```
tmeyer2115: "duplicates of other filter nodes removed"? They won't be duplicates of static filters, necessarily.

src/core/filters/simplefilternode.js
```diff
@@ -73,4 +73,29 @@ export default class SimpleFilterNode extends FilterNode {
...
+    }
+    const thisMatchersToValues = thisFilter[thisFieldId];
+    const otherMatchersToValues = otherFilter[otherFieldId];
+    const thisMatchers = Object.keys(thisMatchersToValues);
+    const otherMatchers = Object.keys(otherMatchersToValues);
+    if (thisMatchers.length !== otherMatchers.length) {
+      return false;
+    }
+    return thisMatchers.every(m =>
+      otherMatchersToValues.hasOwnProperty(m) && otherMatchersToValues[m] === thisMatchersToValues[m]
```
tmeyer2115: LL

oshi97: 102 characters 😭

tmeyer2115 (approved):

PR #922, opened 2020-06-22
ResultsHeader: changeFiltersText config option + underline

oshi97:
> Current: the "Change Filters" link is hardcoded, https://github.com/yext/answers/blob/master/src/ui/templates/results/resultsheader.hbs#L59
Expected:
>1) We had an (undocumented) config option, changeFiltersText, and we need to make this backwards compatible – https://github.com/yext/answers/blob/master/src/ui/templates/results/resultssectionheader.hbs#L38
>2) Need an underline by default (for wcag)

TEST=manual
check that has underline by default
test using changeFiltersText and appliedFilters.changeFiltersText can
set the change filters link text, and that
appliedFilters.changeFiltersText has higher priority

### DIFF
--- a/src/ui/components/results/resultsheadercomponent.js
+++ b/src/ui/components/results/resultsheadercomponent.js
@@ -12,6 +12,7 @@ const DEFAULT_CONFIG = {
   resultsCountSeparator: '|',
   verticalURL: undefined,
   showChangeFilters: false,
+  changeFiltersText: 'change filters',
   removable: false,
   delimiter: '|',
   isUniversal: false,
--- a/src/ui/components/results/resultsheadercomponent.js
+++ b/src/ui/components/results/resultsheadercomponent.js
@@ -12,7 +12,6 @@ const DEFAULT_CONFIG = {
   resultsCountSeparator: '|',
   verticalURL: undefined,
   showChangeFilters: false,
-  changeFiltersText: 'change filters',
   removable: false,
   delimiter: '|',
   isUniversal: false,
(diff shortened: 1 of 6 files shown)


src/ui/components/results/resultsheadercomponent.js
```diff
@@ -12,6 +12,7 @@ const DEFAULT_CONFIG = {
   resultsCountSeparator: '|',
   verticalURL: undefined,
   showChangeFilters: false,
+  changeFiltersText: 'change filters',
```
tmeyer2115: It seems that the default for the hidden, top-level changeFiltersText was just 'change'. Do we need to use that default here?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -85,6 +85,8 @@ export default class UniversalResultsComponent extends Component {
         resultsCountSeparator: defaultConfigOption(config, ['appliedFilters.resultsCountSeparator', 'resultsCountSeparator'], '|'),
         // Whether to show a 'change filters' link, linking back to verticalURL.
         showChangeFilters: defaultConfigOption(config, ['appliedFilters.showChangeFilters', 'showChangeFilters'], false),
+        // The text for the change filters link.
+        changeFiltersText: defaultConfigOption(config, ['appliedFilters.changeFiltersText', 'changeFiltersText'], 'change filters'),
```
tmeyer2115: Same q

src/ui/components/results/resultsheadercomponent.js
oshi97: we can keep the old accordion results behavior. the new default text for should be 'change filters', though

oshi97: another breaking change, I'll ask Rose if its okay

tmeyer2115 (approved):

PR #923, opened 2020-06-22
ResultsHeader: use old scss vars for titlebar

oshi97:
Use the old scss variables for yxt-Results-titleBar where possible.
The only value that cannot use a preexisting scss variable is the
font-weight, which in the mock was changed from 700 to 600.

>"Current:
.yxt-Results-Titlebar {
background: var(--yxt-color-background-highlight)
}
.yxt-Results-title {
  color: var(--yxt-color-text-primary)
}
Expected -- use the previous variables instead, unless there was a reason we changed them (see scss from 1.3.0)
  .yxt-Results-titleBar {
    background-color: var(--yxt-results-title-bar-background);
  }
  .yxt-Results-title {
    color: var(--yxt-results-title-bar-text-color);
  }
for yxt-Results-title:
>- we just aren't targetting the scss correctly? I see it here
>- it looks like this defaults to var(--yxt-color-text-primary) so we should be good if we update"

TEST=manual
check that background-color of yxt-Results-titleBar is the same
check that yxt-Results-title has some color, font-weight, line-height,
and font-size as before.
do quick scan for other preexisting scss variables

### DIFF
--- a/src/ui/sass/modules/_Results.scss
+++ b/src/ui/sass/modules/_Results.scss
@@ -198,7 +198,7 @@ $results-cards-margin: calc(var(--yxt-base-spacing) / 2) !default;
     display: flex;
     padding: calc(var(--yxt-base-spacing) / 2) var(--yxt-base-spacing);
     align-items: center;
-    background-color:  var(--yxt-color-background-highlight);
+    background-color:  var(--yxt-results-title-bar-background);
   }
 
   &-title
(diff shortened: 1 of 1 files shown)


tmeyer2115 (commented):
LGTM, but the hotfix branch isn't named correctly. Hotfix branches should have semantic versioned name (v.a.b.c)

oshi97:
*updated to merge into hotfix/v1.4.1 instead of master

tmeyer2115 (approved):

PR #924, opened 2020-06-22
DirectAnswer: add readme for custom DA card, fix fieldType

oshi97:
This commit adds a readme section for how to add a custom DA
card to the SDK. It also fixes the way fieldType is accessed when
calculating which cardOverride to use. One out of the 2 places
it is mentioned in the SDK spec is wrong and I only read the incorrect
one, and fieldType was not part of the query responses  at the time this
code was originally written.

J=SPR-2451
TEST=manual
test that I can specify a cardOverride for fieldType: 'phone',
that will be applied to 'amani farooque phone' but not
'amani farooque employee department'
test that custom card used in readme works

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,12 +469,12 @@ ANSWERS.addComponent('DirectAnswer', {
 })
 ```
 
-## Custom Direct Answer Card
+## Creating a Custom Direct Answer Card
 
-You can customize the look and behavior of your DirectAnswer, you can create a custom Direct Answer card.
+You can customize the look and behavior of your Direct Answer by creating a custom Direct Answer card.
 
-A custom Direct Answer card is given the directAnswer data from the query, the same data the built-in component
-operates on. That object will look something like the below:
+A custom Direct Answer card is given the same data as the built-in card.
+That data will look something like the below:
 
 ```js
 {
@@ -503,7 +503,8 @@ operates on. That object will look something like the below:
 }
 ```
 
-A Direct Answer custom card needs a corresponding template. This can be added either inline by changing the component's constructor to:
+A custom Direct Answer card needs a corresponding template.
+This can be added either inline by changing the component's constructor to:
 
 ```js
     constructor(config, systemConfig) {
@@ -516,13 +517,13 @@ Or by including a custom template bundle, and adding:
 
 ```js
   static defaultTemplateName () {
-    return 'CustomDirectAnswerTemplateName';
+    return 'CustomDirectAnswerTemplate';
   }
 ```
 
-Where 'YourTemplateName' is the name of the template registered onto the Handlebars renderer.
+Where 'CustomDirectAnswerTemplate' is the name the template is registered under.
 
-The example custom Direct Answer card will use the Handlebars template below.
+We will use the following template for our example card.
 
 ```hbs
   <div class="customDirectAnswer">
@@ -561,7 +562,7 @@ The example custom Direct Answer card will use the Handlebars template below.
   {{/inline}}
 ```
 
-Some css to flip the thumbs up icon the right way.
+This specific example needs some css to flip the thumbs up icon the right way.
 
 ```css
   .customDirectAnswer-thumbsUpIcon svg {
@@ -569,57 +570,23 @@ Some css to flip the thumbs up icon the right way.
   }
 ```
 
-This is an example of a possible custom Direct Answer card. It has custom rendering that depends on
-the fieldType of the directAnswer, as well as a custom analytics event.
+This is the javascript class for our custom Direct Answer card.
+It applies custom formatting to the Direct Answer, registers analytics events
+to the thumbs up/down icons, and passes custom event options into the template.
 
 ```js
-  class CustomDirectAnswerClassName extends ANSWERS.Component {
+  class CustomDirectAnswerClass extends ANSWERS.Component {
     constructor(config, systemConfig) {
       // If you need to override the constructor, make sure to call super(config, systemConfig) first.
       super(config, systemConfig);
 
-      // For simplicity's sake, this custom Direct Answer card sets the template using setTemplate(), as opposed to
+      // For simplicity's sake, we set this card's template using setTemplate(), as opposed to
       // a custom template bundle.
-      this.setTemplate(`
-        <div class="customDirectAnswer">
-          <div class="customDirectAnswer-type">
-            {{type}}
-          </div>
-          <div class="customDirectAnswer-value">
-            {{#each customValue}}
-            {{#if url}}
-              {{> valueLink }}
-            {{else}}
-              {{{this}}}
-            {{/if}}
-            {{/each}}
-          </div>
-          {{> feedback}}
-        </div>
-
-        {{#*inline 'feedback'}}
-        <span class="customDirectAnswer-thumbsUpIcon js-customDirectAnswer-thumbsUpIcon"
-          data-component="IconComponent"
-          data-opts='{"iconName": "thumb"}'
-        ></span>
-        <span class="customDirectAnswer-thumbsDownIcon js-customDirectAnswer-thumbsDownIcon"
-          data-component="IconComponent"
-          data-opts='{"iconName": "thumb"}'
-        ></span>
-        {{/inline}}
-
-        {{#*inline 'valueLink'}}
-        <a class="customDirectAnswer-fieldValueLink" href="{{{url}}}"
-          {{#if @root/eventType}}data-eventtype="{{@root/eventType}}"{{/if}}
-          {{#if @root/eventOptions}}data-eventoptions='{{{ json @root/eventOptions }}}'{{/if}}>
-          {{{displayText}}}
-        </a>
-        {{/inline}}
-      `);
+      this.setTemplate(`<div> your template here </div>`)
     }
 
     /**
-     * The setState method allows you to pass variables directly into your template.
+     * setState() lets you pass variables directly into your template.
      * Here, data is the directAnswer data from the query.
      * Below, we pass through a custom direct answers value, customValue.
      * @param {Object} data
@@ -653,7 +620,7 @@ the fieldType of the directAnswer, as well as a custom analytics event.
     }
 
     /**
-     * reportQuality will send the quality feedback to analytics
+     * reportQuality() sends an analytics event (either THUMBS_UP or THUMBS_DOWN).
      * @param {boolean} isGood true if the answer is what you were looking for
      */
     reportQuality(isGood) {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -468,6 +469,72 @@ ANSWERS.addComponent('DirectAnswer', {
...
+     *     fieldName: "Phone Number",
+     *     fieldApiName: "mainPhone",
+     *     value: "+11234567890",
+     *     fieldType: "phone" 
+     *   },
+     *   relatedItem: { 
+     *     verticalConfigId: 'people',
+     *     data: { ... }
+     *   }
+     * }
```
tmeyer2115: I think you should separate this out. Maybe start the section by showing what the directAnswer data looks like. Then, have a blurb about how you can base a custom DA card around this object with an example (like the one you have currently).

README.md
```diff
@@ -468,6 +469,72 @@ ANSWERS.addComponent('DirectAnswer', {
...
+## Custom Direct Answer Card
+
+```js
+  class MyClassName extends ANSWERS.Component {
+    constructor(config = {}, systemConfig = {}) {
+      super(config, systemConfig);
+
+      // Add your Handlebars template here.
+      this.setTemplate(`
+        <div class="myDirectAnswer">
```
tmeyer2115: I think we should separate out this HTML into another code block. People can provide a template in-line like this, but they can also provide it in other ways. I think the best practice is to not have it inline like this. I think we want a bit more complexity in this example as well. Maybe show a basic switch on the type to do some very simple formatting of the answer value. One of the switch cases could further branch by checking for existence of a particular fieldName and altering formatting accordingly.

oshi97: What is best practice for how to provide a custom template? I'm not aware of any other ways to do this. This is how it's currently being done in the theme.

oshi97: I feel a little uncomfortable just glossing over how to add a template especially since the "Creating Custom Components" section of the readme also kind of glosses over this. I guess the other way would be to add your template directly to the handlebars renderer? Think we should add documentation on the best way to add a template.

README.md
oshi97: tried to separate things, as well as include similar logic to what is currently in the theme

README.md
oshi97: I wrote a small blurb that you can include your own template bundle and reference the template by name. Does this work?

README.md
```diff
@@ -468,6 +469,158 @@ ANSWERS.addComponent('DirectAnswer', {
 })
 ```
 
+## Custom Direct Answer Card
+
+If you need the full power of an SDK component, you can specificy a custom Direct Answers card.
```
tmeyer2115: I'm not sure I follow this opening sentence. We want to highlight that you can fully customize/configure the look of your DA with a custom DA card.

README.md
```diff
@@ -468,6 +469,158 @@ ANSWERS.addComponent('DirectAnswer', {
...
+  answer: {
+    entityName: "Entity Name",
+    fieldName: "Phone Number",
+    fieldApiName: "mainPhone",
+    value: "+11234567890",
+    fieldType: "phone" 
+  },
+  relatedItem: { 
+    verticalConfigId: 'people',
+    data: { ... }
```
tmeyer2115: Flesh this out a bit. The data contains all the other fieldValues, right? People will want to know how to access those.

README.md
```diff
@@ -468,6 +469,158 @@ ANSWERS.addComponent('DirectAnswer', {
...
+  }
+}
+```
+
+An SDK component needs a corresponding template. This can be added either inline by changing the constructor to
```
tmeyer2115: I would say "The new Direct Answers card component". The README states elsewhere that components, in general, need HBS templates. These new instructions reference "SDK component" and "component", but don't mention that the component they are creating is for a new Direct Answers card. I think that's important to explicitly include early on.

README.md
oshi97: got it updated

README.md
```diff
@@ -468,6 +469,170 @@ ANSWERS.addComponent('DirectAnswer', {
...
+  <a class="customDirectAnswer-fieldValueLink" href="{{{url}}}"
+    {{#if @root/eventType}}data-eventtype="{{@root/eventType}}"{{/if}}
+    {{#if @root/eventOptions}}data-eventoptions='{{{ json @root/eventOptions }}}'{{/if}}>
+    {{{displayText}}}
+  </a>
+  {{/inline}}
+```
+
+This is an example of a possible custom Direct Answer card. It has custom rendering that depends on
+the fieldType of the directAnswer, as well as a a custom analytics event.
```
alexisgrow: nit: Extra "a" in: `as well as a a custom analytics`

README.md
```diff
@@ -468,6 +469,170 @@ ANSWERS.addComponent('DirectAnswer', {
...
+    /**
+     * The name of your custom direct answer card. THIS is the value you will use in any config,
+     * such as defaultCard, when you want to specify this custom Direct Answer card.
+     * @returns {string}
+     */
+    static get type() {
+      return 'MyCustomDirectAnswerCard';
+    }
+  }
+
```
alexisgrow, line 680: I think custom components also need the 

```
  static areDuplicateNamesAllowed() {...}
```

func, is that right?

README.md
```diff
@@ -468,6 +469,170 @@ ANSWERS.addComponent('DirectAnswer', {
...
+```
+
+This is an example of a possible custom Direct Answer card. It has custom rendering that depends on
+the fieldType of the directAnswer, as well as a a custom analytics event.
+
+```js
+  class CustomDirectAnswerClassName extends ANSWERS.Component {
+    constructor(config, systemConfig) {
+      // If you need to override the constructor, make sure to call super(config, systemConfig) first.
+      super(config, systemConfig);
```
alexisgrow, line 581: I know setting the template is discussed above the example, but I don't think this example component is setting the template.

README.md
```diff
@@ -468,6 +469,170 @@ ANSWERS.addComponent('DirectAnswer', {
...
+        customValue: this.getCustomValue(answer),
+        eventType: 'CUSTOM_EVENT',
+        eventOptions: {
+          searcher: "UNIVERSAL",
+          verticalConfigId: this.verticalConfigId,
+          entityId: this.associatedEntityId,
+        }
+      });
+    }
+
```
alexisgrow, line 610: For the Direct Answer card feedback, they'd need to set up the click listeners in an onMount. Do you think it's worth adding one here in this example? This is what the theme onMount looks like: https://github.com/yext/answers-hitchhiker-theme/blob/master/directanswercards/card_component.js#L40

README.md
oshi97: oof, thank you!

README.md
oshi97: the built in DirectAnswer doesn't currently set this (which defaults to false) so I think it should be fine.

README.md
oshi97: yeah I talked with Tom about this, he didn't like calling setTemplate inline. @tmeyer2115 ?

README.md
oshi97: sure added a basic onMount. I tried doing the same form thing that the theme does, but then after looking at it again it just seemed way too complicated for this case. (mostly just the custom event part)

README.md
tmeyer2115: The reason I suggested not including that explicitly was because there are multiple ways to provide a custom component with a template. I didn't want folks thinking they had to do it that way. Most probably will. I'm fine if you want to add it back in (with a comment indicating this is one of a few ways of supplying the template).

README.md
alexisgrow: oh nice ok sgtm!

README.md
alexisgrow: Amazing lgtm

alexisgrow (commented):
LGTM, but waiting for Tom for the +2!

README.md
oshi97: sure, added it back with a comment saying this is done for simplicity's sake. we don't provide instructions on how to provide a custom template bundle so I was hesitant to do it like that in the example.

README.md
```diff
@@ -468,6 +469,252 @@ ANSWERS.addComponent('DirectAnswer', {
 })
 ```
 
+## Custom Direct Answer Card
+
+You can customize the look and behavior of your DirectAnswer, you can create a custom Direct Answer card.
```
tmeyer2115: ".. of your DirectAnswer by creating a custom Direct Answer card"

README.md
```diff
@@ -468,6 +469,252 @@ ANSWERS.addComponent('DirectAnswer', {
...
+
+A custom Direct Answer card is given the directAnswer data from the query, the same data the built-in component
```
tmeyer2115: We should have consistency in how we refer to things. The terms "Direct Answer", "DirectAnswer", and "directAnswer" are used here. I think we should just stick to calling it "Direct Answer". You and I understand all 3 phrases have a subtly different meaning. Most people reading this won't have that insight (and it's not too important that they do).

README.md
```diff
@@ -468,6 +469,252 @@ ANSWERS.addComponent('DirectAnswer', {
...
+
+Or by including a custom template bundle, and adding:
+
+```js
+  static defaultTemplateName () {
+    return 'CustomDirectAnswerTemplateName';
+  }
+```
+
+Where 'YourTemplateName' is the name of the template registered onto the Handlebars renderer.
```
tmeyer2115: This name is not used above.

README.md
```diff
@@ -468,6 +469,252 @@ ANSWERS.addComponent('DirectAnswer', {
...
+
+```js
+  class CustomDirectAnswerClassName extends ANSWERS.Component {
+    constructor(config, systemConfig) {
+      // If you need to override the constructor, make sure to call super(config, systemConfig) first.
+      super(config, systemConfig);
+
+      // For simplicity's sake, this custom Direct Answer card sets the template using setTemplate(), as opposed to
+      // a custom template bundle.
+      this.setTemplate(`
```
tmeyer2115: I don't think you need to write out the entire template again here. Just do something like:

this.setTemplate(`TEMPLATE FROM ABOVE`)

There's already a lot going on with this code snippet.

README.md
```diff
@@ -468,6 +469,252 @@ ANSWERS.addComponent('DirectAnswer', {
...
+
+Some css to flip the thumbs up icon the right way.
+
+```css
+  .customDirectAnswer-thumbsUpIcon svg {
+    transform: rotate(180deg);
+  }
+```
+
+This is an example of a possible custom Direct Answer card. It has custom rendering that depends on
```
tmeyer2115: "This is an example of the JS for a custom Direct Answer card".

README.md
oshi97: sorry

README.md
oshi97: sure

tmeyer2115 (approved):

Issue #915, opened 2020-06-17
RADIUS_FILTER 0 value documentation clarity


bhainesva:
Hi all, 

I struggled a little using the RADIUS_FILTER so opening this issue to get people's opinions on whether the documentation could be clarified. 

The comment for the `value` property of the RADIUS_FILTER options says
> Required, the value of the radius to apply (in meters). If this value is 0, will not filter by radius.

Not sure this is wrong but I found it misleading. It suggests to me if you're trying to create an option that removes any distance limits you should set the value to 0. But what actually happens in that case is that it falls back to the normal answers radius logic which applies a limit based on the query. To really remove radius filtering you need to specify a really large number as the value.

Thanks for consideration.

oshi97:
Hey Ben,

This is definitely misleading, just put up a PR (https://github.com/yext/answers/pull/925) to address this. Please take a look when you get the chance!

oshi97:
PR merged in! closing this

PR #925, opened 2020-06-22
Readme: update radius: 0 readme explanation.

oshi97:
Addresses https://github.com/yext/answers/issues/915.
The readme was not being entirely truthful with what happens
When a radius filter has a value of 0 selected.
In this case no locationRadius param is passed to the backend,
but the backend is still free to do whatever custom thing they want.

TEST=none

### DIFF
--- a/README.md
+++ b/README.md
@@ -1175,6 +1175,7 @@ ANSWERS.addComponent('FilterOptions', {
 ```
 
 The options config varies depending on whether the optionType is 'STATIC_FILTER' or 'RADIUS_FILTER'.
+A STATIC_FILTER allows you to filter on a custom specified field, while a RADIUS_FILTER will add separate `locationRadius` parameter to your request, set to the currently selected value.
 
 ##### STATIC_FILTER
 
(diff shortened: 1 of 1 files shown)


tmeyer2115 (commented):
I think these README comments are a bit too literal. The locationRadius parameter is added and removed, but someone reading this doc is not necessarily going to know what that does. That requires knowledge of the back-end. We want to describe things in the README according to the behavior they'll see.

README.md
```diff
@@ -1214,7 +1215,7 @@ A STATIC_FILTER allows you to filter on a custom specified field, while a RADIUS
 {    
   options: [
     {
-      // Required, the value of the radius to apply (in meters). If this value is 0, no explicit locationRadius parameter will be applied to the request.
+      // Required, the value of the radius to apply (in meters). If this value is 0, the SDK will not add explicit radius filtering to the request.
```
tmeyer2115: Might be good to quickly mention that the backend (?) might still do some distance filtering of it's own

oshi97: sure, added a blurb. I wanted to avoid saying specifically what the backend does because 1. I don't know and 2. they're (mostly) free to change whenever they like.

tmeyer2115 (approved):

PR #926, opened 2020-06-22
Add activeMobile and activeDesktop classes to style pagination background color

sawong99:
Changed active state styling when there is only one page in the pagination. Previously, the active page in the pagination would have a different background color. Now, if the pagination only has 1 visible page, the active page inherits its background color.

J=SPR-2467

TEST=manual

Tested on local test site. Checked that when maxVisiblePagesDesktop/maxVisiblePagesMobile is 1, the active page in the pagination has no background color. Changed size of window to test behavior on desktop/mobile.

### DIFF
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -255,15 +255,6 @@ export default class PaginationComponent extends Component {
         if (this._maxVisiblePagesMobile > 1) {
           num.activeMobile = true;
         }
-        // if (this._maxVisiblePagesDesktop === 1 && this._maxVisiblePagesMobile === 1) {
-        //   num.onePageAll = true;
-        // } else if (this._maxVisiblePagesMobile === 1) {
-        //   num.activeDesktop = true;
-        // } else if (this._maxVisiblePagesDesktop === 1) {
-        //   num.activeMobile = true;
-        // } else {
-        //   num.activeAll = true;
-        // }
       } else {
         if (i <= mobileBackLimit || i > mobileFrontLimit) {
           num.mobileHidden = true;
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -249,10 +249,10 @@ export default class PaginationComponent extends Component {
       const num = { number: i };
       if (i === pageNumber) {
         num.active = true;
-        if (this._maxVisiblePagesMobile == 1) {
+        if (this._maxVisiblePagesMobile === 1) {
           num.onePageMobile = '-onepage-mobile';
         }
-        if (this._maxVisiblePagesDesktop == 1) {
+        if (this._maxVisiblePagesDesktop === 1) {
           num.onePageDesktop = '-onepage-desktop';
         }
       } else {
--- a/src/ui/components/results/paginationcomponent.js
+++ b/src/ui/components/results/paginationcomponent.js
@@ -249,6 +249,12 @@ export default class PaginationComponent extends Component {
       const num = { number: i };
       if (i === pageNumber) {
         num.active = true;
+        if (this._maxVisiblePagesMobile == 1) {
+          num.onePageMobile = '-onepage-mobile';
+        }
+        if (this._maxVisiblePagesDesktop == 1) {
+          num.onePageDesktop = '-onepage-desktop';
+        }
       } else {
         if (i <= mobileBackLimit || i > mobileFrontLimit) {
           num.mobileHidden = true;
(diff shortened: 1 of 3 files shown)


src/ui/templates/results/pagination.hbs
```diff
@@ -56,7 +56,8 @@
 
     {{#each pageNumbers}}
       {{#if this.active}}
-        <span id="active-page" class="yxt-Pagination-page" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span>
+        <span id="active-page" class="yxt-Pagination-page yxt-Pagination-active{{this.onePageDesktop}}{{this.onePageMobile}}" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span>
+        {{!-- <span id="active-page" class="yxt-Pagination-page{{this.onePage}}" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span> --}}
```
creotutar: Can we remove the comment here?

src/ui/components/results/paginationcomponent.js
```diff
@@ -249,6 +249,12 @@ export default class PaginationComponent extends Component {
       const num = { number: i };
       if (i === pageNumber) {
         num.active = true;
+        if (this._maxVisiblePagesMobile === 1) {
+          num.onePageMobile = '-onepage-mobile';
```
creotutar: Could we turn this into a boolean (like below)? I am of the camp that there is more value in knowing whether this is true or false, and then in the template, we can add classes and logic as needed.

src/ui/templates/results/pagination.hbs
```diff
@@ -56,7 +56,8 @@
 
     {{#each pageNumbers}}
       {{#if this.active}}
-        <span id="active-page" class="yxt-Pagination-page" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span>
+        <span id="active-page" class="yxt-Pagination-page yxt-Pagination-active{{this.onePageDesktop}}{{this.onePageMobile}}" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span>
```
creotutar: We try to use SUIT style classes in SDK (minus the `yxt-` prefix). Could we add modifiers instead? e.g.

`yxt-Pagination-active yxt-Pagination-active--onePageMobile yxt-Pagination-active--onePageDesktop`

src/ui/sass/modules/_Pagination.scss
```diff
@@ -55,10 +55,28 @@ $pagination-color-hover: var(--yxt-color-text-secondary) !default;
     }
   }
 
-  #active-page {
+  &-active {
     background-color: var(--yxt-pagination-color-active-page);
   }
 
+  &-active-onepage-desktop {
```
creotutar: Aligned with the change below to add more classes, could we only add this `background-color: inherit` for the classes that need them?

e.g.

```
&-active {
  &--onePageMobile {
    @media (max-width: $breakpoint-mobile-max) {
      background-color: inherit;
    }
  }
}
```

creotutar: As an afterthought, perhaps we now have a different meaning for `active`. Instead of adding this reset styling to remove the background-color, maybe we can start passing in classes to give active styling. That is, instead of always having the styling on line 59, perhaps we can pass in activeDesktop or activeMobile with corresponding classes to add background-color with their respective media breakpoints. What do you think?

src/ui/sass/modules/_Pagination.scss
```diff
@@ -55,10 +55,28 @@ $pagination-color-hover: var(--yxt-color-text-secondary) !default;
     }
   }
 
-  #active-page {
+  &--active-all {
     background-color: var(--yxt-pagination-color-active-page);
   }
 
+  &--active-mobile {
```
creotutar: I think we want this all to be one word camelCase to fit the BEM style e.g. `--activeMobile`

src/ui/templates/results/pagination.hbs
```diff
@@ -56,7 +56,9 @@
 
     {{#each pageNumbers}}
       {{#if this.active}}
-        <span id="active-page" class="yxt-Pagination-page" aria-label="Currently on page {{../pageNumber}}">{{../pageLabel}} {{../pageNumber}}</span>
+        <span id="active-page" class="yxt-Pagination-page {{#if this.activeAll}}yxt-Pagination-active-all{{/if}} {{#if this.onePageAll}}yxt-Pagination--onepage-all{{/if}}
+        {{#if this.activeMobile}}yxt-Pagination--active-mobile{{/if}}{{#if this.activeDesktop}}yxt-Pagination--active-desktop{{/if}}"
```
creotutar: Can we add spaces here for when the if statements evaluate to true? Are they one class if we don't add this?

e.g. `{{#if this.activeDesktop}} yxt-Pagination--active-desktop{{/if}}`

creotutar (approved):
:orange: :orange: :orange: LGTM

PR #927, opened 2020-06-23
Default initial search readme

rosiegrant:
adding documentation for the default initial search on the search bar (previously missing). not sure if this is the right branch, lmk!


rosiegrant:
this was definitely the wrong thing to do ignore me :D

PR #928, opened 2020-06-23
Add referrerPageUrl to the query endpoint requests and persistentstorage

creotutar:
We add the referrerPageUrl as part of the query endpoint requests to
support query rules in the Answers backend. The URL of the current page
is the source of truth for this parameter. We make three changes to
support this:

* On ANSWERS.init, look at the current URL in persistent storage to get
  the initial value of referrerPageUrl.
* If there is no referrerPageUrl query param already, source it from
  document.referrer of the current page.
* Change the view more link to include referrerPageUrl in the
  navigation.

Note: This makes it such that referrerPageUrl is a query parameter that
is always on the page. This is because referrerPageUrl is determined
when the page loads and is added as a query parameter to maintain the
parameter as you navigate within the experience.

J=SPR-2452
TEST=manual

Test entry of page with URL param referrerPageUrl already set.

Test all search requests contain the referrerPageUrl in the request.

Test works when navigating to an answers page from a standalone page.
Add link to a standalone page to an answers root page. Click the link. Make sure
referrerPageUrl is the original standalone page.

Test as you navigate through the experience, the referrerPageUrl is
carried
 * Change filters
 * Navigation tabs
 * Alternative verticals
 * View more button

Test works in an iFrame, as you navigate within an iFrame, the
referrerPageUrl persists.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -169,8 +169,8 @@ class Answers {
       console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
     }
 
-    if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null
-        && typeof document.referrer === 'string') {
+    if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null &&
+         typeof document.referrer === 'string') {
       const referrer = urlWithoutQueryParamsAndHash(document.referrer);
       persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer, true);
       globalStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer);
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -31,6 +31,7 @@ import { SANDBOX, PRODUCTION } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
 import RichTextFormatter from './core/utils/richtextformatter';
 import { isValidContext } from './core/utils/apicontext';
+import { urlWithoutQueryParamsAndHash } from './core/utils/urlutils';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
@@ -166,6 +167,12 @@ class Answers {
       persistentStorage.delete(StorageKeys.API_CONTEXT, true);
       globalStorage.delete(StorageKeys.API_CONTEXT);
       console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
+    }
+
+    if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null) {
+      const referrer = urlWithoutQueryParamsAndHash(document.referrer || '');
+      persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer, true);
+      globalStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer);
     }
 
     this._masterSwitchApi = statusPage
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -169,8 +169,7 @@ class Answers {
       console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
     }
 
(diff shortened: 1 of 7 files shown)


src/core/utils/urlutils.js
```diff
@@ -55,3 +55,7 @@ export function addParamsToUrl (url, params = {}) {
   }
   return url.split('?')[0] + '?' + urlParams;
 }
+
+export function urlWithoutQueryParamsAndHash (url) {
+  return url.split('?')[0].split('#')[0];
```
oshi97, line 60: would window.location work here? maybe something like window.location.href + window.location.pathname?

oshi97: nevermind please ignore!

oshi97 (commented):
+1 from me!

src/answers-umd.js
```diff
@@ -168,6 +169,12 @@ class Answers {
       console.error(`Context parameter "${context}" is invalid, omitting from the search.`);
     }
 
+    if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null) {
+      const referrer = urlWithoutQueryParamsAndHash(document.referrer || '');
+      persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer, true);
```
tmeyer2115, line 174: Should we even set a value in persistent/global storage if there is no document.referrer?

tmeyer2115 (commented):
Are the navigation tabs and alternative vertical links already constructed in such a way that they'll pass along the referrerPageUrl?

creotutar:
@tmeyer2115 Yup! They source from the current window.location.search when they create their URLs. Because this is done on ANSWERS.init, before the addComponents, the params will be in the URL.

src/answers-umd.js
creotutar: I suppose the `||''` on L173 is unnecessary, since `document.referrer` should always be on the window. I added that mostly in case we add ANSWERS.init tests that don't not have a `window.document`. What do you think?

tmeyer2115: I think we can cross that bridge when we come to it. For now, I'd rather we not set anything if there is no referrer. Conceptually, that feels a bit better to me.

creotutar: Could you explain the reasoning here so I know for next time? If there is no referrer, we want the ANSWERS.init to fail? Or we just don't want to assume the document.referrer if it doesn't exist?

tmeyer2115: If there's no referrer, we don't want Answers.init to fail. I don't think there will always be a referrer. I'd say we should not set anything in globalStorage or persistentStorage at all in this case, since there's not really anything to store.

tmeyer2115: On second thought, might be worth asking the PMs about this situation. If there is no referrer, should we default to no query param or use the current URL? If someone were to bookmark a site and use that to navigate to it, for instance, there'd be no referrer.

creotutar: Hello! If there is no referrer, we want that to be explicit in the URL. The PM stressed that we want the referrer to be carried within an iFrame such that if you navigate to a different tab or generally within the experience, then the referrer is still the same. Thus, if it is empty, we should explicitly say it is empty so that is carried (otherwise we would get the home page as the referrer when we navigate with the tabs, for example).

Did I understand that correctly?

creotutar: If this is a question about what value we should use to determine if someone came to the page directly, I'd be curious about why we would want to stray from the `document.referrer` value.

I asked by giving an example a while ago with `''` as the referrer and the PM was good with it, just confirmed again and they said it works! How do you feel?

tmeyer2115: Okay, so if they navigated to the Answers page directly, we'd want to have a 'referrerPageUrl=' in the URL? Then that would be persisted across all pages of the experience (or iframe). Based on this: https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer, I think the typeof will always be string. So, that check might be redundant. Once we get rid of that, seems good to me!

creotutar: Ah, misunderstanding here. When I asked "when there is no referrer", I should have specified I meant "null/undefined" not `""`. Changed!

tmeyer2115 (approved):

PR #929, opened 2020-06-23
Remove the hold stage for canary builds push to s3

creotutar:
We should not be holding the builds for every commit merged into
develop. Instead, we should push to s3 with the associated canary
version as soon as it passes unit tests and builds correctly.

J=SPR-2453
TEST=manual

circleci config validate


alexisgrow (approved):
LGTM

PR #931, opened 2020-06-24
Refactor Acceptance Tests

oshi97:
This commit:
splits acceptancesuice.js into its constituent tests
uses bespoke experience key sdkacceptancetests instead of rosetest
use VerticalPage for the existing Facets test. The FacetsPage fixture is just a vertical page with Facets
so it felt weird having a whole separate page block for it.
TEST=auto
reran acceptance tests


oshi97:
rebased onto hotfix/v1.4.1 instead of develop

oshi97:
no longer renames facetspage to filterspage, will make a new removable filters html fixture instead

oshi97:
no longer renames facetspage to filterspage

PR #940, opened 2020-06-25
FilterOptions: fix clearSearch button's vertical alignment

oshi97:
J=SPR-2506
TEST=manual
Tested that on page
https://synergic-health_yextdemos_com.yextpages.net/providers.html
adding this line of css correctly vertically aligns the clear button

Also tested that locally, I can manually set the text input to arbitrary
heights and the clear button will still be aligned

test that extra space above and below clear button is not clickable

### DIFF
--- a/src/ui/sass/modules/_FilterOptions.scss
+++ b/src/ui/sass/modules/_FilterOptions.scss
@@ -101,6 +101,7 @@
     position: absolute;
     right: 0;
     top: 50%;
+    height: 100%;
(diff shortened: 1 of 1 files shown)


tmeyer2115 (approved):
LGTM! Was this tested in IE11 as well?

oshi97:
> LGTM! Was this tested in IE11 as well?

... it is now! noticed a separate issue where ie11 automatically gives the input a clear button (added css to not display this)

PR #942, opened 2020-06-26
Use pixels instead of rem/em for filter options caret sizing

alexisgrow:
TEST=manual

Use local test site with facets and filterbox. Confirm that caret sizes look the same before and after change.


creotutar (approved):
💯 💯 💯 LGTM

PR #943, opened 2020-06-26
Increase package size limit for legacy bundle

alexisgrow:
TEST=auto

Run npm run build locally, see package size is under limit. That is not the same limit used in CircleCI, so wait for the "build" step to finish there as well, see it was successful.


tmeyer2115 (approved):

oshi97 (approved):
🌳

PR #946, opened 2020-06-27
ResultsHeader: fix hiddenFields to work on NLP filters

oshi97:
The hiddenFields config option was only being applied to non-nlp applied
filters. NLP filters were ignoring this config option. Also added this
to my applied filters acceptance tests PR.

T=335040
TEST=manual
test that work for BOTH NLP applied filters and applied filters from
a Facets or FilterBox component.

### DIFF
--- a/src/ui/components/results/resultsheadercomponent.js
+++ b/src/ui/components/results/resultsheadercomponent.js
@@ -47,7 +47,7 @@ export default class ResultsHeaderComponent extends Component {
      * Array of nlp filters in the search response.
      * @type {Array<AppliedQueryFilter>}
      */
-    this.nlpFilterNodes = this._convertNlpFiltersToFilterNodes(data.nlpFilters || []);
+    this.nlpFilterNodes = convertNlpFiltersToFilterNodes(data.nlpFilters || []);
 
     /**
      * TODO (SPR-2455): Ideally, we would be able to set moduleId to DYNAMIC_FILTERS, the actual data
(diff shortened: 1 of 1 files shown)


src/ui/components/results/resultsheadercomponent.js
```diff
@@ -86,6 +86,15 @@ export default class ResultsHeaderComponent extends Component {
     });
   }
 
+  /**
+   * Converts an array of {@link AppliedQueryFilter}s into equivalent {@link FilterNode}s, then
+   * removes filternodes that are either blank, or have a fieldId listed in config.hiddenFields.
+   * @param {Array<AppliedQueryFilter>} nlpFilters
+   */
+  _convertNlpFiltersToFilterNodes (nlpFilters) {
+    return pruneFilterNodes(convertNlpFiltersToFilterNodes(nlpFilters), this._config.hiddenFields);
```
tmeyer2115: Small nit, but can we have all the pruning occur in this method (both filtering out hidden fields and duplicates)? It's kinda weird how we do one step of pruning here, then another prune of nlp filters in _groupAppliedFilters.

oshi97: Pruning nlp duplicates of facets/static filters needs to happen after/within setState, since whether or not an nlp filter is a duplicate depends on the other applied filter nodes, which are updated within setState so that they can use the dynamic filters received from the response, not the facets sent in the request. I think moving the hiddenFields pruning into there makes sense though, even if it's very slightly less efficient.

tmeyer2115 (approved):

PR #951, opened 2020-06-30
Add referrerPageUrl to the query endpoint requests and persistentstorage

creotutar:
We add the referrerPageUrl as part of the query endpoint requests to
support query rules in the Answers backend. The URL of the current page
is the source of truth for this parameter. We make three changes to
support this:

* On ANSWERS.init, look at the current URL in persistent storage to get
  the initial value of referrerPageUrl.
* If there is no referrerPageUrl query param already, source it from
  document.referrer of the current page.
* Change the view more link to include referrerPageUrl in the
  navigation.

Note: This makes it such that referrerPageUrl is a query parameter that
is always on the page. This is because referrerPageUrl is determined
when the page loads and is added as a query parameter to maintain the
parameter as you navigate within the experience.

When we redirect, we do not want to passthrough the referrerPageUrl
query param of the page with the redirectUrl search bar.

To ground in an example
Page X: initial page, navigate to Page Y
Page Y: contains a searchbar with redirectUrl, redirects to Page Z
Page Z: our answers experience

The referrerPageUrl of Z is currently X instead of Y because redirectUrl
passes all query params from page Y and page Y generates referrerPageUrl
on ANSWERS.init.

J=SPR-2452
TEST=manual

Test entry of page with URL param referrerPageUrl already set.

Test all search requests contain the referrerPageUrl in the request.

Test works when navigating to an answers page from a standalone page.
Add link to a standalone page to an answers root page. Click the link. Make sure
referrerPageUrl is the original standalone page.

Test as you navigate through the experience, the referrerPageUrl is
carried
 * Change filters
 * Navigation tabs
 * Alternative verticals
 * View more button

Test works in an iFrame, as you navigate within an iFrame, the
referrerPageUrl persists.

* Do not default document.referrer, check if exists before setting so we do not block ANSWERS.init

* semistandard, && at end of line

* Assume document.referrer always exists, error if it does not

Test on a QA site with a searchbar that redirects to a site. Make sure
the site has referrerPageUrl in the query params already. Make sure
that the referrerPageUrl of the resulting answers page has
referrerPageUrl as the QA site, not the value of the QA site's
referrerPageUrl.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -32,6 +32,7 @@ import { SANDBOX, PRODUCTION } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
 import RichTextFormatter from './core/utils/richtextformatter';
 import { isValidContext } from './core/utils/apicontext';
+import { urlWithoutQueryParamsAndHash } from './core/utils/urlutils';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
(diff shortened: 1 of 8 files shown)


oshi97 (approved):
lgtm! I think in general we don't need PRs when approving feature branches, since the 2 constituent commits have already been approved.

PR #953, opened 2020-06-30
Fix empty search behavior when searchbar is initialized with redirectURL

sawong99:
Fix behavior of empty search when redirect url is provided. Previously, when a searchBar is initialized with a redirectUrl param, it's possible to submit an empty search. Now, if it's a search bar for a vertical search (verticalKey has been included), we use the allowEmptySearch boolean to control empty search behavior. If it's a search bar for a universal search, an empty search is not be allowed.

J=SPR-2470

TEST=manual

Tested on local site. Tested behavior when search is empty/not empty, verticalKey is included/not included and when allowEmptySearch is true/false.

### DIFF
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -414,9 +414,11 @@ export default class SearchComponent extends Component {
 
     // If we have a redirectUrl, we want the form to be
     // serialized and submitted.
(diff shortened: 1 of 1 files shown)


oshi97 (approved):

oshi97 (approved):
just a reminder to cherry-pick this to develop as well!

PR #957, opened 2020-07-01
SearchBar: add yxt-SearchBar-AnimatedIcon class w/display: flex

oshi97:
Below from the jira ticket:

> https://www.screencast.com/t/M340mRJKD0D.
Add a new class to the animated icon that can be used to target styling
(right now we just have js-yxt-AnimatedReverse component
Add display: flex; to this newly added class by default.
See here for further information and context (Chris came up with the original solution)
Note: The search bar icon appears centered when you use the SDK directly
– drives me truly nuts I can’t figure this out 😱  
I don’t think it hurts to have this styling in the SDK by default (doesn’t seem to impact anything)
but definitely let me know if not

J=SPR-2558
TEST=manual
test that animated icon has new css class, which has display: flex attached to it
check that the icon out of the box does not look different

### DIFF
--- a/src/ui/templates/search/search.hbs
+++ b/src/ui/templates/search/search.hbs
@@ -49,12 +49,12 @@
           data-opts='{ "iconUrl": "{{_config.customIconUrl}}" }'
       ></div>
     {{else}}
-      <div class="js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon--paused
+      <div class="yxt-SearchBar-AnimatedIcon js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon--paused
         {{#if autoFocus}} yxt-SearchBar-AnimatedIcon--inactive{{/if}}"
         data-component="IconComponent"
         data-opts="{{json forwardIconOpts}}">
       </div>
-      <div class="js-yxt-AnimatedReverse yxt-SearchBar-AnimatedIcon--paused
+      <div class="yxt-SearchBar-AnimatedIcon js-yxt-AnimatedReverse yxt-SearchBar-AnimatedIcon--paused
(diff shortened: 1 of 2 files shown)


tmeyer2115 (commented):
If the issue does not appear when using the SDK directly, does it appear when using the SDK via the Theme? Rose should be able to provide more context here. I think we also want to test this in the situation where they were seeing the problem to see if this fix works there.

oshi97:
> If the issue does not appear when using the SDK directly, does it appear when using the SDK via the Theme? Rose should be able to provide more context here. I think we also want to test this in the situation where they were seeing the problem to see if this fix works there.

Checked again, yes it appears when using the SDK via theme (with only default theme css) and yes it is fixed with display: flex.

src/ui/templates/search/search.hbs
```diff
@@ -49,12 +49,12 @@
           data-opts='{ "iconUrl": "{{_config.customIconUrl}}" }'
       ></div>
     {{else}}
-      <div class="js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon--paused
+      <div class="yxt-SearchBar-AnimatedIcon js-yxt-AnimatedForward yxt-SearchBar-AnimatedIcon--paused
```
creotutar: can we have the `yxt-SearchBar-AnimatedIcon` and `yxt-SearchBar-AnimatedIcon--paused` right next to each other, with latter after the former. it makes it easier to know the effects when trying to determine how `--paused` overrides the base `AnimatedIcon` class

creotutar:
Hello! This looks good to me! Can we update the commit message to be 100 characters long per line?

creotutar (approved):
🍨 🍨 🍨 LGTM

PR #958, opened 2020-07-02
UniversalResults: only default verticalURL inside VerticalResults

oshi97:
Defaulting occurs in VerticalResults already, because this defaulting was occurring
before the defaulting in VerticalResults, none of the VerticalResults defaults were
being applied, preventing it from trying to default to a matching value in the global
VerticalPages config before <verticalKey>.html.

J=SPR-2559
TEST=manual

Test that the view more url will default to a matching entry in
verticalPages before defaulting to \<verticalKey\>.html


creotutar:
This looks good to me!! Can we update desc here too? <100 characters. I'm having trouble understanding the TESTs, can we make them whole sentences? Can we also add a newline after the `TEST` like we do in other commits?

creotutar (approved):
🎩 🎩 🎩 LGTM

PR #960, opened 2020-07-03
query rules: Set persistentStorage on search instead of on init or setContext

creotutar:
Similar to the query, we only set referrerPageUrl or context for the
first time in the URL after a search. This is to mitigate changes to
client subdomains. Before, referrerPageUrl would be added to the URL on
ANSWERS.init, even when the client only had a redirectUrl searchbar.

We change it so that after /any/ search to the API, we update the query
params. A search on universal, vertical, pagination, facets, will all
change the query parameters to any updated values. Note: they are still
available on globalStorage. We are moving the persistentStorage updates.

J=SPR-2554
TEST=manual

Test a searchbar with a redirectUrl. When searching, the referrerPageUrl
and the context do not show on the URL, only on the redirected URL.

Test that setContext context is not applied on a regular experience
until search or pagination or facet apply (essentially any re-search).

Test that referrerPageUrl is still added to links in navigation, view
more, alternative verticals, even if it is not currently in the URL.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -177,9 +177,10 @@ class Answers {
     }
 
     if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null) {
-      const referrer = urlWithoutQueryParamsAndHash(document.referrer);
-      persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer, true);
-      globalStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer);
+      globalStorage.set(
+        StorageKeys.REFERRER_PAGE_URL,
+        urlWithoutQueryParamsAndHash(document.referrer)
+      );
     }
 
     this._masterSwitchApi = statusPage
@@ -476,7 +477,6 @@ class Answers {
       return;
     }
 
-    this.core.persistentStorage.set(StorageKeys.API_CONTEXT, contextString, true);
     this.core.globalStorage.set(StorageKeys.API_CONTEXT, contextString);
   }
 }
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -144,6 +144,19 @@ export default class Core {
       this.filterRegistry.setFacetFilterNodes([], []);
     }
 
+    const { setQueryParams } = options;
+    const context = this.globalStorage.getState(StorageKeys.API_CONTEXT);
+    const referrerPageUrl = this.globalStorage.getState(StorageKeys.REFERRER_PAGE_URL);
+
+    if (setQueryParams) {
+      if (context) {
+        this.persistentStorage.set(StorageKeys.API_CONTEXT, context, true);
+      }
+      if (referrerPageUrl !== null) {
+        this.persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrerPageUrl, true);
+      }
+    }
+
     const searchConfig = this.globalStorage.getState(StorageKeys.SEARCH_CONFIG) || {};
     if (!searchConfig.verticalKey) {
       this.globalStorage.set(StorageKeys.SEARCH_CONFIG, {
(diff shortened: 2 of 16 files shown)


tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -109,8 +109,15 @@ describe('navigation component configuration', () => {
     return {
       globalStorage: {
         getState: (storageKey) => {
-          expect(storageKey).toEqual(StorageKeys.VERTICAL_PAGES_CONFIG);
-          return verticalsConfig;
+          switch (storageKey) {
+            case StorageKeys.REFERRER_PAGE_URL:
```
oshi97: could you make the default return null and only add a case for StorageKeys.VERTICAL_PAGES_CONFIG?

src/answers-umd.js
```diff
@@ -177,9 +177,10 @@ class Answers {
     }
 
     if (globalStorage.getState(StorageKeys.REFERRER_PAGE_URL) === null) {
-      const referrer = urlWithoutQueryParamsAndHash(document.referrer);
-      persistentStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer, true);
-      globalStorage.set(StorageKeys.REFERRER_PAGE_URL, referrer);
+      globalStorage.set(
```
oshi97, line 180: Could you explain again why we need to set referrer page url and api context every time a search is run? Don't really understand why we don't want these on a redirectUrl searchbar. Even after a redirect, shouldn't referrer page url be correctly updated to the new document.referrer?  And wouldn't api context have to be manually set, so even if the searchbar had a redirectUrl wouldn't we want to carry it over?

creotutar: If we are on a client page like https://www.yext.com/ then we would automatically change the parent URL from `https://www.yext.com/` to `https://www.yext.com/?referrerPageUrl=` on ANSWERS.init. This is not the desired behavior from the PM because we should aim to not modify a client's domain (especially if it is their homepage).

Also API context can be set with `ANSWERS.setContext` on any page with Answers. Thus someone with a redirectUrl searchbar can set context by using `ANSWERS.setContext`

tests/ui/components/navigation/navigationcomponent.js
creotutar: Definitely could, but then there is no point for the expect line right?
```
expect(storageKey).toEqual(StorageKeys.VERTICAL_PAGES_CONFIG);
```

What do you think?

oshi97: yeah definitely don't need it

src/core/core.js
```diff
@@ -220,12 +233,25 @@ export default class Core {
...
 
-  search (queryString, urls) {
+  search (queryString, urls, options = {}) {
+    const { setQueryParams } = options;
+    const context = this.globalStorage.getState(StorageKeys.API_CONTEXT);
+    const referrerPageUrl = this.globalStorage.getState(StorageKeys.REFERRER_PAGE_URL);
+
+    if (setQueryParams) {
+      if (context) {
+        this.persistentStorage.set(StorageKeys.API_CONTEXT, context);
```
oshi97, line 248: do we need to set persistentStorage here for api context? Can we not do it in core.setContext?

src/answers-umd.js
oshi97: I see makes sense to me!

oshi97: document.referrer is something that does not ever change on the page, so does it need to be put into globalStorage? It's also something that is recomputed on init, regardless of what the previous value in the url was, so does it need to go into persistentStorage? If we don't store it in either global or persistentStorage, maybe we can remove it from the setQueryParams part of core.verticalSearch()/search() and simplify things a good deal. We then also don't have to pass it through to any static methods, and instead have those static methods call a static getter. Just some thoughts let me know if they sound dumb to you.

src/core/core.js
creotutar: If we set persistentStorage in setContext, then the context will be applied as soon as `ANSWERS.setContext` is called, but the PM would like it to be applied only right before a search. That is, if we are on a redirectUrl searchbar and we use ANSWERS.setContext, it will not affect the client's URL, and will only show up on the experience after the search.

src/answers-umd.js
creotutar: `document.referrer` sadly will change as we navigate within an experience. For example if you land on the universal search, the referrerPageUrl will be `''` but if you navigate to FAQs, `document.referrer` for that page will be universal now instead of the beginning `''`. We would like to propagate the first `document.referrer` through clicks in the experience (esp. for iframes).

oshi97: in that case isnt the faq page a separate page so answers.init gets called again?

oshi97: ok sooo nevermind I didn't understand that we wanted referrer page url to be the original referrer to the experience, sgtm!

oshi97 (approved):

PR #962, opened 2020-07-06
UniversalResults: add top level appliedFilters config

oshi97:
This commit adds a top level appliedFilters config
to UniversalResults. This allows users to easily specify
shared config for appliedFilters, and override it
per vertical.

J=SPR-2540
TEST=manual

Tested that each config option specified in
getChildConfig can be set in the global config,
as well as overridden for a specific vertical.


creotutar:
Can we add this to the README?

oshi97:
> Can we add this to the README?

@creotutar added!

creotutar (approved):
📙 📙 📙  LGTM

PR #965, opened 2020-07-07
Ensure Filter Options are centered

sawong99:
Change align-items: flex-start to align-items: center in .yxt-FilterOptions-options. Previously, the text next to the filter options buttons were slightly too low. Now they are centered.

J=SPR-2561

TEST=manual

Tested on local test site.


oshi97 (approved):
if you've discussed with Rose already lgtm!

PR #967, opened 2020-07-09
Fix back-button for query rules by only updating url when different

creotutar:
Our persistent storage updates history in batches. Imagine the following
scenario:

```
persistentStorage.set(QUERY, ..., replaceHistory = false);
persistentStorage.set(REFERRER_PAGE_URL, ..., replaceHistory = false);
persistentStorage.set(CONTEXT, ..., replaceHistory = true);
```

Because of our implementation oof batched updates, all three will be set
at the same time with replaceHistory = true (Even if the first two were
applied with false).

Our implementation of query rules applies the query params to
the url right before a search. This is at the same time query is
applied. Thus, we should keep replaceHistory to match what query is
doing (replaceHistory = false).

This puts in a problem when we land on an experience with a `query`
query param. Because replaceHistory must be false and because we set
context/referrerPageUrl before any search, we push another state
when we land on this page. This makes it so you cannot go backwards in
browser history from this page. (See test example below).

By only updating the URL when query params are different, we do not push
that extra state on page land. This is also useful for avoiding pushing
extra unnecessary state.

J=SPR-2554
TEST=manual

Test that you can make three queries in succession and you are not stuck
in the second query. For example.

Land on empty universal.
Search for virginia
Search for all
Search for christian

If you go back in your browser history twice, you should be back on the
`virginia` search. Not the `all` search.

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -59,3 +59,24 @@ export function addParamsToUrl (url, params = {}) {
 export function urlWithoutQueryParamsAndHash (url) {
   return url.split('?')[0].split('#')[0];
 }
+
+/**
+ * returns if two SearchParams objects have the same key,value entries
+ * @param {SearchParams} params1
+ * @param {SearchParams} params2
+ * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
+ */
+export function equivalentParams (params1, params2) {
+  if (params1.entries().length !== params2.entries().length) {
+    return false;
+  }
+  for (let entry of params1.entries()) {
+    if (!params2.has(entry[0])) {
+      return false;
+    }
+    if (entry[1] !== params2.get(entry[0])) {
+      return false;
+    }
+  }
+  return true;
+}
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -70,11 +70,8 @@ export function equivalentParams (params1, params2) {
   if (params1.entries().length !== params2.entries().length) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


oshi97:
"// batch update calls across components to avoid updating the url too much"
beautiful

src/core/utils/urlutils.js
```diff
@@ -59,3 +59,24 @@ export function addParamsToUrl (url, params = {}) {
...
+ * returns if two SearchParams objects have the same key,value entries
+ * @param {SearchParams} params1
+ * @param {SearchParams} params2
+ * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
+ */
+export function equivalentParams (params1, params2) {
+  if (params1.entries().length !== params2.entries().length) {
+    return false;
+  }
+  for (let entry of params1.entries()) {
```
oshi97: we could probably use Array.every() to be a little more concise here

creotutar: Native `URLSearchParams` returns an iterator (not an Array). This would mean we would need to cast it into an Array first. I'm not sure that work is beneficial for conciseness. What do you think?

oshi97: Oh okay that's a good point. I think you can still make it more concise by removing the 

```js
if (!params2.has(entry[0])) {
  return false;
}
```

if params2 does not have the key entry[0] it will return null, so since entry[1] will always be a string we can skip that check. (nits: maybe we can also change entry to be [key, value], and make it const instead of let?)

oshi97 (approved):

PR #968, opened 2020-07-13
circleci: Remove max-age cache for canary/latest

creotutar:
When we make a change to develop, we want the changes to reflect
immediately, even if we had already loaded canary/latest on a page.

Change to no-cache instead of a large max-age so that a page must
re-validate the context before using a cached version.

J=SPR-2549
TEST=manual

circleci config validate


oshi97 (approved):
do we need to specify cache-control since the default is no-cache?

oshi97 (approved):

PR #969, opened 2020-07-13
VerticalResults: add resultsCountTemplate config option

oshi97:
This commit allows users to specify a resultsCountTemplate
in VerticalResults, which lets them specify a custom results count

J=SPR-2619
TEST=manual

test that I can specify a custom results count template, using variables 
passed to the main template, and with arbitrary html tags (I tested `<button>, <table>, and <div>`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,8 +556,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
-  // Optional, custom text for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
-  resultsCountText: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
+  // Optional, a custom template for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount.
+  resultsCountTemplate: '<div>{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}</div>',
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, the card used to display each individual result, see the Cards section for more details,
--- a/README.md
+++ b/README.md
@@ -556,8 +556,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
-  // Optional, a custom template for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
-  resultsCountTemplate: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
+  // Optional, custom text for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
+  resultsCountTemplateString: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, the card used to display each individual result, see the Cards section for more details,
--- a/README.md
+++ b/README.md
@@ -556,6 +556,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
+  // Optional, a custom template for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
+  resultsCountTemplate: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, the card used to display each individual result, see the Cards section for more details,
--- a/README.md
+++ b/README.md
@@ -557,7 +557,7 @@ ANSWERS.addComponent('VerticalResults', {
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
   // Optional, custom text for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
-  resultsCountTemplateString: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
+  resultsCountText: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, the card used to display each individual result, see the Cards section for more details,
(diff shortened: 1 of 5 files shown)


src/ui/templates/results/resultsheader.hbs
```diff
@@ -14,11 +14,15 @@
   {{#if (and _config.showResultCount resultsCount)}}
     <div class="yxt-ResultsHeader-resultsCount" aria-label="{{resultsCountStart}} through {{resultsCountEnd}} of {{resultsCount}}">
       <div aria-hidden="true">
-        <span class="yxt-ResultsHeader-resultsCountStart">{{resultsCountStart}}</span>
-        <span class="yxt-ResultsHeader-resultsCountDash">-</span>
-        <span class="yxt-ResultsHeader-resultsCountEnd">{{resultsCountEnd}}</span>
-        <span class="yxt-ResultsHeader-resultsCountOf">of</span>
-        <span class="yxt-ResultsHeader-resultsCountTotal">{{resultsCount}}</span>
+        {{#if _config.resultsCountTemplate}}
```
oshi97: I thought about whether we want to override the entire "yxt-ResultsHeader-resultsCount" div or just this portion, but I think the use case here is mostly for overriding the text. We probably still want to let them add arbitrary html in case they have some specific styling in mind.

src/ui/templates/results/resultsheader.hbs
```diff
@@ -14,11 +14,17 @@
   {{#if (and _config.showResultCount resultsCount)}}
     <div class="yxt-ResultsHeader-resultsCount" aria-label="{{resultsCountStart}} through {{resultsCountEnd}} of {{resultsCount}}">
       <div aria-hidden="true">
```
creotutar: Not specific to this change, but why is this aria-hidden=true? It seems like this is always shown and information for assistive technology. 

Should this still be true if we are allowing this template to be changed as config?

creotutar: I think we will want a class to target the innermost parent of either custom resultsCountText or default. That way if we need to make a change that affects both, we can using their most direct parent. Is there a reason we have this classless aria-hidden div?

oshi97: Basically, we wanted to override the screen reader text. The goal was to have it only read the custom text we gave it (the aria label) and ignore the text in the spans. What's the best way to do this?

README.md
```diff
@@ -556,6 +556,8 @@ ANSWERS.addComponent('VerticalResults', {
   maxNumberOfColumns: 1,
   // Optional, whether to display the total number of results, default true
   showResultCount: true,
+  // Optional, custom text for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount with the double bracket notation e.g. {{resultsCount}}.
+  resultsCountText: '{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}',
```
creotutar: it seems weird to me that this is classified as a text rather than a template (as it gets compiled). are we 100% about the name `resultsCountText`?

(others are called template like noResults, etc, but I understand those are actually changing the entire component template)

creotutar: ah i read the Slack thread, it seems "plain string" is what we are going for

oshi97: I originally called this resultsCountTemplateString. No, I'm not 100% sure, but since Tom and Rose have decided we don't want to let people pass in arbitrary html, it's difficult to use the word template somewhere since we then have to add additional language explaining that it's not really a template since you can't add random html.

oshi97: Just updated the PR so this isn't confusing. Tom and I discussed off github that we don't want people to insert html here (which I originally thought was functionality we wanted). Sorry I didn't get a chance to do this earlier I was looking at some issues (which turned out to not be issues) with the 1.4.3 and 1.3.3 cut.

src/ui/templates/results/resultsheader.hbs
oshi97: @tmeyer2115 convinced me that we don't want to let them add random html here (also confirmed with rose)

src/ui/templates/results/resultsheader.hbs
creotutar: It seems aria-label does not do anything for the div element (as screen-readers focus on interactable elements not static elements)

creotutar: What we have now SGTM, I would be careful about classless divs, though, for easy parent targeting.

creotutar (approved):
👍 👍 👍 LGTM

tmeyer2115 (commented):
So, we're not letting them inject HTML, but they are having to provide us a HBS template. Thinking about this some more, I believe this kind of hybrid approach could cause confusion. The interface we have suggests to them that they can inject whatever HTML they want, because the string looks very similar to one they'd provide when overriding a component's template. 

I think we can make this clearer by accepting something that doesn't look like an HBS template, something more akin to a JS template literal. For instance, something like:

resultsCountText: '${resultCountStart} - ${resultCountEnd} of ${totalPages})'

 I don't think we should rely on HBS to do the variable substitution for us. 

I'm definitely not opposed to allowing McGin & co. to override the DOM for the ResultsCount. That may ultimately be what they want (so they can have different styling for the different bits of text). But, as you and I discussed yesterday, resultsCountText is not the best avenue for doing this. I think what we want is a nice, general way to override the template for a component's child component. @rosiegrant, do you think what we have here is sufficient or should we instead prioritize overriding a child component's template?

src/ui/templates/results/resultsheader.hbs
```diff
@@ -12,6 +12,9 @@
 
 {{#*inline "resultscount"}}
   {{#if (and _config.showResultCount resultsCount)}}
+  {{#if _config.resultsCountTemplate}}
```
tmeyer2115: I think we should indent this nested if block. Having it at the same level of indentation as the conditional on line 14 makes it seem like this is not nested.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -207,5 +207,9 @@ export default class HandlebarsRenderer extends Renderer {
       return getInverted ? self.SafeString(highlightedVal.getInverted())
         : self.SafeString(highlightedVal.get());
     });
+
+    self.registerHelper('renderTemplate', function (template) {
```
tmeyer2115: Do we need this helper? It's a bit weird to have a HBS helper that compiles and renders an arbitrary HBS template. That seems contrary to the conventional flow of things. Is there a way to avoid this? Could the results header component have something like:

this._compiledCountTemplate = this._renderer.compile(this._resultsCountTemplate);

Then in the HBS we'd have:

{{#if compiledCountTemplate}}
  {{ > compiledCountTemplate }}
{{/if}}

oshi97: Sure it probably doesn't make sense to recompile the template every time

src/ui/components/results/resultsheadercomponent.js
```diff
@@ -44,6 +44,12 @@ export default class ResultsHeaderComponent extends Component {
      */
     this.resultsLength = data.resultsLength || 0;
 
+    /**
+     * The compiled custom results count template, if the user specifies one.
+     * @type {Function}
+     */
+    this._compiledResultsCountTemplate = this._renderer.compile(this._config.resultsCountTemplate);
```
tmeyer2115, line 51: If no resultsCountTemplate is provided, does this compile to undefined or ''?

oshi97: when resultsCountTemplate is not provided, resultsCountTemplate will default to ""
in that case, this compiled template will be a function that returns ""

tmeyer2115 (approved):

PR #974, opened 2020-07-15
persistentstorage: Remove persistent storage history batch updates

creotutar:
* persistentstorage: Remove persistent storage history batch updates

The persistent storage batching has an interesting bug where you cannot
have updates to history with BOTH replaceHistory = false and
replaceHistory = true within a batch. That is, if you called

updateHistory(replaceHistory = true);
updateHistory(replaceHistory = true);
updateHistory(replaceHistory = false);

The batch would update the URL with replaceHistory = false (the last
value to be requested). This is a problem for query rules, where we
always want to replace the URL with the query params (and not add to the
history stack).

Updating the history stack was causing problems with going back in
history. Because we add query params on load with a search (like
referrerPageUrl), we get stuck on a page when we go back in history to a
page without a referrerPageUrl, it gets re-added on load, and if we go
back, we repeat the cycle.

This fix fixes the backspacing problem. There should be no changes
associated with history state.

J=SPR-2554
TEST=manual

Test that our sequential usage of URL params does not affect the normal
flow of searching.

On a local Jambo site,
Make a search A
Make a search B
In Chrome console ANSWERS.setContext({'state': 'hi'});
Make a search C
Go back in history
Go back in history
Make sure you can reach search A again and before search A.

From a redirectUrl searchbar pointing to answers page A
Make a search on the searchbar
Click search
Takes you to page A
Make sure you can go back in history to the redirectUrl searchbar.

Assert facets and pagination interactions (commonly add to the browser
history) work as intended with back and forward navigation.

* Fix equivalentParams to use correct length function

Before, equivalent params was trying to use the length function on an
iterator which will always return undefined. Update the params entries
to arrays to check for this length, instead of an undefined iterator
length.

J=SPR-2554
TEST=manual

Check that equivalentParams(params1, params2)
params1 = []
params2 = [{key1: val1}]

Will return false instead of true

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -67,7 +67,10 @@ export function urlWithoutQueryParamsAndHash (url) {
  * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
  */
 export function equivalentParams (params1, params2) {
-  if (params1.entries().length !== params2.entries().length) {
+  const entries1 = Array.from(params1.entries());
+  const entries2 = Array.from(params2.entries());
+
+  if (entries1.length !== entries2.length) {
     return false;
   }
   for (const [key, val] of params1.entries()) {
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -214,11 +214,6 @@ export default class SearchComponent extends Component {
 
   onCreate () {
     if (this.query != null && !this.redirectUrl) {
-      this.core.persistentStorage.set(
-        StorageKeys.REFERRER_PAGE_URL,
-        this.core.globalStorage.getState(StorageKeys.REFERRER_PAGE_URL),
-        true
-      );
       this.core.setQuery(this.query);
     }
   }
--- a/tests/core/storage/persistentstorage.js
+++ b/tests/core/storage/persistentstorage.js
@@ -7,10 +7,8 @@ describe('adding and removing data', () => {
 
   beforeEach(() => {
     delete global.window.location;
-    global.window = Object.create(global.window);
     global.window.location = {};
     global.window.location.search = '';
-    global.window.history = {};
 
     updateCb = jest.fn();
     storage = new PersistentStorage({ updateListener: updateCb });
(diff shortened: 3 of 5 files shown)


tests/core/storage/persistentstorage.js
```diff
@@ -6,11 +6,19 @@ describe('adding and removing data', () => {
   let updateCb;
 
   beforeEach(() => {
+    delete global.window.location;
```
creotutar, line 9: This is necessary to make changes to window.location.search within tests, as they are not natively changeable

src/core/utils/urlutils.js
```diff
@@ -67,7 +67,10 @@ export function urlWithoutQueryParamsAndHash (url) {
  * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
  */
 export function equivalentParams (params1, params2) {
-  if (params1.entries().length !== params2.entries().length) {
+  const entries1 = Array.from(params1.entries());
+  const entries2 = Array.from(params2.entries());
+
+  if (entries1.length !== entries2.length) {
```
tmeyer2115, line 73: Was this calculation not working correctly without the array conversion?

src/ui/components/search/searchcomponent.js
```diff
@@ -214,11 +214,6 @@ export default class SearchComponent extends Component {
 
   onCreate () {
     if (this.query != null && !this.redirectUrl) {
-      this.core.persistentStorage.set(
-        StorageKeys.REFERRER_PAGE_URL,
-        this.core.globalStorage.getState(StorageKeys.REFERRER_PAGE_URL),
-        true
-      );
```
tmeyer2115, line 221: Is this logic handled in core now?

src/core/utils/urlutils.js
oshi97: Yeah, apparently params.entries() returns an iterator not an array, which doesn't have a length attribute (so was always undefined)

creotutar: That is right. That's on me. Created SPR-2663 to create unit tests for our urlutils function suite.

src/ui/components/search/searchcomponent.js
creotutar: Yes, this was to band-aid the problem of not being able to batch different replaceHistory calls. It was essentially setting with replaceHistory=true before it ever got to a batch that had replaceHistory=false.

tmeyer2115 (approved):

oshi97 (approved):

PR #975, opened 2020-07-16
gulp: Add IIFE template bundle for dist

creotutar:
In response to clients having trouble using the current UMD template
bundle that is imported automatically, we are offering a non-UMD
alternative bundle (IIFE). One can use it by importing

```html
<script src="/answerstemplates-iife.compiled.min.js"></script>
```

Then you should specify the template bundle in ANSWERS.init with:

```js
templateBundle: TemplateBundle
```

Note that this is different from our standard, because we are importing
a different format of module

```js
templateBundle: TemplateBundle.default
```

J=SPR-2555
TEST=manual

Test on a local Jambo site that the expereince is showing templates
correctly (if templates are not imported correctly then you would not
see an experience).

Test to see both outputs are valid in a local site after a `gulp`
answerstemplates-iife.compiled.min.js / TemplateBundle
answerstemplates.compiled.min.js / TemplateBundle.default

Test to see `gulp dev` correctly builds a template bundle and changing
an hbs template in the library triggers another successful template bundle
build.

### DIFF
--- a/conf/gulp-tasks/templates.gulpfile.js
+++ b/conf/gulp-tasks/templates.gulpfile.js
@@ -1,6 +1,7 @@
 const { parallel, series, src, dest, watch } = require('gulp');
 
 const fs = require('fs');
+const del = require('del');
 const insert = require('rollup-plugin-insert');
 
 const rollup = require('gulp-rollup-lightweight');
@@ -18,6 +19,7 @@ const wrap = require('gulp-wrap');
 
 const source = require('vinyl-source-stream');
 
+const filenamePrecompiled = 'answerstemplates.precompiled.min.js';
 const filenameUMD = 'answerstemplates.compiled.min.js';
 const filenameIIFE = 'answerstemplates-iife.compiled.min.js';
 
@@ -58,7 +60,7 @@ function precompileTemplates () {
         return declare.processNameByPath(path, '').replace('.', '/');
       }
     }))
-    .pipe(concat('answerstemplates.precompiled.min.js'))
+    .pipe(concat(filenamePrecompiled))
     .pipe(wrap({ src: './conf/templates/handlebarswrapper.txt' }))
     .pipe(dest('dist'));
 }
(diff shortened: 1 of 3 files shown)


conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -100,5 +126,11 @@ function watchTemplates (cb) {
   }, series(precompileTemplates, bundleTemplates));
 }
```
tmeyer2115, line 135: Should we update the watchTemplates task as well?

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -55,19 +58,15 @@ function precompileTemplates () {
         return declare.processNameByPath(path, '').replace('.', '/');
       }
     }))
-    .pipe(concat('answerstemplates.compiled.min.js'))
+    .pipe(concat('answerstemplates.precompiled.min.js'))
```
tmeyer2115: For my own curiosity, what's this and how is it generated?

conf/gulp-tasks/templates.gulpfile.js
creotutar: Very good call, updating

conf/gulp-tasks/templates.gulpfile.js
creotutar: It's precompiling the templates (adding js wrapper code, compiling the handlebars) and then concatenating all into one file. This is the source file for both of the bundles, so I made it so precompileTemplates only happens once on a `gulp`. It is generated into a file `answerstemplates.precompiled.min.js` in the dist folder that is read by both the bundle types. This is just common js code that is not transpiled into a specific bundle type.

tmeyer2115: Is this something that will appear in the CDN now? Without a specific bundle type, is this an asset someone could actually use? If not, I'm not sure we want it in the CDN.

alexisgrow (commented):
🔥 LGTM, but want to confirm that the question is resolved about precompiled templates being in the CDN. It seems that before now, if there was a file for the precompiled templates it was likely being overwritten since we re-used the same file name.

tmeyer2115 (approved):

PR #976, opened 2020-07-16
Add left and right border to map on universal page

sawong99:
Add left and right border to map on universal page. border-left and border-right are set to var(--yxt-results-border).

J=SPR-2622

TEST=manual

Tested on local site to ensure left and right border were added. Tested on IE11 to make sure changes apply in IE11 as well.


alexisgrow (approved):
lg 🔥

PR #980, opened 2020-07-21
urlutils: Add urlutils unit tests

creotutar:
Make small changes urlutils functions to make them testable (outside of
the global variable window). Make tests that their main usages and edge
cases.

J=SLAP-182
TEST=npm run test

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -53,6 +53,11 @@ export function replaceUrlParams (url, params = new SearchParams()) {
   return url.split('?')[0] + '?' + params.toString();
 }
 
+/**
+ * Returns the given url without query params and hashes
+ * @param {string} url Full url e.g. https://yext.com/?query=hello#Footer
+ * @returns {string} Url without query params and hashes e.g. https://yext.com/
+ */
 export function urlWithoutQueryParamsAndHash (url) {
   return url.split('?')[0].split('#')[0];
 }
--- a/tests/core/utils/urlutils.js
+++ b/tests/core/utils/urlutils.js
@@ -5,7 +5,7 @@ import {
   getCachedLiveApiUrl,
   getKnowledgeApiUrl,
   getAnalyticsUrl,
-  addParamsToUrl,
+  replaceUrlParams,
   urlWithoutQueryParamsAndHash,
   equivalentParams
 } from '../../../src/core/utils/urlutils';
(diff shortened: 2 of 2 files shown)


src/core/utils/urlutils.js
```diff
@@ -43,19 +43,26 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
...
- * @param {string} url
+ * passed in params, appended to it.
+ * @param {string} url Full url, if there are query params in the url, they are removed
  * @param {Object} params
+ * @param {SearchParams} currentParams The params of the current URL, kept in the final url
  * @returns {string}
  */
-export function addParamsToUrl (url, params = {}) {
-  const urlParams = new SearchParams(window.location.search.substring(1));
+export function addParamsToUrl (url, params = {}, currentParams) {
```
tmeyer2115: This new function signature is a bit confusing. It feels like params and currentParams are doing pretty similar things. Both are for specifying params that will end up in the final URL. Their differentiator is the type? currentParams is typeof SearchParams and params is an object. 

Instead of this, could we mock new SearchParams in Jest? I think that would make the unit tests a bit more accurate as well, since we'd be ensuring that calls to this method invoke new SearchParams properly. All real calls to this method will be using new SearchParams and not passing in currentParams.

https://stackoverflow.com/questions/52886079/mocking-urlseachparams

creotutar: Ah, I agree, the function signature is confusing.

Originally, this was added so that this function is not always relying on a global variable (ie `window.location.search`). This is so we do not have to do [this hack](https://github.com/yext/answers/blob/master/tests/core/storage/persistentstorage.js#L9) to get that variable modifiable by the code.

I think ideally, this util function would not rely on the current URL params and instead just apply the params to the given URL. I did not want to refactor everything, in case things were relying on this fact, though. I think we set ourselves up for unknown side-effects if we are relying on this, though, because the URL changes with any call to persistentStorage. 

What do you think? Should we refactor the function and make it so that params is `SearchParams` instead and includes all of the current params (so that way it is easier to reproduce behavior and makes unit tests easier)? Or should that be another item and instead use hack above?

tests/core/utils/urlutils.js
```diff
@@ -0,0 +1,111 @@
...
+  it('checks when they have different param values', () => {
+    const params2 = new SearchParams('?query=hello&referrerPageUrl=');
+    const params3 = new SearchParams('?query=hello&referrerPageUrl=');
+    params3.set('referrerPageUrl', 'https%3A%2F%2Fwww.yext.com%2F');
+    expect(equivalentParams(params2, params3)).toEqual(false);
+    expect(equivalentParams(params3, params2)).toEqual(false);
+  });
+
+  it('checks when they are the exact same', () => {
+    const params2 = new SearchParams('query=all&referrerPageUrl=&Facets.filterbox.filter0=%5B%5D&Facets.filterbox.filter1=%5B%5D&Facets.filterbox.filter2=%5B%5D&Facets.filterbox.filter3=%5B%5D&Facets.filterbox.filter4=%5B%5D&Facets.filterbox.filter5=%5B%5D&Facets.filterbox.filter6=%5B%5D&Facets.filterbox.filter7=%5B%5D&Facets.filterbox.filter8=%5B%5D&Facets.filterbox.filter9=%5B%5D&Facets.filterbox.filter10=%5B%5D&context=%7B"state"%3A"hx"%7D&tabOrder=index.html%2CKM%2Cevents%2Cfaq%2Cjob%2Clinks%2Cpeople%2Crestaurant');
```
oshi97: nit but can we use a variable for the copied params strings (here and also above). maybe we can give the SearchParams instances slightly more descriptive names as well?

src/core/utils/urlutils.js
tmeyer2115: I think the refactor you're proposing would be good! This method takes in a baseUrl and adds whatever is in params to it. If you want all the current URL params added, ensure they are part of params. There are no surprises that way, we add exactly what you ask us to.

 If it's easier for now, I'd say lets keep the method as it was (with the previous signature) and add the hack(s) to the Jest tests. We can make an item for the refactor.

tests/core/utils/urlutils.js
```diff
@@ -0,0 +1,111 @@
...
+  it('adds params when currentParams is undefined', () => {
+    expect(addParamsToUrl('https://yext.com/', { referrerPageUrl: '' }))
+      .toEqual('https://yext.com/?referrerPageUrl=');
+  });
+
+  it('adds params when params already exist', () => {
+    expect(addParamsToUrl(
+      'https://yext.com/',
+      { query: 'all', referrerPageUrl: '' },
+      new SearchParams('?page=10&facets=true')
```
alexisgrow: nit: Since SearchParams takes "window.location.search.substring(1)" in the urlutils.js code, should this have the question mark prefix? Functionally it may not make a difference?

src/core/utils/urlutils.js
creotutar: This has been refactored to help with another item!

tests/core/utils/urlutils.js
creotutar: These should be functionally the same! URLSearchParams will strip any leading `?`.

oshi97 (approved):
!!!

alexisgrow (approved):
LGTM

PR #981, opened 2020-07-22

sawong99:
Added tab order calculations in core.js instead of navigationcomponent.js so that both navigationcomponent.js and verticalresultscomponent.js can access the functions. Updated getVerticalURL in verticalresultscomponent.js so that it adds the tabOrder to the end of the URLs for the View All buttons. Initially, tab order was not preserved when clicking View All from the universal page. Now, tab order is preserved.

J=SLAP-445

TEST=manual


creotutar:
@sawong99 It feels like this isn't the right base branch or that we started vertical-pages on the wrong branch

Ideally the PR would only add the changes described in the commit message -- could we cherry-pick your specific commit into a branch based off of hotfix/v1.4.4?

PR #982, opened 2020-07-23

sawong99:
Added tab order calculations in core.js instead of navigationcomponent.js so that both navigationcomponent.js and verticalresultscomponent.js can access the functions. Updated getVerticalURL in verticalresultscomponent.js so that it adds the tabOrder to the end of the URLs for the View All buttons. Initially, tab order was not preserved when clicking View All from the universal page. Now, tab order is preserved.

J=SLAP-445

TEST=manual


creotutar:
This seems to have a lot of conflicts with my item (removing facets from URLs). Let's coordinate on these items so we don't get a ton of merge conflicts 👯

src/core/core.js
```diff
@@ -9,6 +9,8 @@ import QuestionSubmission from './models/questionsubmission';
 import AnalyticsEvent from './analytics/analyticsevent';
 import FilterRegistry from './filters/filterregistry';
 
+import SearchParams from '../../src/ui/dom/searchparams';
```
creotutar: you shouldn't need to leave and include src again, core is already in src
e.g. `../src/ui/dom/searchparams`

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -378,9 +377,9 @@ export default class NavigationComponent extends Component {
     return el.closest(closestElSelector);
   }
 
-  getUrlParams () {
-    return new SearchParams(window.location.search.substring(1));
-  }
+  // getUrlParams () {
```
creotutar: can we remove all comments?

src/core/core.js
```diff
@@ -452,4 +454,87 @@ export default class Core {
   on (evt, moduleId, cb) {
     return this.globalStorage.on(evt, moduleId, cb);
   }
+
+  getUrlParams () {
```
creotutar: it /feels/ to me like we need our own `ui/tools/urlutils.js` for these functions. it doesnt feel like any of this code is related to the core itself (it references UI classes, is only used in UI components).

i've been thinking about this further and it seems like most of our new `core/utils/urlutils` (that deal with search params) should belong in the `ui` folder and not the `core` folder.

what do you think?

tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -123,7 +127,7 @@ describe('vertical results component', () => {
     });
 
     it('if unset defaults to vertical key', () => {
-      expect(component.getVerticalURL()).toEqual('key.html?query=my-query&otherParam=123');
+      expect(component.getVerticalURL()).toEqual('defaultTabUrl');
```
creotutar: i would think we want to keep the expected the output instead of using this `defaultTabUrl`. is there a reason we changed all of them?

src/core/core.js
```diff
@@ -452,4 +454,87 @@ export default class Core {
...
+        tabOrder.unshift(tabConfig);
+      } else {
+        tabOrder.push(tabConfig);
+      }
+    }
+
+    return tabOrder;
+  }
+
+  generateTabUrl (baseUrl, params, tabs, tabOrder) {
```
creotutar: it seems this generateTabUrl is different from the one removed from navigationcomponent. it looks like we added context/referrerPageUrl stuff (which should only belong in v1.5.0+). can we make sure all of these moved functions are the same as what they were before, except for necessary changes?

creotutar:
If v1.5.0 is already out, should this still be in a hotfix branch or a support branch? My guess is the latter.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +241,9 @@ export default class VerticalResultsComponent extends Component {
   getVerticalURL (data = {}) {
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
     const verticalURL = this._config.verticalURL || verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
-    return addParamsToUrl(verticalURL, { query: this.query });
+
+    let tabOrder = this.core.getDefaultTabOrder(this._verticalsConfig, this.core.getUrlParams());
```
creotutar: nit: const please

tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -123,7 +123,7 @@ describe('vertical results component', () => {
     });
 
     it('if unset defaults to vertical key', () => {
-      expect(component.getVerticalURL()).toEqual('key.html?query=my-query&otherParam=123');
+      expect(component.getVerticalURL()).toEqual('key.html?query=virginia&otherParam=123&tabOrder=');
```
creotutar: i would expect the query to still be `my-query`. is this not the case?

tests/ui/tools/urlutils.js
```diff
@@ -0,0 +1,107 @@
+import DOM from '../../../src/ui/dom/dom';
+import { generateTabUrl, mergeTabOrder, getDefaultTabOrder, getUrlParams } from '../../../src/ui/tools/urlutils';
+// The DOM doesn't exist within components in the JEST environment,
+// so we have to provide it to our DOM API properly.
+DOM.setup(
```
oshi97: We shouldn't need to setup the DOM, since none of the methods in urlutils.js do anything with it

src/ui/tools/urlutils.js
```diff
@@ -0,0 +1,64 @@
+import SearchParams from '../dom/searchparams';
+
```
oshi97: Sorry this is my fault for only seeing this now, but I think a lot of this logic could go into src/core/models/navigation. (This is the model that is stored in StorageKeys.NAVIGATION). We could do these calculations each time a new Navigation is created in SearchDataTransformer, which is the class that helps transform search data after a search. Then, components like the nav or vertical results components wouldn't need redo these calculations, or even have to know about any of this fancy logic, they would just ask for the tabOrder stored in globalStorage and that would be it.

oshi97 (commented):
Think we want to move some of the logic into src/core/models/navigation

oshi97:
Could you also ask Bryan whether this item is supposed to address the full tab order spec (also linked in the jira item)? https://docs.google.com/document/d/1MuVnxmP08xdQaxs8Ic0Q5jSkIJVylME5QLqv2-h--7E/edit#heading=h.91nix9nsma8a

rosiegrant:
hi! the only new behavior in the spec is: "User clicks “View More” on the Universal Search Page for a specific vertical section."

Bryan (rightfully) asked for a spec for this behavior since it was never really documented.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +242,25 @@ export default class VerticalResultsComponent extends Component {
...
+      tabOrder = mergeTabOrder(dataTabOrder, tabOrder, this._verticalsConfig);
+    }
+
+    const params = { query: this.query };
+    const urlParams = getUrlParams();
+    for (const paramKey in params) {
+      urlParams.set(paramKey, params[paramKey]);
+    }
+    return generateTabUrl(verticalURL, urlParams, tabOrder);
+    // return generateTabUrl(verticalURL, getUrlParams(), tabOrder);
```
cea2aj: Can we remove this commented-out return?

creotutar:
Did we decide to move forward with moving the functions to the navigation model in core? It doesn't look like that is reflected in the changed files.

I'm curious, though, why the URL generation should go into core when it's specific to the UI

cc @sawong99 @oshi97

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +242,24 @@ export default class VerticalResultsComponent extends Component {
   getVerticalURL (data = {}) {
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
     const verticalURL = this._config.verticalURL || verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
-    return addParamsToUrl(verticalURL, { query: this.query });
+
+    let tabOrder = getDefaultTabOrder(this._verticalsConfig, getUrlParams());
```
oshi97: Could we try to generalize all of this this logic into a helper method that generates the tabOrder param? We could then add it to verticalURL with addParamsToUrl (like before). Ideally, VerticalResults wouldn't contain any tabOrder logic. It also looks like we're not accounting for cases when verticalPages is specified in the nav component rather than in ANSWERS.init. It's very possible we don't need to handle this case, because it never worked before it's not a breaking change (and I think we've deprecated specifying verticalPages through the nav component anyways). I would double check with Rose that it's ok that we don't handle correctly merging the tabOrder with verticalPages if verticalPages is specified in the navigation component and not in ANSWERS.init.

src/core/utils/taborder.js
```diff
@@ -0,0 +1,64 @@
...
+    if (tabs[tabConfig] && tabs[tabConfig].isFirst) {
+      tabOrder.unshift(tabConfig);
+    } else {
+      tabOrder.push(tabConfig);
+    }
+  }
+  return tabOrder;
+}
+
+export function generateTabUrl (baseUrl, params, tabOrder) {
```
oshi97: Could we use addParamsToUrl instead, and remove this method? It looks like they do the same thing

oshi97: It looks like Christian renamed addParamsToUrl to replaceUrlParams so would want to rebase first

oshi97:
> Did we decide to move forward with moving the functions to the navigation model in core? It doesn't look like that is reflected in the changed files.
> 
> I'm curious, though, why the URL generation should go into core when it's specific to the UI
> 
> cc @sawong99 @oshi97

This is just what we've been doing, if we want to refactor it we probably want to do a full refactor for all of the src/core/*utils files. Also not sure if that's something we want, I was kind of envisioning that this tabOrder transformation would all happen within core, and would not be something any of the components would need to worry about. It's just the sdk is a huge hairball (at least with regards to these kinds of things) so we didn't really have a choice but to add this logic to the components for now.

src/core/utils/taborder.js
```diff
@@ -0,0 +1,64 @@
+import SearchParams from '../../ui/dom/searchparams';
```
tmeyer2115: Now that I have a bit more context on the change, I agree with @creotutar, this should probably live in the UI portion of the SDK. The Core is reserved for those parts of the SDK that aren't UI-specific. For instance, we have classes in Core for interacting with the various back-ends. We also keep our data models here. These helpers are much more related to the front-end, UI portion of the SDK. They are only used to help two Components render the proper links. 

Maybe we add this file under the src/ui/tools directory?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +242,24 @@ export default class VerticalResultsComponent extends Component {
...
+    let dataTabOrder = [];
+    if (this.core.globalStorage.getState(StorageKeys.NAVIGATION)) {
+      dataTabOrder = this.core.globalStorage.getState(StorageKeys.NAVIGATION).tabOrder;
+    }
+    // We want to persist the params from the existing URL to the new
+    // URLS we create.
+    if (tabOrder !== undefined && dataTabOrder !== undefined) {
+      tabOrder = mergeTabOrder(dataTabOrder, tabOrder, this._verticalsConfig);
```
tmeyer2115: We're passing in this._verticalsConfig as tabs. Are these things interchangeable?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +242,24 @@ export default class VerticalResultsComponent extends Component {
...
+    }
+
+    const params = { query: this.query };
+    const urlParams = getUrlParams();
+    for (const paramKey in params) {
```
tmeyer2115: This can be streamlined a bit. I don't think we need the loop here since params is an object with a single attribute. Can we just do:
urlParams.set('query', this.query) directly?

src/ui/components/results/verticalresultscomponent.js
oshi97: ok, sounds like we don't have to worry about doing weird magic in the sdk, but we will have to rejigger the theme to get the full benefits of your change. Still would prefer for this logic to be hidden from vertical results as much as possible, but think the overall idea is what we want!

src/ui/components/results/verticalresultscomponent.js
oshi97: I can answer this one, yes they are the same thing! I think one of Alexis' earliest contributions from a whiel back was updating tabConfig to verticalPages (and I think adding some additional properties? Don't remember exactly)

src/core/utils/taborder.js
creotutar: Just a heads up! This is getting a bit more beefed up in #983. Might be less merge work if we keep it in this PR

PR #983, opened 2020-07-23
urls: remove facets/pagination/filters from inter-experience links

creotutar:
We remove query parameters dealing with facets, pagination, or filters
from the inter-experience links. This is because this encoded
information should not persist as we move throughout the experience
(ie to another vertical).

For example: if you land on a vertical page with facets, these facets are
automatically added to the URL. When you try to navigate to another
tab, these query parameters persist (encoded within the nav URL) and
can lead to confusion when another vertical also has facets with similar
facet names.

re Implementation: Because name is a configurable option for components, and because
facets/filters use this as a basis for the persistentStorage key, we
must account for the different types of possible names. Thus, we store
the componentNames mapped to a component type, for easier access.

To avoid a circular dependency (Component -> urlUtils -> Component.type),
we split out the types into a different ComponentTypes class that allow
one source of truth for the component type for both the new urlutils function
and the components themselves.

J=SLAP-499
TEST=manual

Test that navigation links on a page with filters, pagination, and
facets do not contain those query parameters.
  - filter search
  - geolocation
  - facets
  - filterbox with specific CF filter

Test that View More links, Change Filter links on universal do not
contain those query parameters.

Test that Alternative Verticals vertical and universal links do not
contain those query parameters.

Add acceptance test for going from page with facets to universal
and to vertical page and make sure those do not contain the
facet/filter/pagination params.

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -1,5 +1,6 @@
 import { PRODUCTION, SANDBOX } from '../constants';
 import SearchParams from '../../ui/dom/searchparams';
+import StorageKeys from '../storage/storagekeys';
 
 /**
  * Returns the base url for the live api backend in the desired environment.
@@ -84,17 +85,52 @@ export function equivalentParams (params1, params2) {
 }
 
 /**
- * Removes params that begin with the given prefixes
- * Note: modifies the params parameter
+ * Creates a new params object without params that begin with the given prefixes
  * @param {SearchParams} params The parameters to remove from
  * @param {string[]} prefixes The prefixes of parameters to remove
+ * @return {SearchParams} A new instance of SearchParams without removed params
+ *   from the params parameter
  */
 export function removeParamsWithPrefixes (params, prefixes) {
-  for (const [key] of Array.from(params.entries())) {
+  const newParams = new SearchParams();
+  for (const [key, val] of params.entries()) {
+    let includeEntry = true;
     for (const prefix of prefixes) {
       if (key.startsWith(prefix)) {
-        params.delete(key);
+        includeEntry = false;
       }
     }
+    if (includeEntry) {
+      newParams.set(key, val);
+    }
   }
+  return newParams;
+}
+
+/**
+ * Removes parameters for filters, facets, sort options, and pagination
+ * from the params provided. This is useful for constructing
+ * inter-experience answers links.
+ * @param {SearchParams} params The parameters to remove from
+ * @param {function} getComponentNamesForComponentTypes Given string[]
+ *   component types, returns string[] component names for those types
+ * @return {SearchParams} Parameters that have filtered out params that
+ *   should not persist across the answers experience
+ */
+export function filterParamsForExperienceLink (
+  params,
+  getComponentNamesForComponentTypes
+) {
+  let prefixes = getComponentNamesForComponentTypes([
+    'Facets', 'FilterBox', 'FilterOptions', 'RangeFilter', 'DateRangeFilter', 'SortOptions'
+  ]);
+  prefixes = prefixes.concat(
+    getComponentNamesForComponentTypes(['GeoLocationFilter', 'FilterSearch'])
+      .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
+  );
+  prefixes.push(StorageKeys.FILTER);
+
+  const newParams = removeParamsWithPrefixes(params, prefixes);
+  newParams.delete(StorageKeys.SEARCH_OFFSET);
+  return newParams;
 }
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -126,7 +126,7 @@ export function filterParamsForExperienceLink (
     ComponentTypes.DATE_RANGE_FILTER,
     ComponentTypes.SORT_OPTIONS
   ];
-  const queryPrefixComponentComponentTypes = [
+  const queryPrefixComponentTypes = [
     ComponentTypes.GEOLOCATION_FILTER,
     ComponentTypes.FILTER_SEARCH
   ];
@@ -135,7 +135,7 @@ export function filterParamsForExperienceLink (
   ];
   prefixes = prefixes.concat(getComponentNamesForComponentTypes(prefixComponentTypes));
   prefixes = prefixes.concat(
-    getComponentNamesForComponentTypes(queryPrefixComponentComponentTypes)
+    getComponentNamesForComponentTypes(queryPrefixComponentTypes)
       .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
   );
 
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -1,6 +1,7 @@
 import { PRODUCTION, SANDBOX } from '../constants';
 import SearchParams from '../../ui/dom/searchparams';
 import StorageKeys from '../storage/storagekeys';
+import ComponentTypes from '../../ui/components/componenttypes';
 
 /**
  * Returns the base url for the live api backend in the desired environment.
@@ -121,11 +122,23 @@ export function filterParamsForExperienceLink (
   params,
   getComponentNamesForComponentTypes
 ) {
-  let prefixes = getComponentNamesForComponentTypes([
-    'Facets', 'FilterBox', 'FilterOptions', 'RangeFilter', 'DateRangeFilter', 'SortOptions'
-  ]);
+  const prefixComponentTypes = [
+    'FACETS',
+    'FILTER_BOX',
+    'FILTER_OPTIONS',
+    'RANGE_FILTER',
+    'DATE_RANGE_FILTER',
+    'SORT_OPTIONS'
+  ].map((component) => {
+    return ComponentTypes[component];
+  });
+  const queryPrefixComponentComponentTypes = [
+    ComponentTypes['GEOLOCATION_FILTER'],
+    ComponentTypes['FILTER_SEARCH']
+  ];
+  let prefixes = getComponentNamesForComponentTypes(prefixComponentTypes);
   prefixes = prefixes.concat(
-    getComponentNamesForComponentTypes(['GeoLocationFilter', 'FilterSearch'])
+    getComponentNamesForComponentTypes(queryPrefixComponentComponentTypes)
       .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
   );
   prefixes.push(StorageKeys.FILTER);
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -118,28 +118,26 @@ export function filterParamsForExperienceLink (
   params,
   getComponentNamesForComponentTypes
 ) {
-  const prefixComponentTypes = [
+  const componentTypesToExclude = [
     ComponentTypes.FACETS,
     ComponentTypes.FILTER_BOX,
     ComponentTypes.FILTER_OPTIONS,
     ComponentTypes.RANGE_FILTER,
     ComponentTypes.DATE_RANGE_FILTER,
-    ComponentTypes.SORT_OPTIONS
-  ];
-  const queryPrefixComponentTypes = [
+    ComponentTypes.SORT_OPTIONS,
     ComponentTypes.GEOLOCATION_FILTER,
     ComponentTypes.FILTER_SEARCH
   ];
-  let prefixes = [
-    StorageKeys.FILTER
-  ];
-  prefixes = prefixes.concat(getComponentNamesForComponentTypes(prefixComponentTypes));
-  prefixes = prefixes.concat(
-    getComponentNamesForComponentTypes(queryPrefixComponentTypes)
-      .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
-  );
+  let paramsToFilter = componentTypesToExclude.flatMap(type => {
+    let params = getComponentNamesForComponentTypes([type]);
+    if (type === ComponentTypes.GEOLOCATION_FILTER || type === ComponentTypes.FILTER_SEARCH) {
+      params = params.map(param => `${StorageKeys.QUERY}.${param}`);
+    }
+    return params;
+  });
+  paramsToFilter = paramsToFilter.concat([StorageKeys.FILTER]);
 
-  const newParams = removeParamsWithPrefixes(params, prefixes);
+  const newParams = removeParamsWithPrefixes(params, paramsToFilter);
   newParams.delete(StorageKeys.SEARCH_OFFSET);
   return newParams;
 }
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -86,7 +86,8 @@ export function equivalentParams (params1, params2) {
 }
 
 /**
- * Creates a new params object without params that begin with the given prefixes
+ * Creates a copy of the provided {@link SearchParams}, with the specified
+ * attributes filtered out
  * @param {SearchParams} params The parameters to remove from
  * @param {string[]} prefixes The prefixes of parameters to remove
  * @return {SearchParams} A new instance of SearchParams without removed params
@@ -95,12 +96,7 @@ export function equivalentParams (params1, params2) {
 export function removeParamsWithPrefixes (params, prefixes) {
   const newParams = new SearchParams();
   for (const [key, val] of params.entries()) {
-    let includeEntry = true;
-    for (const prefix of prefixes) {
-      if (key.startsWith(prefix)) {
-        includeEntry = false;
-      }
-    }
+    const includeEntry = prefixes.every(prefix => !key.startsWith(prefix));
     if (includeEntry) {
       newParams.set(key, val);
     }
@@ -110,7 +106,7 @@ export function removeParamsWithPrefixes (params, prefixes) {
 
 /**
  * Removes parameters for filters, facets, sort options, and pagination
- * from the params provided. This is useful for constructing
+ * from the provided {@link SearchParams}. This is useful for constructing
  * inter-experience answers links.
  * @param {SearchParams} params The parameters to remove from
  * @param {function} getComponentNamesForComponentTypes Given string[]
@@ -123,25 +119,25 @@ export function filterParamsForExperienceLink (
   getComponentNamesForComponentTypes
 ) {
   const prefixComponentTypes = [
-    'FACETS',
-    'FILTER_BOX',
-    'FILTER_OPTIONS',
-    'RANGE_FILTER',
-    'DATE_RANGE_FILTER',
-    'SORT_OPTIONS'
-  ].map((component) => {
-    return ComponentTypes[component];
-  });
-  const queryPrefixComponentComponentTypes = [
-    ComponentTypes['GEOLOCATION_FILTER'],
-    ComponentTypes['FILTER_SEARCH']
+    ComponentTypes.FACETS,
+    ComponentTypes.FILTER_BOX,
+    ComponentTypes.FILTER_OPTIONS,
+    ComponentTypes.RANGE_FILTER,
+    ComponentTypes.DATE_RANGE_FILTER,
+    ComponentTypes.SORT_OPTIONS
   ];
-  let prefixes = getComponentNamesForComponentTypes(prefixComponentTypes);
+  const queryPrefixComponentComponentTypes = [
+    ComponentTypes.GEOLOCATION_FILTER,
+    ComponentTypes.FILTER_SEARCH
+  ];
+  let prefixes = [
+    StorageKeys.FILTER
+  ];
+  prefixes = prefixes.concat(getComponentNamesForComponentTypes(prefixComponentTypes));
   prefixes = prefixes.concat(
     getComponentNamesForComponentTypes(queryPrefixComponentComponentTypes)
       .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
   );
-  prefixes.push(StorageKeys.FILTER);
 
   const newParams = removeParamsWithPrefixes(params, prefixes);
   newParams.delete(StorageKeys.SEARCH_OFFSET);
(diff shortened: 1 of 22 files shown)


src/core/utils/urlutils.js
```diff
@@ -80,3 +80,19 @@ export function equivalentParams (params1, params2) {
   }
   return true;
 }
+
+/**
+ * Removes params that begin with the given prefixes
```
creotutar: TODO: add unit tests

src/core/utils/urlutils.js
```diff
@@ -80,3 +80,19 @@ export function equivalentParams (params1, params2) {
...
+ * Note: modifies the params parameter
+ * @param {SearchParams} params The parameters to remove from
+ * @param {string[]} prefixes The prefixes of parameters to remove
+ */
+export function removeParamsWithPrefixes (params, prefixes) {
+  for (const [key, val] of Array.from(params.entries())) {
```
creotutar: Array.from because deleting in loop changes iterator pointer

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -472,8 +473,23 @@ export default class NavigationComponent extends Component {
     }
 
     // We want to persist the params from the existing URL to the new
-    // URLS we create.
+    // URLS we create, except for facets/filters/pagination
     params.set('tabOrder', this._tabOrder);
+
+    params.delete(StorageKeys.SEARCH_OFFSET);
+    params.delete(StorageKeys.FILTER);
+    removeParamsWithPrefixes(
```
creotutar: TODO add acceptance tests

src/ui/components/componentmanager.js
```diff
@@ -177,4 +186,14 @@ export default class ComponentManager {
   getActiveComponent (type) {
     return this._activeComponents.find(c => c.constructor.type === type);
   }
+
+  getComponentNamesForComponentType (type) {
```
creotutar: TODO: add javadoc

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -472,8 +473,23 @@ export default class NavigationComponent extends Component {
...
+      params,
+      this.componentManager.getComponentNamesForComponentTypes([
+        'Facets',
```
oshi97: missing Geolocation filter and FilterSearch here. Because the SDK is beautiful those actually use StorageKeys.GEOLOCATION and StorageKeys.FILTER, instead of the name they get in config. On top of that it will probably be a very long time before we refactor these because of url backwards compatibility. hooray for the sdk!

creotutar: OOF. What is the difference between geolocationcomponent.js and locationbiascomponent.js. I've been looking at the former (which also sadly does not use the name they get in the config).

oshi97: geolocationcomponent.js lets you either search in its own searchbar for a location to filter by, or you can click a button and it will use your current location.
image

locationbiascomponent is the location thing you see at the footer of every jambo page. It just says what your current location used in the sdk is, and lets you update it with the "update your location" button. I don't think this affects persistent storage at all but would have to double check
image

creotutar: Ah nice ok, that's what I thought. On my global search for `StorageKeys.GEOLOCATION`, I only see it set in globalStorage. I don't think it's something that needs to be removed from params for filters, right?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -472,8 +473,22 @@ export default class NavigationComponent extends Component {
     }
 
     // We want to persist the params from the existing URL to the new
-    // URLS we create.
+    // URLS we create, except for facets/filters/pagination
     params.set('tabOrder', this._tabOrder);
+
+    params.delete(StorageKeys.SEARCH_OFFSET);
+
+    let prefixes = this.componentManager.getComponentNamesForComponentTypes([
```
creotutar: This logic is matching the logic detailed in each component. That is, the persistentStorage key started with `this.name`. Should that change in any of these components, it would need to be changed here as well. I did not know if we wanted to invest time in this item for some new system `getStorageKey` so there is one source of truth for these URLs.

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -39,7 +39,7 @@ export default class AlternativeVerticalsComponent extends Component {
      * This gets updated based on the server results
      * @type {AlternativeVertical[]}
      */
-    this.verticalSuggestions = AlternativeVerticalsComponent._buildVerticalSuggestions(
+    this.verticalSuggestions = this._buildVerticalSuggestions(
```
creotutar: This was necessary to get the componentManager. This was done in another PR by Tom.

src/ui/components/navigation/navigationcomponent.js
oshi97: yup not needed for filters

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -123,6 +123,17 @@ export default class AlternativeVerticalsComponent extends Component {
       params.set(StorageKeys.REFERRER_PAGE_URL, referrerPageUrl);
     }
 
+    let prefixes = this.componentManager.getComponentNamesForComponentTypes([
```
creotutar: This entire block was going originally to be in a function in `urlutils`, until I got circular dependency errors because of a componentManager import. It seems we cannot access componentManager from urlutils and instead must rely on the class' property. I could add the function in urlutils to take `componentManager` as a parameter, but thought that was weird, given its specific to its implementation. What do y'all think?

I think ideally it'd be part of some persistentStorage revamp item

src/core/utils/urlutils.js
```diff
@@ -82,3 +82,19 @@ export function equivalentParams (params1, params2) {
...
   return true;
 }
+
+/**
+ * Removes params that begin with the given prefixes
+ * Note: modifies the params parameter
+ * @param {SearchParams} params The parameters to remove from
+ * @param {string[]} prefixes The prefixes of parameters to remove
+ */
+export function removeParamsWithPrefixes (params, prefixes) {
```
tmeyer2115, line 96: Instead of mutating the params object, could we create a copy and update that one? I always prefer to avoid mutation when possible (and when I remember, haha).

src/ui/components/componentmanager.js
```diff
@@ -41,6 +41,11 @@ export default class ComponentManager {
      * A local reference to the analytics reporter dependency
      */
     this._analyticsReporter = null;
+
+    /**
+     * A mapping from component types to component names, as these may be configured by a user
+     */
+    this._componentTypeToComponentNames = {};
```
tmeyer2115, line 48: This is a super useful thing to have 👍 in general.

src/ui/components/componentmanager.js
```diff
@@ -177,4 +186,24 @@ export default class ComponentManager {
...
+  getComponentNamesForComponentType (type) {
+    return this._componentTypeToComponentNames[type];
+  }
+
+  /**
+   * Returns a concatenated list of all names associated with the given component types
+   * @param {string[]} type The types of the component
+   * @returns {string[]} The component names for the component types
+   */
+  getComponentNamesForComponentTypes (types) {
```
tmeyer2115, line 195: If someone calls this with a single type, it's equivalent to a call to getComponentNamesForComponentType, right? Because of that, I'd say lets just keep this method and get rid of the other one. It doesn't feel like the other method is adding much.

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -472,8 +473,22 @@ export default class NavigationComponent extends Component {
...
+      'Facets', 'FilterBox', 'FilterOptions', 'RangeFilter', 'DateRangeFilter', 'SortOptions'
```
tmeyer2115: Could we use FacetsComponent.type, FilterOptionsComponent.type, etc. here? That way, if for some reason we change the type name, we don't have to change it here as well.

src/ui/components/results/alternativeverticalscomponent.js
tmeyer2115: I think it'd be okay to make a function in urlutils that takes in a componentManager instance and spits out the list of bad prefixes. We definitely don't want to duplicate this logic in multiple places. Actually, if it makes you feel better about the interface, your new urlutils method could just take in a function mapping component type to names. It could be totally agnostic about how that function is implemented or where it comes from. That would also make unit tests easier for it.

tests/acceptance/acceptancesuite.js
```diff
@@ -105,3 +106,65 @@ test(`Facets load on the page, and can affect the search`, async t => {
   actualResultsCount = await verticalResultsComponent.getResultsCountTotal();
   await t.expect(actualResultsCount).eql(initialResultsCount);
 });
+
+fixture`Experience links work as expected`
+  .before(setupServer)
+  .after(shutdownServer)
+  .page`http://localhost:9999/tests/acceptance/fixtures/html/facets`;
+
+test('Facets, pagination, and filters do not persist accross experience links', async t => {
```
tmeyer2115, line 115: Are we verifying that static filters and sorts are also not preserved in this test?

src/core/utils/urlutils.js
creotutar: Done, had it like that before but I was 50/50 on balancing

1. time to create a new SearchParams for every call
2. keeping function pure

Found a way to do it without have to copy over the array before doing any filtering

src/ui/components/componentmanager.js
creotutar: Done

tests/acceptance/acceptancesuite.js
creotutar: Added

src/core/utils/urlutils.js
```diff
@@ -82,3 +84,66 @@ export function equivalentParams (params1, params2) {
   }
   return true;
 }
+
+/**
+ * Creates a new params object without params that begin with the given prefixes
```
tmeyer2115: How about "Creates a copy of the provided {@link SearchParams}, with the specified attributes filtered out"? The phrasing "Creates a new params object without params..." is a bit confusing.

src/core/utils/urlutils.js
```diff
@@ -82,3 +84,66 @@ export function equivalentParams (params1, params2) {
...
+ * @param {SearchParams} params The parameters to remove from
+ * @param {string[]} prefixes The prefixes of parameters to remove
+ * @return {SearchParams} A new instance of SearchParams without removed params
+ *   from the params parameter
+ */
+export function removeParamsWithPrefixes (params, prefixes) {
+  const newParams = new SearchParams();
+  for (const [key, val] of params.entries()) {
+    let includeEntry = true;
+    for (const prefix of prefixes) {
```
tmeyer2115: Minor nit, but we can streamline this a bit:
includeEntry = prefixes.some(prefix => key.startsWith(prefix));
Whenever possible, we try and do things in the more functional style (a holdover from our Java days).

src/core/utils/urlutils.js
```diff
@@ -82,3 +84,66 @@ export function equivalentParams (params1, params2) {
...
+  const queryPrefixComponentComponentTypes = [
+    ComponentTypes['GEOLOCATION_FILTER'],
+    ComponentTypes['FILTER_SEARCH']
+  ];
+  let prefixes = getComponentNamesForComponentTypes(prefixComponentTypes);
+  prefixes = prefixes.concat(
+    getComponentNamesForComponentTypes(queryPrefixComponentComponentTypes)
+      .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
+  );
+  prefixes.push(StorageKeys.FILTER);
```
tmeyer2115: Is there anything we can do to simplify this code block a bit and enhance readability? One idea I have is to have:
const filterPrefixes = [
  ComponentTypes.Facets,
   ...,
 `${StorageKeys.QUERY}.${ComponentTypes.Filter_SEARCH}`,
  StorageKeys.FILTER
]
I think that could make this method a bit less complex. It's a bit difficult to follow all the maps, concats, etc. going on and what we're building is a relatively simple list.

src/ui/components/componenttypes.js
```diff
@@ -0,0 +1,13 @@
+/** @module */
+
+// TODO add all component types
```
tmeyer2115: Add a little JSDoc explaining what this enum is (don't need much, fairly self explanatory)

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -479,8 +480,13 @@ export default class NavigationComponent extends Component {
     }
 
     // We want to persist the params from the existing URL to the new
-    // URLS we create.
+    // URLS we create, except for facets/filters/pagination
     params.set('tabOrder', this._tabOrder);
-    return baseUrl + '?' + params.toString();
+    const filteredParams = filterParamsForExperienceLink(
+      params,
+      this.componentManager.getComponentNamesForComponentTypes.bind(this.componentManager)
```
tmeyer2115: Minor nit: let's use an inline function here, that will save us the 'this' binding and is a bit more readable (I think). types => this.componentManager.getComponentNamesForComponentTypes(types). 
Same thing for the other invocations of filterParamsForExperienceLink.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -237,9 +237,26 @@ export default class VerticalResultsComponent extends Component {
     if (!universalConfig.url) {
       return undefined;
     }
+    return this._getExperienceURL(universalConfig.url);
+  }
 
+  getVerticalURL (data = {}) {
+    const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
```
tmeyer2115: LL

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -237,9 +237,26 @@ export default class VerticalResultsComponent extends Component {
...
+    const verticalURL = this._config.verticalURL || verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
```
tmeyer2115: LL

tests/acceptance/acceptancesuite.js
```diff
@@ -105,3 +106,77 @@ test(`Facets load on the page, and can affect the search`, async t => {
...
+  await searchComponent.clearQuery();
+  await searchComponent.enterQuery('virginia');
+  await searchComponent.submitQuery();
+
+  const changeFiltersLink = await Selector('.yxt-ResultsHeader-changeFilters')
+    .nth(0).getAttribute('href');
+  await t.expect(changeFiltersLink).contains('referrerPageUrl');
+  await verifyCleanLink(changeFiltersLink);
+
+  const viewAllLink = await Selector('.yxt-ResultsHeader-changeFilters')
```
tmeyer2115: Is this the correct selector?

src/core/utils/urlutils.js
```diff
@@ -82,3 +84,66 @@ export function equivalentParams (params1, params2) {
...
+ * inter-experience answers links.
+ * @param {SearchParams} params The parameters to remove from
+ * @param {function} getComponentNamesForComponentTypes Given string[]
+ *   component types, returns string[] component names for those types
+ * @return {SearchParams} Parameters that have filtered out params that
+ *   should not persist across the answers experience
+ */
+export function filterParamsForExperienceLink (
+  params,
+  getComponentNamesForComponentTypes
```
oshi97, line 119: maybe this method can be moved to somewhere that already can see component manager, so we don't have to pass in the component manager method? I thought core.js would have a reference to component manager but apparently it doesn't, maybe it would make sense to add a reference there and then add this method to core too? or maybe we add a method to component manager that gets all of the prefixes we want? Maybe that makes some sense, since component manager is for managing component it can contain the logic for getting all of the persistent storage keys that we don't want in inter-experience links?

src/core/utils/urlutils.js
creotutar: I don't think that particular prefix map will work because we need to get the names of the component types (ie it would be `${StorageKeys.QUERY}.${getComponentNamesForComponentTypes(['ComponentTypes.Filter_SEARCH'])}` and similarly for all of the other types).

I can remove the map on the prefixes and just spell out all ComponentTypes.<Component> (and also structure the prefixes array a little different). Let me know if we want to simplify further.

I think the root cause is that we're manually recreating the persistentStorage key prefixes instead of sourcing from one place. IE I think these prefixes should be kept somewhere instead of requiring me to construct the component name and then the specific format of this name. However I'm not sure if this is a now thing or a refactor for later on.

src/core/utils/urlutils.js
creotutar: This is an interesting thought. Are you proposing that we move the entire function into componentManager or `getFilterPrefixes` maybe? It _feels_ like it is not the job of the component manager to generate these prefixes but I'm also not opposed to moving it.

Thoughts? @tmeyer2115 @alexisgrow

src/ui/components/componenttypes.js
```diff
@@ -0,0 +1,18 @@
+/** @module */
+
+/**
+ * The component types is a map that contains the types for all components
+ * given some unique key.
```
tmeyer2115: We've effectively created an enum here, so I think it's okay to use that terminology. "An enum listing the different Component types supported in the SDK"

src/core/utils/urlutils.js
tmeyer2115: Adding a `getFilterPrefixes` method to the ComponentManager seems outside its responsibility. Its role is to essentially be a registry of the components on the page. It adds them, deletes them, etc. The `getFilterPrefixes` method feels related to a different concern (the construction of cross-page links) and on a different level of abstraction.

src/core/utils/urlutils.js
```diff
@@ -82,3 +84,62 @@ export function equivalentParams (params1, params2) {
...
+  ];
+  prefixes = prefixes.concat(getComponentNamesForComponentTypes(prefixComponentTypes));
+  prefixes = prefixes.concat(
+    getComponentNamesForComponentTypes(queryPrefixComponentComponentTypes)
+      .map((name) => { return `${StorageKeys.QUERY}.${name}`; })
+  );
+
+  const newParams = removeParamsWithPrefixes(params, prefixes);
+  newParams.delete(StorageKeys.SEARCH_OFFSET);
+  return newParams;
```
tmeyer2115, line 142: I think we should try and streamline this method a bit. It is a bit difficult to read as is. One thing that would help readability is potentially some different variable names.

prefixComponentTypes -> componentTypesToExclude

componentTypesToExclude = [
    ComponentTypes.FACETS,
    ComponentTypes.FILTER_BOX,
    ComponentTypes.FILTER_OPTIONS,
    ComponentTypes.RANGE_FILTER,
    ComponentTypes.DATE_RANGE_FILTER,
    ComponentTypes.SORT_OPTIONS,
    ComponentTypes.GEOLOCATION_FILTER,
    ComponentTypes.FILTER_SEARCH
  ];

let paramsToFilter = componentTypesToExclude.map(type => {
    let typeFilter = getComponentNamesForComponentTypes(type)
    if (type === ComponentTypes.GEOLOCATION_FILTER || type === ComponentTypes.FILTER_SEARCH) {
       typeFilter === `${StorageKeys.QUERY}.${typeFilter}`;
    }
   return typeFilter;
});
paramsToFilter.push(StorageKeys.FILTER);

I think the above might be a bit more readable. There's only one concat and map and we don't have two partial lists that need to be run through getComponentNamesForComponentTypes, then re-combined.

src/ui/components/componenttypes.js
creotutar: For context, I was copying the registry jsdoc. I'll update

src/core/utils/urlutils.js
creotutar: So getComponentNamesForComponentTypes returns an array, as there can be many names, so we'd need to reduce otherwise we'd have an array of arrays.

I'm not sure how much this lends itself to being readable. What do you think?

```
  const componentTypesToExclude = [
    ComponentTypes.FACETS,
    ComponentTypes.FILTER_BOX,
    ComponentTypes.FILTER_OPTIONS,
    ComponentTypes.RANGE_FILTER,
    ComponentTypes.DATE_RANGE_FILTER,
    ComponentTypes.SORT_OPTIONS
    ComponentTypes.GEOLOCATION_FILTER,
    ComponentTypes.FILTER_SEARCH
  ];
  let prefixes = [
    StorageKeys.FILTER
  ];
  prefixes = prefixes.concat(
    componentTypesToExclude
    .reduce((currentPrefixes, type) => {
      let prefixesToAdd = getComponentNamesForComponentTypes([type]);
      if (type === ComponentTypes.GEOLOCATION_FILTER || type === ComponentTypes.FILTER_SEARCH) {
        prefixesToAdd = prefixesToAdd.map(p => `${StorageKeys.QUERY}.${p}`);
      }
      return currentPrefixes.concat(prefixesToAdd);
    }, []);
  );
```

tmeyer2115: Passed this along on Slack as well, but what about:
```
  const componentTypesToExclude = [
    ComponentTypes.FACETS,
    ComponentTypes.FILTER_BOX,
    ComponentTypes.FILTER_OPTIONS,
    ComponentTypes.RANGE_FILTER,
    ComponentTypes.DATE_RANGE_FILTER,
    ComponentTypes.SORT_OPTIONS
    ComponentTypes.GEOLOCATION_FILTER,
    ComponentTypes.FILTER_SEARCH
  ];
  let paramsToFilter = componentTypesToExclude.flatMap(type => {
     let params = getComponentNamesForComponentTypes(type);
      if (type === ComponentTypes.GEOLOCATION_FILTER || type === ComponentTypes.FILTER_SEARCH) {
        params = params.map(p => `${StorageKeys.QUERY}.${p}`);
      }
      return params;
  }
  paramsToFilter.concat([StorageKeys.FILTER]);
```

tmeyer2115 (approved):

oshi97 (approved):
lets gooooo one less jank filter interaction

PR #984, opened 2020-07-23
Add word-break to yxt-FilterOptions-optionLabel

sawong99:
Add word-break: break-word; and hyphens: auto; to yxt-FilterOptions-optionLabel. Previously, some Facets/Filter options with long words were not getting wrapped. Now, long words will continue on the next line with a hyphen.

J=SLAP-100

TEST=manual

Tested on local test site by ensuring that long words in facets/filter options didn't overflow and got wrapped.

### DIFF
--- a/src/ui/sass/modules/_FilterOptions.scss
+++ b/src/ui/sass/modules/_FilterOptions.scss
@@ -123,6 +123,8 @@
 
   &-radioButtonLabel {
     cursor: pointer;
+    word-break: break-word;
+    hyphens: auto;
   }
 
   &-optionLabel {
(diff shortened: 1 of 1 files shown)


src/ui/sass/modules/_FilterOptions.scss
```diff
@@ -132,6 +132,8 @@
     align-items: center;
     margin-left: 22px;
     cursor: pointer;
+    word-break: break-word;
+    hyphens: auto;
```
alexisgrow, line 138: `hyphens: auto` is only partially supported in some modern browsers (like Chrome), confirming that's OK?

alexisgrow (approved):
Question about browser support, otherwise LGTM!

src/ui/sass/modules/_FilterOptions.scss
sawong99: Confirmed with Rose that it's ok!

PR #985, opened 2020-07-23
urlutils: refactor addParamsToUrl

creotutar:
We refactor the addParamsToUrl function for two reasons:

1. The function relies on window.location.search, a global variable
   usually in the browser. This is not available for change in our jest
   tests without extra work. To make the function easier to work with in
   unit tests, we factor out the global variable dependence.

2. In the cases we use addParamsToUrl, we would like to remove certain
   parameters in this generated URL. This makes it easier to pass in
   exactly what parameters should be generated with the url. SLAP-499

There should be no consumer-facing changes as a result of this refactor.
Because the function is not additive but rather replaces the search params,
we rename the function as well.

J=SLAP-182
TEST=manual

In a local Jambo repository, test all usages of addParamsToUrl
Change Filters in Universal
View All Link in Universal
Alternative Vericals in Vertical

Test that current parameters are being added to the links, correctly.
Test that new parameters are being added to the links, correctly.

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -42,18 +42,15 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
 }
 
 /**
- * Returns the passed in url, with all url params from the current url, as well as any
- * pasased in params, appended to it.
+ * Returns the passed in url with the passed in params appended as query params
+ * Note: query parameters in the url are stripped, you should include those query parameters
+ * in `params` if you want to keep them
  * @param {string} url
- * @param {Object} params
+ * @param {SearchParams} params to add to the url
  * @returns {string}
  */
-export function addParamsToUrl (url, params = {}) {
-  const urlParams = new SearchParams(window.location.search.substring(1));
-  for (const paramKey in params) {
-    urlParams.set(paramKey, params[paramKey]);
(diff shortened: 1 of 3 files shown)


src/core/utils/urlutils.js
```diff
@@ -42,18 +42,15 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
...
+ * @param {SearchParams} params to add to the url
  * @returns {string}
  */
-export function addParamsToUrl (url, params = {}) {
-  const urlParams = new SearchParams(window.location.search.substring(1));
-  for (const paramKey in params) {
-    urlParams.set(paramKey, params[paramKey]);
-  }
-  return url.split('?')[0] + '?' + urlParams;
+export function addParamsToUrl (url, params = new SearchParams()) {
```
alexisgrow: Since this strips existing query params and so isn't purely additive anymore, do you think the function name should change?

creotutar: was also previously stripping existing query params. can rename to `combineParamsWithUrl` how does that sound?

creotutar: changed to replaceUrlParams at your suggestion

alexisgrow (approved):
LGTM!

PR #986, opened 2020-07-23
Persist sessionsOptIn value across page navigation.

tmeyer2115:
If the setSessionsOptIn method is called, the provided value now persists
across page navigation. This includes navigation via the nav bar, alternative
verticals, or 'View All' links. Note that the value set in the init config's
sessionTrackingEnabled will not be persisted across page navigation. The data
is persisted by storing it in the URL. When loading an experience on a page,
if there is sessionsOptIn data in the URL, that will take precedecne over
the sessionTrackingEnable config option.

Updates had to be made to the VerticalResults, Navigation, and Alternative
Verticals components to ensure they re-rendered properly whenever the
setSessionsOptIn method was called.

The session tracking data in GlobalStorage is now structured differently. It
is an object with a value and a setDynamically attribute. The setDynamically
attribute allows us to differentiate between a session tracking value set on
ANSWERS.init and one set with ANSWERS.setSessionsOptIn. If it was set using the
former, we do not persist the value across pages.

I did some minor code cleanup as well. Specifically, the AlternativeVerticals
class had a static _buildVerticalSuggestions method. I do not beleive this
method needs to be static anymore. It simplifies things to just be a normal,
private method of the class.

J=SLAP-15
TEST=manual

Performed the following tests:
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  nav bar. Verified that session tracking was disabled there.
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  'View All' links. Verified that session tracking was diabled there.
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  AlternativeVerticals links. Verfied that session tracking was disabled there.
- Verified that if sessions-opt-in is present in the URL, it overrides the
  sessionTrackingEnabled config option.
- Verified that if sessions-opt-in is not present in the URL, the session
  TrackingEnabled value is used, as before.
- Called setSessionsOptIn(false) on a page. Navigated to another page and
  called setSessionsOptIn(true) there. Verified that session tracking was then
  enabled there.
- Ensured that ApiRequests are sent with the correct value.
- Ensured that calling setContext triggeres a correct re-render of the
  NavigationComponent.
- Ensured that AlternativeVerticals were still generated as expected.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -166,7 +166,22 @@ class Answers {
     globalStorage.set(StorageKeys.SEARCH_CONFIG, parsedConfig.search);
     globalStorage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
-    globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
+
+    // Check if sessionsOptIn data is stored in the URL. If it is, prefer that over
+    // what is in parsedConfig.
+    const sessionOptIn = globalStorage.getState(StorageKeys.SESSIONS_OPT_IN);
+    if (!sessionOptIn) {
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: parsedConfig.sessionTrackingEnabled, setDynamically: false });
+    } else {
+      // If sessionsOptIn was stored in the URL, it was stored only as a string.
+      // Parse this value and add it back to globalStorage.
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: (/true/i).test(sessionOptIn), setDynamically: true });
+    }
+
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
 
     const context = globalStorage.getState(StorageKeys.API_CONTEXT);
@@ -402,7 +417,8 @@ class Answers {
    * @param {boolean} optIn
    */
   setSessionsOptIn (optIn) {
-    this.core.globalStorage.set(StorageKeys.SESSIONS_OPT_IN, optIn);
+    this.core.globalStorage.set(
+      StorageKeys.SESSIONS_OPT_IN, { value: optIn, setDynamically: true });
   }
 
   /**
--- a/src/core/search/searchapi.js
+++ b/src/core/search/searchapi.js
@@ -76,28 +76,31 @@ export default class SearchApi {
       version: this._version,
       environment: this._environment,
       params: {
-        'input': input,
-        'experienceKey': this._experienceKey,
-        'version': this._experienceVersion,
-        'filters': filter,
-        'facetFilters': facetFilter,
-        'verticalKey': verticalKey,
-        'limit': limit,
-        'offset': offset,
-        'location': geolocation ? `${geolocation.lat},${geolocation.lng}` : null,
-        'queryId': id,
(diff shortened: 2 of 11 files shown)


src/answers-umd.js
```diff
@@ -166,7 +166,22 @@ class Answers {
...
+    if (!sessionOptIn) {
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: parsedConfig.sessionTrackingEnabled, setDynamically: false });
+    } else {
+      // If sessionsOptIn was stored in the URL, it was stored only as a string.
+      // Parse this value and add it back to globalStorage.
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: (/true/i).test(sessionOptIn), setDynamically: true });
```
oshi97, line 182: nit but do we want to use ^ and $ in the regex to ensure it's a full word match?

src/core/search/searchapi.js
```diff
@@ -76,28 +76,31 @@ export default class SearchApi {
...
-        'retrieveFacets': isDynamicFiltersEnabled,
-        'locale': this._locale,
-        'skipSpellCheck': skipSpellCheck,
-        'queryTrigger': queryTrigger,
-        'sessionTrackingEnabled': sessionTrackingEnabled,
-        'sortBys': sortBys,
-        'locationRadius': locationRadius,
-        'context': context,
-        'referrerPageUrl': referrerPageUrl
+        input: input,
```
oshi97, line 79: yes!! these quotes always tick me off

oshi97 (approved):

creotutar:
Can we separate the refactoring (`const` vs `let`, single quotes) into a separate commit? It feels like they can be separate atomic commits. (Also helps for reading in the PR, so I can see the lines directly affected for the given change, and also not have to worry about changes in const and let).

Should this be in a `dev/<name>` branch?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -470,6 +473,10 @@ export default class NavigationComponent extends Component {
     if (referrerPageUrl !== null) {
       params.set(StorageKeys.REFERRER_PAGE_URL, referrerPageUrl);
     }
+    const sessionsOptIn = this.core.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN);
+    if (sessionsOptIn.setDynamically) {
```
creotutar, line 477: Do we want to null check for `sessionsOptIn`?

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -121,8 +118,11 @@ export default class AlternativeVerticalsComponent extends Component {
     if (typeof referrerPageUrl === 'string') {
       params[StorageKeys.REFERRER_PAGE_URL] = referrerPageUrl;
     }
+    if (sessionsOptIn && sessionsOptIn.setDynamically) {
```
creotutar, line 121: Is the purpose of `setDynamically` to save us from this one params assignment? Was this causing a bug?

tmeyer2115:
> Can we separate the refactoring (`const` vs `let`, single quotes) into a separate commit? It feels like they can be separate atomic commits. (Also helps for reading in the PR, so I can see the lines directly affected for the given change, and also not have to worry about changes in const and let).
> 
> Should this be in a `dev/<name>` branch?

Yep, should be named with a `dev` prefix! Old habits die hard. Closing this PR and opening another one here: https://github.com/yext/answers/pull/991. This has the origin as the correct `dev`-prefixed branch. As for the smaller changes, those were automatically applied when I ran semistandard on the files. I needed to fix some of the file's code styling before the tests would run. When I ran it, it seemed to fix some other things in the files as well.

PR #987, opened 2020-07-23
urlutils: refactor addParamsToUrl, add unit tests

creotutar:
We refactor the addParamsToUrl function for two reasons:

1. The function relies on window.location.search, a global variable
   usually in the browser. This is not available for change in our jest
   tests without extra work. To make the function easier to work with in
   unit tests, we factor our the global variable dependence.

2. In the cases we use addParamsToUrl, we would like to remove certain
   parameters in this generated URL. This makes it easier to pass in
   exactly what parameters should be generated with the url. SLAP-499

There should be no consumer-facing changes as a result of this refactor.

Make unit tests that test their main usages and edge cases.

J=SLAP-182
TEST=manual

In a local Jambo repository, test all usages of addParamsToUrl
Change Filters in Universal
View All Link in Universal
Alternative Vericals in Vertical

Test that current parameters are being added to the links, correctly.
Test that new parameters are being added to the links, correctly.

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -42,18 +42,15 @@ export function getAnalyticsUrl (env = PRODUCTION, conversionTrackingEnabled = f
 }
 
 /**
- * Returns the passed in url, with all url params from the current url, as well as any
- * pasased in params, appended to it.
+ * Returns the passed in url with the passed in params appended as query params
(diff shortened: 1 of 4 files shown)


creotutar:
Merging the feature branch, all code has been reviewed in #980 and #985

PR #989, opened 2020-07-24
browser default is used when no target is specifed to RTF formatter

rosiegrant:
see https://yext.slack.com/archives/GREFVUD7V/p1595606176016000?thread_ts=1595605288.013200&cid=GREFVUD7V

### DIFF
--- a/README.md
+++ b/README.md
@@ -2143,7 +2143,7 @@ targetConfig = { url: '_blank', phone: '_self', email: '_parent' }
 targetConfig = '_blank'
 ```
 
-When `targetConfig` is a string, it is assumed that any link, regardless of type, has the specified `target` behavior. This parameter, like `eventOptionsFieldName`, is optional. When not provided, no `target` attribute is supplied to the links.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2143,7 +2143,7 @@ targetConfig = { url: '_blank', phone: '_self', email: '_parent' }
 targetConfig = '_blank'
 ```
 
-When `targetConfig` is a string, it is assumed that any link, regardless of type, has the specified `target` behavior. This parameter, like `eventOptionsFieldName`, is optional. When not provided, no `target` attribute is supplied to the links.
+When `targetConfig` is a string, it is assumed that any link, regardless of type, has the specified `target` behavior. This parameter, like `eventOptionsFieldName`, is optional. When not provided, no `target` attribute is supplied to the links, and it will use the browser default of `_self`.
```
alexisgrow, line 2146: This wording is slightly confusing to me, does it omit the target attribute on the link or does it supply `target="_self"`?

oshi97:
hiii closing this for now, feel free to reopen though!

PR #990, opened 2020-07-24
Added a connorSearch and connorNumberSearch methods to the searchapi

cea2aj:
Added connorSearch and connorNumberSearch methods in accordance with the codelab. 

connorSearch simply searches for the string 'Connor', and connorNumberSearch will search for 'Connor' appended by the number passed in as the parameter. For example connorNumberSearch(9398) will search for 'Connor9398'. These methods are simply for the codelab and should not be merged into master.

J=none
T=auto

Ran "npm run test" and confirmed tests passed

### DIFF
--- a/src/core/search/searchapi.js
+++ b/src/core/search/searchapi.js
@@ -158,4 +158,24 @@ export default class SearchApi {
     return request.get()
       .then(response => response.json());
   }
+
+  /**
+   * Initiate a search for Connor
+   */
+  connorSearch () {
+    const query = 'Connor';
+    return this.universalSearch(query, {});
+  }
+
+  /**
+   * Initiate a search for Connor concatenated by a number
+   * @param {number} number the number which is concatenated to the string 'Connor'
+   */
+  searchForConnorNumber (number) {
+    if (number && typeof number !== 'number') {
+      throw new AnswersCoreError('number is not of type number', 'searchForConnorNumber');
+    }
+    const query = `Connor${number.toString()}`;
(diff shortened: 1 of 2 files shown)


cea2aj:
When in doubt, should we stick to naming conventions for change them if it improves clarity?

cea2aj:
I made this two functions in an attempt to stick to the Clean Code principle of having each functions do just one thing

oshi97:
> When in doubt, should we stick to naming conventions for change them if it improves clarity?

I think it really depends on the situation. If it's something small and obvious it's definitely better to use the "better" convention and also change other usages of it. On the other hand if it's a big refactor, most of the time we don't want to scope creep it into an unrelated PR, and it might be better to have a separate PR for it. That also lets everybody have more discussion on the best way to refactor it. Are there any patterns you were thinking about in particular?

cea2aj:
> > When in doubt, should we stick to naming conventions for change them if it improves clarity?
> 
> I think it really depends on the situation. If it's something small and obvious it's definitely better to use the "better" convention and also change other usages of it. On the other hand if it's a big refactor, most of the time we don't want to scope creep it into an unrelated PR, and it might be better to have a separate PR for it. That also lets everybody have more discussion on the best way to refactor it. Are there any patterns you were thinking about in particular?

Makes sense! The question came up because I was refactoring connorNumberSearch, but it ended up being both clearer and aligned with the conventions. I'll let you know if if this pops up again for a specific pattern in the future.

oshi97 (approved):

PR #991, opened 2020-07-27
Persist sessionsOptIn value across page navigation.

tmeyer2115:
If the setSessionsOptIn method is called, the provided value now persists
across page navigation. This includes navigation via the nav bar, alternative
verticals, or 'View All' links. Note that the value set in the init config's
sessionTrackingEnabled will not be persisted across page navigation. The data
is persisted by storing it in the URL. When loading an experience on a page,
if there is sessionsOptIn data in the URL, that will take precedecne over
the sessionTrackingEnable config option.

Updates had to be made to the VerticalResults, Navigation, and Alternative
Verticals components to ensure they re-rendered properly whenever the
setSessionsOptIn method was called.

The session tracking data in GlobalStorage is now structured differently. It
is an object with a value and a setDynamically attribute. The setDynamically
attribute allows us to differentiate between a session tracking value set on
ANSWERS.init and one set with ANSWERS.setSessionsOptIn. If it was set using the
former, we do not persist the value across pages.

I did some minor code cleanup as well. Specifically, the AlternativeVerticals
class had a static _buildVerticalSuggestions method. I do not believe this
method needs to be static anymore. It simplifies things to just be a normal,
private method of the class. Also ran semistandard on a few files to clean up some
usages of 'let' and unnecessary quotes. 

J=SLAP-15
TEST=manual

Performed the following tests:
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  nav bar. Verified that session tracking was disabled there.
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  'View All' links. Verified that session tracking was diabled there.
- Called setSessionsOptIn(false) on a page. Navigated to another page via the
  AlternativeVerticals links. Verfied that session tracking was disabled there.
- Verified that if sessions-opt-in is present in the URL, it overrides the
  sessionTrackingEnabled config option.
- Verified that if sessions-opt-in is not present in the URL, the session
  TrackingEnabled value is used, as before.
- Called setSessionsOptIn(false) on a page. Navigated to another page and
  called setSessionsOptIn(true) there. Verified that session tracking was then
  enabled there.
- Ensured that ApiRequests are sent with the correct value.
- Ensured that calling setContext triggeres a correct re-render of the
  NavigationComponent.
- Ensured that AlternativeVerticals were still generated as expected.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -179,7 +179,7 @@ class Answers {
       // Parse this value and add it back to globalStorage.
       globalStorage.set(
         StorageKeys.SESSIONS_OPT_IN,
-        { value: (/true/i).test(sessionOptIn), setDynamically: true });
+        { value: (/^true$/i).test(sessionOptIn), setDynamically: true });
     }
 
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -166,7 +166,22 @@ class Answers {
     globalStorage.set(StorageKeys.SEARCH_CONFIG, parsedConfig.search);
     globalStorage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
-    globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
+
+    // Check if sessionsOptIn data is stored in the URL. If it is, prefer that over
+    // what is in parsedConfig.
+    const sessionOptIn = globalStorage.getState(StorageKeys.SESSIONS_OPT_IN);
+    if (!sessionOptIn) {
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: parsedConfig.sessionTrackingEnabled, setDynamically: false });
+    } else {
+      // If sessionsOptIn was stored in the URL, it was stored only as a string.
+      // Parse this value and add it back to globalStorage.
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
+        { value: (/true/i).test(sessionOptIn), setDynamically: true });
+    }
+
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
 
     const context = globalStorage.getState(StorageKeys.API_CONTEXT);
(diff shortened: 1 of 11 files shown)


oshi97 (approved):

cea2aj:
Would it improve clarity if we refactor sessionsOptIn to sessionCookieOptIn? It might not be necessary, however I had to refer to the comment above setSessonsOptIn to understand what it was.

tmeyer2115:
> Would it improve clarity if we refactor sessionsOptIn to sessionCookieOptIn? It might not be necessary, however I had to refer to the comment above setSessonsOptIn to understand what it was.

You're right, setSessionsOptIn is not the most descriptive name on the planet! For now, I'd like to keep the method name as is. Changing it's signature (including the name) would be considered a breaking change to the SDK (as this method is used directly by consumers). But, your suggestion is definitely something to keep in mind when we release a new major version!

src/answers-umd.js
```diff
@@ -166,7 +166,22 @@ class Answers {
...
     globalStorage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
-    globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
+
+    // Check if sessionsOptIn data is stored in the URL. If it is, prefer that over
+    // what is in parsedConfig.
+    const sessionOptIn = globalStorage.getState(StorageKeys.SESSIONS_OPT_IN);
+    if (!sessionOptIn) {
+      globalStorage.set(
+        StorageKeys.SESSIONS_OPT_IN,
```
cea2aj, line 175: This comment makes me think that the code in the first block of the if statement will refer to the case when sessionsOptIn is true, and we are preferring that to parsedConfig, however that is not the case. Can we change the comment to better reflect that? Also, we might be able to get away with a shorter comment by changing sessionOptIn to sessionOptInStoredInURL

src/answers-umd.js
```diff
@@ -166,7 +166,22 @@ class Answers {
     globalStorage.set(StorageKeys.SEARCH_CONFIG, parsedConfig.search);
     globalStorage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
-    globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
+
+    // Check if sessionsOptIn data is stored in the URL. If it is, prefer that over
+    // what is in parsedConfig.
+    const sessionOptIn = globalStorage.getState(StorageKeys.SESSIONS_OPT_IN);
+    if (!sessionOptIn) {
```
cea2aj, line 173: This comment makes me think that the code in the first block of the if statement will refer to the case when sessionsOptIn is true, and we are preferring that to parsedConfig, however that is not the case. Can we change the comment to better reflect that? Also, we might be able to get away with a shorter comment by changing sessionOptIn to sessionOptInStoredInURL

PR #996, opened 2020-07-29
Answers init returns a promise

cea2aj:
This PR changes ANSWERS.init() such that it returns a promise which resolves when initialization is complete. Both the onReady callback and the promise may be used. If both are included, the onReady callback is called first, and then the promise is resolved.

Usage looks like this:
```javascript
ANSWERS.init({
  //config
}).then(() => {
  //onReady code 
});
```

This commit also changes _handlePonyfillCssVariables such that it returns a promise.

J=SLAP-504
TEST=manual

I tested this with html that initializes the SearchBar and Universal results. I tested onReady code just in the callback, in both the callback and in the promise, and in just the promise. I confirmed onReady executes before the promise resolves.

I tested the code with MasterSwitchAPI and with ponyfillCss manually disabled.

I also tested on IE11 and confirmed that the existing polylfills work so long as arrow functions are not used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,8 @@ function initAnswers() {
     apiKey: '<API_KEY_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
-    // Required, initialize components here, invoked when the Answers component library is loaded/ready.
-    //   Components may instead be initialized after the init promise resolves.
+    // Optional, initialize components here, invoked when the Answers component library is loaded/ready.
+    //    If components are not added here, they must be added when the init promise resolves
     onReady: function() {},
     // Optional*, Yext businessId, *required to send analytics events
     businessId: 'businessId',
--- a/README.md
+++ b/README.md
@@ -85,6 +85,24 @@ function initAnswers() {
 }
 ```
 
+ANSWERS.init() returns a promise which optionally allows components to be added using promise syntax
+```js
+function initAnswers() {
+  ANSWERS.init({
+    apiKey: '<API_KEY_HERE>', // See [1]
+    experienceKey: '<EXPERIENCE_KEY_HERE>',
+  }).then(function() {
+    ANSWERS.addComponent('SearchBar', {
+      container: '#SearchBarContainer',
+    });
+
+    ANSWERS.addComponent('UniversalResults', {
+      container: '#UniversalResultsContainer',
+    });
+  });
+}
+```
+
 [1] Learn more about [getting your API key](https://developer.yext.com/docs/guides/get-started/).
 
 # ANSWERS.init Configuration Options
@@ -98,7 +116,8 @@ function initAnswers() {
     apiKey: '<API_KEY_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
-    // Required, initialize components here, invoked when the Answers component library is loaded/ready
+    // Required, initialize components here, invoked when the Answers component library is loaded/ready.
+    //   Components may instead be initialized after the init promise resolves.
     onReady: function() {},
     // Optional*, Yext businessId, *required to send analytics events
     businessId: 'businessId',
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ function initAnswers() {
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
     // Optional, initialize components here, invoked when the Answers component library is loaded/ready.
-    //    If components are not added here, they must be added when the init promise resolves
+    //    If components are not added here, they can also be added when the init promise resolves
     onReady: function() {},
     // Optional*, Yext businessId, *required to send analytics events
     businessId: 'businessId',
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -434,17 +434,18 @@ class Answers {
    * @return {Promise} resolves after ponyfillCssVariables, or immediately if disabled
    */
   _handlePonyfillCssVariables (ponyfillDisabled) {
+    if (ponyfillDisabled) {
+      return Promise.resolve();
+    }
     return new Promise((resolve, reject) => {
-      if (!ponyfillDisabled) {
-        this.ponyfillCssVariables()
-          .then(resolve());
-      }
-      resolve();
+      this.ponyfillCssVariables({
+        onFinally: resolve()
+      });
     });
   }
 
   /*
-   * A promise that updates the css styles with new current variables. This is useful when the css
+   * Updates the css styles with new current variables. This is useful when the css
    * variables are updated dynamically (e.g. through js) or if the css variables are
    * added after the ANSWERS.init
    *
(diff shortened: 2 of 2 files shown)


cea2aj:
We could clean up this code a bit more by removing _handlePonyfillCssVariables altogether if we simply call await ponyfillCssVariables inside the init, but we would need to make the init async, and I'm not sure if that's a good idea

src/answers-umd.js
```diff
@@ -259,52 +259,28 @@ class Answers {
...
     }
 
-    // Templates are currently downloaded separately from the CORE and UI bundle.
-    // Future enhancement is to ship the components with templates in a separate bundle.
-    this.templates = new DefaultTemplatesLoader(templates => {
-      this.renderer.init(templates);
-      this._handlePonyfillCssVariables(
-        parsedConfig.disableCssVariablesPonyfill,
-        this._invokeOnReady.bind(this));
+    this.promise = new Promise((resolve, reject) => {
```
creotutar: Does this stop us from being backwards compatible? It seems like people _must_ use the Promise with these changes. Is this what we wanted, considering a lot of current customers are using a non-promised version?

cea2aj: Nope, it's still backwards compatible with onReady in the config. Right before the promise is resolved onReady is executed, even you if aren't using the promise version.

creotutar: Love the new promises around css-vars-ponyfill

oshi97: This is definitely a creative solution! I would still rather just return the promise directly, but if Rose thinks we need to keep returning the ANSWERS instance then this looks good to me. I thought about maybe naming it onReadyPromise but that doesn't seem as elegant for some reason?

src/answers-umd.js
```diff
@@ -259,50 +259,29 @@ class Answers {
...
-    // Future enhancement is to ship the components with templates in a separate bundle.
-    this.templates = new DefaultTemplatesLoader(templates => {
-      this.renderer.init(templates);
-      this._handlePonyfillCssVariables(
-        parsedConfig.disableCssVariablesPonyfill,
-        this._invokeOnReady.bind(this));
+    const onReadyPromise = new Promise((resolve, reject) => {
+      this._handlePonyfillCssVariables(parsedConfig.disableCssVariablesPonyfill)
+        .then(this._masterSwitchApi.isDisabled())
+        .then((isDisabled) => {
```
oshi97: nit but I think our style is to normally omit the parenthesis around a javascript arrow function'
s param if there's only one param

src/answers-umd.js
```diff
@@ -259,50 +259,29 @@ class Answers {
...
+          if (!isDisabled) {
```
oshi97: do we need to reject if isDisabled is true?

src/answers-umd.js
cea2aj: Will do!

src/answers-umd.js
```diff
@@ -259,50 +259,29 @@ class Answers {
...
     }
 
-    // Templates are currently downloaded separately from the CORE and UI bundle.
-    // Future enhancement is to ship the components with templates in a separate bundle.
-    this.templates = new DefaultTemplatesLoader(templates => {
-      this.renderer.init(templates);
-      this._handlePonyfillCssVariables(
-        parsedConfig.disableCssVariablesPonyfill,
-        this._invokeOnReady.bind(this));
+    const onReadyPromise = new Promise((resolve, reject) => {
```
oshi97: oh hmm sorry I missed this during the first round but could we just return the Promise returned by _handlePonyfillCssVariables? instead of wrapping it inside a new promise

src/answers-umd.js
cea2aj: I didn't call reject because I was trying to match the original logic flow as closely as possible, however I think you're right. It should be called.

src/answers-umd.js
cea2aj: I wrapped it in a promise because the init resolving depends on the results of _handlePonyfillCssVariables promise and the MasterSwitchApi.isDisabled(), which is also a promise. If you can simplify it, I'm happy to take a look

cea2aj: Update: We managed to simplify this!

src/answers-umd.js
```diff
@@ -452,33 +427,57 @@ class Answers {
...
   }
 
+  /**
+   * A promise that resolves when ponyfillCssVariables resolves,
+   * or resolves immediately if ponyfill is disabled
+   * @param {boolean} option to opt out of the css variables ponyfill
+   * @return {Promise} resolves after ponyfillCssVariables, or immediately if disabled
+   */
+  _handlePonyfillCssVariables (ponyfillDisabled) {
+    return new Promise((resolve, reject) => {
```
oshi97, line 440: nit but I think we can shorten this one slightly too, by doing
```js
if (!ponyfillDisabled) {
  return Promise.resolve();
}
return this.ponyfillCssVariables()
```
Another option, though, might be to have ponyfillCssVariables() go back to just being a pass through to cssVars(), and have all of the promise logic go into this _handle function? I think that divides up responsibility a little more neatly (let me know if you have any other thoughts). Then this _handle method could return a Promise.resolve() if ponyfills are disabled, or a new Promise to passes resolve into the onFinally of ponyfillCssVariables if not disabled

cea2aj: I like that idea. It's definitely a little cleaner

README.md
```diff
@@ -98,7 +116,8 @@ function initAnswers() {
     apiKey: '<API_KEY_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
-    // Required, initialize components here, invoked when the Answers component library is loaded/ready
+    // Required, initialize components here, invoked when the Answers component library is loaded/ready.
+    //   Components may instead be initialized after the init promise resolves.
```
oshi97: since this isn't required anymore, remove the "Required" and maybe reword this slightly

src/answers-umd.js
```diff
@@ -452,6 +427,23 @@ class Answers {
...
+   * @param {boolean} option to opt out of the css variables ponyfill
+   * @return {Promise} resolves after ponyfillCssVariables, or immediately if disabled
+   */
+  _handlePonyfillCssVariables (ponyfillDisabled) {
+    if (ponyfillDisabled) {
+      return Promise.resolve();
+    }
+    return new Promise((resolve, reject) => {
+      this.ponyfillCssVariables({
+        onFinally: resolve()
```
oshi97: here, we want to pass in a function. currently what this does (I believe at least) is execute the resolve as this line as evaluated, and pass the return value of resolve, as oppose to pass the resolve function itself into ponyfilCssVariables

oshi97: either just resolve or () => resolve() should work (assuming I'm not bonkers)

cea2aj: Ahh, good catch. Just changed it to resolve.

README.md
```diff
@@ -98,7 +116,8 @@ function initAnswers() {
     apiKey: '<API_KEY_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
-    // Required, initialize components here, invoked when the Answers component library is loaded/ready
+    // Optional, initialize components here, invoked when the Answers component library is loaded/ready.
+    //    If components are not added here, they must be added when the init promise resolves
```
oshi97: definitely just my opinion but "must be added" feels a little too strong for me here. technically, you can also add components in both places and things will work fine. Maybe just "they can also" instead of "they must be"?

cea2aj: Sure, that sounds good

oshi97 (approved):
lgtm!

PR #999, opened 2020-07-30

sawong99:
Added tab order calculations in src/ui/tools/taborder.js so that both navigationcomponent.js and verticalresultscomponent.js can access the functions. Updated getVerticalURL in verticalresultscomponent.js so that it adds the tabOrder to the end of the URLs for the View All buttons. Initially, tab order was not preserved when clicking View All from the universal page. Now, tab order is preserved.

J=SLAP-445

TEST=manual


src/ui/tools/taborder.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+  return tabOrder;
+}
+
+export function getTabOrder (tabsConfig, dataTabOrder) {
+  let tabOrder = getDefaultTabOrder(tabsConfig, getUrlParams());
+  // We want to persist the params from the existing URL to the new
+  // URLS we create.
+  if (tabOrder !== undefined && dataTabOrder !== undefined) {
```
oshi97: nit but I think this can just be `if (tabOrder && dataTabOrder)`

sawong99: will fix this!

src/ui/tools/taborder.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * mergeTabOrder merges two arrays into one
+ * by appending additional tabs to the end of the original array
+ * @param {string[]} tabOrder Tab order provided by the server
+ * @param {string[]} otherTabOrder Tab order provided by configuration
+ * @return {string[]}
+ */
+export function mergeTabOrder (tabOrder, otherTabOrder, tabs) {
+  for (let i = 0; i < otherTabOrder.length; i++) {
```
oshi97: nit but can we change this to a `for (const tabConfig of otherTabOrder)` or a `otherTabOrder.forEach(tabConfig => ...`?

src/ui/tools/taborder.js
```diff
@@ -0,0 +1,69 @@
...
+export function getDefaultTabOrder (tabsConfig, urlParams) {
+  let tabOrder = [];
+
+  // Use the ordering from the URL as the primary configuration
+  // And then merge it with the local configuration, if provided.
+  if (urlParams && urlParams.has('tabOrder')) {
+    tabOrder = urlParams.get('tabOrder').split(',');
+  }
+
+  for (let i = 0; i < tabsConfig.length; i++) {
```
oshi97: same nit as the below (for const of or forEach loop)

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -241,7 +242,8 @@ export default class VerticalResultsComponent extends Component {
   getVerticalURL (data = {}) {
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
     const verticalURL = this._config.verticalURL || verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
-    return addParamsToUrl(verticalURL, { query: this.query });
+    const dataTabOrder = this.core.globalStorage.getState(StorageKeys.NAVIGATION) ? this.core.globalStorage.getState(StorageKeys.NAVIGATION).tabOrder : [];
+    return addParamsToUrl(verticalURL, { query: this.query, tabOrder: getTabOrder(this._verticalsConfig, dataTabOrder) });
```
oshi97: nit but can we have the getTabOrder call on another line? I think it's a little easier to read that way (not squishing multiple calls into one line and you also get to name the variable something descriptive)

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -5,8 +5,9 @@
 import Component from '../component';
 import { AnswersComponentError } from '../../../core/errors/errors';
 import StorageKeys from '../../../core/storage/storagekeys';
-import SearchParams from '../../dom/searchparams';
 import DOM from '../../dom/dom';
+import { mergeTabOrder, getDefaultTabOrder, getUrlParams } from '../../tools/taborder';
```
oshi97, line 9: @creotutar you're right this file path definitely makes more sense

oshi97 (approved):
lgtm! 🔥

PR #1001, opened 2020-07-30

sawong99:
Added tab order calculations in src/ui/tools/taborder.js so that both navigationcomponent.js and verticalresultscomponent.js can access the functions. Updated getVerticalURL in verticalresultscomponent.js so that it adds the tabOrder to the end of the URLs for the View All buttons. Initially, tab order was not preserved when clicking View All from the universal page. Now, tab order is preserved.

J=SLAP-445

TEST=manual


src/ui/components/results/verticalresultscomponent.js
```diff
@@ -152,10 +153,12 @@ export default class VerticalResultsComponent extends Component {
     }
   }
 
-  getVerticalURL (data) {
+  getVerticalURL (data = {}) {
     const verticalConfig = this._verticalsConfig.find(config => config.verticalKey === this.verticalKey) || {};
-    const verticalURL = verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
-    return addParamsToUrl(verticalURL, { query: this.query });
+    const verticalURL = this._config.verticalURL || verticalConfig.url || data.verticalURL || this.verticalKey + '.html';
```
oshi97: nit but config.verticalURL doesn't exist yet

sawong99: ah ok, I'll remove it!

oshi97 (approved):

PR #1004, opened 2020-07-31
urls: remove facets/pagination/filters from inter-experience links

creotutar:
NOTE: This includes all refactors assumed during its development in the
develop branch, including the urlutils refactor for replaceUrlParams and
the alternativeVerticals buildVerticalSuggestions class function.

We remove query parameters dealing with facets, pagination, or filters
from the inter-experience links. This is because this encoded
information should not persist as we move throughout the experience
(ie to another vertical).

For example: if you land on a vertical page with facets, these facets are
automatically added to the URL. When you try to navigate to another
tab, these query parameters persist (encoded within the nav URL) and
can lead to confusion when another vertical also has facets with similar
facet names.

re Implementation: Because name is a configurable option for components, and because
facets/filters use this as a basis for the persistentStorage key, we
must account for the different types of possible names. Thus, we store
the componentNames mapped to a component type, for easier access.

To avoid a circular dependency (Component -> urlUtils -> Component.type),
we split out the types into a different ComponentTypes class that allow
one source of truth for the component type for both the new urlutils function
and the components themselves.

J=SLAP-499
TEST=manual

Test that navigation links on a page with filters, pagination, and
facets do not contain those query parameters.

filter search
geolocation
facets
filterbox with specific CF filter
Test that View More links, Change Filter links on universal do not
contain those query parameters.

Test that Alternative Verticals vertical and universal links do not
contain those query parameters.

Add acceptance test for going from page with facets to universal
and to vertical page and make sure those do not contain the
facet/filter/pagination params.


src/core/utils/urlutils.js
```diff
@@ -80,3 +84,60 @@ export function equivalentParams (params1, params2) {
   }
   return true;
 }
+
+/**
+ * Creates a copy of the provided {@link SearchParams}, with the specified
+ * attributes filtered out
+ * @param {SearchParams} params The parameters to remove from
+ * @param {string[]} prefixes The prefixes of parameters to remove
+ * @return {SearchParams} A new instance of SearchParams without removed params
```
oshi97: nit but can we reword slightly we say params 3 times in a row here lol

oshi97 (approved):

src/core/utils/urlutils.js
creotutar: 😂

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #1011, opened 2020-08-03
SearchComponent: fix urls sent to core.search()

oshi97:
SearchComponent was relying on the verticalKeys in
the navigation component's "tab" state having a
configId property. This was changed to verticalKey around 5
months ago.

This commit also changes resultssectionheader.hbs to hide
the view all link if viewAllText is not specified. Previously,
the link element would still be there, just with no text. This
should prevent issues with the view all text dom element
magically appearing for somebody upgrading from 1.3.3 to 1.3.4
due to this fix.

TEST=manual

Test that view all links will now show up in universal without
doing a translateData.

Test that the link will not show up if no viewAllText is specified.

### DIFF
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -570,7 +570,8 @@ export default class SearchComponent extends Component {
             if (params.toString().length > 0) {
               url += '?' + params.toString();
             }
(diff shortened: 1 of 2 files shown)


oshi97:
This has been broken since at least 0.13.3 if not even before that, so we can probably ignore this

PR #1000, opened 2020-07-30
Preserve Tab Order from View All Links on Universal 

sawong99:
Added tab order calculations in src/ui/tools/taborder.js so that both navigationcomponent.js and verticalresultscomponent.js can access the functions. Updated getVerticalURL in verticalresultscomponent.js so that it adds the tabOrder to the end of the URLs for the View All buttons. Initially, tab order was not preserved when clicking View All from the universal page. Now, tab order is preserved.

J=SLAP-445

TEST=manual

Tested on local site with 4 pages - universal, employees, faqs and locations. Used different search queries that changed the order of the tabs and made sure that clicking View All preserved the order.

PR #1013, opened 2020-08-04
Version 1.5.2

oshi97:
## Styling Fixes
* Add `word-break` and `hyphens` styling to the labels of Filter options. This allows us to gracefully wrap options with long names (#984).
## Changes
* The `View All` links on universal now include the correct `tabOrder` parameter (#1000).
* A new method was added to `ANSWERS`: `registerTemplate`. This method compiles the provided template and merges it, locally, with the SDK template bundle (#1002).
* URL parameters for Facets, Filters, Pagination, and Sorts are now removed from Navigation, Alternative Vertical, and `View All` links (#1004).
* The `SearchBar` submit button has `type="button"` when the `useForm` attribute is false. This is necessary to get the component working properly on ASP.net sites (#1003).


alexisgrow (approved):
LGTM!

PR #1014, opened 2020-08-05
Filters/Facets - Do not update URL Until Apply

cea2aj:
When searchOnChange inside the filter box component isn't set, don't update the URL (persistent storage) until the apply button is hit.

This behavior wasn't happening because getFilterNode() inside filterOptionsComponent had the side effect of calling persistentStorage.set(). To fix this I created a new function called saveSelectedToPersistentStorage() which is now called inside filterBoxComponent._saveFilterNodesToStorage() and filterOptionsComponent.apply().

J=SLAP-16
TEST=manual, auto

For manual testing I created a local answers site with a FilterBox, FilterOptions and a Facets component and observed the URL as I selected options and hit the apply button. I did this when searchOnChange was both true and false for the various components.

For automatic testing I added unit tests for the FilterBox and FilterOptions components that verify changes to persistent storage as the options are selected and as the apply button is pressed.


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -642,6 +642,12 @@ export default class FilterOptionsComponent extends Component {
     this.updateListeners(true, true);
     this.setState();
   }
+  /**
```
oshi97, line 649: nit: can we have a line break here

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -658,7 +664,6 @@ export default class FilterOptionsComponent extends Component {
         remove: () => this._clearSingleOption(o)
       }));
 
-    this.core.persistentStorage.set(this.name, this.config.options.filter(o => o.selected).map(o => o.label));
```
oshi97, line 662: FilterOptions is intended to also work as a standalone component, so it still needs to store stuff in persistent storage in that case as well. I don't think many people (if any) use it standalone, but it's still something we support

src/ui/components/filters/filteroptionscomponent.js
cea2aj: Sure

src/ui/components/filters/filteroptionscomponent.js
cea2aj: Good catch. I'll take a look

tests/acceptance/acceptancesuite.js
```diff
@@ -59,13 +60,19 @@ test(`Facets load on the page, and can affect the search`, async t => {
   const facets = FacetsPage.getFacetsComponent();
   const filterBox = facets.getFilterBox();
 
+  const getPageUrl = ClientFunction(() => window.location.href);
```
tmeyer2115: I'm not sure this should be part of an acceptance test. I think we've been kinda blurring the line between our unit and acceptance tests. Part of the reason for that, I think, is that it can be hard for us to write unit tests for our components. A lot of things need to be mocked that are difficult to mock. 

I think we should try and make this into a unit test of FilterOptions, if possible. Essentially, what we want to test is that if storeOnChange is false, we don't call persistentStorage.set until the apply button is also clicked. With acceptance tests, we want to be testing normal customer flows that span multiple pages and components. This particular test is scoped to just the FilterOptions component. 

It seems like you may have already added unit test coverage for this. If so, I think we can just remove your changes from this file.

cea2aj: Yep, I do have unit test coverage so I'll remove the acceptance test changes

README.md
```diff
@@ -1375,12 +1375,10 @@ ANSWERS.addComponent('FilterOptions', {
   options: [
     /** Depends on the above optionType, either 'STATIC_FILTER' or 'RADIUS_FILTER', see below. **/
   ],
-  // Optional, if true, the filter value is saved on change and sent with the next search. Defaults to false.
-  storeOnChange: false,
+  // Optional, if true, the filter value is saved on change and sent with the next search. Defaults to true.
```
oshi97, line 1378: beautiful

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -64,7 +64,7 @@ class FilterOptionsConfig {
     this.onChange = config.onChange || function () { };
 
     /**
-     * If true, stores the filter to storage on each change
+     * If true, stores the filter to global and persistent storage on each change
```
oshi97, line 67: 👍

oshi97 (approved):
I love it, sounds like you tested all 3 components as well which is awesome

PR #1016, opened 2020-08-06
Acceptance tests use answers.js instead of answers.min.js

cea2aj:
Use answers.js instead of answers.min.js in the acceptance tests. This lets us run "npm run dev" to keep the builds that the acceptance tests rely on up to date. This means that we no longer have to run "npm run build" before running acceptance tests.

TEST=auto

I confirmed the acceptance tests still run

### DIFF
--- a/tests/acceptance/fixtures/html/universal.html
+++ b/tests/acceptance/fixtures/html/universal.html
@@ -3,7 +3,7 @@
         <meta charset="utf-8" />
     
         <script src="http://localhost:9999/dist/answerstemplates.compiled.min.js"></script>
-        <script src="http://localhost:9999/dist/answers.min.js"></script>
+        <script src="http://localhost:9999/dist/answers.js"></script>
         <link rel="stylesheet" type="text/css" href="http://localhost:9999/dist/answers.css">
     </head>
     <body>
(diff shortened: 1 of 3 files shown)


cea2aj:
I opted for this solution because a mandatory build step for the acceptance tests added 13 seconds to the tests, which I felt was too long

tests/acceptance/fixtures/html/universal.html
```diff
@@ -3,7 +3,7 @@
         <meta charset="utf-8" />
     
         <script src="http://localhost:9999/dist/answerstemplates.compiled.min.js"></script>
-        <script src="http://localhost:9999/dist/answers.min.js"></script>
+        <script src="http://localhost:9999/dist/answers.js"></script>
         <link rel="stylesheet" type="text/css" href="http://localhost:9999/dist/answers.css">
```
tmeyer2115, line 7: Just double checking, but does npm run dev keep the other two assets listed up to date?

creotutar (approved):
👍 👍 👍 LGTM

tests/acceptance/fixtures/html/universal.html
oshi97: Yup! both of these files are currently output by the dev gulp task

oshi97 (approved):

PR #1017, opened 2020-08-11
gulp: Remove global polyfills

creotutar:
We remove the global polyfills in our SDK builds. Previously, we prepended our
polyfill-prefix file to every file in our builds. This is for both template
builds and our answers library builds. As a result of this, any page that
imported these assets would have their global polyfills replaced with the
polyfills defined in the libraries we imported. In particular, this is the
core-js/stable/* and regnenerator-runtime libraries.

Instead of using these globally, we transpile our code with babel and
associated plugins to make up for the loss of the polyfills. This decision is
in response to many techops where client sites, seeing that our polyfills clash
with their implementations. This is in effort to make our library and
templatebundle self-contained, not to touch any code outside of our
implementation.

Implementation detail: we must also run one of our node_modules
(whatwg-fetch polyfill) through babel as well because it relies on a
Promise implementation (noted in their documentation).

J=SLAP-555
TEST=manual

Test that an answers experience (not Jambo) using a local reference to
the SDK will run basic search functions. Not a Jambo site because the
HH theme is running code that requires a polyfill outside
of the SDK (Object.assign). Note: The core-js code will be added to the
theme in the short-term.

In IE11, Edge (Windows) Firefox, Chrome (MacOS) iPhone Safari (iOS)
Test search on universal
Test search on vertical
Test facets/filters/pagination on vertical
Test map on universal
Test view more links / change filters on universal
Test direct answers on universal and feedback (error on IE11 on develop,
see SLAP-556)
Test autocomplete shows on both universal / vertical

### DIFF
--- a/conf/gulp-tasks/bundle/bundletaskfactory.js
+++ b/conf/gulp-tasks/bundle/bundletaskfactory.js
@@ -1,12 +1,9 @@
-const fs = require('fs');
-
 const { dest } = require('gulp');
 
 const rollup = require('gulp-rollup-lightweight');
 const babel = require('rollup-plugin-babel');
 const resolve = require('rollup-plugin-node-resolve');
 const commonjs = require('rollup-plugin-commonjs');
-const insert = require('rollup-plugin-insert');
 
 const source = require('vinyl-source-stream');
 const replace = require('gulp-replace');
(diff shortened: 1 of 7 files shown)


alexisgrow (approved):
LG!

cea2aj (approved):
Looks good

PR #1018, opened 2020-08-12
alternativeverticals: update universal URL on context change

creotutar:
ANSWERS.setContext should update the experience URLs on the page to
include the context so that it persists throughout the experience. This
was done for the vertical suggestions in the Alternative Verticals
component but not for the universal link. Update this link as well.

Implementation detail: Because the full universal URL was previously passed
in from the Vertical Results (parent of Alternative Verticals) as a static string, we
instead pass in the base universal URL to the AV component and calculate
the query params dynamically every time the context is changed.

Nothing is meant to change with the universal URL behavior except for adding
context when the context is changed.

J=SLAP-529
TEST=manual

On a local Jambo site and a local site using the SDK directly, check to
see that the a call to `ANSWERS.setContext({'state': 'hi'})` adds the
context query parameter to the links in the Alternative Verticals
universal link, "view results across all search categories", by default.

### DIFF
--- a/src/ui/components/results/alternativeverticalscomponent.js
+++ b/src/ui/components/results/alternativeverticalscomponent.js
@@ -47,10 +47,19 @@ export default class AlternativeVerticalsComponent extends Component {
     );
 
     /**
-     * The url to the universal page to link back to with current query
+     * The url to the universal page to link back to without query params
      * @type {string|null}
      */
-    this._universalUrl = opts.universalUrl || '';
+    this._baseUniversalUrl = opts.baseUniversalUrl || '';
+
+    /**
+     * The url to the universal page to link back to with current query params
+     * @type {string|null}
+     */
+    this._universalUrl = this._getUniversalURL(
+      this._baseUniversalUrl,
+      new SearchParams(window.location.search.substring(1))
+    );
 
     /**
      * Whether or not results are displaying, used to control language in the info box
@@ -64,6 +73,10 @@ export default class AlternativeVerticalsComponent extends Component {
         this._verticalsConfig,
         this.core.globalStorage.getState(StorageKeys.API_CONTEXT),
         this.core.globalStorage.getState(StorageKeys.REFERRER_PAGE_URL)
+      );
+      this._universalUrl = this._getUniversalURL(
+        this._baseUniversalUrl,
+        new SearchParams(window.location.search.substring(1))
       );
       this.setState(this.core.globalStorage.getState(StorageKeys.ALERNATIVE_VERTICALS));
     });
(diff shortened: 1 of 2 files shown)


src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -150,4 +163,36 @@ export default class AlternativeVerticalsComponent extends Component {
...
+
+  /**
+   * Adds parameters that are dynamically set. Removes parameters for facets,
+   * filters, and pagination, which should not persist across the experience.
+   * @param {string} baseUrl The url append the appropriate params to. Note:
+   *    params already on the baseUrl will be stripped
+   * @param {SearchParams} params The parameters to include in the experience URL
+   * @return {string} The formatted experience URL with appropriate query params
+   */
+  _getUniversalURL (baseUrl, params) {
```
creotutar, line 175: This is copied from vertical results. I believe this a not-fun example of having to copy these transformations in a few places (context/referrerPageUrl/filtering params).

This exists in a few places: VerticalResults, Navigation, AV now. I didn't refactor into its own function because of this dependency on the `this.componentManager`. There is also a weird inconsistency in the lib where `tabOrder` is added in all those places except for the universal results for both VR and AV. So what goes into this new "`getExperienceLink`" function wouldn't include tabOrder.

LMK if you want this made into another function to DRY up, I also didn't know if this was in scope for a hotfix or if we wanted to get it into develop instead later (as it would mean changing navigationcomponent code outside of this fix).

cea2aj: I think it's best to DRY this and put it in urlutils. We can get around the this.componentManager dependency with a bind, and we can make a second function that includes tabOrder.

oshi97: maybe the shared method could also just take in getComponentNamesForComponentTypes as a param? (we're already passing it as a param into filterParamsForExperienceUrl anyways)

src/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -150,4 +163,36 @@ export default class AlternativeVerticalsComponent extends Component {
 
     return verticals;
   }
+
+  /**
+   * Adds parameters that are dynamically set. Removes parameters for facets,
+   * filters, and pagination, which should not persist across the experience.
+   * @param {string} baseUrl The url append the appropriate params to. Note:
+   *    params already on the baseUrl will be stripped
```
tmeyer2115: Indent so flush with "The url ..."

tmeyer2115 (approved):

src/ui/components/results/alternativeverticalscomponent.js
creotutar: This will be addressed in SLAP-570

PR #1019, opened 2020-08-12

creotutar:

When you land on a page, previously a query param would be added to

When you land on a page with no facet query parameters in the URL,

parameters.

J=SLAP-529
TEST=manual

Test multiple searches and make sure you can back nav through all
searches
Test search, apply facet, search, make sure you can back nav through
all states

Test multiple searches and make sure you can back nav through all
searches
Test search, apply facet, search, make sure you can back nav through
all states
Test search, apply filter, search, make sure you can back nav through
all states
On a universal page

Test multiple searches and make sure you can back nav through all
searches
Test landing with a query
page before landing.

On a vertical page ("people" verticalKey for me)
        Clicking on a Facet filter option and then clicking apply pushes
        on state to history.
        Landing on the page with Facets and a query does not push state
        when the facet query parameters are added. You can back nav from
        the landed page to whatever page you were on before in only one
        click.

        Searching in a sequence produces the correct history state:
        On State 1
        Navigate and land on empty vertical,
        On State 2
        Search all
        On State 3
        Search max
        On State 4
        Search all
        On State 5
        Click on Facet
        On State 6
        Click Apply
        On State 7

        If you back nav in the browser, you should be able to get to all
        7 states.

### DIFF
--- a/src/ui/storage/persistentstorage.js
+++ b/src/ui/storage/persistentstorage.js
@@ -110,8 +110,9 @@ export default class PersistentStorage {
 
   /**
    * Get a value for a given key in storage
+   * @param {string} key The unique key to get value for
    */
-  get (query) {
-    return this._params.get(query);
+  get (key) {
+    return this._params.get(key);
   }
 }
--- a/src/ui/storage/persistentstorage.js
+++ b/src/ui/storage/persistentstorage.js
@@ -1,5 +1,6 @@
 import SearchParams from '../dom/searchparams';
 import { AnswersStorageError } from '../../core/errors/errors';
+import { equivalentParams } from '../../core/utils/urlutils';
 
 /** @module PersistentStorage */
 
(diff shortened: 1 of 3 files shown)


src/ui/storage/persistentstorage.js
```diff
@@ -102,4 +107,11 @@ export default class PersistentStorage {
     }
     return allParams;
   }
+
+  /**
+   * Get a value for a given key in storage
+   */
+  get (query) {
```
alexisgrow: nit: should this param be renamed? Is it always the query or is it a given key?

creotutar: copied from `ui/com/searchparams.js` but can change to key if that is more useful!

alexisgrow (approved):

alexisgrow (approved):

PR #1020, opened 2020-08-13

creotutar:

1. When you land on a page, previously a query param would be added to

2. When you land on a page with no facet query parameters in the URL,

parameters.

J=SLAP-529
TEST=manual

 - Test multiple searches and make sure you can back nav through all
   searches
 - Test search, apply facet, search, make sure you can back nav through
   all states

 - Test multiple searches and make sure you can back nav through all
   searches
 - Test search, apply facet, search, make sure you can back nav through
   all states
 - Test search, apply filter, search, make sure you can back nav through
   all states

On a universal page
 - Test multiple searches and make sure you can back nav through all
   searches

Test landing with a query
page before landing.

### DIFF
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -674,6 +674,10 @@ export default class FilterOptionsComponent extends Component {
 
   saveFilterToPersistentStorage () {
     const replaceHistory = (this.core.persistentStorage.get(this.name) === null);
-    this.core.persistentStorage.set(this.name, this.config.options.filter(o => o.selected).map(o => o.label), replaceHistory);
+    this.core.persistentStorage.set(
+      this.name,
+      this.config.options.filter(o => o.selected).map(o => o.label),
+      replaceHistory
+    );
   }
 }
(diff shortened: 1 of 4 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -671,4 +671,9 @@ export default class FilterOptionsComponent extends Component {
     // AND all of the ORed together nodes.
     return FilterNodeFactory.and(...totalFilterNodes);
   }
+
+  saveFilterToPersistentStorage () {
+    const replaceHistory = (this.core.persistentStorage.get(this.name) === null);
+    this.core.persistentStorage.set(this.name, this.config.options.filter(o => o.selected).map(o => o.label), replaceHistory);
```
alexisgrow: nit: line length, would it be possible to put each param on a new line?

alexisgrow (approved):
LGTM, one nit

src/ui/storage/persistentstorage.js
```diff
@@ -66,6 +66,9 @@ export default class PersistentStorage {
       return;
     }
 
+    console.log('updating history w replaceHistory ' + replaceHistory);
+    console.log('before ' + currentParams);
+    console.log('after ' + this._params);
```
alexisgrow: Are these still needed?

creotutar: why are those there.........

alexisgrow (approved):

PR #1021, opened 2020-08-13

creotutar:

When you land on a page, previously a query param would be added to

When you land on a page with no facet query parameters in the URL,

parameters.

J=SLAP-529
TEST=manual

Test multiple searches and make sure you can back nav through all
searches
Test search, apply facet, search, make sure you can back nav through
all states

Test multiple searches and make sure you can back nav through all
searches
Test search, apply facet, search, make sure you can back nav through
all states
Test search, apply filter, search, make sure you can back nav through
all states
On a universal page

Test multiple searches and make sure you can back nav through all
searches
Test landing with a query
page before landing.

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -55,3 +55,24 @@ export function addParamsToUrl (url, params = {}) {
   }
   return url.split('?')[0] + '?' + urlParams;
 }
+
+/**
+ * returns if two SearchParams objects have the same key,value entries
+ * @param {SearchParams} params1
+ * @param {SearchParams} params2
+ * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
+ */
+export function equivalentParams (params1, params2) {
+  const entries1 = Array.from(params1.entries());
+  const entries2 = Array.from(params2.entries());
+
+  if (entries1.length !== entries2.length) {
+    return false;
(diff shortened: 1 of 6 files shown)


src/core/utils/urlutils.js
```diff
@@ -55,3 +55,24 @@ export function addParamsToUrl (url, params = {}) {
...
   return url.split('?')[0] + '?' + urlParams;
 }
+
+/**
+ * returns if two SearchParams objects have the same key,value entries
+ * @param {SearchParams} params1
+ * @param {SearchParams} params2
+ * @return {boolean} true if params1 and params2 have the same key,value entries, false otherwise
+ */
+export function equivalentParams (params1, params2) {
```
alexisgrow, line 65: Would it make sense for there to be an "isEqual" function in the "SearchParams" class?

creotutar: sadly SearchParams returns either a native implementation or our own implementation. Thus, we would only have this `isEqual` function if we were on an unsupported browser (e.g. IE11).

alexisgrow: Oh I didn't know that 😮  in that case, this works!

alexisgrow (approved):

alexisgrow (approved):

PR #1022, opened 2020-08-13
search: respect back navigation with empty search

creotutar:
Consider the example below

Land on /index.html (e.g. https://answers.yext.com/?)
In State 1
Search "all"
In State 2
Back navigation in browser history
In State 3

You would expect state 3 == state 1 (in terms of results shown on the page and the query surfaced in the search bar). However, this is not the case, we do not make changes when navigating back into an empty query. Instead state 3 == state 2 in the above scenario. We make fixes here to make it so that state 3 == state 1 (as closely as possible).

There were a few things we need to change.

First, `window.onpopstate` updates current globalStorage by adding any new/updated query params to the globalStorage. It does not remove query params that were in e.g. state 2 but not in state 1 (in our case, `query`). In `setAll`, if we see that there was a query and no longer a query, we remove it. This is something we can iterate on for all possible parameters that should be reset onpopstate.

Second, we must differentiate between an empty search (purposeful) and a null search (landing). This is because these are two separate states. When searchcomponent is listening for updates to StorageKeys.QUERY, query can be `null` when a user lands on the page or when they back nav into the landing page. It can be empty when a user sets that as the `defaultInitialSearch`, the `query` param is `query=`, the user purposefully searches with no query in the searchbar. The latter can show all results, depending on the backend configuration whereas the former should always show a blank experience. We were conflating these definitions.

Third, we must re-do a defaultInitialSearch, if it exists. This requires making searchcomponent cognizant of the ANSWERS.init searchConfig.

Note: currently, (in master)
1. on UniversalResults, if `allowEmptySearch: true`, clicking search with an empty query does *not* show all results
2. on VerticalResults, if landing with `query=` in the parameters and `allowEmptySearch: true`, does *not* show all results

J=SLAP-562
TEST=manual

Test the above example.
Test on a local Jambo site, test on a local client Jambo site, test on a local site referencing the SDK directly.

Test starting from an empty universal search, starting from an empty vertical search:
* Searching empty returns all results if `allowEmptySearch` is true
* Searching for a direct answer and then moving back to empty works
* Searching for alternative verticals and then moving back to empty works
* Searching for spell check and then moving back to empty works
* Searching for a direct answer and then moving back to empty works

Test landing with `query=test` `query=` and no `query` query parameter works as expected.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -8,9 +8,7 @@ import QuestionSubmission from './models/questionsubmission';
 import SearchIntents from './models/searchintents';
 import Navigation from './models/navigation';
 import AlternativeVerticals from './models/alternativeverticals';
-import SpellCheck from './models/spellcheck';
 import DirectAnswer from './models/directanswer';
-import DynamicFilters from './models/dynamicfilters';
 import LocationBias from './models/locationbias';
 
 import StorageKeys from './storage/storagekeys';
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -233,6 +233,7 @@ export default class Core {
   }
 
   clearResults () {
+    this.globalStorage.set(StorageKeys.QUERY, null);
     this.globalStorage.set(StorageKeys.QUERY_ID, '');
     this.globalStorage.set(StorageKeys.RESULTS_HEADER, {});
     this.globalStorage.set(StorageKeys.SPELL_CHECK, {}); // TODO has a model but not cleared w new
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -2,10 +2,18 @@
 
 import SearchDataTransformer from './search/searchdatatransformer';
 
-import StorageKeys from './storage/storagekeys';
 import VerticalResults from './models/verticalresults';
 import UniversalResults from './models/universalresults';
 import QuestionSubmission from './models/questionsubmission';
+import SearchIntents from './models/searchintents';
+import Navigation from './models/navigation';
+import AlternativeVerticals from './models/alternativeverticals';
+import SpellCheck from './models/spellcheck';
+import DirectAnswer from './models/directanswer';
+import DynamicFilters from './models/dynamicfilters';
+import LocationBias from './models/locationbias';
+
+import StorageKeys from './storage/storagekeys';
 import AnalyticsEvent from './analytics/analyticsevent';
 import FilterRegistry from './filters/filterregistry';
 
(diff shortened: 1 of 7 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -138,6 +138,15 @@ export default class SearchComponent extends Component {
       if (this.queryEl) {
         this.queryEl.value = q;
       }
+      if (q === null) {
+        this.core.globalStorage.set(StorageKeys.VERTICAL_RESULTS, {
```
creotutar: This is not an exhaustive list. There are probably a million filters I'm missing. Do we have a better way of clearing everything for this case? Note: we use `searchState: 'search-complete'` o/w the VerticalResults don't update. (Can use the const later, just hardcoded string for PoC)

cea2aj: Thoughts on creating a method setEmptySearch inside global storage that performs the clearing?

creotutar: 👍 
put in core because core has most of the logic about which globalStorage keys are changed on search

cea2aj (approved):
LGTM!

src/core/core.js
```diff
@@ -226,6 +232,22 @@ export default class Core {
       });
   }
 
+  clearResults () {
+    this.globalStorage.set(StorageKeys.QUERY, null);
```
oshi97, line 236: does using globalStorage.delete work here?

src/ui/components/search/searchcomponent.js
```diff
@@ -138,6 +145,12 @@ export default class SearchComponent extends Component {
       if (this.queryEl) {
         this.queryEl.value = q;
       }
+      if (q === null) {
```
oshi97, line 151: What happens if somebody has a default initial search, makes a search, then clears the searchbar? I think we also just want the check to be
```js
if (!q && this._defaultInitialSearch)
```
it feels a little weird to care about things being null vs undefined

src/core/core.js
creotutar: it does not. currently `delete`s do not emit an update, only `set`s currently.

src/ui/components/search/searchcomponent.js
creotutar: I'm not sure I understand.

If somebody has a default initial search, makes a search, then clears the searchbar, the initial default search is re-done. I'm wary of the `!q` because it stops us from allowing empty search. If someone were to press enter on an empty searchbar we would want to search with an empty query as normal (note the `return`). We would not want to search, though, if the searchbar is emptied through the clear button or through a back nav, right?

oshi97 (approved):

src/ui/components/search/searchcomponent.js
oshi97: oh shoot my bad

oshi97 (approved):

oshi97 (approved):

PR #1025, opened 2020-08-17
alternativeverticals: update universal URL on context change (#1018)

creotutar:
* alternativeverticals: update universal URL on context change

ANSWERS.setContext should update the experience URLs on the page to
include the context so that it persists throughout the experience. This
was done for the vertical suggestions in the Alternative Verticals
component but not for the universal link. Update this link as well.

Implementation detail: Because the universal URL was previously passed
in from the Vertical Results (parent of Alternative Verticals), we
instead pass in the base universal URL to the AV component and calculate
the query params dynamically every time the context is changed.

Nothing is meant to change with the universal URL behavior except for adding
context when the context is changed.

J=SLAP-529
TEST=manual

On a local Jambo site and a local site using the SDK directly, check to
see that the a call to `ANSWERS.setContext({'state': 'hi'})` adds the
context query parameter to the links in the Alternative Verticals
universal link, "view results across all search categories", by default.

* Fix multiline jsdoc comment

### DIFF
--- a/src/ui/components/results/alternativeverticalscomponent.js
+++ b/src/ui/components/results/alternativeverticalscomponent.js
@@ -47,10 +47,19 @@ export default class AlternativeVerticalsComponent extends Component {
     );
 
     /**
-     * The url to the universal page to link back to with current query
+     * The url to the universal page to link back to without query params
      * @type {string|null}
      */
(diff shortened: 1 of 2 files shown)


creotutar:
Approved in #1018

PR #1028, opened 2020-08-18
handlebarsrenderer: add runtimeTranslation helper

oshi97:
this commit adds the runtimeTranslation helper to the handlebars
renderer. this is a simple helper that combines the passed in
translation phrase, which contains the phrase to be translated as
either an object containing the phrases plural forms or just a string,
with interpolated values.

J=SLAP-558
TEST=auto

Added unit tests for handlebarsrenderer


oshi97:
will rebase this to use #1029 after #1029 is merged in to avoid duplicating code

cea2aj (approved):

PR #1033, opened 2020-08-19
translator.js: handle multiple plural forms

oshi97:
update translate helpers to reflect changes
from https://github.com/yext/jambo/pull/85

J=SLAP-575
TEST=auto

updated unit tests

### DIFF
--- a/src/core/i18n/translator.js
+++ b/src/core/i18n/translator.js
@@ -1,4 +1,4 @@
-import { getNPlurals, getPluralFunc } from 'plural-forms/dist/minimal';
+import { getNPlurals, getPluralFunc, hasLang } from 'plural-forms/dist/minimal';
 
 export default class Translator {
   /**
@@ -36,7 +36,10 @@ export default class Translator {
    * @returns {string}
    */
   static _selectPluralForm (parsedTranslations, count) {
-    const locale = parsedTranslations.locale;
+    let locale = parsedTranslations.locale;
+    if (!hasLang(locale)) {
+      locale = 'en';
+    }
     const oneToNArray = this._generateArrayOneToN(locale);
     const pluralFormIndex = getPluralFunc(locale)(count, oneToNArray);
     return parsedTranslations[pluralFormIndex];
--- a/src/core/i18n/translator.js
+++ b/src/core/i18n/translator.js
@@ -1,3 +1,5 @@
+import { getNPlurals, getPluralFunc } from 'plural-forms/dist/minimal';
+
 export default class Translator {
   /**
    * Performs a translation which supports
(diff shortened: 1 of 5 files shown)


src/core/i18n/translator.js
```diff
@@ -11,13 +13,42 @@ export default class Translator {
...
   }
 
+  /**
+   * If translations is json, parse it and choose the correct plural form.
+   * Otherwise just return the non-interpolated translation string.
+   * @param {string} translations
+   * @param {number} count
+   * @returns {string}
+   */
   static _selectPluralization (translations, count) {
```
cea2aj: Maybe this could be renamed to _selectStringToInterpolate since the selection of the plural form is now happening inside _selectPluralForm

src/core/i18n/translator.js
```diff
@@ -11,13 +13,42 @@ export default class Translator {
...
+   * Returns the correct plural form given a parsed translations object and count.
+   * @param {Object} parsedTranslations
+   * @param {number} count
+   * @returns {string}
+   */
+  static _selectPluralForm (parsedTranslations, count) {
+    const locale = parsedTranslations.locale;
+    const oneToNArray = this._generateArrayOneToN(locale);
+    const pluralFormIndex = getPluralFunc(locale)(count, oneToNArray);
+    return parsedTranslations[pluralFormIndex];
```
cea2aj: Thoughts on utilizing the hasLang function from plural-forms to validate the locale?

src/core/i18n/translator.js
oshi97: sure, updated!

src/core/i18n/translator.js
oshi97: sure I kind of like that. I hope we don't ever need this but we did have a case in the past where some client wanted to use a "fake" locale in our pages system

cea2aj (approved):
LG!

src/core/i18n/translator.js
```diff
@@ -7,17 +9,49 @@ export default class Translator {
...
+  /**
+   * Returns the correct plural form given a parsed translations object and count.
+   * @param {Object} parsedTranslations
+   * @param {number} count
+   * @returns {string}
+   */
+  static _selectPluralForm (parsedTranslations, count) {
+    let locale = parsedTranslations.locale;
+    if (!hasLang(locale)) {
+      locale = 'en';
```
cea2aj, line 41: nit: We could make a default locale property in the Translator class and refer to that

oshi97: oh hmm right now the entire class is static so I think we would need to add a constructor and make this an instance method?

PR #1038, opened 2020-08-24
Create a BundleAbstractFactory class that provides bundle() factory m…

tmeyer2115:
…ethods.

This PR adds the BundleAbstractFactory class. This class provides factory
methods for building the different types of SDK bundle: modern, legacy-iife,
and legacy-umd. With this class, we can simplify the library.gulpfile.js
file. That file was getting too bulky and the work required for creating
translated SDK bundles would have only made this worse.

In the future, we can further improve things by creating a Bundler class. This
class will have a bundle() method. The super class can have any shared methods.
Each bundle type will have its own Bundler implementation. The abstract factory
will then provide a Bundler.

TEST=manual

For the same ref, built all the SDK assets with and without my changes. Using
a diff, ensured that the generated assets were the same in both cases.

### DIFF
--- a/conf/gulp-tasks/bundle/bundleabstractfactory.js
+++ b/conf/gulp-tasks/bundle/bundleabstractfactory.js
@@ -23,18 +23,18 @@ Object.freeze(BundleType);
  * A class providing factories for building different types of SDK bundle.
  */
 class BundleAbstractFactory {
-  constructor() {
+  constructor () {
     this._namespace = 'ANSWERS';
   }
 
   /**
    * Creates a factory for building a specific type of SDK bundle.
-   * 
+   *
    * @param {BundleType} bundleType The type of SDK bundle to build.
    * @returns {Function} Factory method for building an SDK bundle of the
    *                     the specified type.
    */
-  create(bundleType) {
+  create (bundleType) {
     switch (bundleType) {
       case BundleType.MODERN:
         return () => this._modernBundle();
@@ -49,10 +49,10 @@ class BundleAbstractFactory {
 
   /**
    * A factory method for producing the modern SDK bundle.
-   * 
+   *
    * @returns {File} The modern SDK bundle.
    */
-  _modernBundle() {
+  _modernBundle () {
     const rollupConfig = {
       input: './src/answers-umd.js',
       output: {
@@ -81,10 +81,10 @@ class BundleAbstractFactory {
 
   /**
    * A factory method for producing the legacy, IIFE-style SDK bundle.
-   * 
+   *
    * @returns {File} The legacy, IIFE-style SDK bundle.
    */
-  _legacyBundleIIFE() {
+  _legacyBundleIIFE () {
     const legacyBundleConfig = {
       format: 'iife',
       name: this._namespace,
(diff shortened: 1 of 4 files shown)


conf/gulp-tasks/bundle/bundleabstractfactory.js
```diff
@@ -0,0 +1,188 @@
...
+      .pipe(dest('dist'));
+  }
+
+  /**
+   * Attempts to compute the correct library version for an SDK bundle.
+   *
+   * @returns {string} The SDK bundle's library version.
+   */
+  _getLibVersion () {
+    try {
```
oshi97: Nit but can we assign library version to a variable so it's not recomputed every time time it's called, or maybe just have it be something that's passed into this class?

conf/gulp-tasks/bundle/bundleabstractfactory.js
```diff
@@ -0,0 +1,188 @@
...
+class BundleAbstractFactory {
+  constructor () {
+    this._namespace = 'ANSWERS';
+  }
+
+  /**
+   * Creates a factory for building a specific type of SDK bundle.
+   *
+   * @param {BundleType} bundleType The type of SDK bundle to build.
+   * @returns {Function} Factory method for building an SDK bundle of the
```
oshi97: Would it be more accurate to call these gulp tasks instead of factory methods?

conf/gulp-tasks/bundle/bundleabstractfactory.js
```diff
@@ -0,0 +1,188 @@
...
+   *                     the specified type.
+   */
+  create (bundleType) {
```
oshi97: Would it be better just to have individual methods for these, instead of a switch statement + enum?

conf/gulp-tasks/bundle/bundleabstractfactory.js
```diff
@@ -0,0 +1,188 @@
...
+    return this._legacyBundle(legacyBundleConfig, 'answers-umd.js');
+  }
+
+  /**
+   * A helper method that can produce either variant of the legacy SDK bundle.
+   *
+   * @param {Object<string, ?>} outputConfig Any variant-specific configuration
+   *                                         for the legacy bundle.
+   * @param {string} fileName The filename of the created bundle.
+   * @returns {File} The legacy SDK bundle.
```
oshi97: I don't think this returns a file, it looks like it returns whatever .pipe() returns, which it looks like is a stream.Writable? (not too familiar with node streams)

conf/gulp-tasks/bundle/bundleabstractfactory.js
tmeyer2115: In the Factory design pattern, there's usually a single entry point, like this one. The method takes in a type and spits out the correct version of whatever it is the factory is building. One of the ideas behind this is that it really simplifies the interface for the caller. There's a single method only, they give it a type, it gives them the right object of that type. They don't need to know which particular individual method to call. In this way, the Factory is a bit like the facade design pattern.

oshi97 (approved):

PR #1042, opened 2020-08-25
Add preliminary infrastructure for providing localized assets.

tmeyer2115:
This PR adds a number of classes that will support the creation of localized
SDK assets. The classes added are:

- Translator: A class, backed by i18next, that supplies translations. These
  translations can be used either at run-time or build-time.
- LocalFileParser: A class that can read in a .PO file for a locale and convert
  it into a form understood by i18next and the Translator.
- TranslationPlaceholder: A data model representing content in SDK source that
  needs translation.
- TranslationResolver: A class that resolves TranslationPlaceholders to either
  a translated string or a run-time call to provide that string.

Eventually, test coverage will be added for these various classes. Tests were
omitted for now because these classes may be subject to change as we're just
starting our SDK localization effort.

TEST=manual

Manually generated a French version of the SDK (with a single run-time
translation) using this infrastructure.

### DIFF
--- a/conf/i18n/translator.js
+++ b/conf/i18n/translator.js
@@ -8,10 +8,10 @@ const i18next = require('i18next');
 class Translator {
   /**
    * Creates a new {@link Translator} that wraps the provided {@link i18next} instance.
-   * 
+   *
    * @param {i18next} i18nextInstance The instance to wrap.
    */
-  constructor(i18nextInstance) {
+  constructor (i18nextInstance) {
     this._i18next = i18nextInstance;
   }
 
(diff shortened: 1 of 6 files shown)


conf/i18n/translator.js
```diff
@@ -0,0 +1,156 @@
...
+            pluralForms[pluralFormIndex] = localeTranslations[translationKey];
+            return pluralForms;
+          },
+          { 1: localeTranslations[phrase] });
+    }
+
+    // If no translations can be found, we return a map containing the provided
+    // singular and plural forms.
+    return {
+      1: phrase,
```
oshi97: oh this is slightly out of date (probably at least part of why runtimeTranslation interpolation wasn't working) I think copying and pasting the translator.js in the most recent jambo localized-pages branch should work though

tmeyer2115: Womp. That makes sense, the danger of duplicating code. Updated!

alexisgrow (approved):
this LGTM!

oshi97 (approved):

PR #1043, opened 2020-08-25
Remove inputAndButtons inline partial from SearchBar's template.

tmeyer2115:
This was done to make it such that the SearchBar component can be used on a
site that does not allow unsafe-eval in its Content Security Policy. We found
that Handlebars makes use of eval in a few cases. One of those is when a
template uses an in-line helper. We did not update other usages of inline
partials in the SDK. This is because most Answers implementations use an
iframe'd experience. In these situations, the SearchBar component would be the
only one subject to the client's CSP.

J=SLAP-585
TEST=auto

Disabled eval on a local site. Verified that I could still interact with the
SearchBar.

### DIFF
--- a/src/ui/templates/search/search.hbs
+++ b/src/ui/templates/search/search.hbs
@@ -7,62 +7,103 @@
   <div class="yxt-SearchBar-container">
     {{#if useForm}}
       <form class="yxt-SearchBar-form">
-        {{> inputAndButtons}}
-      </form>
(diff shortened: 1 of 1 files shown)


alexisgrow (approved):
LGTM!

oshi97:
inline partials are my favorite though...

bryanreed (approved):

PR #1048, opened 2020-08-26
Create deep copies of Component state using lodash.cloneDeep.

tmeyer2115:
This PR updates the Component class so that copies of the state, made prior
to calls to transformData, are made with lodash.cloneDeep. Previously, copies
were made using JSON.parse(JSON.stringify). This old approach had some issues.
For starters, it is limited by what JSON can serialize intelligently. Second,
if the object has a circular structure, JSON serialization will fail. We found
that when the SearchBar's container was managed with React, its child
components had a state with a circular strucutre. This caused errors when
trying to make copies of their state. Effectively, this blocked the SDK from
being used with React.

The lodash.cloneDeep method is able to handle circular structures by default.
It is also not limited by what can be serialized intelligibly with JSON.

J=SLAP-572
TEST=auto,manual

Reran all unit tests. Created a test site where the SearchBar's container was
created by React. The SearchBar redirected to an iframe'd experience. I was
able to use the SearchBar and all its child components without issue. As a
sanity check, I recreated the same site without React to ensure that worked
as expected.

To test that the cloneDeep worked as anticipated, I added some print statements
to the SDK. These ensured that I was not altering component state when using
transformData. I was operating of a full copy.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -5,7 +5,7 @@ module.exports = [
   },
   {
     path: 'dist/answers-modern.min.js',
-    limit: '100kb'
+    limit: '105kb'
   },
   {
     path: 'dist/answerstemplates.compiled.min.js',
--- a/package-lock.json
+++ b/package-lock.json
@@ -10506,6 +10506,11 @@
       "integrity": "sha1-DM8tiRZq8Ds2Y8eWU4t1rG4RTZ0=",
       "dev": true
     },
+    "lodash.clonedeep": {
+      "version": "4.5.0",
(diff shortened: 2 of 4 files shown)


alexisgrow (approved):
🔥🔥🔥

oshi97 (approved):
Would it be weird to add a really small react jest test with the searchbar?

tmeyer2115:
> Would it be weird to add a really small react jest test with the searchbar?

Nope, wouldn't be weird at all! If it's alright, I will make an item for this as we want to get this shipped.

PR #1049, opened 2020-08-26
Version 1.5.4

tmeyer2115:
# Version 1.5.4

## Changes
* We made a fix to the `SearchBar`'s template to ensure the component could be used on a site whose Content Security Policy disallows `unsafe-eval`. (#1043)
* We changed how the `Component` class makes a copy of internal state. The previous algorithm clashed with React and effectively prevented the SDK from being used with it. (#1048)

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -5,7 +5,7 @@ module.exports = [
   },
   {
     path: 'dist/answers-modern.min.js',
-    limit: '100kb'
+    limit: '105kb'
   },
   {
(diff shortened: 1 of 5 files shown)


oshi97 (approved):
lgtm! for the official release notes could we add the commit number next to each change?

tmeyer2115:
> lgtm! for the official release notes could we add the commit number next to each change?

Yep! Done.

PR #1052, opened 2020-08-27
localized template build process

oshi97:
This commit does some refactoring of the template gulpfile,
and also implements preprocessing of translate helpers.
This is done using the processAST option of gulp-handlebars,
to mutate the source templates, along with Handlebars.Visitor.
Each desired locale gets its own gulp chain.

in order to handle gulp tasks that were both async and
returned a stream, `.on('end', callback)` was added to
necessary streams.

J=SLAP-605
TEST=manual

added the below to the standard card template
```hbs
      <div class="eat some snails">
        {{translate phrase='Eat some snails'}}
      </div>
      <div class="profile desc">
        {{translate phrase='Profile Description [[desc]]' desc=verticalKey}}
      </div>
      <div class="some items">
        {{translate phrase='Some item [[count]]' pluralForm='Some items [[count]]' count=count}}
      </div>
      <div class="child">
        {{translate phrase='I am looking for my child named [[name]]' context='male' name=_config.subtitle}}
      </div>
```
and used the below po file
```po
msgid "Eat some snails"
msgstr "manger des escargots"

msgid "Some item [[count]]"
msgid_plural "Some items [[count]]"
msgstr[0] "Un articles [[count]]"
msgstr[1] "Certain articles [[count]]"

msgid "Profile Description [[desc]]"
msgstr "Description du profil [[desc]]"

msgctxt "male"
msgid "I am looking for my child named [[name]]"
msgstr "Je cherche mon fils nommé [[name]]"
```

saw that things were translated correctly (also added count=42 to
the cards setState()) i.e. the card had text

manger des escargots
Description du profil people
Certain articles 42
Je cherche mon fils nommé Samuel Stanton

tested that npm run dev still watches files correctly

tested that above translations also work when locale is english without a po file


conf/gulp-tasks/template/bundle.js
```diff
@@ -0,0 +1,60 @@
...
+    {
+      format: 'umd',
+      name: 'TemplateBundle',
+      exports: 'named'
+    },
+    filenameUMD,
+    filenamePrecompiled
+  );
+}
+exports.bundleTemplatesIIFE = function (done, filenameIIFE, filenamePrecompiled) {
```
tmeyer2115: We weren't building IIFE template bundles before, were we? Thanks for adding! This should help.

conf/gulp-tasks/template/bundle.js
```diff
@@ -0,0 +1,60 @@
...
+    done,
+    {
+      format: 'iife',
+      name: 'TemplateBundle'
+    },
+    filenameIIFE,
+    filenamePrecompiled
+  );
+}
+function bundleTemplates (done, outputConfig, fileName, filenamePrecompiled) {
```
tmeyer2115: Add spaces in between methods. Could we add some JSDoc here? It's relatively clear what the last three args do, but it's not too clear what `done` does. Might also be worth renaming `done` to be more descriptive.

conf/gulp-tasks/template/precompile.js
```diff
@@ -0,0 +1,51 @@
+const { src, dest } = require('gulp');
+
+const handlebars = require('gulp-handlebars');
+const concat = require('gulp-concat');
+const declare = require('gulp-declare');
+const wrap = require('gulp-wrap');
+
+function precompileTemplates (done, outputFile, processAST) {
```
tmeyer2115: Add some JSDoc here.

conf/gulp-tasks/template/templatetaskfactory.js
```diff
@@ -0,0 +1,80 @@
...
+    this.locale = locale;
+    this.translationResolver = translationResolver;
+    if (locale === 'en') {
+      this.filenamePrecompiled = 'answerstemplates.precompiled.min.js';
+      this.filenameUMD = 'answerstemplates.compiled.min.js';
+      this.filenameIIFE = 'answerstemplates-iife.compiled.min.js';
+    } else {
+      this.filenamePrecompiled = `${locale}-answerstemplates.precompiled.min.js`;
+      this.filenameUMD = `${locale}-answerstemplates.compiled.min.js`;
+      this.filenameIIFE = `${locale}-answerstemplates-iife.compiled.min.js`;
```
tmeyer2115: Minor nit, but let's refactor this to something like { filenamePrecompiled, filenameUMD, filenameIIFE } = _getFilenames(locale); The_getFilenames method can handle this conditional as well.

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+const { parallel, series, watch } = require('gulp');
+const path = require('path');
+const LocalFileParser = require('../i18n/localfileparser');
+const TemplateTaskFactory = require('./template/templatetaskfactory');
+const Translator = require('../i18n/translator');
+const TranslationResolver = require('../i18n/translationresolver');
+
+let devTaskFactory;
+const localFileParser = new LocalFileParser(path.join(__dirname, '../i18n/translations'));
+const devLocale = 'en';
```
tmeyer2115: Lets move this to some kind of constants file. I bet Alexis is going to want the same quantity: DEV_LOCALE. I was thinking we'd have a constants file where we define the support locales (and maybe have a map to their translation files). What do you think?

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+
+async function createTaskFactory (locale) {
+  const translation = await localFileParser.fetch(locale);
+  const translator = await Translator.create(locale, [], { [locale]: { translation } });
+  const translationResolver = new TranslationResolver(translator, translationResult => translationResult);
```
tmeyer2115: LL

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+  return new TemplateTaskFactory(translationResolver, locale);
+}
+
+async function watchTemplates () {
+  devTaskFactory = devTaskFactory || await createTaskFactory(devLocale);
```
tmeyer2115: Can't we just assign devTaskFactory = await createTaskFactory(devLocale) on line 8? That would save us this line and line 31

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+    devTaskFactory.bundleTemplatesUMD,
+    devTaskFactory.cleanFiles,
+    watchTemplates
+  )();
+}
+
+exports.dev = devTemplates;
+
+async function defaultTask () {
+  const locales = ['en'];
```
tmeyer2115: Yeah, I definitely think we want this in some shared constants file.

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+    devTaskFactory.precompileTemplates,
+    devTaskFactory.bundleTemplatesUMD,
+    devTaskFactory.cleanFiles,
+    watchTemplates
+  )();
+}
+
+exports.dev = devTemplates;
+
+async function defaultTask () {
```
tmeyer2115: Add some JSDoc to this

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -1,145 +1,58 @@
...
+    return series(
+      taskFactory.precompileTemplates,
+      parallel(
+        series(taskFactory.bundleTemplatesIIFE, taskFactory.minifyTemplatesIIFE),
+        series(taskFactory.bundleTemplatesUMD, taskFactory.minifyTemplatesUMD)
+      ),
+      taskFactory.cleanFiles
+    );
+  }));
+  return new Promise(resolve => parallel(...localizedTemplateTasks)(resolve));
```
tmeyer2115: Such cool guy. To make sure I follow:
- You create an array of Promises. Each promise resolves to a complete, localized template bundling task.
- The Promise.all forces all those promises to resolve, so we end up with a list of tasks, each for building localized template bundles.
- We execute all those tasks in parallel here. 

Is that right? Do we need to wrap the parallel call at the end in a Promise? I think if you have an async function, the return value is automagically wrapped in a Promise.

conf/i18n/models/translationplaceholder.js
```diff
@@ -61,6 +61,54 @@ class TranslationPlaceholder {
...
+   */
+  hasNoInterpolation () {
+    return Object.keys(this._interpolationValues).length === 0;
+  }
+
+  /**
+   * Creates a {@link TranslateInvocation} from a Handlebars MustacheStatement.
+   * @param {hbs.AST.MustacheStatement} mustacheStatement
+   */
+  static fromMustacheStatementNode (mustacheStatement) {
```
tmeyer2115: I think we might consider moving these two methods elsewhere. I think, as a rule, we should not add static constructors to a class unless they're relatively short. This one requires an additional helper method. With complex, involved static constructors, I feel the class begins to do too much and have too much responsibility. Maybe a class specifically for building placeholders?

conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * _MustacheStatement() is automatically called on any MustacheStatement nodes by the visitor.
+   * Returning undefined leaves the node unaffected, otherwise it replaces it with the
+   * returned value.
+   * @param {hbs.AST.MustacheStatement} statement
+   * @returns {hbs.AST.MustacheStatment|undefined}
+   */
+  _MustacheStatement (statement) {
+    if (!this._validHelpers.includes(statement.path.original)) {
```
tmeyer2115: Lets do some renaming here to make it clearer what's happening. We're looking to see if the statement's "path" corresponds to a translate helper

conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,163 @@
+const Handlebars = require('handlebars');
+const TranslationPlaceholder = require('./models/translationplaceholder');
+
+/**
+ * TranslateHelperVisitor accepts a handlebars AST, and replaces all translate placeholders
+ * with either static translations, or a runtime translation helper.
+ */
+class TranslateHelperVisitor {
```
tmeyer2115, line 9: Can this class extend Handlebars.Visitor? Or will that not work

conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,163 @@
...
+      typeof translatedPhrase === 'string' && placeholder.hasNoInterpolation();
+    if (canBeTranslatedStatically) {
+      return this._replaceHelperWithStaticTranslation(statement, translatedPhrase);
+    } else {
+      const renamedStatement = this._renameHelperToRuntimeTranslation(statement);
+      return this._updateHashPairsForRuntimeTranslations(renamedStatement, translatedPhrase);
+    }
+  }
+
+  _resolveTranslationPlaceholder (placeholder) {
```
tmeyer2115, line 65: JSDoc

conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,163 @@
...
+    this._contextParam = 'context';
+    this._pluralFormParam = 'pluralForm';
+  }
+
+  /**
+   * Dispatches TranslateHelperVisitor on a Handlebars AST.
+   * @param {hbs.AST.Program} ast
+   */
+  accept (ast) {
+    this._visitor.accept(ast);
```
tmeyer2115, line 31: Such dispatch.

conf/gulp-tasks/template/bundle.js
oshi97: Oh no Christian added this before! I just factored it out and made some small tweaks

conf/gulp-tasks/templates.gulpfile.js
oshi97: we cant do top level async await :(

conf/i18n/translatehelpervisitor.js
oshi97: I wrote it that way originally but then I remembered our hopy mantra of composition over inheritance. I was pretty split on which way was better, so if you prefer extending here I’ll swap it!

tmeyer2115: Gotcha. Would extending that class allow us to remove some boilerplate or not really change the class much? Yeah, we do usually favor composition over inheritance. If extending the class wouldn't really get us anything in terms of removing boilerplate, than yeah, let's compose.

conf/gulp-tasks/templates.gulpfile.js
oshi97: yup that looks right to me!

I wasn't getting proper async completion without passing `resolve` as a parameter to parallel on line 55. I doubt what I have is the best way to write it though, so I'll try to play around a little more and see if there's a way to avoid wrapping it in a promise constructor

conf/i18n/translatehelpervisitor.js
oshi97: it lets use remove the accept() passthrough but not much else is changed besides that

tmeyer2115: Gotcha, let's leave as is then.

conf/i18n/translatehelpervisitor.js
oshi97: yeah boi

conf/i18n/translatehelpervisitor.js
oshi97: added a variable!

conf/i18n/models/translationplaceholder.js
oshi97: moved to TranslationPlaceholderFactory

conf/gulp-tasks/templates.gulpfile.js
oshi97: Ok made some tweaks and I think what I have right now is pretty much the best I can come up with, now the the only async await function is the createTaskFactory() helper, and everything else uses regular promises (which feel more concrete to me)

conf/gulp-tasks/template/bundle.js
```diff
@@ -0,0 +1,79 @@
...
+    },
+    filenameUMD,
+    filenamePrecompiled
+  );
+}
+
+/**
+ * @param {Function} callback called when the stream ends
+ * @param {string} filenameIIFE the IIFE output filename
+ * @param {string} filenamePrecompiled the precompiled template file
```
tmeyer2115: returns here and below

conf/gulp-tasks/template/bundle.js
```diff
@@ -0,0 +1,79 @@
...
+const rollup = require('gulp-rollup-lightweight');
+const babel = require('rollup-plugin-babel');
+const resolve = require('rollup-plugin-node-resolve');
+const commonjs = require('rollup-plugin-commonjs');
+const builtins = require('rollup-plugin-node-builtins');
+
+const source = require('vinyl-source-stream');
+
+/**
+ * @param {Function} callback called when the stream ends
```
tmeyer2115: Is this callback invoked when the bundleTemplates task is done? I'm not sure which stream is being referred to.

conf/gulp-tasks/template/templatetaskfactory.js
```diff
@@ -14,67 +14,87 @@ class TemplateTaskFactory {
...
-
     this.cleanFiles = this.cleanFiles.bind(this);
+  }
+
+  /**
+   * Customize the gulp task names, e.g. "bound precompileTemplates" becomes "en-precompileTemplates".
+   * @param {string} locale
+   */
+  _customizeTaskNames (locale) {
+    Object.defineProperty(this.precompileTemplates, 'name', { value: `${locale}-precompileTemplates` });
```
tmeyer2115: LLs here for all the `defineProperty` calls. I think our LL is about 80ish chars (although this is not codified anywhere)

conf/gulp-tasks/template/templatetaskfactory.js
```diff
@@ -14,67 +14,87 @@ class TemplateTaskFactory {
...
-    Object.defineProperty(this.minifyTemplatesIIFE, 'name', { value: `${locale}-minifyTemplatesIIFE` });
-
     this.minifyTemplatesUMD = this.minifyTemplatesUMD.bind(this);
-    Object.defineProperty(this.minifyTemplatesUMD, 'name', { value: `${locale}-minifyTemplatesUMD` });
-
     this.cleanFiles = this.cleanFiles.bind(this);
+  }
+
+  /**
+   * Customize the gulp task names, e.g. "bound precompileTemplates" becomes "en-precompileTemplates".
```
tmeyer2115: LL

conf/gulp-tasks/template/templatetaskfactory.js
```diff
@@ -14,67 +14,87 @@ class TemplateTaskFactory {
   constructor (translationResolver, locale = 'en') {
```
tmeyer2115: I could have sworn I left this comment in the initial review, but now I can't find it anywhere. 

As it stands, this class doesn't follow the archetype of a Factory. The Factory design pattern is a bit prescriptive. Specifically, it exposes a single method to the consumer. This method takes in a type or some other enum and returns an object corresponding to that type. 

I think we need to update this class to fully follow the Factory pattern or name it differently. If we wanted a factory, you could create a TemplateTask enum or something like that. There'd be a method that takes in that enum and spits out the right gulp task for you.

conf/gulp-tasks/template/templatetaskfactory.js
oshi97: oh I thought it was 100 chars (these lines are also over 100 though)

conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * _MustacheStatement() is automatically called on any MustacheStatement nodes by the visitor.
+   * Returning undefined leaves the node unaffected, otherwise it replaces it with the
+   * returned value.
+   * @param {hbs.AST.MustacheStatement} statement
+   * @returns {hbs.AST.MustacheStatment|undefined}
+   */
+  _MustacheStatement (statement) {
+    const isTranslationHelper = this._validHelpers.includes(statement.path.original);
```
cea2aj: Can we add a verb to this function name?

conf/gulp-tasks/template/templatetaskfactory.js
oshi97: refactored into 3 smaller factories! after some thinking the idea of 1 factory doing one kind of thing makes a lot of sense (even if its more annoying to write initially)

conf/i18n/translatehelpervisitor.js
oshi97: sure I was trying to follow the method signature of Handlebars.Visitor but we don't have to be restrained by them

conf/gulp-tasks/template/filenameutils.js
```diff
@@ -0,0 +1,39 @@
...
+  return addLocalePrefix(fileName, locale);
+};
+
+/**
+ * Gets the precompiled templates fileName for the given locale.
+ * @param {string} locale
+ * @returns {string}
+ */
+exports.getPrecompiledFileName = function (locale) {
+  const precompiledFileName = 'answerstemplates.precompiled.min.js';
```
alexisgrow, line 40: Would it make sense for this file name to be in the `fileNames` const at the top of this file?

oshi97: I originally had it like that but it felt a little weird to have a TemplateType.PRECOMPILED since the precompiled template file isn't really a template type it's just an intermediary file. What do you think?

conf/gulp-tasks/template/bundletemplatestaskfactory.js
```diff
@@ -0,0 +1,78 @@
...
+  _createBundleFunction (templateType) {
+    const bundleConfig = this._getOutputConfig(templateType);
+    const outputFile = getFileName(templateType, this._locale);
+    const precompiledFile = getPrecompiledFileName(this._locale);
+    return callback =>
+      bundleTemplates(callback, bundleConfig, precompiledFile, outputFile);
+  }
+
+  /**
+   * Returns the output config for the specified {@link TemplateType}.
```
tmeyer2115, line 54: Space in between function description and params

conf/gulp-tasks/template/bundletemplatestaskfactory.js
```diff
@@ -0,0 +1,78 @@
+const bundleTemplates = require('./bundletemplates');
+const TemplateType = require('./templatetype');
+const {
+  addLocalePrefix,
+  getFileName,
+  getPrecompiledFileName
+} = require('./filenameutils');
+
+/**
+ * BundleTemplatesTaskFactory is a factory for creating SDK template bundling tasks.
```
tmeyer2115, line 19: What would you think about moving bundleTemplates into here? It feels a bit weird that we have this factory to create a template bundle task, but someone could also call that method directly to get such a task.

conf/gulp-tasks/template/createcleanfilestask.js
```diff
@@ -0,0 +1,32 @@
...
+  });
+  return cleanFilesTask;
+}
+
+/**
+ * The cleanFiles task for a specified locale.
+ * @param {Function} callback
+ * @param {string} locale
+ */
+function _cleanFiles (callback, locale) {
```
tmeyer2115: I'm not super familiar with gulp/async. For my own edification, why do we need all these `callback` parameters now?

conf/gulp-tasks/template/minifytemplatestaskfactory.js
```diff
@@ -0,0 +1,68 @@
...
+    return callback => this._minify(callback, fileName);
+  }
+
+  /**
+   * Minifies the designated js file.
+   * @param {Function} callback
+   * @param {string} filename
+   * @returns {stream.Readable}
+   */
+  _minify (callback, filename) {
```
tmeyer2115, line 49: Is there any real difference between the minification of the template bundles and the SDK bundles? If not, I think we should look to have a single minify factory. That way, we can consolidate the number of different factories somewhat.

conf/gulp-tasks/template/precompiletemplatestaskfactory.js
```diff
@@ -0,0 +1,44 @@
...
+    this._locale = locale;
+    this._translator = translator;
+  }
+
+  /**
+   * Creates the precompileTemplates task for this factory's locale,
+   * and customizes its name.
+   * @returns {Function}
+   */
+  create () {
```
tmeyer2115: Would it be simpler to just have a method that people could invoke directly (a la bundletemplates.js) and get rid of this factory? One argument against having a factory for this is it only builds a single thing. It can't provide you with different kinds of tasks, just the one. The Factory pattern is typically used when you can build multiple kinds of an object.

conf/gulp-tasks/template/createcleanfilestask.js
oshi97: I think what's happening is that we had to change our gulp tasks to be async, to handle initializing i18next. Previously, these tasks would return streams, and gulp already knows how to handle async completion for a stream, so we didn't have to add these callback function.

Now, these tasks return promises that resolve to streams. My guess is that if gulp sees that your tasks returns a promise, it will mark the task as finished as soon as the promise resolves. I don't believe (based purely off experimentation for the past few days) that gulp has logic that says if your task returns a promise BUT it's also a stream, then wait for the stream to finish.

So in order to only have the promise the task returns resolve AFTER the returned stream is finished, I added these callbacks, and also wrapped the returned promise to only resolve after the stream is finished. Gulp was already passing these callbacks as the first argument to every task, we just weren't using them.

conf/gulp-tasks/template/filenameutils.js
alexisgrow: Ah, true, yeah since it doesn't persist I can see how it would be misleading if there was a "TemplateType" for it! Better to leave out of that!

conf/gulp-tasks/template/minifytemplatestaskfactory.js
oshi97: sure, could we make an item for that after the js bundle item is done (since they both have a minify factory)? it looks like the logic is almost the same

alexisgrow: 👀 not Oliver but I was looking into this! There are two things that are different, the SDK bundle

1. renames the minified file (the template bundle I think overwrites(?)) 
2. doesn't mangle "ANSWERS" when uglifying

The renaming (1) I think is the larger difference.

conf/gulp-tasks/template/bundletemplatestaskfactory.js
oshi97: I kind of liked having the two in separate files, but I guess they're pretty coupled together (and we probably shouldn't pretend like they aren't)

conf/gulp-tasks/template/minifytemplatestaskfactory.js
tmeyer2115: I think we should look to consolidate if possible. Between all the new factories floating around and the `async` complexities, I think our Gulp architecture is getting sufficiently complicated. I think we should simplify wherever possible. I'm fine if we make an item to address!

conf/gulp-tasks/template/bundletemplatestaskfactory.js
oshi97: I feel like for a one liner it doesn't really make sense to add an extra line? will do it for now though and we can codify linting rules for this later. I assume eslint has something for jsdoc

conf/gulp-tasks/template/precompiletemplatestaskfactory.js
oshi97: removed the factory! we shouldn't ever be having different versions of the precompile step, as far as I can tell anyways, so the factory is pretty pointless

alexisgrow (approved):
LGTM!

PR #1053, opened 2020-08-27
Generate localized JS bundles

alexisgrow:
Refactors the JS bundle gulpfile and BundleTaskFactory, separating the rollup config and bundling logic from the factory logic. This PR also implements preprocessing of JS files. Currently the preprocessing is done using a regex, but we have plans to move to a JS parser that constructs ASTs. Due to some of the preprocessing logic, our gulp tasks now contain synchronous and asynchronous code, so this PR wraps some of the return values in Promises and adds callback functions to streams.

TEST=manual
J=SLAP-606

```js
const test = TranslationFlagger.flag({
  phrase: 'Item'
});
const test1 = TranslationFlagger.flag({
  phrase: 'There is [[number]] item [[name]]',
  pluralPhrase: 'There are [[number]] items [[name]]',
  count: 2,
  interpolationValues: {
    name: this._config.title,
    number: this._config.details.length
  }
});
const test2 = TranslationFlagger.flag({
  phrase: 'Call',
  context: 'Call is a verb'
});
const test3 = TranslationFlagger.flag({
  phrase: 'I am looking for my child named [[name]]',
  interpolationValues: {
    name: this._config.title,
  }
});
```

Using this PO file:
```po
msgid "Item"
msgid_plural "Item"
msgstr[0] "Article"
msgstr[1] "Articles"

msgid "There is [[number]] item [[name]]"
msgid_plural "There are [[number]] items [[name]]"
msgstr[0] "Il y a [[number]] article [[name]]"
msgstr[1] "Il y a [[number]] articles [[name]]"

msgid "Call"
msgstr "Appelle"
```

And received this output:
```js
var test = "Article";
var test1 = ANSWERS.translateJS({"0":"Il y a [[number]] article [[name]]","1":"Il y a [[number]] articles [[name]]","locale":"fr-FR"}, {name:this._config.title,number:this._config.details.length}, 2);
var test2 = "Appelle";
var test3 = ANSWERS.translateJS("I am looking for my child named [[name]]", {name:this._config.title}, undefined); // (This isn't in the PO file so it is still English)
```


conf/gulp-tasks/bundle/bundle.js
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * The Gulp task for minifying a version of the SDK bundle.
+ *
+ * @param {function} callback
+ * @param {string} bundleName The name of the JS bundle
+ * @returns {stream.Writable} A {@link Writable} stream containing the minified
+ *                            SDK bundle.
+ */
+exports.minifyBundle = function (callback, bundleName) {
```
oshi97: I like this refactor a lot!

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -52,26 +92,19 @@ function compileCSS () {
...
-function watchJS (cb) {
+/**
+ * Creates a new legacy bundle on JS file updates
+ *
+ * @param {BundleTaskFactory} bundleTaskFactory
+ */
+function watchJS (bundleTaskFactory) {
   return watch(['./src/**/*.js'], {
     ignored: './dist/'
-  }, parallel(legacyBundleIIFE));
```
oshi97, line 58: I think we can just remove the parallel here (I think I wrote this dingus line lol)

conf/i18n/translationresolver.js
```diff
@@ -16,6 +16,32 @@ class TranslationResolver {
...
 
+  /**
+   * Converts the {@link TranslationPlaceholder} of into either an exact translation
+   *  or the appropriate run-time call.
+   *
+   * @param {TranslationPlaceholder} placeholder The {@link TranslationPlaceholder}.
+   * @returns {string} A fully translated string or the correct call for getting the
+   *                   translated string at run-time.
+   */
+  resolve (placeholder) {
```
oshi97, line 27: ah yeah putting this here this makes a lot more sense

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -1,46 +1,86 @@
...
-}
+/**
+ * Creates the un-minified legacy JS bundle, compiles CSS, and kicks off the watch task.
+ * @returns {Promise<Function>}
+ */
+exports.dev = function devJSBundle () {
+  return createTaskFactory(DEV_LOCALE).then(devTaskFactory => {
+    return new Promise(resolve => {
+      return parallel(
+        series(devTaskFactory.create(BundleType.LEGACY_IIFE), watchJS(devTaskFactory)),
```
oshi97: Tried this out locally because it felt a little weird that the resolve here wasn't being used. here I think if you return parallel(........)(resolve) instead of just parallel(............) and have watchJS return a function or just add a function wrapper around it, I think things will work? here `watchJS(devTaskFactory)` is a stream not a function, which I believe is why gulp starts crying about `Task never defined: [object Object]`

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -0,0 +1,144 @@
...
+    .pipe(uglify({
+      mangle: { reserved: ['ANSWERS'] }
+    }))
+    .pipe(dest('dist'))
+    .on('end', callback);
+}
+
+/**
+ * The Gulp task for producing the modern version of the SDK bundle.
+ * @param {Function} callback
```
tmeyer2115, line 18: Space in between function desc and params

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Bundles the JS based on the given Rollup config.
+ *
+ * @param {Function} callback
+ * @param {Object} rollupConfig config for the Rollup plugin used for JS module bundling
+ * @param {string} bundleName The filename of the created bundle.
+ * @param {string} libVersion The current JS library version
+ * @param {TranslationResolver} translationResolver for the given locale
+ * @returns {stream.Writable} A {@link Writable} stream cotaining the legacy
```
tmeyer2115: Update docs here as this will contain some arbitrary bundle, not just legacy.

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -0,0 +1,144 @@
...
+ *                            SDK bundle.
+ */
+function _buildBundle (callback, rollupConfig, bundleName, libVersion, translationResolver) {
+  return rollup(rollupConfig)
+    .pipe(source(`${bundleName}.js`))
+    .pipe(replace('@@LIB_VERSION', libVersion))
+    .pipe(replace(/replaceWithTranslation\([^;]+\);/g, translateCall => {
```
tmeyer2115: Should we add a non-greedy qualifier to this regex? I think the ([^;]+); will get us most of the way there, but it's possible someone could have something like: { a: replaceWithTranslation(...), b: replaceWithTranslation(...) }. I guess to that end, we probably can't also look for an ending semicolon.

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -0,0 +1,144 @@
...
+      const placeholder = new TranslateCallParser().parse(translateCall);
+      const translatedValue = translationResolver.resolve(placeholder);
```
tmeyer2115: I think we should name this differently than `translatedValue`. That name suggests a translated string (not run-time call).

conf/gulp-tasks/bundle/bundletaskfactory.js
```diff
@@ -35,16 +26,47 @@ class BundleTaskFactory {
...
+    return bundleFunction;
+  }
+
+  /**
+   * Provides a Gulp task to minify an SDK bundle of the specified type.
+   *
+   * @param {BundleType} bundleType The type of SDK bundle to minify.
+   * @returns {Function} Gulp task for minifying the requested SDK bundle.
+   */
+  minify (bundleType) {
```
tmeyer2115: I think this method gives the Factory two responsibilities. It provides bundling tasks and minifying tasks. It doesn't seem like you can mix and match bundlers and minifiers. Given that that's the case, would it make more sense for this factory to return a Gulp task that both bundles & minifies a `bundleType`? If we make that change, probably worth thinking about if this Factory needs renaming.

conf/gulp-tasks/bundle/bundletaskfactory.js
```diff
@@ -53,31 +75,30 @@ class BundleTaskFactory {
...
-        commonjs({
-          include: './node_modules/**'
-        }),
-        babel({
-          babelrc: false,
-          exclude: 'node_modules/**',
-          presets: ['@babel/env']
-        })
-      ]
+  _modernBundle (callback) {
```
tmeyer2115, line 57: Here and below, update JSDoc to mention new param.

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -1,46 +1,86 @@
...
-function minifyLegacy () {
-  return src('./dist/answers.js')
-    .pipe(rename('answers.min.js'))
-    .pipe(uglify())
-    .pipe(dest('dist'));
-}
+  const translator = await Translator.create(locale, [], { [locale]: { translation } });
+  const translationResolver = new TranslationResolver(
+    translator,
+    (translationResult, interpValues, count) => {
```
tmeyer2115, line 81: In some cases, translationResult will already be a format string (if there's interpolation but no pluralization). Does this runtime call generator handle that case?

src/ui/i18n/translator.js
```diff
@@ -0,0 +1,17 @@
...
+ * bundling tasks, this functions simply returns the phrase passed in.
+ *
+ * @param {string} phrase
+ * @param {string} pluralPhrase
+ * @param {string | number} count
+ * @param {string} context
+ * @param {Object} interpolationValues
+ * @returns {string}
+ */
+export function replaceWithTranslation ({ phrase, pluralPhrase, count, context, interpolationValues }) {
```
tmeyer2115: Minor nit: With this, we have 3 files in the SDK called translator.js. Two in src and one in conf. I think it would be good to rename at least one of the usages in src so we can differentiate more easily between those two.

conf/gulp-tasks/bundle/bundle.js
alexisgrow: Updated to `translationResult`, what do you think?

conf/gulp-tasks/bundle/bundletaskfactory.js
alexisgrow: Yeah, I was thinking that as I was writing this! Bundling and minifying are similar because they both are gulp tasks and use the "this._localePrefix" and "BundleType", but otherwise they are distinct things. 

I originally had the "create" do both bundling and minifying, but our dev task currently doesn't minify. Since minifying adds time and it'd be nice to keep the time to rebuild on `npm run dev` as fast as possible, it seemed like keeping the tasks separate was a better option. I can split the bundle minifying into a separate class, like a `MinifyTaskFactory`?

conf/gulp-tasks/library.gulpfile.js
alexisgrow: lolll I'll remove!

conf/gulp-tasks/bundle/bundletaskfactory.js
alexisgrow: Added a MinifyTaskFactory class!

src/ui/i18n/translator.js
alexisgrow: Spoke offline and renamed to translationflagger.js! 🎏

conf/gulp-tasks/library.gulpfile.js
alexisgrow: Yes! I added a test to the PR description:
```js
const test3 = replaceWithTranslation({
  phrase: 'I am looking for my child named [[name]]',
  interpolationValues: {
    name: this._config.title,
  }
});
```

is changed to:
```js
var test3 = ANSWERS.translateJS("I am looking for my child named [[name]]", {name:this._config.title}, undefined);
```

conf/gulp-tasks/bundle/bundle.js
alexisgrow: Oh totally! Updated to `\([^;]+?\);?`

conf/gulp-tasks/bundle/minifytaskfactory.js
```diff
@@ -0,0 +1,26 @@
...
+    this._locale = locale;
+  }
+
+  /**
+   * Provides a Gulp task to minify an SDK bundle of the specified type.
+   *
+   * @param {BundleType} bundleType The type of SDK bundle to minify.
+   * @returns {Function} Gulp task for minifying the requested SDK bundle.
+   */
+  minify (bundleType) {
```
tmeyer2115, line 21: I'd say lets move minifyBundle into here if possible. I think that fits with this class's responsibility. Additionally, it allows us to clarify the role of bundle.js: it only provides Gulp tasks for bundling.

alexisgrow: Totally, moved it!

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -1,46 +1,90 @@
...
-const bundleTaskFactory = new BundleTaskFactory(getLibraryVersion());
-
-function bundle () {
-  return bundleTaskFactory.create(BundleType.MODERN)();
-}
+/**
+ * Creates the un-minified legacy JS bundle, compiles CSS, and kicks off the watch task.
+ * @returns {Promise<Function>}
+ */
+exports.dev = function devJSBundle () {
```
oshi97, line 18: I think npm run dev still isn't quite functional but besides that everything lgtm!

conf/gulp-tasks/library.gulpfile.js
alexisgrow: Ah thanks!! That did it!

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -1,46 +1,90 @@
...
 
-function legacyBundleIIFE () {
-  return bundleTaskFactory.create(BundleType.LEGACY_IIFE)();
+/**
+ * Creates modern, legacy and legacyUMD bundles in parallel.
+ *
+ * @param {BundleTaskFactory} bundleTaskFactory
+ * @param {MinifyTaskFactory} minifyTaskFactory
+ */
+async function createBundles (bundleTaskFactory, minifyTaskFactory) {
```
tmeyer2115: Does this function need to be `async`? IIRC, `BundleTaskFactory#create` and `MinifyTaskFactory#minify` aren't `async` functions.

src/ui/i18n/translationflagger.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param {string} phrase
+ * @param {string} pluralPhrase
+ * @param {string | number} count
+ * @param {string} context
+ * @param {Object} interpolationValues
+ * @returns {string}
+ */
+export function replaceWithTranslation ({ phrase, pluralPhrase, count, context, interpolationValues }) {
+  return phrase;
```
tmeyer2115: As Bryan noted, we want to handle things gracefully if these calls don't get transpiled out. Obviously, we won't give them a translated version of their phrase, but I think we could still pluralize/interpolate here. That would allow us to handle the no-transpilation case even more gracefully.

conf/gulp-tasks/library.gulpfile.js
alexisgrow: Ah, you're right. Updated to be synchronous.

oshi97 (approved):
lgtm!

PR #1054, opened 2020-09-01
Add translation extractor

oshi97:
This commit adds the npm script extract-translations, which
extracts sdk translations to a .pot file. Under the hood, gulp is used
to stream source files to an extractor class, which extracts messages
to an instance of gettext-extractor, which manages creation of the
actual .pot file. Also tweaked the js regex to allow for **most** punctuation
(for some reason colons still do not work)

TEST=manual

tested plain translation, translation with interpolation, translation with context, 
and plural translation with context for both js and hbs

### DIFF
--- a/conf/gulp-tasks/extracttranslations.gulpfile.js
+++ b/conf/gulp-tasks/extracttranslations.gulpfile.js
@@ -1,0 +1,33 @@
+const { src } = require('gulp');
+const { Transform } = require('readable-stream');
+const path = require('path');
+const TranslationExtractor = require('../i18n/extract/translationextractor');
+
+module.exports = function precompileTemplates (callback) {
+  const extractor = new TranslationExtractor();
+
+  const transform = new Transform({
+    objectMode: true,
+    transform (chunk, encoding, callback) {
+      const contents = chunk.contents.toString();
+      const extname = path.extname(chunk.path);
+      const pathFromBase = path.relative('./', chunk.path);
+      switch (extname) {
+        case '.js':
+          extractor.extractFromJS(contents, pathFromBase);
+          break;
+        case '.hbs':
+          extractor.extractFromHBS(contents, pathFromBase);
+          break;
+      }
+      callback();
+    }
+  });
+
+  const templatesStream = src(['./src/ui/templates/**/*.hbs', './src/**/*.js'])
+    .pipe(transform)
+    .on('end', () => {
+      extractor.savePotFile('./conf/i18n/translations/messages.pot');
+    });
+  return templatesStream;
+};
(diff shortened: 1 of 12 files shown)


conf/gulp-tasks/extracttranslations.gulpfile.js
```diff
@@ -0,0 +1,43 @@
...
+  const extractor = new TranslationExtractor();
+
+  /**
+   * Extracts messages from a given source file.
+   * 
+   * @param {vinyl.File} file 
+   * @param {string} encoding 
+   * @param {Function} next 
+   */
+  function extractMessagesFromFile(file, encoding, next) {
```
alexisgrow: Do you think it would make sense for this to be the public method of the `TranslationExtractor` class? I was thinking it maybe nice to have the abstraction but I'm sure there are also reasons to have the function here!

conf/i18n/extract/translationextractor.js
```diff
@@ -0,0 +1,108 @@
...
+      this._extractFromJSCall(match, filepath);
+    }
+  }
+
+  /**
+   * Returns the extracted messages as a string.
+   *
+   * @returns {Array<Object>}
+   */
+  getPotString () {
```
alexisgrow: Is this used?

conf/i18n/extract/translationextractor.js
```diff
@@ -0,0 +1,108 @@
...
+    return this._extractor.getPotString();
+  }
+
+  /**
+   * Saves the currently extracted messages to te given output path.
+   * Creates any parent directories as necessary.
+   *
```
cea2aj: small spelling fix: "te" should be "the"

conf/i18n/translatecallparser.js
```diff
@@ -10,17 +10,19 @@ class TranslateCallParser {
    * to return a {@link TranslationPlaceholder}
    *
    * @param {string} translateCall
+   * @param {number}
    * @returns {TranslationPlaceholder}
    */
-  parse (translateCall) {
+  parse (translateCall, lineNumber) {
```
cea2aj: nit: Add "lineNumber" after @param {number}

conf/i18n/extract/translationextractor.js
oshi97: teh

conf/i18n/extract/translationextractor.js
oshi97: nope was using this just for testing! will remove

conf/gulp-tasks/extracttranslations.gulpfile.js
oshi97: yeah I think we can move this logic in there, that seems more straightforward!

alexisgrow (approved):
LG 🔥

PR #1055, opened 2020-09-01
Add translation helpers to JS

alexisgrow:
J=SLAP-547
TEST=compile,auto,manual

Test that the `TranslationFlagger.flag` interpolation works with unit test coverage.
Test that the `TranslationFlagger.flag` usages are being replaced with valid JS with `npm run build` successful compilation.
Test that the usages are being replaced with correctly preprocessed content by visually inspecting, e.g.
Before processing:
```js
if (!max) {
  displayValue = this._isExclusive
    ? TranslationFlagger.flag({
      phrase: 'After [[date]]',
      context: 'After a date. Example: After [August 15th]',
      interpolationValues: {
        date: min
      }
    })
    : TranslationFlagger.flag({
      phrase: '[[date]] or later',
      context: 'Beginning at a date (with no end date). Example: [August 15th] or later',
      interpolationValues: {
        date: min
      }
    });
} else if (!min) {
  displayValue = this._isExclusive
    ? TranslationFlagger.flag({
      phrase: 'Before [[date]]',
      context: 'Before a date. Example: Before [August 15th]',
      interpolationValues: {
        date: max
      }
    })
    : TranslationFlagger.flag({
      phrase: '[[date]] and earlier',
      context: 'Ending at a date with (no start date). Example: [August 15th] or earlier',
      interpolationValues: {
        date: max
      }
    });
} else if (min === max) {
  displayValue = this._isExclusive ? '' : min;
} else {
  displayValue = TranslationFlagger.flag({
    phrase: '[[start]] - [[end]]',
    context: 'Start date to end date. Example: [August 15th] - [August 16th]',
    interpolationValues: {
      start: min,
      end: max
    }
  });
}
```
After processing:
```js
if (!max) {
  displayValue = this._isExclusive ? ANSWERS.translateJS("After [[date]]", {date:min}, undefined) : ANSWERS.translateJS("[[date]] or later", {date:min}, undefined);
} else if (!min) {
  displayValue = this._isExclusive ? ANSWERS.translateJS("Before [[date]]", {date:max}, undefined) : ANSWERS.translateJS("[[date]] and earlier", {date:max}, undefined);
} else if (min === max) {
  displayValue = this._isExclusive ? '' : min;
} else {
  displayValue = ANSWERS.translateJS("[[start]] - [[end]]", {start:min,end:max}, undefined);
}
```

### DIFF
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -2,4 +2,4 @@ exports.DEV_LOCALE = 'en';
 
 exports.BUILD_LOCALES = ['en'];
 
-exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\([^;]+?\)/g;
+exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s\S)*\{[^;]+?\}(\s\S)*\)/g;
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -2,4 +2,4 @@ exports.DEV_LOCALE = 'en';
 
 exports.BUILD_LOCALES = ['en'];
 
-exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s\S)*\{[^;]+?\}(\s\S)*\)/g;
+exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s)*\{[^;]+?\}(\s)*\)/g;
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -197,7 +197,7 @@ export default class DateRangeFilterComponent extends Component {
           }
         })
         : TranslationFlagger.flag({
-          phrase: `[[dateOrTime]] or later`,
+          phrase: '[[dateOrTime]] or later',
           context: 'Beginning at a date (with no end date). Example: [August 15th] or later',
           interpolationValues: {
             dateOrTime: min
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -6,6 +6,7 @@ import DOM from '../../dom/dom';
 import FilterNodeFactory from '../../../core/filters/filternodefactory';
 import FilterMetadata from '../../../core/filters/filtermetadata';
 import ComponentTypes from '../../components/componenttypes';
+import TranslationFlagger from '../../i18n/translationflagger';
 
 /**
  * A filter for a range of dates
@@ -188,16 +189,47 @@ export default class DateRangeFilterComponent extends Component {
     let displayValue;
     if (!max) {
       displayValue = this._isExclusive
-        ? `After ${min}`
-        : `${min} or later`;
+        ? TranslationFlagger.flag({
+          phrase: 'After [[dateOrTime]]',
+          context: 'Starting date/time with no end date/time',
+          interpolationValues: {
+            dateOrTime: min
+          }
+        })
+        : TranslationFlagger.flag({
+          phrase: `[[dateOrTime]] or later`,
+          context: 'Starting date/time with no end date/time',
+          interpolationValues: {
+            dateOrTime: min
+          }
+        });
     } else if (!min) {
       displayValue = this._isExclusive
-        ? `Before ${max}`
-        : `${max} and earlier`;
+        ? TranslationFlagger.flag({
+          phrase: 'Before [[dateOrTime]]',
+          context: 'Ending date/time with no start date/time',
+          interpolationValues: {
+            dateOrTime: max
+          }
+        })
+        : TranslationFlagger.flag({
+          phrase: '[[dateOrTime]] and earlier',
+          context: 'Ending date/time with no start date/time',
+          interpolationValues: {
+            dateOrTime: max
+          }
+        });
     } else if (min === max) {
       displayValue = this._isExclusive ? '' : min;
     } else {
-      displayValue = `${min} - ${max}`;
+      displayValue = TranslationFlagger.flag({
+        phrase: '[[start]] - [[end]]',
+        context: 'Start date/time to end date/time',
+        interpolationValues: {
+          start: min,
+          end: max
+        }
+      });
     }
     return new FilterMetadata({
       fieldName: this._title,
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -190,33 +190,33 @@ export default class DateRangeFilterComponent extends Component {
     if (!max) {
       displayValue = this._isExclusive
         ? TranslationFlagger.flag({
-          phrase: 'After [[dateOrTime]]',
+          phrase: 'After [[date]]',
           context: 'After a date. Example: After [August 15th]',
           interpolationValues: {
-            dateOrTime: min
+            date: min
           }
         })
         : TranslationFlagger.flag({
-          phrase: '[[dateOrTime]] or later',
+          phrase: '[[date]] or later',
           context: 'Beginning at a date (with no end date). Example: [August 15th] or later',
           interpolationValues: {
-            dateOrTime: min
+            date: min
           }
         });
     } else if (!min) {
       displayValue = this._isExclusive
         ? TranslationFlagger.flag({
-          phrase: 'Before [[dateOrTime]]',
+          phrase: 'Before [[date]]',
           context: 'Before a date. Example: Before [August 15th]',
           interpolationValues: {
-            dateOrTime: max
+            date: max
           }
         })
         : TranslationFlagger.flag({
-          phrase: '[[dateOrTime]] and earlier',
+          phrase: '[[date]] and earlier',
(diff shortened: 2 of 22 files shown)


conf/i18n/constants.js
```diff
@@ -2,4 +2,4 @@ exports.DEV_LOCALE = 'en';
 
 exports.BUILD_LOCALES = ['en'];
 
-exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\([^;]+?\)/g;
+exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s\S)*\{[^;]+?\}(\s\S)*\)/g;
```
tmeyer2115: With this regex, don't we want only whitespace-type characters before and after the curly braces? For that matter, do we want all whitespace characters or just the ones for a new line? I think this regex would allow something like: `TranslationFlagger.flag( b a{ phrase: 'Stuff' });` Our regex should make sure we can find all valid function calls of `TranslationFlagger.flag`.

src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -188,16 +189,47 @@ export default class DateRangeFilterComponent extends Component {
...
+          interpolationValues: {
+            date: max
+          }
+        });
     } else if (min === max) {
       displayValue = this._isExclusive ? '' : min;
     } else {
-      displayValue = `${min} - ${max}`;
+      displayValue = TranslationFlagger.flag({
+        phrase: '[[start]] - [[end]]',
```
tmeyer2115, line 226: This one, I'm not totally sure if we need a translation as there's no text. I guess it depends if other languages represent pure date ranges differently. Is this something we'd translate in alpha?

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -30,43 +31,62 @@ const DEFAULT_CONFIG = {
...
+  }),
 
   /**
    * An optional label to use for the question
    * @type {string}
    */
-  'questionLabel': 'Question',
+  'questionLabel': TranslationFlagger.flag({
+    phrase: 'Question',
+    context: 'Label for an input field'
```
tmeyer2115: Is `Label for an input field` the sort of context we give in Alpha?

src/ui/components/results/paginationcomponent.js
```diff
@@ -91,7 +92,12 @@ export default class PaginationComponent extends Component {
      * @type {string}
      * @private
      */
-    this._pageLabel = config.pageLabel === undefined ? 'Page' : config.pageLabel;
+    this._pageLabel = config.pageLabel !== undefined
```
tmeyer2115, line 95: I think we may need to make an additional item for internationalizing pagination. Right now, the component creates text of the form `${_pageLabel}: ${pageNum}`. This assumes that all languages follow the `label: num` syntax and I'm not sure that's true. I think we may need to do some interpolation here.

src/ui/components/results/universalresultscomponent.js
```diff
@@ -85,7 +88,10 @@ export default class UniversalResultsComponent extends Component {
       // Show a view more link by default, which also links to verticalURL.
       viewMore: true,
       // By default, the view more link has a label of 'View More'.
-      viewMoreLabel: defaultConfigOption(config, ['viewMoreLabel', 'viewAllText'], 'View More'),
+      viewMoreLabel: defaultConfigOption(config, ['viewMoreLabel', 'viewAllText'], TranslationFlagger.flag({
```
tmeyer2115: LL. Let's also put `config` and the array on their own lines. That way, all three args start on a newline.

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -128,20 +129,28 @@ class VerticalResultsConfig {
        * The aria-label given to the applied filters bar. Defaults to 'Filters applied to this search:'.
        * @type {string}
        **/
-      labelText: defaultConfigOption(config, ['appliedFilters.labelText'], 'Filters applied to this search:'),
+      labelText: defaultConfigOption(config, ['appliedFilters.labelText'], TranslationFlagger.flag({
```
tmeyer2115: LL. If all args can't fit on the same line as the function name, see if they can all fit on a separate newline after the function name. If that won't work either, each arg should begin on its own newline.

src/ui/components/search/locationbiascomponent.js
```diff
@@ -1,11 +1,21 @@
...
 import DOM from '../../dom/dom';
+import TranslationFlagger from '../../i18n/translationflagger';
 
 const DEFAULT_CONFIG = {
-  ipAccuracyHelpText: 'based on your internet address',
-  deviceAccuracyHelpText: 'based on your device',
-  updateLocationButtonText: 'Update your location'
+  ipAccuracyHelpText: TranslationFlagger.flag({
+    phrase: 'based on your internet address',
+    context: 'Describes the accuracy [of location]. Example: [Salt Lake City, Utah] based on your internet address'
```
tmeyer2115: Will having `[of location]` in brackets cause confusion? Could that be misinterpreted as referring to interpolation?

src/ui/components/filters/daterangefiltercomponent.js
alexisgrow: Yeah I was going back and forth on this one. There are several examples in the .po file I was using from alpha that translate ranges like this, e.g.
```
#. Time range
#. EXAMPLE: 10:00 AM - 12:30 PM
#: ....js:71
msgctxt "datetime"
msgid "{0} - {1}"
msgstr "{0} - {1}"
```

src/ui/components/questions/questionsubmissioncomponent.js
alexisgrow: The original PO files I found didn't have any good examples of this 🙈, I dug through some more files, and a similar example is: `Labels the minute value of the time that a business opens on the day of the week provided as an argument.` I can update these to align more closely with that example.

src/ui/components/results/universalresultscomponent.js
alexisgrow: Updating! What is our line length for this repo? This one flew under my radar since it's significantly shorter than lines ~104-123 of this file.

src/ui/components/results/verticalresultscomponent.js
alexisgrow: Updating! Same question as in this comment about consistent LL in this repo https://github.com/yext/answers/pull/1055/files#r482162852. This line is <100 characters.

src/ui/components/results/directanswercomponent.js
```diff
@@ -60,7 +67,10 @@ export default class DirectAnswerComponent extends Component {
      * The display text for the View Details click to action link
      * @type {string}
      */
-    this._viewDetailsText = config.viewDetailsText || 'View Details';
+    this._viewDetailsText = config.viewDetailsText || TranslationFlagger.flag({
+      phrase: 'View Details',
+      context: 'Button label, call to action'
+    });
```
cea2aj: What do you think about the context: 'Button label, view is a verb'? The translator might not be familiar with a call to action.

conf/i18n/constants.js
alexisgrow: Agh! Yes! I don't know what I was thinking, the `/S` should definitely not be in there!

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -129,28 +129,39 @@ class VerticalResultsConfig {
...
 
       /**
        * The aria-label given to the removable filter buttons.
        * @type {string}
        */
-      removableLabelText: defaultConfigOption(config, ['appliedFilters.removableLabelText'], TranslationFlagger.flag({
-        phrase: 'Remove this filter',
-        context: 'Button label'
-      }))
+      removableLabelText: defaultConfigOption(config,
```
tmeyer2115: `config` on new line

src/ui/components/filters/daterangefiltercomponent.js
tmeyer2115: Okay! Sounds like we may need this translated then.

src/ui/components/results/verticalresultscomponent.js
tmeyer2115: I try and stick to 90 characters or less. We have not actual standard in this repo (which is a problem). Oliver and I have been discussing some linting we could add to really codify this.

src/ui/components/results/paginationcomponent.js
alexisgrow: Could you clarify slightly more what you're seeing? 

In the component's .hbs file, it has the form `[pageLabel] [pageNum]`, for example `Page 1` (no colon, code here: https://github.com/yext/answers/blob/feature/localized-assets/src/ui/templates/results/pagination.hbs#L60). On some non-english search engines, I've seen the label "Page" omitted entirely and this JS currently supports that use-case. Is the concern that some languages may use a different convention entirely (e.g. `1 Page`, or `Page #1`)?

alexisgrow: If it's a convention thing, I can totally add it to an interpolation helper - like "[pageLabel] [pageNum]" - in this PR https://github.com/yext/answers/pull/1057

tmeyer2115 (approved):

PR #1060, opened 2020-09-03
pagination: Fix disappearing pagination on back nav and refresh

creotutar:
Previously, there were two pagination persistence bugs.

1. Land on a vertical page with a query with results. Go to page 2. Go
   to page 3. Try to navigate backwards in browser history. You will go
   to page 1 instead of page 3.

2. Land on a vertical page w a query w results. Go to page 2. Refresh
   the page. You will be on page 1 instead of page 2.

Both of these related bugs were because we were defaulting to resetting
pagination on any searchbar search. In 1, backwards navigation updates the
query and the searchbar takes the updated query and searches. In 2,
landing  with a `query` query parameter is determined in the SearchBar
component onCreate.

To fix this, we add another value for the pre-existing `queryTrigger`
global storage key, QUERY_PARAMETER, meaning the query was triggered by
a query parameter.

Note: as a result of this fix, we were no longer able to go from Page 2
to Page 1. This is because of another bug, where navigating backwards in
history from presence of a query param to absence of the param will not
erase the parameter in storage. This is fixed for the pagination param here
but should be addressed more widely in SLAP-616.

J=SLAP-610
TEST=manual

Test on a local Jambo HH Theme site attached to a local dist of the SDK.
Test on a local site using the SDK directly.

Test that navigating on a vertical to Page 2 and then Page 3 allows you
to navigate back in browser history to Page 2 and Page 1 respectively.

Test that navigating to a page and then refreshing maintains pagination
offset. (IE Test that query params for search-offset work as intended).

Test that navigating to a page and then searching another query in the
searchbar /will/ reset pagination (and remove the associated query
parameter).

Test that queryTrigger in the liveapi request in all the situations
are null.

Test that with defaultInitialSearch set, queryTrigger still sends as
'initialize' in the liveapi request.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -169,7 +169,7 @@ class Answers {
         }
 
         if (!data[StorageKeys.SEARCH_OFFSET]) {
-          this.core.globalStorage.delete(StorageKeys.SEARCH_OFFSET);
+          this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, 0);
         }
         globalStorage.setAll(data);
       }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -19,6 +19,7 @@ import GlobalStorage from './core/storage/globalstorage';
 import { AnswersComponentError } from './core/errors/errors';
 import AnalyticsEvent from './core/analytics/analyticsevent';
 import StorageKeys from './core/storage/storagekeys';
+import QueryTriggers from './core/models/querytriggers';
 import SearchConfig from './core/models/searchconfig';
 import AutoCompleteApi from './core/search/autocompleteapi';
 import MockAutoCompleteService from './core/search/mockautocompleteservice';
@@ -163,6 +164,12 @@ class Answers {
       resetListener: data => {
         if (!data[StorageKeys.QUERY]) {
           this.core.clearResults();
+        } else {
+          this.core.globalStorage.set(StorageKeys.QUERY_TRIGGER, QueryTriggers.QUERY_PARAMETER);
+        }
+
+        if (!data[StorageKeys.SEARCH_OFFSET]) {
+          this.core.globalStorage.delete(StorageKeys.SEARCH_OFFSET);
         }
         globalStorage.setAll(data);
       }
@@ -173,6 +180,9 @@ class Answers {
     globalStorage.set(StorageKeys.LOCALE, parsedConfig.locale);
     globalStorage.set(StorageKeys.SESSIONS_OPT_IN, parsedConfig.sessionTrackingEnabled);
     parsedConfig.noResults && globalStorage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
+    if (globalStorage.getState(StorageKeys.QUERY)) {
+      globalStorage.set(StorageKeys.QUERY_TRIGGER, QueryTriggers.QUERY_PARAMETER);
+    }
 
     const context = globalStorage.getState(StorageKeys.API_CONTEXT);
     if (context && !isValidContext(context)) {
@@ -434,7 +444,7 @@ class Answers {
     if (prepopulatedQuery != null) {
       return;
     }
-    this.core.globalStorage.set('queryTrigger', 'initialize');
+    this.core.globalStorage.set(StorageKeys.QUERY_TRIGGER, QueryTriggers.INITIALIZE);
     this.core.setQuery(searchConfig.defaultInitialSearch);
   }
 
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -10,6 +10,7 @@ import Navigation from './models/navigation';
 import AlternativeVerticals from './models/alternativeverticals';
 import DirectAnswer from './models/directanswer';
 import LocationBias from './models/locationbias';
+import QueryTriggers from './models/querytriggers';
 
 import StorageKeys from './storage/storagekeys';
 import AnalyticsEvent from './analytics/analyticsevent';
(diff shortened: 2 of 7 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -547,29 +552,28 @@ export default class SearchComponent extends Component {
...
   }
 
   /**
    * Performs a query using the provided string input.
    * @param {string} query The string to query against.
+   * @param {Object} searchOptionsOverrides The options to pass for core search
    * @returns {Promise} A promise that will perform the query and update globalStorage accordingly.
    */
-  search (query) {
+  search (query, searchOptionsOverrides) {
```
creotutar: This is a weird function signature. I was considering making it just `searchOptions` and passing it from the initial call of `debouncedSearch` but I was worried about moving the logic of `if verticalKey searchOptions1 else searchOptions2` to every occurrence.

creotutar: Still on the fence.

creotutar: I changed it

src/answers-umd.js
```diff
@@ -163,6 +164,12 @@ class Answers {
       resetListener: data => {
         if (!data[StorageKeys.QUERY]) {
           this.core.clearResults();
+        } else {
+          this.core.globalStorage.set(StorageKeys.QUERY_TRIGGER, QueryTriggers.QUERY_PARAMETER);
+        }
+
+        if (!data[StorageKeys.SEARCH_OFFSET]) {
```
creotutar, line 171: See note in the commit message. I made SLAP-616 to address going from presence to absence of query param on back nav.

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -49,7 +49,7 @@ export default class SpellCheckComponent extends Component {
     let params = new SearchParams(window.location.search.substring(1));
     params.set('query', query.value);
     params.set('skipSpellCheck', true);
```
cea2aj, line 51: nit: Maybe we should use StorageKeys here as well? Everything else LGTM!

src/core/models/querytriggers.js
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * QueryTriggers is an ENUM of the possible triggers for a
+ * query update.
+ *
+ * @enum {string}
+ */
+export default {
+  INITIALIZE: 'initialize',
+  QUERY_PARAMETER: 'query-parameter'
```
tmeyer2115, line 11: I think we should consider renaming `QUERY_PARAMETER`. In the context of our system, query is already a loaded term (search query), so I think it's a bit confusing for us to adopt another meaning for query here.

alexisgrow (approved):
this LGTM!

src/ui/components/search/searchcomponent.js
```diff
@@ -150,11 +160,22 @@ export default class SearchComponent extends Component {
...
           this.core.setQuery(this._defaultInitialSearch);
         }
         return;
       }
-      this.debouncedSearch(q);
+
+      const queryTrigger = this.core.globalStorage.getState(StorageKeys.QUERY_TRIGGER);
+      const resetPagination = this._verticalKey &&
+        queryTrigger !== QueryTriggers.QUERY_PARAMETER &&
+        queryTrigger !== QueryTriggers.INITIALIZE;
```
tmeyer2115, line 172: Do we anticipate making a `QueryTrigger` for manual input in the search bar? If not, could we simplify this logic by saying if `verticalKey` and if `!queryTrigger`, then `resetPagination`?

src/core/core.js
```diff
@@ -497,6 +503,18 @@ export default class Core {
     this.filterRegistry.clearLocationRadiusFilterNode();
   }
 
+  /**
+   * Returns the query trigger for the search API given the SDK query trigger
+   * @param {QueryTriggers} queryTrigger SDK query trigger
+   * @returns {QueryTriggers} query trigger if accepted by the search API, null o/w
+   */
+  getQueryTriggerForSearchApi (queryTrigger) {
```
oshi97, line 511: I don't think we want to use queryTrigger both for persistent storage magic and also as something completely separate we send to the api, I think it would be easier to just have them be two separate things.

src/ui/components/search/spellcheckcomponent.js
creotutar: Agree! But I think that would be a change that should not be in this commit (as that is outside of the scope of this work). I can make an item for the more sweeping change of using StorageKeys wherever possible with global/persistent storage. SLAP-621

src/ui/components/search/searchcomponent.js
creotutar: I was not sure what to anticipate! We currently have another query trigger (for SpellCheck component, which is more dynamic, relying on the `type` of the data returned). I believe we would want spellcheck query triggers to continue to reset pagination. It felt easier here to enumerate these two than try to figure out the possible types for SpellCheck. How does that sound to you?

oshi97 (commented):
Would it be possible to, instead of making these query trigger changes, update the resetListener (really just onpopstate) in answers-umd.js? I was thinking of maybe setting an IMPORTANT_SEARCH_OFFSET storage key onpopstate that would override the regular search_offset and would be deleted after every search. I was just thinking we want the fix this the easiest way that is to refactor, and if we can scope this change to just when the history changes I think that would be ideal?

tmeyer2115 (approved):

oshi97 (approved):

src/core/core.js
creotutar: Talked offline. This is definitely not an ideal way to get context about entry for the query update. This is necessary because we have /two/ ways we need to fix (on a refresh and on a popstate from back nav). Both of these initiate a search through a query update.

Discussed w @tmeyer2115 a bit yesterday about the second point. QueryTriggers will be a first class _SDK_ enum. We will guide behavior based on this storage key and if it has a certain value, we will also pass it in the LiveAPI request.

Keeping this implementation for now, I understand the refactoring worries:(

Issue #369, opened 2019-11-18
Default Query - Should be Submitting with "retrieveFacets = true"


rosiegrant:
If we submit a default query and the page has a facets component, the `retrieveFacets` parameter on the API call should always be set to `true` 

Otherwise the default queries that we submit won't show facets:
https://answers-js-demo.netlify.com/sites/rose/employees.html
![image](https://user-images.githubusercontent.com/22160788/69088799-921a2780-0a15-11ea-8c52-d76a5ff0b5e5.png)

data-enabler:
Looks like `retrieveFacets` doesn't get enabled until a Facets component gets added to the page, meaning that there's an ordering dependency here 😞. Haven't confirmed, but I suspect it would work if the Facets component were added before the SearchBar.

data-enabler:
Noticed that it stopped happening when I tried to reproduce it in 0.10.1, and it's because the location permission check added in #387 causes the initial search to happen asynchronously (and therefore after both components have been added to the page).

There are plans to add a `retrieveFacets` option to the top-level search config, but in the meantime I think making initial search always asynchronous (currently it isn't in browsers that don't support the permissions API) should fix this for 99% of cases (and make the behavior more consistent). Basically, as long as the Facets component isn't added asynchronously after the SearchBar, it should work as expected.

data-enabler:
Fixed in v0.10.1

PR #1058, opened 2020-09-02
FilterOptions: replace new Event() with DOM.trigger()

oshi97:
The event constructor does not work in ie11, but we have an
ie11 compliant implementation with DOM.trigger()

T=https://yextops.zendesk.com/agent/tickets/348126
TEST=manual

Test that before, in browserstack ie11,
 clicking the X button in a searchable facet has console error

With code change, no console error

and that it still clears the facet search input on chrome + ie11

PR #1062, opened 2020-09-03
downgrade regenerator-runtime due to unsafe eval usages

oshi97:
This commit downgrades regenerator-runtime from
^0.13.3 (currently 0.13.7) to being pinned to exactly
0.13.1. This is because there is a change in 0.13.2 which
adds back in a Function() constructor to define
regenerator runtime globally.
See #369 https://github.com/facebook/regenerator/commits/master/packages/regenerator-runtime

T=https://yextops.zendesk.com/agent/tickets/347915
TEST=manual

tested with a local apache httpd server on a searchbar only page
added below line to .htaccess, which sets the CSP to
the same as Syncreon

also test csp with a `<meta>` tag

tested both answers.js, answers.min.js, answers-modern.js, and answers-modern.min.js

double checked that our code still runs correctly in ie11, for both a searchbar only
page with no unsafe-eval csp, and a regular page with facets and results
that allows unsafe eval

### DIFF
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -68,7 +68,8 @@ function legacyBundleIIFE () {
   return legacyBundle({
     format: 'iife',
     name: NAMESPACE,
-    sourcemap: true
+    sourcemap: true,
+    strict: false
   },
   'answers.js'
   );
@@ -79,7 +80,8 @@ function legacyBundleUMD () {
     format: 'umd',
     name: NAMESPACE,
     export: 'default',
-    sourcemap: true
+    sourcemap: true,
+    strict: false
   },
   'answers-umd.js'
   );
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -68,8 +68,7 @@ function legacyBundleIIFE () {
   return legacyBundle({
     format: 'iife',
     name: NAMESPACE,
-    sourcemap: true,
-    strict: false
+    sourcemap: true
   },
   'answers.js'
   );
@@ -80,8 +79,7 @@ function legacyBundleUMD () {
     format: 'umd',
     name: NAMESPACE,
     export: 'default',
-    sourcemap: true,
-    strict: false
+    sourcemap: true
   },
   'answers-umd.js'
   );
--- a/package-lock.json
+++ b/package-lock.json
@@ -3866,7 +3866,8 @@
     "binaryextensions": {
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/binaryextensions/-/binaryextensions-2.3.0.tgz",
-      "integrity": "sha512-nAihlQsYGyc5Bwq6+EsubvANYGExeJKHDO3RjnvwU042fawQTQfM3Kxn7IHUXQOz4bzfwsGYYHGSvXyW4zOGLg=="
+      "integrity": "sha512-nAihlQsYGyc5Bwq6+EsubvANYGExeJKHDO3RjnvwU042fawQTQfM3Kxn7IHUXQOz4bzfwsGYYHGSvXyW4zOGLg==",
+      "dev": true
     },
     "bl": {
       "version": "1.2.2",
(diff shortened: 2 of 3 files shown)


oshi97:
ci isn't working for me for some reason (couldn't figure out why so filed a support ticket), manually ran npm run build and npm test
edit: now working no idea what was happening

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -1,7 +1,5 @@
 /** @module FilterOptionsComponent */
 
-/* global Event */
```
oshi97, line 3: Should have been removed in https://github.com/yext/answers/pull/1058 to make semistandard happy, but circleci was not running my builds and I missed this

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -68,7 +68,8 @@ function legacyBundleIIFE () {
   return legacyBundle({
     format: 'iife',
     name: NAMESPACE,
-    sourcemap: true
+    sourcemap: true,
```
tmeyer2115: Do we need to fix this for the modern bundle as well?

oshi97: nope! just checked, it's not an issue for the modern bundle. I believe it's because the modern bundle doesn't have the regenerator-runtime polyfill (or any of the polyfill-prefix stuff)?

oshi97: updated the test desc

tmeyer2115 (commented):
I'm a bit wary of just disabling strict mode. How much research have we done into it? I took a quick look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode. It sounds like strict mode has some benefits:

- Can produce more optimized JS.
- Can help produce more secure JS.

Are we sure we want to get rid of this? Is it standard for libraries such as ours to use strict mode? Others have encountered this same CSP issue: https://github.com/mozilla/pdf.js/issues/11036. One recommendation was to downgrade regenerator-runtime (apparently an older version(s) did not have this problem with strict mode). Is that a viable option here?

tmeyer2115:
> I'm a bit wary of just disabling strict mode. How much research have we done into it? I took a quick look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode. It sounds like strict mode has some benefits:
> 
> * Can produce more optimized JS.
> * Can help produce more secure JS.
> 
> Are we sure we want to get rid of this? Is it standard for libraries such as ours to use strict mode? Others have encountered this same CSP issue: [mozilla/pdf.js#11036](https://github.com/mozilla/pdf.js/issues/11036). One recommendation was to downgrade regenerator-runtime (apparently an older version(s) did not have this problem with strict mode). Is that a viable option here?

I'm not sure that we're getting much performance benefit from strict mode. It sounds like strict mode supports better compiler optimization. Uglify is a compiler of sorts. Is the bundle size different at all between strict and sloppy?

oshi97:
> > I'm a bit wary of just disabling strict mode. How much research have we done into it? I took a quick look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode. It sounds like strict mode has some benefits:
> > 
> > * Can produce more optimized JS.
> > * Can help produce more secure JS.
> > 
> > Are we sure we want to get rid of this? Is it standard for libraries such as ours to use strict mode? Others have encountered this same CSP issue: [mozilla/pdf.js#11036](https://github.com/mozilla/pdf.js/issues/11036). One recommendation was to downgrade regenerator-runtime (apparently an older version(s) did not have this problem with strict mode). Is that a viable option here?
> 
> I'm not sure that we're getting much performance benefit from strict mode. It sounds like strict mode supports better compiler optimization. Uglify is a compiler of sorts. Is the bundle size different at all between strict and sloppy?

The bundle size is identical. Since the SDK doesn't have regenerator runtime in develop anymore, we don't need this change in future versions, at least right now. I do want to add an acceptance test for no unsafe-eval csp on a searchbar only page, though (created slap-619 last week for this)

here are the two builds I looked at
https://app.circleci.com/pipelines/github/yext/answers/4300/workflows/09e8f06a-39bb-4732-a6a8-4cc591c0d94b/jobs/11833
https://app.circleci.com/pipelines/github/yext/answers/4299/workflows/56f52e95-829c-4a49-8a2d-b348bdfbf418/jobs/11832

tmeyer2115 (approved):

PR #1063, opened 2020-09-03
AlternativeVerticals: add :focus styling to links

oshi97:
This commit adds default :focus styling to the vertical
and universal links in AlternativeVerticals aka NoResults.

J=SLAP-588
TEST=manual

Check that I can tab over to the links, and they will show the new
styling when tabbed over to.

test that if I change the focus color to pink, the pink takes priority
over the regular text color

### DIFF
--- a/src/ui/sass/modules/_AlternativeVerticals.scss
+++ b/src/ui/sass/modules/_AlternativeVerticals.scss
@@ -70,11 +70,6 @@ $alternative-verticals-emphasized-font-weight: var(--yxt-font-weight-medium) !de
     }
   }
 
-  &-suggestionLink:focus, &-universalLink:focus {
-    color: var(--yxt-color-brand-primary);
-    text-decoration: underline; 
-  }
-
   &-suggestionLink {
     display: flex;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


src/ui/sass/modules/_AlternativeVerticals.scss
```diff
@@ -70,6 +70,11 @@ $alternative-verticals-emphasized-font-weight: var(--yxt-font-weight-medium) !de
     }
   }
 
+  &-suggestionLink:focus, &-universalLink:focus {
```
alexisgrow: For the suggestionLink, it looks like there is currently logic to only underline the "yxt-AlternativeVerticals-suggestionLink--copy" on hover, should the same thing happen on focus?

For the universalLink, since the focus is the same as the hover CSS, could this styling be colocated with the hover CSS below (currently L113)?

oshi97: yeah that's a great suggestion, I was just going as close to the item's css as possible but what you said makes a lot of sense. The color is already set to --yxt-brand-primary too so there's really no reason to reinforce that on focus, either (I can only imagine this annoying somebody)

alexisgrow (approved):
LGTM! 🔥

PR #1064, opened 2020-09-04
filteroptions: Explicitly pass replaceHistory based on facet update

creotutar:
Consider the following situation

In State 0
Land on a vertical page
In State 1
Search for "all"
In State 2
Select and apply a facet "Product"
In State 3
Search for "people" (any query that will give results)
In State 4
Back nav in Chrome history
In State 5

You would expect state 5 === state 3. That is, you would expect state 5
to have the search "all" and the facet "Product". Instead, state 5 has the search
"people" and the facet "Product". This is a result of how we update
facets. When you are searching for "people", the query is updated and
pushes a state. Then we get the results back from liveapi, get updated
facets, and push /another/ state with the updated facets. (IE we are
pushing two states when we make a search when facets are already being
replaced).

This change makes replaceHistory explicitly set. Whenever a new set of
facets are mounted to the page (e.g. we get new facets from liveapi, we
landed on the page), they we replaceHistory. This is with the knowledge
that a state was already pushed for the search, and we don't want to
push another state for the updated facets.

J=SLAP-612
TEST=manual

Test on a local HH theme site and a local raw site referencing a
local SDK dist

Test the above situation.
Test that applying a facet still pushes a state.
Test that you can apply a facet and then apply another facet and have
two different states.
Test that landing without any facet parameters replaces history when the
facet parameters are added. (ie landing on
http://localhost:5000/public/people.html?query=all). When navigating
back, you should go to the previous page, not the landing page without
parameters.

### DIFF
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -217,7 +217,7 @@ export default class FilterBoxComponent extends Component {
       this._filterComponents.push(component);
       this._filterNodes[i] = component.getFilterNode();
     }
-    this._saveFilterNodesToStorage();
+    this._saveFilterNodesToStorage(true);
 
     // Initialize apply button
     if (!this.config.searchOnChange) {
@@ -225,7 +225,7 @@ export default class FilterBoxComponent extends Component {
 
       if (button) {
         DOM.on(button, 'click', () => {
-          this._saveFilterNodesToStorage();
+          this._saveFilterNodesToStorage(false);
           this._search();
         });
       }
@@ -257,7 +257,7 @@ export default class FilterBoxComponent extends Component {
   onFilterNodeChange (index, filterNode, saveFilterNodes, searchOnChange) {
     this._filterNodes[index] = filterNode;
     if (saveFilterNodes || searchOnChange) {
-      this._saveFilterNodesToStorage();
+      this._saveFilterNodesToStorage(false);
     }
     if (searchOnChange) {
       this._search();
@@ -275,14 +275,16 @@ export default class FilterBoxComponent extends Component {
   /**
    * Save current filters to storage to be used in the next search
    * @private
+   * @param {boolean} replaceHistory Whether we replace or push a new history
+   *                                 state for the associated changes
    */
-  _saveFilterNodesToStorage () {
+  _saveFilterNodesToStorage (replaceHistory) {
     if (this.config.isDynamic) {
       const availableFieldIds = this.config.filterConfigs.map(config => config.fieldId);
       this.core.setFacetFilterNodes(availableFieldIds, this._getValidFilterNodes());
-      this._filterComponents.forEach(fc => fc.saveSelectedToPersistentStorage());
+      this._filterComponents.forEach(fc => fc.saveSelectedToPersistentStorage(replaceHistory));
     } else {
-      this._filterComponents.forEach(fc => fc.apply());
+      this._filterComponents.forEach(fc => fc.apply(replaceHistory));
     }
   }
 
--- a/src/ui/components/filters/filterboxcomponent.js
+++ b/src/ui/components/filters/filterboxcomponent.js
@@ -217,7 +217,7 @@ export default class FilterBoxComponent extends Component {
       this._filterComponents.push(component);
       this._filterNodes[i] = component.getFilterNode();
     }
-    this._saveFilterNodesToStorage(true);
+    this._saveFilterNodesToStorage(this.config.isDynamic);
 
     // Initialize apply button
     if (!this.config.searchOnChange) {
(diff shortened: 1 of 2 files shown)


src/ui/components/filters/filterboxcomponent.js
```diff
@@ -275,14 +275,16 @@ export default class FilterBoxComponent extends Component {
   /**
    * Save current filters to storage to be used in the next search
    * @private
+   * @param {boolean} replaceHistory Whether we replace or push a new history
+   *                                 state for the associated changes
    */
-  _saveFilterNodesToStorage () {
+  _saveFilterNodesToStorage (replaceHistory) {
```
oshi97, line 281: instead of using this method directly, would it make sense to have two separate methods that call this one? one for when the component is mounted and the other for updates? Otherwise it's less clear to me why do some calls replaceHistory and others push it? Or maybe instead of calling this replaceHistory, which leaks implementation details, you could call it something like isNewSearch?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -571,7 +571,12 @@ export default class FilterOptionsComponent extends Component {
     this.updateListeners();
   }
 
-  apply () {
+  /**
+   * Apply filter changes
+   * @param {boolean} replaceHistory Whether we replace or push a new history
+   *                                 state for the associated changes
+   */
+  apply (replaceHistory) {
```
oshi97, line 579: same as the comment above for saveFilterNodes

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -217,15 +217,15 @@ export default class FilterBoxComponent extends Component {
       this._filterComponents.push(component);
       this._filterNodes[i] = component.getFilterNode();
     }
-    this._saveFilterNodesToStorage();
+    this._saveFilterNodesToStorage(true);
```
tmeyer2115: The `FilterBox` component is not only use for Facets. It can be used to represent a collection of static filters as well. Do we want this behavior on that case too? Or should we guard this with `this.config.isDynamic`?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -269,7 +269,7 @@ export default class FilterOptionsComponent extends Component {
     this.showMoreState = this.config.showMore;
 
     if (this.config.storeOnChange) {
-      this.apply();
+      this.apply(true);
```
tmeyer2115: Same thing about wanting this behavior if `FilterOptions` is not being used for a `Facet`.

src/ui/components/filters/filterboxcomponent.js
tmeyer2115: I don't know if there are ideal solutions here. We need to get this out for Mercy and I don't think we're gonna come up with something we all feel good about in the time allotted. The `Facets` and `Filters` components are just a big mess. I'm not blaming any of us for that. They were in bad shape when we inherited them. Over the past couple months, we've been repeatedly asked to quickly patch things for Mercy. This has made the situation worse. It's clear that when these components were designed, there was not much though put into how they'd work with refreshes, back nav, etc. 

I'm advocating that we just get something working for them (that doesn't add regressions or other bugs) and plan to rebuild all of this soon.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -643,13 +648,14 @@ export default class FilterOptionsComponent extends Component {
...
+   *                                 state for the associated changes
    */
-  saveSelectedToPersistentStorage () {
-    const replaceHistory = (this.core.persistentStorage.get(this.name) === null);
+  saveSelectedToPersistentStorage (replaceHistory) {
     this.core.persistentStorage.set(
       this.name,
       this.config.options.filter(o => o.selected).map(o => o.label),
-      replaceHistory
+      replaceHistory || (this.core.persistentStorage.get(this.name) === null)
```
tmeyer2115, line 658: Could we run into a situation here where `replaceHistory` is false and we overwrite it here? Should we be checking that `replaceHistory` is `undefined` instead?

creotutar: So previously, it was always that that value `(this.core.persistentStorage.get(this.name) === null)`. With these changes, we have made it so that it can be overrided to `true` (but never overrided to `false` because of the `||`) based on the `replaceHistory` parameter.

So I guess semantically, this `replaceHistory` parameter is more `overrideReplaceHistoryIfTrue`. If you wanted to check for `undefined` before defaulting, I could remove the `false` from the other function calls and change this logic to

```
(replaceHistory !== undefined) ? replaceHistory : (this.core.persistentStorage.get(this.name) === null)
```

tmeyer2115 (approved):

PR #1065, opened 2020-09-04
performance: Add performance marks at major events

creotutar:
We add performance marks for better metrics into the time between the
initialization of answers and the subsequent major events like the first
query request sent or the query response rendering.

J=None
TEST=manual

Test that on a page load you see all nine performance marks in the
console window.performance.entries() (page with query, depends
on whether you are searching on a vertical or a universal for the
four query marks)

yext.answers.initStart
yext.answers.statusStart
yext.answers.statusEnd
yext.answers.ponyfillStart
yext.answers.ponyfillend
yext.answers.verticalQueryStart
yext.answers.verticalQuerySent
yext.answers.verticalQueryResponseReceived
yext.answers.verticalQueryResponseRendered
yext.answers.universalQueryStart
yext.answers.universalQuerySent
yext.answers.universalQueryResponseReceived
yext.answers.universalQueryResponseRendered

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -135,6 +135,7 @@ export default class Core {
    * @param {boolean} query.append If true, adds the results of this query to the end of the current results, defaults false
    */
   verticalSearch (verticalKey, options = {}, query = {}) {
+    window.performance.mark('yext.answers.queryStart');
     if (!query.append) {
       this.globalStorage.set(StorageKeys.VERTICAL_RESULTS, VerticalResults.searchLoading());
       this.globalStorage.set(StorageKeys.SPELL_CHECK, {});
@@ -232,6 +233,7 @@ export default class Core {
         if (typeof analyticsEvent === 'object') {
           this._analyticsReporter.report(AnalyticsEvent.fromData(analyticsEvent));
         }
+        window.performance.mark('yext.answers.queryResponseRendered');
       });
   }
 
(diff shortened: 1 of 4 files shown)


tests/core/search/searchapi.js
```diff
@@ -10,6 +10,10 @@ describe('vertical searching', () => {
   let searchApi;
 
   beforeEach(() => {
+    delete global.window.performance;
+    global.window.performance = {
+      mark: () => {}
+    };
```
tmeyer2115, line 16: What's this chunk do?

src/core/core.js
```diff
@@ -135,6 +135,7 @@ export default class Core {
    * @param {boolean} query.append If true, adds the results of this query to the end of the current results, defaults false
    */
   verticalSearch (verticalKey, options = {}, query = {}) {
+    window.performance.mark('yext.answers.queryStart');
```
tmeyer2115: Just do have the granularity, let's have separately named marks for vertical vs. universal search. That may tell us something interesting.

tests/core/search/searchapi.js
creotutar: It's putting window.performance.mark on the global namespace for tests (previously not there, got errors about not having it)

This is a similar approach to https://github.com/yext/answers/blob/master/tests/core/storage/persistentstorage.js#L9

src/core/core.js
creotutar: done

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #1066, opened 2020-09-08
Compile CSS only once in gulp build task

alexisgrow:
Previously we were re-compiling the CSS for every localized bundle. Since the CSS doesn't change on a per-locale basis, update to compile the CSS only once.

TEST=manual

Run npm run build with three locales: en, fr, es, and see answer.css built.


oshi97 (approved):
lgtm!

PR #1067, opened 2020-09-08
Add intermixed markdown tests

cea2aj:
Add intermixed markdown tests

J=none
TEST=auto

Add intermixed markdown tests to the translator with interpolation and pluralization

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -511,7 +511,8 @@ class Answers {
    * @param {number} count The count associated with the pluralization
    */
   translateJS (translations, interpolationParams, count) {
-    return Translator.translate(translations, interpolationParams, count);
+    const translation = Translator.translate(translations, interpolationParams, count);
+    return this.renderer.SafeString(translation);
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -509,11 +509,10 @@ class Answers {
    * @param {string} translations The translations, or a stringified JSON of possible translations
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
-   * @returns {SafeString} The translation as a Handlebars SafeString
+   * @returns {string} The translation
    */
   translateJS (translations, interpolationParams, count) {
-    const translation = Translator.translate(translations, interpolationParams, count);
-    return this.renderer.SafeString(translation);
+    return Translator.translate(translations, interpolationParams, count);
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -509,6 +509,7 @@ class Answers {
    * @param {string} translations The translations, or a stringified JSON of possible translations
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
+   * @returns {SafeString} The translation as a Handlebars SafeString
    */
   translateJS (translations, interpolationParams, count) {
     const translation = Translator.translate(translations, interpolationParams, count);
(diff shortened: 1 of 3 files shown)


src/answers-umd.js
```diff
@@ -509,9 +509,11 @@ class Answers {
    * @param {string} translations The translations, or a stringified JSON of possible translations
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
+   * @returns {SafeString} The translation as a Handlebars SafeString
    */
   translateJS (translations, interpolationParams, count) {
-    return Translator.translate(translations, interpolationParams, count);
+    const translation = Translator.translate(translations, interpolationParams, count);
+    return this.renderer.SafeString(translation);
```
alexisgrow: If this helper is called in JS rather than Handlebars, does it still work as expected?

oshi97: Same question here! I was mostly wondering whether this would work properly for analytics events. For context components can report analytics with this.analyticsReporter for example https://github.com/yext/answers/blob/c9f8a6b4329223633adbf32d00edcb95e63a6db0/src/ui/components/questions/questionsubmissioncomponent.js#L248

tmeyer2115: Could you explain a bit why we need this here as well? I think `translateJS` will primarily be used with Javascript source, where we wouldn't be worried about escaping HTML.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -195,7 +195,8 @@ export default class HandlebarsRenderer extends Renderer {
 
     this.registerHelper('runtimeTranslation', function (options) {
       let { phrase, count } = options.hash;
-      return Translator.translate(phrase, options.hash, count);
+      const translation = Translator.translate(phrase, options.hash, count);
+      return self.SafeString(translation);
```
tmeyer2115: I'm a bit concerned about always marking the translations a `SafeString`. My reasons are this:

- This causes us to assume some responsibility for the security of any page using this helper. We decided, for them, that their custom translations are safe and don't need escaping. This might burn us a bit with the security team.
- There might be some use case where someone wants to actual escape all the HTML tags in a custom translation. 

What would you think about having consumers of `runtimeTranslation` explicitly choose whether or not they call it with triple or double braces? The `translate` method in Jambo could contain an `shouldUnescape` boolean or something like that:

`{ translate phrase='<div>Blah [[value]]</div>' value=profile.name shouldUnescape=true }`

that would be preprocessed into:

`{{{ runtimeTranslation phrase='<div>Blah [[value]] translated</div>' value=profile.name }}}`

This puts the responsibility on the invoker of the `translate` helper to choose if the phrase should be unescaped or not. When they add custom translations, they must decide explicitly if they should be escaped. We're not on the hook for that decision (and it's consequences).

src/answers-umd.js
cea2aj: It does when called with JS, however the result is wrapped in a SafeString which is essentially an object that looks like this: {string: "Yext"}. I think I am going to remove this though because of Tom's suggestions below

cea2aj: I added this to support building intermixed markdown inside of the javascript and passing it to the template to render. Not supporting intermixed markdown with the JS helper makes a lot of sense though! I can remove this

src/ui/rendering/handlebarsrenderer.js
cea2aj: I really like that idea! I think it makes a lot of sense. Combining that with your other comment, intermixed markdown then would only be supported with the HBS helper with shouldUnescape=true.

cea2aj:
Switching this PR from "SafeString translations" to "Add intermixed markdown tests" because the changes that Tom suggested should go in Jambo and not in the SDK

tmeyer2115 (approved):

PR #1068, opened 2020-09-09
ResultFactory: Expose 'distanceFromFilter' property of result

alexisgrow:
This item is extremely similar to the work we did to expose the "distance" property of the result in https://github.com/yext/answers/pull/581. 

TEST=manual
J=SLAP-583

Test on a Jambo site, using the new field `Formatter.toMiles(profile, 'd_distanceFromFilter')` in place of current usage `Formatter.toMiles(profile)`; see the distance displayed and also test by logging the entire result object in the console. Before this change, field would not appear. After the change, the distanceFromFilter appears if the search response has the property populated ([example query](https://liveapi.yext.com/v2/accounts/me/answers/vertical/query?v=20190101&api_key=df4b24f4075800e5e9705090c54c6c13&jsLibVersion=v1.5.4-25-g87493130&sessionTrackingEnabled=true&input=new%20york&experienceKey=rosetest&version=STAGING&filters=%7B%7D&facetFilters=%7B%7D&verticalKey=KM&limit=20&offset=0&locale=en&referrerPageUrl=)).

### DIFF
--- a/src/core/models/result.js
+++ b/src/core/models/result.js
@@ -95,5 +95,10 @@ export default class Result {
      * @type {number}
      */
     this.distance = data.distance || null;
+
+    /**
+     * @type {number}
+     */
+    this.distanceFromFilter = data.distanceFromFilter || null;
   }
 }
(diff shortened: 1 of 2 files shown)


oshi97 (approved):
lgtm!

PR #1070, opened 2020-09-14
Support different map languages

cea2aj:
Support different map languages

Add option to specify a locale in the map component. Get the language from the locale in accordance with the map provider. Mapbox uses [Mapbox GL Language](https://github.com/mapbox/mapbox-gl-language) to set the language. Note: Mapbox does not support Italian. The Map also supports the Google languages that include regions (e.g. en-GB or fr-CA)

J=SLAP-639
TEST=manual

Run an internationalized jambo build and observe properly localized MapBox and Google maps. Test the fallback to the locale provided in the answers init.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1838,6 +1838,8 @@ ANSWERS.addComponent('Map', {
   apiKey: '',
   // Optional, can be used for Google Maps in place of the API key
   clientId: '',
+  // Optional, used to determine the language of the map
+  locale: 'en-US',
   // Optional, determines whether or not to collapse pins at the same lat/lng
   collapsePins: false,
   // Optional, the zoom level of the map, defaults to 14
--- a/README.md
+++ b/README.md
@@ -1838,8 +1838,8 @@ ANSWERS.addComponent('Map', {
   apiKey: '',
   // Optional, can be used for Google Maps in place of the API key
   clientId: '',
-  // Optional, used to determine the language of the map
-  locale: 'en-US',
+  // Optional, used to determine the language of the map. Defaults to the locale specified in the ANSWERS init.
+  locale: 'en',
   // Optional, determines whether or not to collapse pins at the same lat/lng
   collapsePins: false,
   // Optional, the zoom level of the map, defaults to 14
--- a/src/ui/components/map/mapcomponent.js
+++ b/src/ui/components/map/mapcomponent.js
@@ -63,12 +63,11 @@ export default class MapComponent extends Component {
 
   // TODO(billy) Make ProviderTypes a factory class
   getProviderInstance (type) {
-    const _config = Object.assign({
-      locale: this.core.globalStorage.getState(StorageKeys.LOCALE)
-    }, {
+    const _config = {
+      locale: this.core.globalStorage.getState(StorageKeys.LOCALE),
       ...this._config,
       noResults: this._noResults
-    });
+    };
 
     return new ProviderTypes[type.toLowerCase()](_config);
   }
--- a/src/ui/components/map/mapcomponent.js
+++ b/src/ui/components/map/mapcomponent.js
@@ -63,11 +63,13 @@ export default class MapComponent extends Component {
 
   // TODO(billy) Make ProviderTypes a factory class
   getProviderInstance (type) {
-    const _config = {
+    const _config = Object.assign({
+      locale: this.core.globalStorage.getState(StorageKeys.LOCALE)
+    }, {
       ...this._config,
-      noResults: this._noResults,
-      initLocale: this.core.globalStorage.getState(StorageKeys.LOCALE)
-    };
+      noResults: this._noResults
+    });
+
     return new ProviderTypes[type.toLowerCase()](_config);
   }
 
--- a/src/ui/components/map/mapcomponent.js
+++ b/src/ui/components/map/mapcomponent.js
@@ -65,7 +65,8 @@ export default class MapComponent extends Component {
   getProviderInstance (type) {
     const _config = {
       ...this._config,
-      noResults: this._noResults
+      noResults: this._noResults,
+      initLocale: this.core.globalStorage.getState(StorageKeys.LOCALE)
     };
     return new ProviderTypes[type.toLowerCase()](_config);
   }
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -41,16 +41,12 @@ export default class GoogleMapProvider extends MapProvider {
     const googleMapsCustomLanguages =
       ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
     const locale = localeStr.replace('_', '-');
-    const language = locale.substring(0, 2);
 
     if (googleMapsCustomLanguages.includes(locale)) {
       return locale;
-    } else if (locale.length < 2) {
-      console.error(`Locale '${locale}' must include at least two characters. Falling back to 'en'`);
-      return 'en';
-    } else if (locale.length > 2) {
-      console.warn(`Locale '${locale}' is not supported by Google Maps as a language. Falling back to '${language}'`);
     }
+
+    const language = locale.substring(0, 2);
     return language;
   }
 
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -56,7 +56,7 @@ export default class GoogleMapProvider extends MapProvider {
         onLoad();
       },
       async: true,
-      src: `https://maps.googleapis.com/maps/api/js?${self.generateCredentials()}&language=${self._locale}`
+      src: `https://maps.googleapis.com/maps/api/js?${self.generateCredentials()}&language=${self._language}`
     });
 
     DOM.append('body', script);
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -23,6 +23,35 @@ export default class GoogleMapProvider extends MapProvider {
     if (!this.hasValidClientCredentials() && !this._apiKey) {
       throw new Error('GoogleMapsProvider: Missing `apiKey` or {`clientId`, `signature`}');
     }
+
+    /**
+     * Language of the map.
+     * @type {string}
+     */
+    this._language = this.getSupportedLanguage(this._locale);
+  }
+
+  /**
+   * Google Maps supports some langauge codes that are longer than two characters. If the
+   * locale matches one of these edge cases, use it. Otherwise, fallback on the first two
+   * characters of the locale and print a warning.
+   * @param {string} localeStr Unicode locale
+   */
+  getSupportedLanguage (localeStr) {
+    const googleMapsCustomLanguages =
+      ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
+    const locale = localeStr.replace('_', '-');
+    const language = locale.substring(0, 2);
+
+    if (googleMapsCustomLanguages.includes(locale)) {
+      return locale;
+    } else if (locale.length < 2) {
+      console.error(`Locale '${locale}' must include at least two characters. Falling back to 'en'`);
+      return 'en';
+    } else if (locale.length > 2) {
+      console.warn(`Locale '${locale}' is not supported by Google Maps as a language. Falling back to '${language}'`);
+    }
+    return language;
   }
 
   loadJS () {
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -28,7 +28,7 @@ export default class GoogleMapProvider extends MapProvider {
      * Language of the map.
      * @type {string}
      */
-    this._language = this.getSupportedLanguage(this._locale);
+    this._language = this.getValidatedLanguage(this._locale);
   }
 
   /**
@@ -37,7 +37,7 @@ export default class GoogleMapProvider extends MapProvider {
    * characters of the locale and print a warning.
    * @param {string} localeStr Unicode locale
    */
-  getSupportedLanguage (localeStr) {
+  getValidatedLanguage (localeStr) {
     const googleMapsCustomLanguages =
       ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
     const locale = localeStr.replace('_', '-');
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -56,7 +56,7 @@ export default class GoogleMapProvider extends MapProvider {
         onLoad();
       },
       async: true,
-      src: `https://maps.googleapis.com/maps/api/js?${self.generateCredentials()}`
+      src: `https://maps.googleapis.com/maps/api/js?${self.generateCredentials()}&language=${self._locale}`
     });
 
     DOM.append('body', script);
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -28,16 +28,16 @@ export default class GoogleMapProvider extends MapProvider {
      * Language of the map.
      * @type {string}
      */
-    this._language = this.getValidatedLanguage(this._locale);
+    this._language = this.getLanguage(this._locale);
   }
 
   /**
    * Google Maps supports some langauge codes that are longer than two characters. If the
    * locale matches one of these edge cases, use it. Otherwise, fallback on the first two
-   * characters of the locale and print a warning.
+   * characters of the locale.
    * @param {string} localeStr Unicode locale
    */
-  getValidatedLanguage (localeStr) {
+  getLanguage (localeStr) {
     const googleMapsCustomLanguages =
       ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
     const locale = localeStr.replace('_', '-');
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -32,7 +32,7 @@ export default class GoogleMapProvider extends MapProvider {
   }
 
   /**
-   * Google Maps supports some langauge codes that are longer than two characters. If the
+   * Google Maps supports some language codes that are longer than two characters. If the
    * locale matches one of these edge cases, use it. Otherwise, fallback on the first two
    * characters of the locale.
    * @param {string} localeStr Unicode locale
--- a/src/ui/components/map/providers/mapboxmapprovider.js
+++ b/src/ui/components/map/providers/mapboxmapprovider.js
@@ -5,6 +5,9 @@ import MapboxLanguage from '@mapbox/mapbox-gl-language';
 import MapProvider from './mapprovider';
 import DOM from '../../../dom/dom';
 
+const supportedLanguages =
+      ['en', 'es', 'fr', 'de', 'ar', 'ja', 'ko', 'pt', 'ru', 'zh'];
+
 /* global mapboxgl */
 
 /**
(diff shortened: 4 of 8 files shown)


src/ui/components/map/providers/mapprovider.js
```diff
@@ -90,6 +90,18 @@ export default class MapProvider {
      * @type {boolean}
      */
     this._collapsePins = config.collapsePins || false;
+
+    /**
+     * Locale of the map. Defaults to 'en'
+     * @type {string}
+     */
+    this._locale = config.locale || 'en';
```
oshi97: do we need to set this? if not could just be `const locale =`

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -3,45 +3,71 @@
...
     });
 
+    let mapboxLanguageScript = DOM.createEl('script', {
+      id: 'yext-map-language-js',
+      onload: () => {
+        this._isMapboxLanguageLoaded = true;
+        this._onScriptLoad(onLoad);
+      },
+      async: true,
+      src: 'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-language/v0.10.1/mapbox-gl-language.js'
```
oshi97: it looks like this script is pretty small, maybe it would be easier and faster to just include it in the bundle? we have the upcoming plans to separate out the components so the hit to the bundle size will be pretty temporary

src/ui/components/map/providers/mapprovider.js
cea2aj: What do you mean here? Not make locale a private property?

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Sure, that works

src/ui/components/map/providers/mapprovider.js
tmeyer2115: We should make a note to update this once the other SDK i18n changes are merged in. I think this should default to the language/locale of the bundle.

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -3,45 +3,71 @@
...
+
+  _onScriptLoad (onLoadCB) {
+    if (!this._isMapboxLoaded || !this._isMapboxLanguageLoaded) {
+      return;
+    }
+
+    this._isLoaded = true;
+    mapboxgl.accessToken = this._apiKey;
+
+    if (typeof onLoad === 'function') {
```
tmeyer2115: Should this be `typeof onLoadCB === 'function'`?

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -3,45 +3,71 @@
...
+      onLoadCB();
```
tmeyer2115: Do we need to do any kind of `this` binding before passing `onLoadCB` to `_onScriptLoad`?

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -3,45 +3,71 @@
...
 /**
  * MapBoxMapProvider is an implementation of a MapProvider
  * that handles the integration with the third party API to expose maps.
  * @extends MapProvider
  */
 export default class MapBoxMapProvider extends MapProvider {
+  constructor (opts = {}, systemOpts = {}) {
+    super(opts, systemOpts);
+
+    this._isMapboxLoaded = false;
```
tmeyer2115: What's the difference between `_isMapboxLoaded` and `isLoaded`?

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Ya, it should

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Possibly, however since implementing Oliver's suggestion of bundling the MapboxLanguage script, this is now outdated

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: The idea was to set `isLoaded` to true once both `_isMapboxLoaded` and `_isMapboxScriptLoaded` were true. However this is now outdated

src/ui/components/map/providers/mapprovider.js
cea2aj: Makes sense! Adding a TODO

oshi97: yeah, it just looks like this is only used in the constructor and not accessed anywhere else

cea2aj: I'll remove the locale property and simply use config.locale directly when setting the language

cea2aj: Actually, I'll use const locale because it is more robust.

README.md
```diff
@@ -1838,6 +1838,8 @@ ANSWERS.addComponent('Map', {
   apiKey: '',
   // Optional, can be used for Google Maps in place of the API key
   clientId: '',
+  // Optional, used to determine the language of the map
```
alexisgrow: Could the default value be listed here in the README comment as well?

src/ui/components/map/providers/mapprovider.js
```diff
@@ -90,6 +90,24 @@ export default class MapProvider {
      * @type {boolean}
      */
     this._collapsePins = config.collapsePins || false;
+
+    /**
+     * Locale of the map. Defaults to 'en'
+     * @type {string}
+     */
+    const locale = config.locale || 'en';
```
alexisgrow: Would it make sense for this to default/fallback to the bundle's locale?

alexisgrow: Oof, I just saw Tom already made this comment here https://github.com/yext/answers/pull/1070#discussion_r488205215. Totally agree about doing it when the i18n changes are in, in the meantime, would it make sense to use the same default as here in the answers-umd.js? https://github.com/yext/answers/blob/master/src/answers-umd.js#L53

README.md
cea2aj: Sure

src/ui/components/map/providers/mapprovider.js
cea2aj: Ya, that makes sense! I wasn't aware of that constant

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -1,8 +1,13 @@
 /** @module MapBoxMapProvider */
 
+import MapboxLanguage from '@mapbox/mapbox-gl-language';
+
 import MapProvider from './mapprovider';
 import DOM from '../../../dom/dom';
 
+const SUPPORTED_LANGUAGES =
+      ['en', 'es', 'fr', 'de', 'ar', 'ja', 'ko', 'pt', 'ru', 'zh'];
```
oshi97: It's kind of annoying that mapbox doesn't expose this array to us https://github.com/mapbox/mapbox-gl-language/blob/master/index.js#L168

I guess we'll just have to manually keep this up to date as mapbox adds new languages?

oshi97 (commented):
+1 from me!

src/ui/components/map/providers/mapprovider.js
cea2aj: After a discussion in slack, we'll default to the locale used in ANSWERS.init() for now

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Ya, I think that's the simplest

src/ui/components/map/providers/mapprovider.js
```diff
@@ -90,6 +90,25 @@ export default class MapProvider {
...
      */
     this._collapsePins = config.collapsePins || false;
+
+    /**
+     * Locale of the map. Defaults to the locale specified in ANSWERS.init() or to 'en'
+     * if a locale is not provided in the init.
+     * @type {string}
+     */
+    const locale = config.locale || config.initLocale || 'en';
+
```
cea2aj: I'm getting the locale that ANSWERS was initialized from global storage and passing it here as the map component constructs the map provider. Is it preferable to access global storage inside this class rather than passing it through the component?

tmeyer2115: The `MapProvider` doesn't have access to `globalStorage`. I think adding that in would be a bit cumbersome. I think it might simplify things to have this class take in only `config.locale` (not `config.initLocale`) and have the `MapComponent` be responsible for providing the right one. Potentially, the `MapComponent` can have a `_locale` member variable that's `config.locale || globalStorageValue` and that's passed to the provider as `config.locale`. Also, I think we can get rid of the `en` default as `globalStorage` should always have `LOCALE` stored.

src/ui/components/map/providers/mapprovider.js
```diff
@@ -90,6 +90,25 @@ export default class MapProvider {
...
+     * Language of the map. Defaults to 'en'
+     * @type {string}
+     */
+    this._language = locale.substring(0, 2) || 'en';
+
+    /**
+     * Fallback which is used if _language is not supported
+     * @type {string}
+     */
+    this._languageFallback = 'en';
```
tmeyer2115: My initial thought is we don't make this a member variable of the `MapProvider` super class as it only impacts the Mapbox implementation. Do you think we will need this for Google at some point?

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -11,6 +16,14 @@ import DOM from '../../../dom/dom';
  * @extends MapProvider
  */
 export default class MapBoxMapProvider extends MapProvider {
+  constructor (opts = {}, systemOpts = {}) {
+    super(opts, systemOpts);
+
+    this._language = SUPPORTED_LANGUAGES.includes(this._language)
+      ? this._language
+      : this._languageFallback;
```
tmeyer2115: Are we going to need more sophisticated `languageFallback` logic than just choosing `en`? I think we may want to discuss further with Product.

src/ui/components/map/providers/mapprovider.js
cea2aj: I like that solution. Updating.

src/ui/components/map/providers/mapprovider.js
cea2aj: Good point. It's unlikely we'll need this for Google since Google supports everything MapBox does and more. Updating

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: It might not be necessary. This was added for languages such as Italian which aren't supported by Mapbox. If Product wants to use a map language other than english for an Italian site then can specify a locale in a jambo VTC.

src/ui/components/map/providers/mapprovider.js
cea2aj: Oops. I deleted Tom's comment because it looks like it was duplicated, but it actually deleted the original comment as well. Here's what it said:

The MapProvider doesn't have access to `globalStorage`. I think adding that in would be a bit cumbersome. I think it might simplify things to have this class take in only `config.locale` (not `config.initLocale`) and have the MapComponent be responsible for providing the right one. Potentially, the MapComponent can have a `_locale` member variable that's `config.locale || globalStorageValue` and that's passed to the provider as `config.locale`. Also, I think we can get rid of the en default as `globalStorage` should always have `LOCALE` stored.

cea2aj: Instead of a `_locale` member, I'm going to use Obect.assign() when creating the config because the MapComponent doesn't seem to have access to the locale in the constructor.

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -58,6 +66,10 @@ export default class MapBoxMapProvider extends MapProvider {
       center: this.getCenterMarker(mapData)
     });
 
+    this._map.addControl(new MapboxLanguage({
+      defaultLanguage: this._language
```
tmeyer2115, line 74: I see there was some more discussion about how the fallback language should be computed. What, ultimately, was the decision there? Did we decide not to fallback?

src/ui/components/map/mapcomponent.js
```diff
@@ -63,10 +63,13 @@ export default class MapComponent extends Component {
 
   // TODO(billy) Make ProviderTypes a factory class
   getProviderInstance (type) {
-    const _config = {
+    const _config = Object.assign({
+      locale: this.core.globalStorage.getState(StorageKeys.LOCALE)
+    }, {
```
alexisgrow: nit: do we need this `}, {` line?

src/ui/components/map/providers/mapprovider.js
```diff
@@ -90,6 +90,19 @@ export default class MapProvider {
...
+     * ANSWERS.init() by default
+     * @type {string}
+     */
+    const locale = config.locale;
+
+    /**
+     * Language of the map. Defaults to 'en'
+     * @type {string}
+     */
+    this._language = locale.substring(0, 2) || 'en';
```
alexisgrow: Is the language always going to be the substring from 0 to 2? I'm used to seeing it parsed by getting the substring until - or _ but I have no idea what the rules are for language codes 🙈

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Rose wanted it to fall back to english and print an error in the console that the locale is not supported. That's actually the default behavior of Mapbox so I removed the fallback code I added

src/ui/components/map/mapcomponent.js
cea2aj: Nope, I'll make this more compact!

src/ui/components/map/providers/mapprovider.js
cea2aj: Yep, there are few specifications for locale strings, but the specs that I've seen always require a language as the first two characters. Both Google and Mapbox support these codes. This will miss a couple locales that Google supports that include region (e.g. en-GB or fr-CA). I can check with product, and if they want that functionality I can tweak this to support that

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Update: Product said to just fallback on english in this case

src/ui/components/map/providers/mapprovider.js
cea2aj: Updating to support these edge cases

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,35 @@ export default class GoogleMapProvider extends MapProvider {
...
+   * @param {string} localeStr Unicode locale
+   */
+  getSupportedLanguage (localeStr) {
+    const googleMapsCustomLanguages =
+      ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
+    const locale = localeStr.replace('_', '-');
+    const language = locale.substring(0, 2);
+
+    if (googleMapsCustomLanguages.includes(locale)) {
+      return locale;
```
cea2aj, line 46: Are we polyfilling includes() here?

tmeyer2115: Most likely. We include all the `core-js` polyfills in the legacy SDK bundles. As a sanity check, you could verify if we use `includes` elsewhere in the SDK.

cea2aj: It's being used elsewhere in the SDK so we are good to go

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,35 @@ export default class GoogleMapProvider extends MapProvider {
...
+    } else if (locale.length < 2) {
+      console.error(`Locale '${locale}' must include at least two characters. Falling back to 'en'`);
```
tmeyer2115: If `locale.length < 2`, would the `locale.substring` call above give some kind of error?

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,35 @@ export default class GoogleMapProvider extends MapProvider {
...
+   * Google Maps supports some langauge codes that are longer than two characters. If the
+   * locale matches one of these edge cases, use it. Otherwise, fallback on the first two
+   * characters of the locale and print a warning.
+   * @param {string} localeStr Unicode locale
+   */
+  getSupportedLanguage (localeStr) {
+    const googleMapsCustomLanguages =
+      ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
+    const locale = localeStr.replace('_', '-');
+    const language = locale.substring(0, 2);
```
tmeyer2115: Small thing, but can we move `const language = ...` into the final `else-if` case? Having it here reads a little strangely. Seeing it above all the conditionals, I expect that the different cases are going to process `language` in some way to produce the correct answer. But, it's just used in the one conditional branch.

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,35 @@ export default class GoogleMapProvider extends MapProvider {
...
+
+    if (googleMapsCustomLanguages.includes(locale)) {
+      return locale;
+    } else if (locale.length < 2) {
+      console.error(`Locale '${locale}' must include at least two characters. Falling back to 'en'`);
+      return 'en';
+    } else if (locale.length > 2) {
+      console.warn(`Locale '${locale}' is not supported by Google Maps as a language. Falling back to '${language}'`);
```
tmeyer2115: I'm not sure if we want a `console.warn` here. This doesn't seem like an exceptional state, it actually seems fairly common. Most locales, save those special ones, correspond to a language that's just the first two characters. I think the Mapbox warning is different. It's rarer and signifies that it's impossible to find a supported language for your locale. Here, we're able to find a supported language.

src/ui/components/map/providers/googlemapprovider.js
cea2aj: Yes, it would. I'll re-order to prevent that case

src/ui/components/map/providers/googlemapprovider.js
cea2aj: Sure, that makes sense

src/ui/components/map/providers/googlemapprovider.js
cea2aj: Got it. I can remove it

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,31 @@ export default class GoogleMapProvider extends MapProvider {
...
+     * Language of the map.
+     * @type {string}
+     */
+    this._language = this.getValidatedLanguage(this._locale);
+  }
+
+  /**
+   * Google Maps supports some langauge codes that are longer than two characters. If the
+   * locale matches one of these edge cases, use it. Otherwise, fallback on the first two
+   * characters of the locale and print a warning.
```
tmeyer2115: Update description. No warning is printed any more.

src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -11,6 +13,16 @@ import DOM from '../../../dom/dom';
...
  */
 export default class MapBoxMapProvider extends MapProvider {
+  constructor (opts = {}, systemOpts = {}) {
+    super(opts, systemOpts);
+
+    /**
+     * Language of the map. Defaults to 'en'
+     * @type {string}
+     */
+    this._language = this._locale.substring(0, 2) || 'en';
```
tmeyer2115: Is the `|| 'en'` default needed here? Will `_getValidatedLocale` take care of that default already? I think at this point in the code, `this._locale` will be defined (and will have at least two characters).

src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -23,6 +23,31 @@ export default class GoogleMapProvider extends MapProvider {
...
+   * @param {string} localeStr Unicode locale
+   */
+  getValidatedLanguage (localeStr) {
```
tmeyer2115: What would you think about renaming this to `getLanguage` or `getLanguageForGoogle`? I don't think this method is so much validating a language as just picking the right language for Google, based on the locale.

src/ui/components/map/providers/googlemapprovider.js
cea2aj: updating

src/ui/components/map/providers/mapboxmapprovider.js
cea2aj: Ya, that's a good point. It's not needed. I'll take it out.

src/ui/components/map/providers/googlemapprovider.js
cea2aj: I was debating this. I erred on the side of using "validate" for both `getValidatedLocale` and `getValidatedLanguage` because they seemed to be doing similar things, however I think that they are different enough to warrant a name change. Updating to `getLanguage`

tmeyer2115 (approved):

PR #1071, opened 2020-09-15
Ensure all styling in the SDK is self contained.

tmeyer2115:
The SDK's styling should affect only the components on the page and nothing else. Unfortunately, we had some global CSS resets that were applied to an entire page, not just the Answers components. This PR corrects that. In order to scope these resets, we added a CSS class name that will be applied to all component containers. The class name chosen was 'yxt-Answers-component'. I also correctly scoped some styling for the Pagination component.

In the future, we should add a CSS linter to the SDK. We should add a linting rule that will detect any styling not scoped to the components.

J=SLAP-571
TEST=manual

Ensured that the CSS resets were still applied to all components. Verified that I could override a reset within a component's styling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,8 +312,8 @@ Every component has the same base configuration options.
     container: 'container',
     // Optional, a unique name for the component
     name: 'name',
-    // Optional, an additional, custom HTML classname for the component. The component will also have a
-    // classname of 'yxt-Answers-component' applied.
+    // Optional, an additional, custom HTML classname for the component. The component will also 
+    // have a classname of 'yxt-Answers-component' applied.
     class: 'class',
     // Optional, handlebars template or HTML to override built-in handlebars template for the component
     template: 'template',
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ Every component has the same base configuration options.
     container: 'container',
     // Optional, a unique name for the component
     name: 'name',
-    // Optional, an additional, custom HTML classname for the component. The component will already have a
+    // Optional, an additional, custom HTML classname for the component. The component will also have a
     // classname of 'component' applied.
     class: 'class',
     // Optional, handlebars template or HTML to override built-in handlebars template for the component
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ Every component has the same base configuration options.
     // Optional, a unique name for the component
     name: 'name',
     // Optional, an additional, custom HTML classname for the component. The component will also have a
-    // classname of 'component' applied.
+    // classname of 'yxt-Answers-component' applied.
     class: 'class',
     // Optional, handlebars template or HTML to override built-in handlebars template for the component
     template: 'template',
--- a/README.md
+++ b/README.md
@@ -312,7 +312,8 @@ Every component has the same base configuration options.
     container: 'container',
     // Optional, a unique name for the component
     name: 'name',
-    // Optional, a custom HTML classname for the component
+    // Optional, an additional, custom HTML classname for the component. The component will already have a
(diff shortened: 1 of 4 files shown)


src/ui/sass/_base.scss
```diff
@@ -1,15 +1,14 @@
-*:focus
```
creotutar, line 1: Are we sure we want to use component? It seems different from our standard scoped classes. Also seems like this would be a very popular class that SDK users would use

tmeyer2115: I picked it because that's the class we were already including by default on all containers. If we want to use a different class name, we can, but we'd still need the `component` default (removing that would be breaking). I thought using this class name, and ensuring it was always present, was less clunky than adding another one on top of it.

README.md
```diff
@@ -312,7 +312,8 @@ Every component has the same base configuration options.
     container: 'container',
     // Optional, a unique name for the component
     name: 'name',
-    // Optional, a custom HTML classname for the component
+    // Optional, an additional, custom HTML classname for the component. The component will already have a
+    // classname of 'component' applied.
```
cea2aj: nit: using "also" instead of "already" seems a little clearer to me. This also applies to the comment in component.js

src/ui/components/component.js
```diff
@@ -109,10 +109,11 @@ export default class Component {
     }
 
     /**
-     * A custom class to be applied to {this._container} node
+     * A custom class to be applied to {this._container} node. Note that the class
+     * 'component' will be included also.
```
cea2aj: tiny nit: can we reword "will be included also" to "will also be included"?

cea2aj: or "will also be applied"

tmeyer2115: If this wording really bothers you, I can change. But, given how close the two phrases are, I'd rather keep as is and not make another commit.

cea2aj: It's totally fine as it is, haha

cea2aj (approved):

src/ui/sass/_base.scss
creotutar: for sure

i agree its easier to use what's already there but i think there's some value to choosing a class that won't interfere with client code (e.g. `yxt-Component` or `yxt-Answers-component`). it feels easier for us to say our contract is "all `yxt` prefaced classes are the classes we are adding styling to", instead of "all yxt + 'component'". i just have the /feeling/ that `component` would be used a lot, esp. in react environments.

i also could not find why we keep the component class (no css referencing the class, not in any js files i saw). do we use that? not important, just something i noticed.

either way, i think the accreting method of adding an answers specific css class makes it easier to talk about our effects. what do you think?

tmeyer2115: I think you're right. We've said before our convention is to prefix classnames with `yxt`. That helps provide collisions and makes it clear to customers where the HTML is coming from. I don't think we've encountered issues so far with our default of `component`, but once we enter React world, we probably will. 

The one thing I don't like is that now people will see divs with the classes `component` and `yxt-Answers-component`. If I were a customer, I'd ask why we need two proprietary class names. Adding the class name `component` was likely a mistake. Unfortunately, it's not one we can remove until a new major version is released. I say we live with this annoyance for now and eye removing the `component` class name in the future. I don't want us to continue adding two classes to every container. 

We don't use the class name of `component` anywhere in the SDK, we provided that as a hook for client styling. Unfortunately, we have no way of know if customers are relying on this class name (they probably are).

creotutar (approved):
💯 💯 💯 LGTM

PR #1072, opened 2020-09-15
Map: add shouldFreezeMap config function

oshi97:
This commit adds the shouldFreezeMap config function, which
allows the user to specify a criteria when the map should not update.

TEST=manual

this is used at http://engblog.yext.com/collapsible-filters-prototype/people.html
to not update the map on mobile phones (since it is hidden in those cases)
Otherwise, for low-end phones, we noticed there was a chance to crash the
page if a user made too many searches too quickly. It also just helps
performance a lot

### DIFF
--- a/README.md
+++ b/README.md
@@ -1847,6 +1847,8 @@ ANSWERS.addComponent('Map', {
   // Optional, determines if an empty map should be shown when there are no results. Defaults to false.
   showEmptyMap: false,
   // Optional, callback to invoke when a pin is clicked. The clicked item(s) are passed to the callback
+  shouldFreezeMap: function() {},
+  // Optional, when this function returns true, the map will not update when new searches are made. When this returns false, it will not have any impact on the map's behavior.
   onPinClick: null,
   // Optional, callback to invoke when a pin is hovered. The clicked item(s) are passed to the callback
   onPinMouseOver: null,
--- a/README.md
+++ b/README.md
@@ -1847,9 +1847,9 @@ ANSWERS.addComponent('Map', {
   // Optional, determines if an empty map should be shown when there are no results. Defaults to false.
   showEmptyMap: false,
   // Optional, callback to invoke when a pin is clicked. The clicked item(s) are passed to the callback
-  shouldFreezeMap: function() {},
-  // Optional, when this function returns true, the map will not update when new searches are made. When this returns false, it will not have any impact on the map's behavior.
   onPinClick: null,
+  // Optional, when this function returns true, the map will not update when new searches are made. When this returns false, it will not have any impact on the map's behavior.
+  shouldFreezeMap: function() {},
   // Optional, callback to invoke when a pin is hovered. The clicked item(s) are passed to the callback
   onPinMouseOver: null,
   // Optional, callback to invoke when a pin is no longer hovered after being hovered. The clicked item(s) are passed to the callback
--- a/src/ui/components/map/mapcomponent.js
+++ b/src/ui/components/map/mapcomponent.js
@@ -46,6 +46,12 @@ export default class MapComponent extends Component {
      * @type {MapProvider}
      */
     this._map = null;
+
+    /**
+     * A configurable method for criteria to not update the map.
+     * @type {Function}
+     */
+    this._shouldFreezeMap = opts.shouldFreezeMap || (() => {});
   }
 
   static get type () {
@@ -82,7 +88,7 @@ export default class MapComponent extends Component {
   }
 
   setState (data, val) {
-    if (Object.keys(data).length === 0) {
+    if (Object.keys(data).length === 0 || this._shouldFreezeMap()) {
       return this;
     }
 


cea2aj:
Can you update the README? I think it could help explain usage

oshi97:
> Can you update the README? I think it could help explain usage

Ah shoot yeah I need to update the readme for all of these changes, thanks for reminding me!

README.md
```diff
@@ -1847,6 +1847,8 @@ ANSWERS.addComponent('Map', {
   // Optional, determines if an empty map should be shown when there are no results. Defaults to false.
   showEmptyMap: false,
   // Optional, callback to invoke when a pin is clicked. The clicked item(s) are passed to the callback
+  shouldFreezeMap: function() {},
```
creotutar: Why is this a function instead of a boolean?

README.md
```diff
@@ -1847,6 +1847,8 @@ ANSWERS.addComponent('Map', {
...
+  // Optional, when this function returns true, the map will not update when new searches are made. When this returns false, it will not have any impact on the map's behavior.
```
creotutar: I think you want the comment above the config option

README.md
creotutar: I see. This is dynamic for that mobile vs desktop case.

creotutar: It is odd to me that we allow this to be a function when a lot of our other boolean config flags are not also functions. It's almost like we are catering this boolean flag specifically for the mobile vs. desktop case. It makes me think we would want a more general way to do mobile vs. desktop so that it can apply to more than just this config option.

oshi97: yeah let me think about this for a bit, I think we also want some config option to support disabling the map when the page is displaying a full screen modal, but currently that wouldn't work because closing the modal wouldn't "unfreeze" the map

oshi97: Actually there's already a number of ways to "unfreeze" the map. 

1. store a reference to the map component, then manually call setState()
```js
let map;
ANSWERS.addComponent('Map', {
    container: '.map-container-google',
    onCreate: function () {
        map = this;
    }
});
```

2. register an additional global storage listener, then manually update the globalStorage state for the key

```js
ANSWERS.addComponent('Map', {
    container: '.map-container-google',
    onCreate: function () {
        this.core.globalStorage.on('update', 'some-custom-key', () => this.setState(this.getState()));
    }
});
```

3. use ANSWERS.components to get a reference to the component, then manually call setState()

```js
ANSWERS.components.getActiveComponent('Map');
```

Anyways, right now if somebody wanted to do something like a full screen modal (and not waste cycles on a hidden map), there's no good way to pause the map. They would have to do something like add onMountOverride, and in that override, include this logic around this._map (in addition to something like one of the above options for unfreezing the map). I think this flag would also make that case a lot easier (in addition to hiding the map on mobile, which is also something we ALREADY do on the vertical-map template).

Does that make this sound slightly less scoped to my specific use case? I'm not sure if it does, but the only other way forward I see would be to export the SDKs map providers, and write a custom map component for them. That's significantly more annoying on the theme side, but if you still feel really adamant that we shouldn't have this config option, then that's fine. Also open to giving it a different name if you have a better one in mind.

oshi97: or tell Jeremy he has to be happy with poor mobile performance

creotutar: I took a look at the other configuration options we offer in the SDK--
* The other `function` config options are usually options that are meant to do actions `onStateChange`, `render`, `transformData`. There is only one that doesn't fit the mold, `pin`, in the Map component which I am assuming is because we want dynamic pins for their multiple states (clicked, hovered, has numbers/not)
* Navigation has mobile-specific config options `mobileOverflowBehavior`, Pagination has mobile vs desktop specific config options e.g. `maxVisiblePagesMobile`
* There are options that could benefit the SDK by having this dynamic function but currently don't like `includeMap` in the UniversalResults config or `limit` in search or `showMoreLimit` in filters

creotutar: I still get the feeling that we want static configuration when we can. I feel /less bad/ about some `ANSWERS.setMapFreeze` but as we talked about online, it's weird that answers would know about the internals of the Map component.

creotutar: As I understand it, we want to have dynamic map freezing/unfreezing on the page (for the case of desktop/mobile and an overlay, to better performance). The reason I suggested above is because the only other cases we have this is for like `ANSWERS.setContext` or `ANSWERS.setGeolocation`. But if want to separate UI from Core Search that is not great.

creotutar: tl;dr I am not sure the way forward, my gut tells me if we expose a function here we give precedence to expose a function in a lot of other places

oshi97: @tmeyer2115 do you have an opinion on this? I was thinking if we don't like this function I could expose the map providers instead, and write a custom map component for the new page template

oshi97: Tom's suggestion of only adding the map when not mobile makes a lot of sense! I previously thought we would have to rely on the screen width to do this, which would cause issues if somebody resized their screen (say from maximizing it or something) BUT we can actually do that with navigator.useragent

oshi97: I did this by forking a tiny method https://github.com/juliangruber/is-mobile/blob/master/index.js

PR #1073, opened 2020-09-15
SearchBar: add onSearchComplete hook

oshi97:
This commit adds the onSearchComplete hook to the SearchBar component.
This allows users to add a callback after ONLY searches made by
the searchbar are completed.
The preexisting callback, onVerticalSearch, occurs after any searches
are made, and does not have context as to whether that search is made
by a filter component, pagination, or the search bar. This removes
the need to manually calculate whether a search originated from the
searchbar.

TEST=manual

used in http://engblog.yext.com/collapsible-filters-prototype/people.html
used to hide the filters panel ONLY when a search is made with the searchbar

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,7 +438,9 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
-  }
+  },
+  // Optional, a callback function that runs after the searchbar completes a search.
+  onSearchComplete: function () { console.log('a search was complete!'); }
 })
 ```
 
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -234,6 +234,8 @@ export default class SearchComponent extends Component {
       message: 'We are unable to determine your location',
       ...config.geolocationTimeoutAlert
     };
+
+    this._onSearchComplete = config.onSearchComplete;
   }
 
   static get type () {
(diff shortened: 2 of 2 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -234,6 +234,8 @@ export default class SearchComponent extends Component {
       message: 'We are unable to determine your location',
       ...config.geolocationTimeoutAlert
     };
+
+    this._onSearchComplete = config.onSearchComplete;
```
creotutar, line 238: jsdoc please

README.md
```diff
@@ -438,7 +438,9 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
-  }
+  },
+  // Optional, a callback function that runs after the searchbar completes a search.
```
creotutar, line 442: What does `completes` a search mean here? Can we add that to the README?

creotutar: We run the `search` function when we do a query on load. Do we want to run `onSearchComplete` in that case? This is true generally whenever the QUERY is updated in globalStorage because of the `update` listener.

oshi97: oh i forgot about this beauty, even the searchbar relies on this query update listener (at least sometimes) it looks like, and debounced search makes things even uglier, I'll just do this a different way

oshi97: where is my cry react github? 😢

creotutar: It's a true cry react :_(

PR #1075, opened 2020-09-15
VerticalResults: add hideResultsHeader config option

oshi97:
This commit adds the 'hideResultsHeader' config option to
VerticalResults. Previously, to hide the results header,
users would have to display: none it. However, the results
header would still do things like perform applied filters
calculations, having a negative impact on page speed.

TEST=manual

used in http://engblog.yext.com/collapsible-filters-prototype/people.html
the above page uses the separated results count and applied filters components,
instead of the normal results header


alexisgrow (approved):
this LGTM!

PR #1077, opened 2020-09-16
Rename translate helpers

cea2aj:
Rename ANSWERS.translateJS() and the HBS translate helper to processTranslation

Rename these helpers to make it clearer that no actual translation is performed. Also rename the Translator class to TranslationProcessor.

J=SLAP-654
TEST=manual, auto

Update unit tests and confirm they pass. Build an internationalized ANSWERS site with the [corresponding jambo branch](https://github.com/yext/jambo/pull/122) and observe a properly translated site.


src/answers-umd.js
```diff
@@ -504,15 +504,15 @@ class Answers {
   }
 
   /**
-   * Performs a translation which supports
+   * Processes a translation which includes performing
    * interpolation, pluralization, or both
```
tmeyer2115: Nit: we broke the line too early above.

src/answers-umd.js
```diff
@@ -504,15 +504,15 @@ class Answers {
...
    * @param {string} translations The translations, or a stringified JSON of possible translations
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
    * @returns {string} The translation
```
tmeyer2115: How about "The translation with any interpolation or pluralization applied"?

src/core/i18n/translationprocessor.js
```diff
@@ -1,14 +1,14 @@
 import { getNPlurals, getPluralFunc, hasLang } from 'plural-forms/dist/minimal';
 
-export default class Translator {
+export default class TranslationProcessor {
   /**
-   * Performs a translation which supports
+   * Processes a translation which includes performing
    * interpolation, pluralization, or both
```
tmeyer2115: Same comment about line break

tests/core/i18n/translator.js
```diff
@@ -1,44 +1,44 @@
-import Translator from '../../../src/core/i18n/translator';
+import TranslationProcessor from '../../../src/core/i18n/translationprocessor';
```
tmeyer2115, line 1: If this file contains the unit tests for the `TranslationProcessor`, it's path should mirror the path to the `TranslationProcessor` source. That is, we should call the file `translationprocessor.js`.

tmeyer2115 (commented):
Do we need to update the calls to `TranslationResolver` in the various Gulp tasks as well? I think those refer to `ANSWERS.translateJS` and `runtimeTranslation`. Do we need to update the README as well?

src/answers-umd.js
cea2aj: fixing

src/answers-umd.js
cea2aj: That works. Updating.

src/core/i18n/translationprocessor.js
cea2aj: fixing

tests/core/i18n/translator.js
cea2aj: Ah, you're right, I overlooked that. Updating

cea2aj:
> Do we need to update the calls to `TranslationResolver` in the various Gulp tasks as well? I think those refer to `ANSWERS.translateJS` and `runtimeTranslation`. Do we need to update the README as well?

The feature/localized-assets branch doesn't look like it's in develop yet. Are we planning on merging that into develop soon? If not, should this PR go into that branch rather than develop?

The translation processor isn't in the README currently. I can add a section describing usage of ANSWERS.processTranslation() and maybe the HBS helper as well, but I think it makes more sense to add that to the README after we change the interface to no longer use JSON.

tmeyer2115:
> > Do we need to update the calls to `TranslationResolver` in the various Gulp tasks as well? I think those refer to `ANSWERS.translateJS` and `runtimeTranslation`. Do we need to update the README as well?
> 
> The feature/localized-assets branch doesn't look like it's in develop yet. Are we planning on merging that into develop soon? If not, should this PR go into that branch rather than develop?
> 
> The translation processor isn't in the README currently. I can add a section describing usage of ANSWERS.processTranslation() and maybe the HBS helper as well, but I think it makes more sense to add that to the README after we change the interface to no longer use JSON.

Ahh, gotcha. We're planning to merge localized-assets in soon, yeah. I'd say Monday most likely. We can merge this PR into develop first. We'll just need to make a note to update feature/localized-assets before it's merged into develop.

tmeyer2115 (approved):

PR #1078, opened 2020-09-16
Add Applied Filters Component

oshi97:
Add Applied Filters Component

This is a factoring out of the applied filters portion of the results header
component. Separating this out allows for basic yet priorly
exhausting combinations. In the past, even wanting to just flip
the order of the results count and applied filters, or insert
simple content in between them, would require a full override
of the results header template.

We can't use this component inside the results header component,
because the SDK is bad at nesting components. Specifically, when
a component is removed (which happens every time a child component
is rerendered) it will remove its moduleId event listener, which has the
side effect of removing all other listeners on that same moduleId.

TEST=manual

This is used in the collapsible filters page
See http://engblog.yext.com/collapsible-filters-prototype/people.html

also added it to a standalone page, and saw that it was the removable filters worked, and it updated at the end of every search

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Outline:
    - [Facets Component](#facets-component)
    - [FilterSearch Component](#filtersearch-component)
    - [Filter Components](#filter-components)
+   - [Applied Filters Component](#applied-filters-component)
    - [Navigation Component](#navigation-component)
    - [QASubmission Component](#qa-submission-component)
    - [SpellCheck Component](#spell-check-component)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1613,6 +1614,33 @@ ANSWERS.addComponent('GeoLocationFilter', {
...
+
+The Applied Filters Component displays your currently applied filters as a row of text tags, labeled
+by filter display value. If the "removable" config option is set to true, these text tags will instead
+be "removable filters", which, when clicked, will remove the clicked filter from the search.
+Only intended for vertical pages.
+
+```js
+ANSWERS.addComponent('AppliedFilters', {
+  container: '.applied-filters-container',
+  // The vertical key of your search, will default to the vertical key specified in the search config.
```
alexisgrow: Could these prop comments be prefixed with "Required," or "Optional"?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -0,0 +1,156 @@
...
+
+  setState (data) {
+    const verticalResults = this.core.globalStorage.getState(StorageKeys.VERTICAL_RESULTS) || {};
+    const nlpFilters = verticalResults.appliedQueryFilters || [];
+
+    /**
+     * Array of nlp filters in the search response.
+     * @type {Array<AppliedQueryFilter>}
+     */
+    this.nlpFilterNodes = convertNlpFiltersToFilterNodes(nlpFilters || []);
```
alexisgrow: Is the `nlpFilters || []` fallback logic duplicated here and on the line above?

alexisgrow:
> We can't use this component inside the results header component,
> because the SDK is bad at nesting components

I saw this comment in the PR description, do you think we'll nest it eventually? When we create this component it looks like we'll have the applied filters code in two places, is that right?

oshi97:
> > We can't use this component inside the results header component,
> > because the SDK is bad at nesting components
> 
> I saw this comment in the PR description, do you think we'll nest it eventually? When we create this component it looks like we'll have the applied filters code in two places, is that right?

Yeah I think once we fix Yext React issues with child components (or swap it for a preact implementation or something) we should be able to nest them!

src/ui/components/results/appliedfilterscomponent.js
oshi97: ah yeah that's duplicated, removed the second fallback!

README.md
oshi97: Sure, they're all optional (except for container) but it's better to be consistent

src/ui/sass/modules/_AppliedFilters.scss
```diff
@@ -0,0 +1,84 @@
...
+    margin-right: calc(var(--yxt-base-spacing) / 4);
+    margin-bottom: 4px;
+    @include Text(
+      $size: var(--yxt-font-size-md),
+      $line-height: var(--yxt-line-height-md),
+      $color: var(--yxt-color-text-secondary),
+    );
+  }
+
+  &-filterValue
```
alexisgrow: Could this and &-filterLabel and &-filterSeparator could be consolidated since they're mostly the same?

oshi97: updated!

alexisgrow (approved):
this lgtm!

PR #1079, opened 2020-09-16
Add results count component

oshi97:
Add ResultsCount component

This is a factoring out of the results count portion of the results header
component. Separating this out allows for basic yet priorly
exhausting combinations. In the past, even wanting to just flip
the order of the results count and applied filters, or insert
simple content in between them, would require a full override
of the results header template.

We can't update the results header component to use this yet,
due to more fundamental issues the sdk has with nested components.

TEST=manual, auto

added jest tests

This is used in the collapsible filters page
See http://engblog.yext.com/collapsible-filters-prototype/people.html


README.md
```diff
@@ -860,6 +861,16 @@ ANSWERS.addComponent('VerticalResults', {
 })
 ```
 
+## Results Count Component
```
alexisgrow: Since the Results Count component only works with the vertical results server response, do you think it would make sense to have "vertical" in its name?

oshi97: ah yeah I like that, updated!

alexisgrow (approved):
lgtm! 🔥

PR #1080, opened 2020-09-17
filteroptions: Fix back navigation behavior by applying previous state

creotutar:
Consider the following answers navigation situation:

In State 1
Search "all"
In State 2
Click on a static filter (e.g. distance "5 miles")
In State 3
Navigate backwards in browser history
In State 4

You would expect state 4 == state 2. However, this is currently not the case.
The results from state 3 still exist in state 4 instead of the results from state 2.
The URL in state 4 reflects the same URL as state 2, however the results are
different. This is because we are not applying the filter changes from the backwards
navigation, even though it is being changed in globalStorage.

This PR aims to update the FilterOptions component when it is updated on a
backwards navigation (ie in globalStorage). We do not fix other static filters like
RangeFilter or DateRangeFilter or SortOptions as that will be part of a bigger rework
aiming to get all Facets/Static Filters manifesting on load/backwards navigation.

J=SLAP-624
TEST=manual

Test on a site using a local SDK dist directly

Test a FilterBox component with RADIUS_FILTER
Test a FilterBox component with STATIC_FILTER
Test a Facet component

Click on a static filter option. Navigate backwards in history. Make sure
the results AND the url reflect correctly in the back nav (that is, the filter
value changes in the URL and the results show the previous state).

Try going from FilterBox.filter0=[[something]] to FilterBox.filter0=[[nothing]]
Try going from FilterBox.filter0=[[something]] to FilterBox.filter0=[[somethingelse]]

### DIFF
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -55,7 +55,7 @@ class FilterOptionsConfig {
      * changing with changes in the persistent storage.
      * @type {object[]}
      */
-    this.options = Array.from(this.initialOptions);
+    this.options = config.options.map(o => ({ ...o }));
 
     /**
      * The label to be used in the legend
@@ -171,28 +171,30 @@ class FilterOptionsConfig {
   }
 
   /**
-   * Sets selected options on load based on options stored in persistent storage and options with selected: true.
-   * If no previous options were stored in persistentStorage, default to options marked
-   * as selected. If multiple options are marked as selected for 'singleoption', only the
-   * first should be selected.
-   * @param {Array<Object>} options
-   * @param {Array<string>} previousOptions
-   * @returns {Array<Object>}
+   * Returns a list of options with `selected` determined by initialOptions and
+   * optionsOverrides. optionsOverrides take precedence over options. If the
+   * control is singleoption and `selected` appears more than once in either
+   * options or previousOptions then the first instance is used.
+   * @param {Array<Object>} initialOptions Options from the component configuration
+   * @param {Array<string>} optionsOverrides Overrides as they appear in persistentStorage
+   * @returns {Array<Object>} The options is the same format as initialOptions with updated
+   *                          selected values
    */
-  setSelectedOptions (options, previousOptions) {
-    if (previousOptions && this.control === 'singleoption') {
+  setSelectedOptions (initialOptions, optionsOverrides) {
+    const options = initialOptions.map(o => ({ ...o }));
+    if (optionsOverrides && this.control === 'singleoption') {
       let hasSeenSelectedOption = false;
       return options.map(o => {
-        if (previousOptions.includes(o.label) && !hasSeenSelectedOption) {
+        if (optionsOverrides.includes(o.label) && !hasSeenSelectedOption) {
           hasSeenSelectedOption = true;
           return { ...o, selected: true };
         }
         return { ...o, selected: false };
       });
-    } else if (previousOptions && this.control === 'multioption') {
+    } else if (optionsOverrides && this.control === 'multioption') {
       return options.map(o => ({
         ...o,
-        selected: previousOptions.includes(o.label)
+        selected: optionsOverrides.includes(o.label)
       }));
     } else if (this.control === 'singleoption') {
       let hasSeenSelectedOption = false;
@@ -282,18 +284,17 @@ export default class FilterOptionsComponent extends Component {
 
     if (!this.config.isDynamic) {
       this.core.globalStorage.on('update', this.name, (data) => {
-        let previousOptions;
         try {
-          previousOptions = JSON.parse(data);
+          const newOptions = JSON.parse(data);
+          this.config.options = this.config.setSelectedOptions(
+            this.config.initialOptions,
+            newOptions
+          );
+          this.updateListeners();
+          this.setState();
         } catch (e) {
           console.warn(`Filter option ${data} could not be parsed`);
         }
-        this.config.options = this.config.setSelectedOptions(
-          this.config.initialOptions,
-          previousOptions
-        );
-        this.updateListeners();
-        this.setState();
       });
     }
   }
--- a/src/ui/components/filters/filteroptionscomponent.js
+++ b/src/ui/components/filters/filteroptionscomponent.js
@@ -44,10 +44,18 @@ class FilterOptionsConfig {
     this.optionType = config.optionType || OptionTypes.STATIC_FILTER;
 
     /**
-     * The list of filter options to display with checked status
+     * The list of filter options to display with checked status as
+     * initially specified in the user configuration
      * @type {object[]}
      */
-    this.options = config.options.map(o => ({ ...o }));
+    this.initialOptions = config.options.map(o => ({ ...o }));
+
+    /**
+     * The list of filter options to display with checked status,
+     * changing with changes in the persistent storage.
+     * @type {object[]}
+     */
+    this.options = Array.from(this.initialOptions);
 
     /**
      * The label to be used in the legend
@@ -270,6 +278,23 @@ export default class FilterOptionsComponent extends Component {
 
     if (this.config.storeOnChange) {
       this.apply(this.config.isDynamic);
+    }
+
+    if (!this.config.isDynamic) {
+      this.core.globalStorage.on('update', this.name, (data) => {
+        let previousOptions;
+        try {
+          previousOptions = JSON.parse(data);
+        } catch (e) {
+          console.warn(`Filter option ${data} could not be parsed`);
+        }
+        this.config.options = this.config.setSelectedOptions(
+          this.config.initialOptions,
+          previousOptions
+        );
+        this.updateListeners();
+        this.setState();
+      });
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -271,6 +279,23 @@ export default class FilterOptionsComponent extends Component {
...
+      this.core.globalStorage.on('update', this.name, (data) => {
+        let previousOptions;
+        try {
+          previousOptions = JSON.parse(data);
+        } catch (e) {
+          console.warn(`Filter option ${data} could not be parsed`);
+        }
+        this.config.options = this.config.setSelectedOptions(
+          this.config.initialOptions,
+          previousOptions
```
alexisgrow: Should this stuff happen if a JSON.parse error was caught?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -44,10 +44,18 @@ class FilterOptionsConfig {
...
      */
-    this.options = config.options.map(o => ({ ...o }));
+    this.initialOptions = config.options.map(o => ({ ...o }));
+
+    /**
+     * The list of filter options to display with checked status,
+     * changing with changes in the persistent storage.
+     * @type {object[]}
+     */
+    this.options = Array.from(this.initialOptions);
```
oshi97: Do we want to do a deep copy here as well?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -271,6 +279,23 @@ export default class FilterOptionsComponent extends Component {
     if (this.config.storeOnChange) {
       this.apply(this.config.isDynamic);
     }
+
+    if (!this.config.isDynamic) {
+      this.core.globalStorage.on('update', this.name, (data) => {
```
oshi97, line 288: I don't have the previous back nav changes down by heart, how exactly does hitting the back button cause this globalStorage module to update?

tmeyer2115:
> Draft
> 
> In State 1
> Search "all"
> In State 2
> Click on a static filter (e.g. distance "5 miles")
> In State 3
> Navigate backwards in browser history
> In State 4
> 
> You would expect state 4 == state 2. However, this is currently not the case.
> Instead, previous filters are not applied, even though they are reflected
> as present in the URL.
> 
> J=SLAP-624
> TEST=manual
> 
> Test on a site using a local SDK dist directly
> 
> Test a FilterBox component with static filters
> Test a SortOptions component

I'm a little confused by the flow outlined above. I'd also expect State 4 to be the same as State 2. But, in State 2, there were no  filters applied. The comment seems to indicate that we would expect filters. Could you clarify a bit? Is it that we'd expect the filters to be selected, but not applied?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -271,6 +279,23 @@ export default class FilterOptionsComponent extends Component {
...
+        let previousOptions;
+        try {
+          previousOptions = JSON.parse(data);
```
tmeyer2115: If the `data` is passed via an `update` listener, would it be the currently selected options? Are we assigning it to `prevOptions` for the next time we hit this code pathway?

src/ui/components/filters/filteroptionscomponent.js
creotutar: Yes! Changed tytyty

src/ui/components/filters/filteroptionscomponent.js
creotutar: We absolutely do, tytyty changed

src/ui/components/filters/filteroptionscomponent.js
creotutar: On back nav in `answers-umd.js` resetListener to persistentStorage, we `globalStorage.setAll()` with all the URL query params (which would include this filter name since filter params are added automatically to the URL)

src/ui/components/filters/filteroptionscomponent.js
creotutar: Yeyeye, I used `previousOptions` as that was how it was described in the function `setSelectedOptions` (and made sense for my case of back nav). But! I think that could be more generalized as `optionsOverrides`. Made that change, considering updates would not /only/ be from a back nav (although I just did a check and am noticing that the only way filteroptions would be updated currently is through a back nav)

creotutar: Updated the README! Lmk what you think

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -159,32 +167,34 @@ class FilterOptionsConfig {
...
-   * If no previous options were stored in persistentStorage, default to options marked
-   * as selected. If multiple options are marked as selected for 'singleoption', only the
-   * first should be selected.
-   * @param {Array<Object>} options
-   * @param {Array<string>} previousOptions
-   * @returns {Array<Object>}
+   * Returns a list of options with `selected` determined by initialOptions and
+   * optionsOverrides. optionsOverrides take precedence over initialOptions. If the
+   * control is singleoption and `selected` appears more than once in either
+   * initialOptions or previousOptions then the first instance is used.
```
alexisgrow: I think `previousOptions` should be `optionsOverrides` here

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -159,32 +167,34 @@ class FilterOptionsConfig {
...
+   * @param {Array<Object>} initialOptions Options from the component configuration
+   * @param {Array<string>} optionsOverrides Overrides as they appear in persistentStorage
```
alexisgrow: I'm not super familiar with persistentStorage, what does "Overrides as they appear in persistentStorage"  mean in this case?

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -44,7 +44,15 @@ class FilterOptionsConfig {
...
-     * The list of filter options to display with checked status
+     * The list of filter options to display with checked status as
+     * initially specified in the user configuration
+     * @type {object[]}
+     */
+    this.initialOptions = config.options.map(o => ({ ...o }));
+
+    /**
+     * The list of filter options to display with checked status,
+     * changing with changes in the persistent storage.
```
tmeyer2115: Is there a way to rephrase "changing with changes in the persistent storage"? I think this might confuse someone because they'll see this comment but that it's getting modified in `globalStorage` update listener.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -271,6 +281,22 @@ export default class FilterOptionsComponent extends Component {
     if (this.config.storeOnChange) {
       this.apply(this.config.isDynamic);
     }
+
+    if (!this.config.isDynamic) {
+      this.core.globalStorage.on('update', this.name, (data) => {
```
tmeyer2115, line 288: I know Clean Code says not to use comments as a crutch, but I think we should have a comment above this block indicating that this listener is fired during back nav. Without a comment, I think it will be unclear to people why this is here. The long term solution is to better architect our storage systems so that this kind of manual reconciliation (from persistentStorage format to globalStorage format) is no longer needed (or we have a better pattern for it). But, that's wayyyyy beyond the scope of this item.

tmeyer2115, line 288: Can this listener get fired outside the back nav case? It really gets triggered whenever `resetListener` is called, right? Can that listener be fired for other kinds of events?

src/ui/components/filters/filteroptionscomponent.js
creotutar: Terribly worded haha, I removed it, I think it's not worth mentioning when an instance variable is dynamic

src/ui/components/filters/filteroptionscomponent.js
creotutar: Ah, I was going for "The options as they are formatted for persistentStorage" (ie how they look in the URL)

src/ui/components/filters/filteroptionscomponent.js
creotutar: Fixed!

src/ui/components/filters/filteroptionscomponent.js
creotutar: It definitely _can_
Does it? It does not. To double check I went through a few interactions with a `console.log` in the listener and found it only logged on back/fwd nav.

This is because we normally work with `static-filter-node.FilterBox.filter0` in global storage when working with static filters. The reason we do not update that `static-filter-node` key here is because we would need to
* Find whether or not that query param was a static vs dynamic filter so we can prepend the correct `x-filter-node` prefix.
* Cast the query parameter into a FilterNode
* Update the key with the FilterNode and then update options based on this an update on this key. However, because this key is updated in other places in filteroptions, it would end up being a bigger refactor of the FilterOptions general workflow to make sure we are only updating options for the back nav case.

creotutar: Agreeeeed. I updated with a comment!

alexisgrow (approved):
this lgtm!

tmeyer2115 (approved):

oshi97 (approved):

PR #1081, opened 2020-09-17
Add the Stylelint-Scss plugin.

tmeyer2115:
This PR adds the Stylelin-Scss plugin and registers it with Stylelint. By
default, Stylelint's rules pertain to standard CSS. It can parse SCSS to ensure
that these rules are being followed. But, it has no rules specific to SCSS
syntax. That's what the Stylelint-Scss plugin allows for. I've selected a few
SCSS rules provided by the plugin:

- dollar-variable-default: This enforces that all $-variable declrations
  include !default.
- at-import-no-partial-leading-underscore: This disallows leading underscores
  in partial names for all @impor statements.
- at-import-partial-extension: This enforces that all @import statements leave
  out the extension.

There was one violation of the dollar-variable-default which I fixed.

J=SLAP-131
TEST=auto

Made sure stylelint checks passed.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,10 +1,16 @@
 {
+    "plugins": [
+        "stylelint-scss"
+    ],
     "rules": {
         "color-no-invalid-hex": true,
         "function-calc-no-invalid": true,
         "unit-no-unknown": true,
         "property-no-unknown": true,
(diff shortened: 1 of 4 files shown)


alexisgrow (approved):
LGTM!

PR #1082, opened 2020-09-18
Add a resultsCountByVertical entry to the onUniversalSearch input.

tmeyer2115:
This PR adds a new entry to the Object passed to onUniversalSearch. This entry,
called resultsCountByVertical, is an Object with values of the form:

{ totalResultsCount: 150, displayedResultsCount: 10}

for each returned vertical. The totalResults count indicates how many results
were found for that vertical. The displayedResultsCount indicates how many
of those results are actually displayed. This additional info was requested
by Mercy. But, it seemed generally useful, so we're adding it to the SDK
officially.

J=SLAP-648
TEST=manual

On a local test site, logged the param passed to onUniversalSearch. I verified
that the resultsCountByVertical was correct by comparing it against the
LiveAPI response.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -15,7 +15,6 @@ import QueryTriggers from './models/querytriggers';
 import StorageKeys from './storage/storagekeys';
 import AnalyticsEvent from './analytics/analyticsevent';
 import FilterRegistry from './filters/filterregistry';
-import Section from './models/section';
 
 /** @typedef {import('./services/searchservice').default} SearchService */
 /** @typedef {import('./services/autocompleteservice').default} AutoCompleteService */
@@ -309,7 +308,7 @@ export default class Core {
         this.globalStorage.delete(StorageKeys.QUERY_TRIGGER);
 
         const exposedParams = this._getOnUniversalSearchParams(
-          data[StorageKeys.UNIVERSAL_RESULTS].sections, 
+          data[StorageKeys.UNIVERSAL_RESULTS].sections,
           queryString);
         const analyticsEvent = this.onUniversalSearch(exposedParams);
         if (typeof analyticsEvent === 'object') {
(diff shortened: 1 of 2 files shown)


src/core/core.js
```diff
@@ -307,9 +307,22 @@ export default class Core {
         this.globalStorage.delete('skipSpellCheck');
         this.globalStorage.delete(StorageKeys.QUERY_TRIGGER);
 
+        const universalSections = data[StorageKeys.UNIVERSAL_RESULTS].sections;
```
creotutar: Do we want the generation of this data in this function? Do we maybe want to split it out into some `getExposedParams`?

cea2aj:
The logic here looks good to me

src/core/core.js
tmeyer2115: Split this out into another function!

cea2aj (approved):

oshi97 (approved):
lgtm!

creotutar (approved):
🛰️ 🛰️ 🛰️ LGTM

src/core/core.js
```diff
@@ -307,17 +307,44 @@ export default class Core {
         this.globalStorage.delete('skipSpellCheck');
         this.globalStorage.delete(StorageKeys.QUERY_TRIGGER);
 
-        const exposedParams = {
-          queryString: queryString,
-          sectionsCount: data[StorageKeys.UNIVERSAL_RESULTS].sections.length
-        };
+        const exposedParams = this._getOnUniversalSearchParams(
+          data[StorageKeys.UNIVERSAL_RESULTS].sections,
+          queryString);
```
creotutar, line 312: Very unimportant, I think it's kinda funny in the SDK we do

```
function(
  param
  param);
```

but
```
let x = {
  field
  field
};
```

PR #1084, opened 2020-09-21
Update translation helper interfaces

cea2aj:
Update the translation helpers to no longer use stringified JSON

Update the TranslationProcessor to expect an object rather than stringified JSON for plural forms. Update the HBS helper to find pluralForm keys and and pass them to the TranslationProcessor in an object as the first parameter. Update fallback behavior surrounding TranslationProcessor by getting the locale if the option is specified and otherwise using the locale from the init. Get the language as first two characters of the locale. If it is supported, use that when selecting plural forms. Otherwise use 'en'.

J=SLAP-638
TEST=auto, manual

Update unit tests and confirm they pass. Run an internationalized jambo build and observe the translations working.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -514,8 +514,10 @@ class Answers {
   }
 
   /**
-   * Processes a translation which includes performing interpolation, pluralization, or both
-   * @param {string} translations The translations, or a stringified JSON of possible translations
+   * Processes a translation which includes performing interpolation, pluralization, or
+   * both
+   * @param {string | Object} translations The translation, or an object containing
+   * translated plural forms
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
    * @returns {string} The translation with any interpolation or pluralization applied
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -273,13 +273,13 @@ class Answers {
 
     if (parsedConfig.useTemplates === false || parsedConfig.templateBundle) {
       if (parsedConfig.templateBundle) {
-        this.renderer.init(parsedConfig.templateBundle, parsedConfig.locale);
+        this.renderer.init(parsedConfig.templateBundle, this._getInitLocale());
       }
     } else {
       // Templates are currently downloaded separately from the CORE and UI bundle.
       // Future enhancement is to ship the components with templates in a separate bundle.
       this.templates = new DefaultTemplatesLoader(templates => {
-        this.renderer.init(templates, parsedConfig.locale);
+        this.renderer.init(templates, this._getInitLocale());
       });
     }
 
@@ -524,10 +524,19 @@ class Answers {
    * @returns {string} The translation with any interpolation or pluralization applied
    */
   processTranslation (translations, interpolationParams, count, language) {
-    const initLocale = this.core.globalStorage.getState(StorageKeys.LOCALE);
+    const initLocale = this._getInitLocale();
     language = language || initLocale.substring(0, 2);
 
     return TranslationProcessor.process(translations, interpolationParams, count, language);
+  }
+
+  /**
+   * Gets the locale that ANSWERS was initialized to. Defaults to 'en'
+   *
+   * @returns {string}
+   */
+  _getInitLocale () {
+    return this.core.globalStorage.getState(StorageKeys.LOCALE) || 'en';
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -524,7 +524,7 @@ class Answers {
    * @returns {string} The translation with any interpolation or pluralization applied
    */
   processTranslation (translations, interpolationParams, count, language) {
-    const initLocale = this.core.globalStorage.get(StorageKeys.LOCALE);
+    const initLocale = this.core.globalStorage.getState(StorageKeys.LOCALE);
     language = language || initLocale.substring(0, 2);
 
     return TranslationProcessor.process(translations, interpolationParams, count, language);
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -273,13 +273,13 @@ class Answers {
 
     if (parsedConfig.useTemplates === false || parsedConfig.templateBundle) {
       if (parsedConfig.templateBundle) {
-        this.renderer.init(parsedConfig.templateBundle);
+        this.renderer.init(parsedConfig.templateBundle, parsedConfig.locale);
       }
     } else {
       // Templates are currently downloaded separately from the CORE and UI bundle.
       // Future enhancement is to ship the components with templates in a separate bundle.
       this.templates = new DefaultTemplatesLoader(templates => {
-        this.renderer.init(templates);
+        this.renderer.init(templates, parsedConfig.locale);
       });
     }
 
@@ -520,10 +520,14 @@ class Answers {
    * translated plural forms
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
+   * @param {string} language The langauge associated with the pluralization
    * @returns {string} The translation with any interpolation or pluralization applied
    */
-  processTranslation (translations, interpolationParams, count) {
-    return TranslationProcessor.process(translations, interpolationParams, count);
+  processTranslation (translations, interpolationParams, count, language) {
+    const initLocale = this.core.globalStorage.get(StorageKeys.LOCALE);
+    language = language || initLocale.substring(0, 2);
+
+    return TranslationProcessor.process(translations, interpolationParams, count, language);
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -531,12 +531,12 @@ class Answers {
   }
 
   /**
-   * Gets the locale that ANSWERS was initialized to. Defaults to 'en'
+   * Gets the locale that ANSWERS was initialized to
    *
    * @returns {string}
    */
   _getInitLocale () {
-    return this.core.globalStorage.getState(StorageKeys.LOCALE) || 'en';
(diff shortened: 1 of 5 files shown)


src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,24 @@ export default class HandlebarsRenderer extends Renderer {
...
+      Object.entries(options.hash).forEach(([key, value]) => {
+        if (key === 'locale') {
+          pluralizationInfo['locale'] = value;
+        } else if (key.length > 10 && key.substring(0, 10) === 'pluralForm') {
+          const pluralFormNum = parseInt(key.substring(10));
+          pluralizationInfo[pluralFormNum] = value;
+        } else {
+          interpolationParams[key] = value;
+        }
+      });
```
cea2aj: Should I extract this into a helper?

src/core/i18n/translationprocessor.js
```diff
@@ -2,46 +2,48 @@ import { getNPlurals, getPluralFunc, hasLang } from 'plural-forms/dist/minimal';
...
-   * If translations is json, parse it and choose the correct plural form.
-   * Otherwise just return the non-interpolated translation string.
-   * @param {string} translations
+   * If translations is a string, return it
+   * Otherwise select the correct plural form based on count
+   * @param {string | Object } translations
    * @param {number} count
    * @returns {string}
    */
   static _selectStringToInterpolate (translations, count) {
```
tmeyer2115: Small thing, but I'm not sure we need this `_selectStringToInterpolate` method. It's not doing much after removing the `try-catch`, it's almost a pass through to `_selectPluralForm`. What would you think of getting rid of it entirely?  Whenever I see methods that have logic that just returns the initial input, my reaction is that the function should just not be called in that case.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,24 @@ export default class HandlebarsRenderer extends Renderer {
     });
 
     this.registerHelper('processTranslation', function (options) {
+      const pluralizationInfo = {};
```
tmeyer2115, line 202: Can we choose another name here? It's not too clear what this contains.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,24 @@ export default class HandlebarsRenderer extends Renderer {
...
+      const interpolationParams = {};
       let { phrase, count } = options.hash;
-      return TranslationProcessor.process(phrase, options.hash, count);
+
+      Object.entries(options.hash).forEach(([key, value]) => {
+        if (key === 'locale') {
+          pluralizationInfo['locale'] = value;
+        } else if (key.length > 10 && key.substring(0, 10) === 'pluralForm') {
```
tmeyer2115: Is `pluralForm` a valid param name? If so, does the check need to be `key.length >= 10`? To make things a bit more readable, I'd suggest making a local variable called `isPluralFormParam` (or some such name). If we use that here, I think it makes the code a bit easier to follow. People could easily see that you treat pluralForm params and locale params specially. If they care, they can dig into how we determine it's a pluralForm param.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,24 @@ export default class HandlebarsRenderer extends Renderer {
...
+          const pluralFormNum = parseInt(key.substring(10));
```
tmeyer2115: Are there languages with more than 9 plural forms?

src/core/i18n/translationprocessor.js
cea2aj: Good point! I can take it out

src/ui/rendering/handlebarsrenderer.js
cea2aj: Sure. This object contains the various plural forms and a locale. What do you think about the name `pluralizations`?  I originally chose `pluralizationInfo` because `pluralizations` may optionally contain a locale, which might make `pluralizations` misleading.

src/ui/rendering/handlebarsrenderer.js
cea2aj: No, `pluralForm` is not a valid param name because if pluralization is being used, the key `pluralForm0` will represent the singular form, and the key `pluralForm1` will represent the pluralForm (for languages such as english). 

I'll add `isPluralFormParam` as you suggested!

src/ui/rendering/handlebarsrenderer.js
cea2aj: According to [this](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals), the most any language has is 6.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,24 @@ export default class HandlebarsRenderer extends Renderer {
...
+          pluralizationInfo[pluralFormNum] = value;
+        } else {
+          interpolationParams[key] = value;
+        }
+      });
+
+      return typeof phrase === 'string'
+        ? TranslationProcessor.process(phrase, interpolationParams, count)
```
tmeyer2115: Do we need `count` here?

src/ui/rendering/handlebarsrenderer.js
tmeyer2115: Gotcha. Would `pluralFormsAndLocale` be too literal, haha? Looking at this again, I'm not too bothered by it. It contains the all the info needed to perform pluralization at run-time. I'm fine keeping this or changing, up to you.

cea2aj: I think it's a little too literal, especially since locale is optional. I'll leave it as is.

src/ui/rendering/handlebarsrenderer.js
cea2aj: Nope. Removing.

src/ui/rendering/handlebarsrenderer.js
cea2aj: Maybe it's not necessary to extract to a helper because Tom's suggestions made this clearer

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -194,8 +194,27 @@ export default class HandlebarsRenderer extends Renderer {
     });
 
     this.registerHelper('processTranslation', function (options) {
+      const pluralizationInfo = {};
+      const interpolationParams = {};
       let { phrase, count } = options.hash;
-      return TranslationProcessor.process(phrase, options.hash, count);
+
+      Object.entries(options.hash).forEach(([key, value]) => {
+        const isPluralFormParam = key.length > 10 && key.substring(0, 10) === 'pluralForm';
```
oshi97, line 202: I think key.startsWith is slightly cleaner here

oshi97: I think key.startsWith is slightly cleaner here

oshi97 (approved):
lgtm!

src/ui/rendering/handlebarsrenderer.js
cea2aj: updated!

src/answers-umd.js
```diff
@@ -514,14 +514,29 @@ class Answers {
...
    * @param {Object} interpolationParams Params to use during interpolation
    * @param {number} count The count associated with the pluralization
+   * @param {string} language The langauge associated with the pluralization
    * @returns {string} The translation with any interpolation or pluralization applied
    */
-  processTranslation (translations, interpolationParams, count) {
-    return TranslationProcessor.process(translations, interpolationParams, count);
+  processTranslation (translations, interpolationParams, count, language) {
+    const initLocale = this._getInitLocale();
+    language = language || initLocale.substring(0, 2);
```
tmeyer2115, line 528: Are there plural forms specific to locales? Does the library we pulled in support that at all? I'm wondering if we want to be stripping the suffix from the locale here.

src/answers-umd.js
```diff
@@ -514,14 +514,29 @@ class Answers {
...
+    return TranslationProcessor.process(translations, interpolationParams, count, language);
+  }
+
+  /**
+   * Gets the locale that ANSWERS was initialized to. Defaults to 'en'
+   *
+   * @returns {string}
+   */
+  _getInitLocale () {
+    return this.core.globalStorage.getState(StorageKeys.LOCALE) || 'en';
```
tmeyer2115: I think we already have some logic in this class that defaults the locale to `en`. We should make sure we're not applying defaults in two places.

src/answers-umd.js
cea2aj: The library we pulled in only supports language codes. It supports some three-character language codes, but we don't use any of those and Rose agreed that using the two-character language code here was sufficient.

src/answers-umd.js
cea2aj: Ah, that's right. Removing this.

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -28,12 +28,15 @@ export default class HandlebarsRenderer extends Renderer {
     this._templates = templates || {};
   }
 
-  init (templates) {
+  init (templates, locale) {
```
tmeyer2115, line 31: Conceptually, I think it's odd that the `HandlebarsRenderer` needs `locale`. The two concepts seem fairly removed from each other. We only have it here because this class is also responsible for creating all the HBS helpers. I think we should make an item to refactor so that we separate that responsibility out from this class. Clean Code hints at this a bit when it talks about cohesion. In this case, `locale` has low cohesion, it's just used to create the one helper. That's a sign this class is doing too much. No need to fix in this PR though! This is definitely beyond the scope of your change.

tmeyer2115 (approved):

PR #1085, opened 2020-09-21
VerticalResultsCount: use VERTICAL_RESULTS storage key

oshi97:
VerticalResultsCount cannot use the StorageKeys.RESULTS_HEADER while a
VerticalResults component (that does not have hideResultsHeader: true)
is on the page, because of the bug that when an event listener is unregistered
(as happens with child components like the ResultsHeader inside VerticalResults)
it will unregister all other listeners for that storage key.

TEST=manual

Tested that I can have both a VerticalResults (with its own results header)
and a VerticalResultsCount on the same page, and both will update.

### DIFF
--- a/src/ui/components/results/verticalresultscountcomponent.js
+++ b/src/ui/components/results/verticalresultscountcomponent.js
@@ -1,10 +1,15 @@
 import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
+import SearchStates from '../../../core/storage/searchstates';
 
 export default class VerticalResultsCountComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(config, systemConfig);
-    this.moduleId = StorageKeys.RESULTS_HEADER;
+    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
+      if (results.searchState === SearchStates.SEARCH_COMPLETE) {
+        this.setState();
+      }
+    });
   }
 
   static areDuplicateNamesAllowed () {
--- a/src/ui/components/results/verticalresultscountcomponent.js
+++ b/src/ui/components/results/verticalresultscountcomponent.js
@@ -5,9 +5,9 @@ import SearchStates from '../../../core/storage/searchstates';
 export default class VerticalResultsCountComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(config, systemConfig);
-    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
-      if (results.searchState === SearchStates.SEARCH_COMPLETE) {
-        this.setState();
+    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, verticalResults => {
+      if (verticalResults.searchState === SearchStates.SEARCH_COMPLETE) {
+        this.setState(verticalResults);
       }
     });
   }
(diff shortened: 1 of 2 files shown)


src/ui/components/results/verticalresultscountcomponent.js
```diff
@@ -1,10 +1,15 @@
...
 import StorageKeys from '../../../core/storage/storagekeys';
+import SearchStates from '../../../core/storage/searchstates';
 
 export default class VerticalResultsCountComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(config, systemConfig);
-    this.moduleId = StorageKeys.RESULTS_HEADER;
+    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
+      if (results.searchState === SearchStates.SEARCH_COMPLETE) {
+        this.setState();
```
creotutar: Is this setState expecting data?

oshi97: no, no data

creotutar: This may be never a case currently, but it seems weird that we are listening for a `VERTICAL_RESULTS` update but not using it and instead re-fetching it in `setState`. It makes me feel like there is a case when it might change in between. Though I'm not sure that would be a problem? What do you think?

oshi97: javascript is single threaded so that wouldn't ever be an issue. this is also how the pagination component works

creotutar: I'm not sure -- what about

AnotherFile.js
```
this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
  this.core.globalStorage.setState(StorageKeys.VERTICAL_RESULTS, 'update it for some reason')
});
```

verticalresultscountcomponent.js
```
this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
  this.setState() // getState !== results
});
```

creotutar: I'm not saying that's a case that should ever happen* Just feels like something we can avoid ourselves from if we passed exactly what we're getting from update?

oshi97: Sure, I can update it if that puts you at ease. I don't see how example you gave is an issue, though, partially because it's an infinite loop but mostly because even if it was just a regular globalStorage setState I don't think there would be any issues?

oshi97: if AnotherComponent updated vertical results, wouldn't the order of execution just be

AnotherComponent updates vertical results -> this triggers all the event listeners, which run synchronously in the order they were registered -> whatever code is after that in AnotherComponent

oshi97: this makes things easier to unit test though which is kind of nice

creotutar: womp-- Sorry, infinite loop, you're right

oshi97: this makes things a little nicer to unit test, but it's a little weird that it relies on two pieces of information (search offset and vertical results) but only one gets to be the special "data" param

oshi97: I think you're right though this is more idiomatic of the SDK (even though I don't really like it)

creotutar (approved):
✨ ✨ ✨  LGTM

PR #1086, opened 2020-09-21
AutoComplete: add unique id to each li option

oshi97:
This commit adds a unique id to each autocomplete li option.
Because the autocomplete component has areDuplicateNamesAllowed
set to false (the default), _config.name should always be unique.
Html ids should not have spaces in them, which is not a restriction
set on component names, but we also use component instance names
in other html ids (see controls/date.hbs and controls/filteroptions.hbs)

J=SLAP-665
TEST=manual

on a universal and also vertical page, checked that each li option
in the searchbar autocomplete has an id that includes both name,
section index, and results index

### DIFF
--- a/src/ui/templates/search/autocomplete.hbs
+++ b/src/ui/templates/search/autocomplete.hbs
@@ -1,3 +1,4 @@
+{{log this}}
 <div class="yxt-AutoComplete-wrapper js-yxt-AutoComplete-wrapper" aria-expanded="{{hasResults}}">
   {{#if hasResults}}
     <div class="yxt-AutoComplete">
@@ -29,6 +30,7 @@
               data-eventtype="AUTO_COMPLETE_SELECTION"
               data-eventoptions='{"suggestedSearchText": "{{value}}"}'
               role="option"
+              id="yxt-AutoComplete-option-{{../../_config.name}}-{{@../index}}-{{@index}}"
           >
               {{highlightValue this true}}
           </li>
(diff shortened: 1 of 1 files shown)


cea2aj:
Would it be beneficial to add a `areDuplicateNamesAllowed` function to the autocomplete component so that this doesn't break if `areDuplicateNamesAllowed` is changed inside of Component?

oshi97:
> Would it be beneficial to add a `areDuplicateNamesAllowed` function to the autocomplete component so that this doesn't break if `areDuplicateNamesAllowed` is changed inside of Component?

I think if change Component's areDuplicateNamesAllowed default to true, we can go back and add this then, but for right now it's kind of just extra noise

src/ui/templates/search/autocomplete.hbs
```diff
@@ -29,6 +29,7 @@
               data-eventtype="AUTO_COMPLETE_SELECTION"
               data-eventoptions='{"suggestedSearchText": "{{value}}"}'
               role="option"
+              id="yxt-AutoComplete-option-{{../../_config.name}}-{{@../index}}-{{@index}}"
```
tmeyer2115, line 32: The `AutoComplete` component is not something people ever add directly. Do we provide a hook in the `SearchBar` for people to name the child `AutoComplete`? If no, I don't think there's really any way someone can set `_config.name` to have spaces in it.

If there is no mechanism to change the name, since `duplicateNamesAllowed` is false, then it shouldn't be possible to add more than one `AutoComplete`.

tmeyer2115 (approved):

src/ui/templates/search/autocomplete.hbs
oshi97: What the searchcomponent does is set the autocomplete's component name to `${this.name}.autocomplete`, so there is kind of a mechanism to change the name you just have to also change your searchbar name

oshi97: the filtersearch component does the same thing when it adds an autocomplete

PR #1087, opened 2020-09-21
acceptance tests: Replace unavailable filter on search

creotutar:
We previously tried to verify filter options by clicking on 'Brands'
with the option 'E' but the Brands filter is no longer returned in the
specific search we try. Replace with a filter option that exists (Puppy
Preference).

J=None
TEST=acceptance


oshi97 (approved):
lgtm! hope we can stop leeching off or roses account soon

PR #1088, opened 2020-09-22
performance: Add performance marks at major events (#1065)

creotutar:
We add performance marks for better metrics into the time between the
initialization of answers and the subsequent major events like the first
query request sent or the query response rendering. Cherry-picked from
`feature/sept-regression-fixes`

J=SLAP-632
TEST=manual

Test that on a page load you see all nine performance marks in the
console window.performance.entries() (page with query, depends
on whether you are searching on a vertical or a universal for the
four query marks)

yext.answers.initStart
yext.answers.statusStart
yext.answers.statusEnd
yext.answers.ponyfillStart
yext.answers.ponyfillend
yext.answers.verticalQueryStart
yext.answers.verticalQuerySent
yext.answers.verticalQueryResponseReceived
yext.answers.verticalQueryResponseRendered
yext.answers.universalQueryStart
yext.answers.universalQuerySent
yext.answers.universalQueryResponseReceived
yext.answers.universalQueryResponseRendered

### DIFF
--- a/tests/acceptance/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuite.js
@@ -180,3 +180,37 @@ test('Facets, pagination, and filters do not persist accross experience links', 
   await t.expect(viewAllLink).contains('referrerPageUrl');
   await verifyCleanLink(viewAllLink);
 });
+
+fixture`Performance marks on search`
+  .before(setupServer)
+  .after(shutdownServer)
+  .page`http://localhost:9999/tests/acceptance/fixtures/html/facets`;
+
+test('window.performance calls are marked for a normal search', async t => {
+  const marksToCheck = [
+    'yext.answers.initStart',
+    'yext.answers.statusStart',
+    'yext.answers.statusEnd',
+    'yext.answers.ponyfillStart',
+    'yext.answers.ponyfillEnd',
+    'yext.answers.verticalQueryStart',
+    'yext.answers.verticalQuerySent',
+    'yext.answers.verticalQueryResponseReceived',
+    'yext.answers.verticalQueryResponseRendered'
+  ];
+  const searchComponent = FacetsPage.getSearchComponent();
+  await searchComponent.submitQuery();
+
+  // Wait for Results to be rendered
+  const resultsSelector = await Selector('.yxt-Results');
+  await resultsSelector.with({ visibilityCheck: true })();
+
+  // All performance marks should be called at least once with a search
+  for (let i = 0; i < marksToCheck.length; i++) {
+    const markName = marksToCheck[i];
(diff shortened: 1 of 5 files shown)


tests/acceptance/acceptancesuite.js
```diff
@@ -180,3 +180,37 @@ test('Facets, pagination, and filters do not persist accross experience links',
   await t.expect(viewAllLink).contains('referrerPageUrl');
   await verifyCleanLink(viewAllLink);
 });
+
+fixture`Performance marks on search`
```
tmeyer2115, line 184: I'm not sure this really fits the criteria for an acceptance test (at least how we typically use them at Yext). We're not testing a particular user flow or interaction here (i.e. I can click on an autocomplete and be redirected to another page with the query auto-populated). We're verifying that performance marks are present and accounted for. I think it'd be more appropriate to have a Jest test suite for this. The suite would have a number of different tests ensuring the various marks are populated. I think these tests would be unit tests. Possibly integration tests. I totally recognize that it's difficult to write good unit tests for the SDK. That's a consequence of a few things, including it's architecture. So, I think it's fine to keep this here for now. We definitely want the coverage. We should make an item to create a unit test suite later on (maybe after the big core refactor). The issue with acceptance tests is that they're slow. They spin up a browser and have a decent amount of overhead. More so than unit tests. It can get to a point where the acceptance test suite is so slow that nobody wants to run it locally. That's what happened with Entity Edit.  So, we want to minimize the number of acceptance tests when possible.

tmeyer2115 (commented):
I think there is a Jira item for this: https://yexttest.atlassian.net/browse/SLAP-632. You should be able to use this for the "J=" in the commit.

tmeyer2115 (approved):

tests/acceptance/acceptancesuite.js
```diff
@@ -180,3 +180,37 @@ test('Facets, pagination, and filters do not persist accross experience links',
...
+  const resultsSelector = await Selector('.yxt-Results');
+  await resultsSelector.with({ visibilityCheck: true })();
+
+  // All performance marks should be called at least once with a search
+  for (let i = 0; i < marksToCheck.length; i++) {
+    const markName = marksToCheck[i];
+    const marksFoundWithName = await t.eval(() => {
+      return JSON.stringify(window.performance.getEntriesByName(markName));
+    }, { dependencies: { markName } });
+    await t.expect(JSON.parse(marksFoundWithName).length > 0).eql(true);
```
oshi97: nit: lets use t.expect.gt here

oshi97 (approved):

oshi97 (approved):

PR #1096, opened 2020-09-24
Update translator interface

cea2aj:
Update the interface of the translate helpers, and update the SDK build step to match it

Update the TranslationProcessor to expect an object rather than stringified JSON for plural forms. Update the HBS helper to find pluralForm keys and and pass them to the TranslationProcessor in an object as the first parameter. Update fallback behavior surrounding TranslationProcessor by getting the locale if the option is specified and otherwise using the locale from the init. Get the language as first two characters of the locale. If it is supported, use that when selecting plural forms. Otherwise use 'en'.

In the SDK, update the build step to output runtime processTranslation calls that match this new interface.

J=SLAP-638
TEST=auto, manual

Update unit tests and confirm they pass. Run a build and observe the properly formatted calls to the runtime translation helper methods. Run an internationalized jambo build and observe the runtime translations working.

### DIFF
--- a/conf/gulp-tasks/templates.gulpfile.js
+++ b/conf/gulp-tasks/templates.gulpfile.js
@@ -94,3 +94,26 @@ async function defaultTemplates () {
 }
 
 exports.default = defaultTemplates;
+
+/**
+ * A test build  task used for testing the precompile task
+ *
+ * @returns {Promise<Function>}
+ */
+async function test () {
+  const bundleTemplatesUMD =
+    new BundleTemplatesTaskFactory(DEV_LOCALE).create(TemplateType.UMD);
+  const cleanFiles = createCleanFilesTask(DEV_LOCALE);
+  const translator = await createTranslator(DEV_LOCALE);
+  const precompileTemplates = createPrecompileTemplatesTask(DEV_LOCALE, translator);
+
+  return new Promise(resolve => {
+    return series(
+      precompileTemplates,
+      bundleTemplatesUMD,
+      cleanFiles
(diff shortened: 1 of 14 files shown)


cea2aj:
The commit "Update TranslationProcessor interface" contains the changes which were already approved in #1084 . We closed the branch that those changes got merged into, so I put the changes here. This simplifies things since the changes to the interface in the SDK should correspond to the changes in the build step.

gulpfile.js
```diff
@@ -13,3 +13,5 @@ exports.dev = parallel(
   library.dev
 );
 exports.extractTranslations = extractTranslations;
+
+exports.test = templates.test;
```
oshi97: 🧪

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -94,3 +94,26 @@ async function defaultTemplates () {
...
+  return new Promise(resolve => {
+    return series(
+      precompileTemplates,
+      bundleTemplatesUMD,
+      cleanFiles
+    )(resolve);
+  });
+}
+
+exports.test = test;
```
oshi97: 🧪

oshi97 (commented):
still has the test command but lgtm besides that!

gulpfile.js
cea2aj: I thought we might want to leave this in here because it's useful for testing. Thoughts?

cea2aj: Removing

oshi97 (approved):

PR #1097, opened 2020-09-24
Wrap additional static text in translate helpers

oshi97:
Upon merging feature/localized-assets with develop,
we added a few static strings that were not wrapped in the 
hbs translate helper (due to it not existing yet).

TEST=manual

Test that on a dev page both autocomplete and verticalresultscount
still have the correct screen reader text and regular text


cea2aj:
Looks good! Just need to fix the unit tests

cea2aj (approved):
Looks good!

PR #1100, opened 2020-09-24
Add translation PO files

cea2aj:
Add PO files for de, it, es, fr, and ja

These are the initial PO files. They will need to be updated as some translations are missing

J=None
TEST=manual

Run a build of an internationalized answers site and observe several of the translations in use across the many languages


oshi97 (commented):
could we also add the VerticalResultsCount translation in for the 5 languages? it should be identical to the yxt-ResultsHeader one except for the line breaks, gettext comment with the line number, and css classes

cea2aj:
> could we also add the VerticalResultsCount translation in for the 5 languages? it should be identical to the yxt-ResultsHeader one except for the line breaks, gettext comment with the line number, and css classes

Oh, that's true. We don't have the yext-ResultsHeader for German, but I can add it to the other languages

conf/i18n/translations/ja.po
```diff
@@ -0,0 +1,428 @@
...
+
+#: src/ui/templates/results/verticalresultscount.hbs:5
+msgctxt "Example: 1-10 of 50"
+msgid ""
+"<span class=\"yxt-VerticalResultsCount-start\">[[start]]</span>\n"
+"<span class=\"yxt-VerticalResultsCount-dash\">-</span>\n"
+"<span class=\"yxt-VerticalResultsCount-end\">[[end]]</span>\n"
+"<span class=\"yxt-VerticalResultsCount-of\">/</span>\n"
+"<span class=\"yxt-VerticalResultsCount-total\">[[resultsCount]]</span>"
+msgstr ""
```
oshi97, line 40: we missed this one!

cea2aj: fixing!

oshi97 (approved):
this lgtm!

PR #1103, opened 2020-09-25
Escape parentheses before translator key lookup

cea2aj:
Escape parentheses before passing the key regex to `_findLocaleWithTranslationKey` inside the translator.

There was a bug where the following translation was not being translated during the SDK build step:
`msgid "([[resultsCount]] result)"
msgid_plural "([[resultsCount]] results)"
msgstr[0] "[[resultsCount]] resultado"
msgstr[1] "[[resultsCount]] resultados"`

The translation was failing due to the open and closing parentheses being interpreted as a regex capture group during the key lookup. This was because we manually check translation keys inside i18next with a regex test for plural translations. Escape parentheses before key lookup to solve this issue.

All the translations needed for 1.6 are valid, however I will put an item in the backlog to investigate if other regex characters might cause problems and if we need to change our key-lookup approach for plural translations.

J=none
Test=manual

Run an internationalized jambo build and observe the proper translation inside of es-answers.js. View the working translation on the local test page.

### DIFF
--- a/conf/i18n/translator.js
+++ b/conf/i18n/translator.js
@@ -55,7 +55,7 @@ class Translator {
    *   https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
    */
   translatePlural (phrase, pluralForm) {
-    const escapedPhrase = this._escapeInterpolationBrackets(phrase);
+    const escapedPhrase = this._escapeBracketsAndParentheses(phrase);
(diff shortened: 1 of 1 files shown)


tmeyer2115 (approved):

oshi97 (approved):

cea2aj:
Closing this because lodash.escapeRegExp is a better solution

PR #1105, opened 2020-09-28
translationprocessor: ie11 fix for regexp constructor

oshi97:
IE11 does not allow you to use a regex literal as the first
argument to the RegExp constructor. I tried targeting ie11
in bundle.js's legacyBundle babel config, but that didn't
seem to fix the issue (didn't spend much time investigating
our babel stuff).

J=SLAP-709
TEST=manual

Test that I can load a page that with components that use
translations (e.g. VerticalResutlsCount) again

### DIFF
--- a/src/core/i18n/translationprocessor.js
+++ b/src/core/i18n/translationprocessor.js
@@ -45,7 +45,7 @@ export default class TranslationProcessor {
   }
 
   static _interpolate (stringToInterpolate, interpolationParams) {
-    const interpolationRegex = new RegExp(/\[\[([a-zA-Z0-9]+)\]\]/, 'g');
+    const interpolationRegex = /\[\[([a-zA-Z0-9]+)\]\]/g;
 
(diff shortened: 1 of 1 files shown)


src/core/i18n/translationprocessor.js
```diff
@@ -45,7 +45,7 @@ export default class TranslationProcessor {
   }
 
   static _interpolate (stringToInterpolate, interpolationParams) {
-    const interpolationRegex = new RegExp(/\[\[([a-zA-Z0-9]+)\]\]/, 'g');
```
tmeyer2115, line 48: Is this our only use of the `RegExp` constructor in the SDK? If not, we may need to update those other instances as well.

oshi97: it's the only usage in the actual source code (there's a few usages in conf but ie11 doesn't care about that)

tmeyer2115 (approved):

PR #1106, opened 2020-09-28
Build SDK assets for KG locales

cea2aj:
Build SDK assets for locales in the KG

Add build-languages and build-locales commands to package.json scripts. The command build-languages creates a build for each language defined in `LANGUAGES_TO_LOCALES` in conf/i18n/constants.js. The command build-locales creates SDK build files for each locale defined in `LANGUAGES_TO_LOCALES`. It does this by creating symlinks to the language builds files for all defined locales. For example, if ALL_LOCALES is defined like this:

`exports.ALL_LOCALES = {
  'en': [
    'en_US'
  ],
  'fr': [
    'fr_CA',
    'fr_FR'
  ]
};`

Localized builds for answers.js will be created with the names en_US-answers.js, fr-answers.js, fr_CA-answers.js, and fr_FR-answers.js. The files answers.js and fr-answers.js will be regular files and the rest will be symlinks to their corresponding language build.

J=SLAP-704
TEST=manual

Run `npm run build-languages` and `npm run build-locales` and confirm the builds are correct. Build an internationalized jambo site and use these new SDK builds. Create a jambo build with the locale fr_CA and confirm that the site works.

### DIFF
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -5,8 +5,8 @@ const sass = require('gulp-sass');
 
 const getLibraryVersion = require('./utils/libversion');
 const { BundleType, BundleTaskFactory } = require('./bundle/bundletaskfactory');
-const { DEV_LOCALE, DEFAULT_LOCALES, ALL_LANGUAGES } = require('../i18n/constants');
-const { createLocaleFilesPerLanguage } = require('../i18n/localebuildutils');
+const { DEFAULT_LOCALES, ALL_LANGUAGES } = require('../i18n/constants');
+const { copyAssetsForLocales } = require('../i18n/localebuildutils');
 const LocalFileParser = require('../i18n/localfileparser');
 const MinifyTaskFactory = require('./bundle/minifytaskfactory');
 const TranslationResolver = require('../i18n/translationresolver');
@@ -18,7 +18,7 @@ const { generateProcessTranslationJsCall } = require('../i18n/runtimecallgenerat
  * @returns {Promise<Function>}
  */
 exports.dev = function devJSBundle () {
-  return createBundleTaskFactory(DEV_LOCALE).then(devTaskFactory => {
+  return createBundleTaskFactory(DEFAULT_LOCALES[0]).then(devTaskFactory => {
     return new Promise(resolve => {
       return parallel(
         series(devTaskFactory.create(BundleType.LEGACY_IIFE), getWatchJSTask(devTaskFactory)),
@@ -29,14 +29,16 @@ exports.dev = function devJSBundle () {
 };
 
 /**
- * Creates a build task for each provided locale and combines them into a series task.
- * @param {Array<string>} locales
+ * Creates a build task for each provided language and combines them into a series task.
+ * This function also supports locales, but it is named to reflect the current use case
+ * of creating bundles for just languages.
+ * @param {Array<string>} languages
  * @returns {Promise<Function>}
  */
-function createJSBundlesForLocales (locales) {
-  const localizedTaskPromises = locales.map(locale => {
-    const minifyTaskFactory = new MinifyTaskFactory(locale);
-    return createBundleTaskFactory(locale)
+function createJSBundlesForLanguages (languages) {
+  const localizedTaskPromises = languages.map(language => {
+    const minifyTaskFactory = new MinifyTaskFactory(language);
+    return createBundleTaskFactory(language)
       .then(bundleTaskFactory => createBundles(bundleTaskFactory, minifyTaskFactory));
   });
   return Promise.all(localizedTaskPromises).then(localizedTasks => {
@@ -45,15 +47,15 @@ function createJSBundlesForLocales (locales) {
 }
 
 exports.default = function defaultJSBundles () {
-  return createJSBundlesForLocales(DEFAULT_LOCALES);
+  return createJSBundlesForLanguages(DEFAULT_LOCALES);
 };
 
 exports.buildLanguages = function allLanguageJSBundles () {
-  return createJSBundlesForLocales(ALL_LANGUAGES);
+  return createJSBundlesForLanguages(ALL_LANGUAGES);
 };
 
 exports.buildLocales = function allLocaleJSBundles () {
-  const baseFileNames = [
+  const assetNames = [
     'answers.js',
     'answers.min.js',
     'answers-modern.js',
@@ -61,8 +63,8 @@ exports.buildLocales = function allLocaleJSBundles () {
     'answers-umd.js',
     'answers-umd.min.js'];
 
-  return createJSBundlesForLocales(ALL_LANGUAGES).then(() => {
-    createLocaleFilesPerLanguage(baseFileNames);
+  return createJSBundlesForLanguages(ALL_LANGUAGES).then(() => {
+    copyAssetsForLocales(assetNames);
   });
 };
 
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -5,7 +5,8 @@ const sass = require('gulp-sass');
 
 const getLibraryVersion = require('./utils/libversion');
 const { BundleType, BundleTaskFactory } = require('./bundle/bundletaskfactory');
-const { DEV_LOCALE, BUILD_LOCALES } = require('../i18n/constants');
+const { DEV_LOCALE, DEFAULT_LOCALES, ALL_LANGUAGES } = require('../i18n/constants');
+const { createLocaleFilesPerLanguage } = require('../i18n/localebuildutils');
 const LocalFileParser = require('../i18n/localfileparser');
 const MinifyTaskFactory = require('./bundle/minifytaskfactory');
 const TranslationResolver = require('../i18n/translationresolver');
@@ -28,17 +29,40 @@ exports.dev = function devJSBundle () {
 };
 
 /**
- * Creates a build task per locale, and combines them into a series task.
+ * Creates a build task for each provided locale and combines them into a series task.
+ * @param {Array<string>} locales
  * @returns {Promise<Function>}
  */
-exports.default = function defaultJSBundle () {
-  const localizedTaskPromises = BUILD_LOCALES.map(locale => {
+function createJSBundlesForLocales (locales) {
+  const localizedTaskPromises = locales.map(locale => {
     const minifyTaskFactory = new MinifyTaskFactory(locale);
     return createBundleTaskFactory(locale)
       .then(bundleTaskFactory => createBundles(bundleTaskFactory, minifyTaskFactory));
   });
   return Promise.all(localizedTaskPromises).then(localizedTasks => {
     return new Promise(resolve => series(compileCSS, ...localizedTasks)(resolve));
+  });
+}
+
+exports.default = function defaultJSBundles () {
+  return createJSBundlesForLocales(DEFAULT_LOCALES);
+};
+
+exports.buildLanguages = function allLanguageJSBundles () {
+  return createJSBundlesForLocales(ALL_LANGUAGES);
+};
+
+exports.buildLocales = function allLocaleJSBundles () {
+  const baseFileNames = [
+    'answers.js',
+    'answers.min.js',
+    'answers-modern.js',
+    'answers-modern.min.js',
+    'answers-umd.js',
+    'answers-umd.min.js'];
+
+  return createJSBundlesForLocales(ALL_LANGUAGES).then(() => {
+    createLocaleFilesPerLanguage(baseFileNames);
   });
 };
 
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -5,7 +5,7 @@ const sass = require('gulp-sass');
 
 const getLibraryVersion = require('./utils/libversion');
 const { BundleType, BundleTaskFactory } = require('./bundle/bundletaskfactory');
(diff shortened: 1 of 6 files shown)


conf/i18n/localebuildutils.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+function createLocaleFilesPerLanguage (baseFileNames) {
+  Object.keys(ALL_LOCALES).forEach((language) => {
+    baseFileNames.forEach((baseFileName) => {
+      const languageBundleName = language !== 'en'
+        ? `${language}-${baseFileName}`
+        : `${baseFileName}`;
+      ALL_LOCALES[language].forEach((locale) => {
+        const localeBundleName = `${locale}-${baseFileName}`;
+        src(`./dist/${languageBundleName}`)
```
oshi97: would a simple fs.copyFileSync suffice here? using a stream seems overkill for a single file. I still really like the symlink idea you brought up, even if it doesn't work in s3, just on the off chance of running out of circle ci memory or something (1 GB is a lot!)

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -28,18 +29,41 @@ exports.dev = function devJSBundle () {
...
 
 /**
- * Creates a build task per locale, and combines them into a series task.
+ * Creates a build task for each provided locale and combines them into a series task.
+ * @param {Array<string>} locales
  * @returns {Promise<Function>}
  */
-exports.default = function defaultJSBundle () {
-  const localizedTaskPromises = BUILD_LOCALES.map(locale => {
+function createJSBundlesForLocales (locales) {
```
tmeyer2115: Are we creating JS bundles per locale here or per language? I was thinking we would only "generate" or "build" assets per language and make copies of that per-locale. If that is the case, we should be careful in our naming here so as not to cause confusion.

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -81,16 +82,32 @@ function createDefaultTask (locale, translator) {
 
 /**
  * Creates a build task per locale, and combines them into a series task.
- *
+ * @param {Array<string>} locales
  * @returns {Promise<Function>}
  */
-async function defaultTemplates () {
-  const localizedTaskPromises = BUILD_LOCALES.map(async locale => {
+async function createTemplatesForLocales (locales) {
```
tmeyer2115: Same question here about if these are locales or languages

conf/i18n/constants.js
```diff
@@ -1,5 +1,110 @@
+exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s)*\{[^;]+?\}(\s)*\)/g;
+
 exports.DEV_LOCALE = 'en';
 
-exports.BUILD_LOCALES = ['en', 'es', 'fr', 'de', 'it', 'ja'];
+exports.DEFAULT_LOCALES = ['en'];
```
tmeyer2115: What's the difference here between `DEFAULT_LOCALES` and `DEV_LOCALE`?

conf/i18n/localebuildutils.js
```diff
@@ -0,0 +1,32 @@
...
+ * ALL_LOCALES.
+ *
+ * For example, if baseFileNames includes 'answers.js' and ALL_LOCALES includes 'fr_CA'
+ * and 'fr_FR' for the language 'fr', the file 'fr-answers.js' will be copied to
+ * 'fr_CA-answers.js' and 'fr_FR-answers.js'. Builds for each language must be created
+ * before this function is run.
+ *
+ * @param {Array<string>} baseFileNames File names used for iteration and file generation
+ */
+function createLocaleFilesPerLanguage (baseFileNames) {
```
tmeyer2115: This function name sounds a bit vague to me. It sounds a bit like we're creating a file of locales per language. Could we update it to be a bit more descriptive? Maybe something like `copyAssetsForLocales`?

conf/i18n/localebuildutils.js
```diff
@@ -0,0 +1,32 @@
...
+  Object.keys(ALL_LOCALES).forEach((language) => {
```
tmeyer2115: Can we do `ALL_LANGUAGES.forEach` here?

conf/i18n/constants.js
```diff
@@ -1,5 +1,110 @@
...
+    'de_DE',
+    'de_EU',
+    'de_LU'
+  ],
+  'ja': [
+    'ja_JP'
+  ]
+};
+
+exports.ALL_LANGUAGES = Object.keys(this.ALL_LOCALES);
```
oshi97: I kind of like having the bespoke ALL_LANGUAGES constant, but at the same time I can't imagine this ever being something other than Object.keys(this.ALL_LOCALES). It feels a little weird having a constant that is just a property of another one, and as a corollary to that having two different ways to specify the same information. Do you think it might make sense to remove ALL_LANGUAGES?

conf/i18n/constants.js
```diff
@@ -1,5 +1,110 @@
+exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s)*\{[^;]+?\}(\s)*\)/g;
+
 exports.DEV_LOCALE = 'en';
 
-exports.BUILD_LOCALES = ['en', 'es', 'fr', 'de', 'it', 'ja'];
+exports.DEFAULT_LOCALES = ['en'];
 
-exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s)*\{[^;]+?\}(\s)*\)/g;
+exports.ALL_LOCALES = {
```
oshi97: I feel like calling this ALL_LOCALES is a little deceiving, since it's a map not an array. Would it be possible to rename this to something that feels slightly more map-like? My brain kept doing a double take because ALL_LANGUAGES is an array but ALL_LOCALES is a map.

conf/gulp-tasks/library.gulpfile.js
cea2aj: The bundles created are simply per language. I used "Locales" in the name to indicate that this function is capable of building bundles for locales. But I see why it can be unclear, so I'll rename it!

conf/gulp-tasks/templates.gulpfile.js
cea2aj: Languages, so i'll rename

conf/i18n/constants.js
cea2aj: `DEV_LOCALE` is the locale used with `npm run dev`. `DEFAULT_LOCALES` are the locales used with `npm run build`. I left `DEFAULT_LOCALES` as an array to allow us to easily create builds for multiple languages. I renamed it from `BUILD_LOCALES` in an attempt to not cause confusion with `ALL_LOCALES` which can also be built. Do you think I should change it back? Or maybe consolidate it with `DEV_LOCALE` ?

conf/i18n/localebuildutils.js
cea2aj: Ya, that works

conf/i18n/localebuildutils.js
cea2aj: Yep! I missed this when I refactored `Object.keys(ALL_LOCALES)` to `ALL_LANGUAGES`

conf/i18n/localebuildutils.js
cea2aj: Ya, I can switch this over to `fs`. I'll test out the symlink and see if it works in CircleCI. Even if it doesn't work in S3, that should really help the circle containers.

conf/i18n/constants.js
cea2aj: Ya, it's a little odd, but I think that using the variable `ALL_LANGUAGES` in code is much clearer than `Object.keys(this.ALL_LOCALES)`

conf/i18n/constants.js
cea2aj: How about `LANGUAGE_LOCALE_MAP`?

conf/i18n/constants.js
tmeyer2115: I think it's confusing to have both `DEFAULT_LOCALES` and `DEV_LOCALE`. Given the two new commands you added (`build locales` and `build languages`), it seems as if `build` will only ever run for a single language. If we eventually want `build` to work across multiple languages/locales, we can revisit. For now, I'd suggest consolidating if possible.

conf/i18n/constants.js
oshi97: does something like LANGUAGES_TO_LOCALES work? or LANGAUGES_TO_ALL_LOCALES?

cea2aj: Lets do `LANGUAGES_TO_LOCALES`

conf/i18n/constants.js
cea2aj: I'll consolidate into `DEFAULT_LOCALES`. For the dev builds we just take the first locale from that array.

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -17,7 +18,7 @@ const { generateProcessTranslationJsCall } = require('../i18n/runtimecallgenerat
  * @returns {Promise<Function>}
  */
 exports.dev = function devJSBundle () {
-  return createBundleTaskFactory(DEV_LOCALE).then(devTaskFactory => {
+  return createBundleTaskFactory(DEFAULT_LOCALES[0]).then(devTaskFactory => {
```
tmeyer2115: I think we should make it so that `DEFAULT_LOCALES` isn't an array. Right now, we just have plans for `build` to run against one locale. It's a bit onerous to have to add `[0]` whenever you want to get the default locale. If we decide later on that we want `build` to generate assets for a couple of locales, we can revisit making this an array. For now, I feel it adds unnecessary complexity for this to be an array.

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -28,18 +29,43 @@ exports.dev = function devJSBundle () {
...
+    const minifyTaskFactory = new MinifyTaskFactory(language);
+    return createBundleTaskFactory(language)
       .then(bundleTaskFactory => createBundles(bundleTaskFactory, minifyTaskFactory));
   });
   return Promise.all(localizedTaskPromises).then(localizedTasks => {
     return new Promise(resolve => series(compileCSS, ...localizedTasks)(resolve));
   });
+}
+
+exports.default = function defaultJSBundles () {
```
tmeyer2115: Would it make sense to call this `defaultLanguageJsBundles`? I'm trying to think of how we can more easily differentiate this from `devJsBundle`

conf/gulp-tasks/library.gulpfile.js
cea2aj: Sure. I can make it singular.

conf/gulp-tasks/library.gulpfile.js
cea2aj: Ya, that works

oshi97 (commented):
+1 from me!

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -46,8 +46,8 @@ function createJSBundlesForLanguages (languages) {
   });
 }
 
-exports.default = function defaultJSBundles () {
-  return createJSBundlesForLanguages(DEFAULT_LOCALES);
+exports.default = function defaultLanguageJSBundle () {
```
tmeyer2115: Minor nit, but I think we want the suffix to be `JSBundles` (as more than one is created).

conf/i18n/localebuildutils.js
```diff
@@ -21,9 +20,8 @@ function copyAssetsForLocales (assetNames) {
         : `${assetName}`;
       LANGUAGES_TO_LOCALES[language].forEach((locale) => {
         const localeBundleName = `${locale}-${assetName}`;
-        src(`./dist/${languageBundleName}`)
-          .pipe(rename(localeBundleName))
-          .pipe(dest('./dist/'));
+        fs.symlinkSync(`./${languageBundleName}`,
+          `./dist/${localeBundleName}`);
```
tmeyer2115: Can we put this on the line above?

conf/i18n/localebuildutils.js
```diff
@@ -21,9 +20,8 @@ function copyAssetsForLocales (assetNames) {
         : `${assetName}`;
       LANGUAGES_TO_LOCALES[language].forEach((locale) => {
         const localeBundleName = `${locale}-${assetName}`;
-        src(`./dist/${languageBundleName}`)
-          .pipe(rename(localeBundleName))
-          .pipe(dest('./dist/'));
+        fs.symlinkSync(`./${languageBundleName}`,
```
tmeyer2115: Out of curiosity, do you have a sense for how much space the `symlink` will save in CircleCI? Does this need to be `./dist/${languageBundleName}`?

conf/gulp-tasks/library.gulpfile.js
cea2aj: Ah, I was thinking singular because this now single-language. Updating

conf/i18n/localebuildutils.js
cea2aj: done!

conf/i18n/localebuildutils.js
cea2aj: Just checked, and it saves about 1GB. Also, the languageBundleName path doesn't need to include dist. The first param in `symLinkSync` is relative to the second params's parent directory according to the [docs](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback). I also confirmed that all the locale builds are available in S3 from the right directory

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #1107, opened 2020-09-29
Update .circleci/config.yml for i18n builds

cea2aj:
Update .circleci/config.yml for i18n builds

We only want to build all locales for the master branch, releases branches matching `/^release\/*/`, tagged releases, support branches matching `/^support\/*/` and feature branches matching `/^feature\/*-i18n/`. For canary (develop) and all other branches, just run the regular npm build which will only build and test the SDK assets for English. This is much quicker (2mins vs 11min), and it saves us a lot of space on AWS S3.

J=SLAP-738
TEST=manual

Create a test support branch and a feature/-TEST-TEST-i18n branch to trigger i18n builds. Observe circle and confirm that builds in S3 are created for all locales. Confirm that this branch triggers a regular build in CircleCI. Run a localized jambo build and point it to the i18n assets in S3. Load the locale fr_CA for the answers site and observe it using the fr_CA build files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ workflows:
   build_and_deploy:
     jobs:
       - build:
-        filters:
+          filters:
             branches:
               ignore:
                 - master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,47 +133,10 @@ workflows:
   version: 2
   build_and_deploy:
     jobs:
-      - build:
-          filters:
-            branches:
-              ignore:
-                - master
-                - /^support.*/
-                - /^feature\/.*-i18n/
-                - /^release.*/
-      - build_i18n:
-          filters:
-            tags:
-              only: /^v.*/
-            branches:
-              only:
-                - master
-                - /^support.*/
-                - /^feature\/.*-i18n/
-                - /^release.*/
+      - build
       - unit_test:
-          filters:
-            branches:
-              ignore:
-                - master
-                - /^support.*/
-                - /^feature\/.*-i18n/
-                - /^release.*/
           requires:
             - build
-      - unit_test:
-          name: unit_test_i18n
-          filters:
-            tags:
-              only: /^v.*/
-            branches:
-              only:
-                - master
-                - /^support.*/
-                - /^feature\/.*-i18n/
-                - /^release.*/
-          requires:
-            - build_i18n
       - deploy_canary:
           filters:
             branches:
@@ -191,6 +154,19 @@ workflows:
                 - /^release.*/
           requires:
             - unit_test
+  build_and_deploy_i18n:
+    jobs:
+      - build_i18n:
+          filters:
+            tags:
+              only: /^v.*/
+      - unit_test:
+          name: unit_test_i18n
+          filters:
+            tags:
+              only: /^v.*/
+          requires:
+            - build_i18n
       - deploy_branch:
           name: deploy_branch_i18n
           filters:
@@ -225,4 +201,4 @@ workflows:
               ignore: /.*/
           requires:
             - hold
-      
+
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,9 +138,9 @@ workflows:
             branches:
               ignore:
                 - master
-                - /^support.*/
+                - /^support\/.*/
                 - /^feature\/.*-i18n/
-                - /^release.*/
+                - /^release\/.*/
       - unit_test:
           requires:
             - build
@@ -154,14 +154,25 @@ workflows:
           filters:
             branches:
               ignore:
-                - master
                 - develop
-                - /^support.*/
-                - /^feature\/.*-i18n/
-                - /^release.*/
           requires:
             - unit_test
   build_and_deploy_i18n:
+    jobs:
+      - build_i18n:
+          filters:
+            branches:
+              only:
+                - /^support\/.*/
+                - /^feature\/.*-i18n/
+                - /^release\/.*/
+      - unit_test:
+          requires:
+            - build_i18n
+      - deploy_branch:
+          requires:
+            - unit_test
+  build_and_deploy_hold:
     jobs:
       - build_i18n:
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -87,6 +100,28 @@ jobs:
           command: echo 'export FORMATTED_CIRCLE_BRANCH=$(echo ${CIRCLE_BRANCH} | sed "s/\//-/g")' >> $BASH_ENV
       - deploy-to-aws:
           subdirectory: '/dev/${FORMATTED_CIRCLE_BRANCH}'
+  # deploys support branch assets to an uncached folder in the S3 bucket named by branch
+  deploy_support:
```
oshi97: what is different about deploy_support and deploy_feature_i18n compared to deploy_branch?

.circleci/config.yml
```diff
@@ -124,6 +159,15 @@ workflows:
           filters:
             tags:
               only: /^v.*/
+      - build_all_locales:
+          filters:
+            tags:
+              only: /^v.*/
```
oshi97: I think this will run build and then build_all_locales for all tagged builds, we can probably just remove tags from the regular build, and filter out branches that would run build_all_locales

.circleci/config.yml
cea2aj: Oh, I can consolidate these

cea2aj: Actually I can consolidate deploy_support and deploy_feature_i18n, but their difference to deploy_branch is that deploy_branch doesn't depend on build_all_locales

.circleci/config.yml
cea2aj: Done

.circleci/config.yml
```diff
@@ -121,50 +134,95 @@ workflows:
   build_and_deploy:
     jobs:
       - build:
+          filters:
+            branches:
+              ignore:
+                - master
```
creotutar, line 140: We are going from building all branches to building all branches except for `master`, `support*`, `feature/*-i18n`, and `release*` branches. What is the reason here? Don't we still want to show `dev/<branch>` assets?

creotutar: I see, it would be an either-or situation with build and build_i18n

creotutar: This looks like it happens a lot. You don't want to have two different workflows?

cea2aj: Yep! The build_i18n takes much longer and takes up more space in S3, so we decided to run the full build only for specific branches and tags.

.circleci/config.yml
```diff
@@ -121,50 +134,95 @@ workflows:
...
+                - /^support.*/
+                - /^feature\/.*-i18n/
+                - /^release.*/
       - unit_test:
           filters:
-            tags:
-              only: /^v.*/
+            branches:
+              ignore:
```
creotutar: The require of `build` should propagate the filters (that is, you don't need to re-include them here as it will not go through the required `build` job unless it meets the filter). Is there another reason we need to keep this?

.circleci/config.yml
cea2aj: I think putting the i18n builds into a  build_and_deploy_i18n workflow would work well

.circleci/config.yml
cea2aj: The [docs](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag) made it seem like we need to include the filter at all levels. I'll do a test without them and see if it works. Are you suggesting that they propagate from build to unit_test?

creotutar: Ah I see, so this is the case of `tags` default being "don't run any tags", so it must be included on every level. However, the default for `branches` is all branches, and since only branches that meet the prior required step (`build`) pass through, using all branches works (Like how previously in our config, we did not copy the filters from the previous steps)

cea2aj: Ah, I missed that distinction between how the `tags` default is to not run and how the `branches` default is to run for everything. So if I'm understanding correctly, I can remove the branch filters from the build and unit test jobs and only have the branch filters for the deploy jobs.

cea2aj: Update: I do need the branch filters for the build steps but not the unit test step

.circleci/config.yml
```diff
@@ -121,15 +134,62 @@ workflows:
...
             tags:
               only: /^v.*/
+            branches:
+              only:
+                - master
+                - /^support.*/
+                - /^feature\/.*-i18n/
+                - /^release.*/
       - unit_test:
+          name: unit_test_i18n
```
oshi97: Is there some way we could simplify the filtering? I've always found it mentally annoying to piece together exactly what is happening, and we've definitely increased that complexity here. I think moving the unit test job into the build and build_i18n step could help a little (and I guess we would rename those jobs slightly)

oshi97: actually eh it's probably fine I think the green diff on github makes it seem more confusing than it is

oshi97: it looks fine when I just look at the raw file

.circleci/config.yml
```diff
@@ -121,15 +134,62 @@ workflows:
...
     jobs:
       - build:
+          filters:
+            branches:
+              ignore:
+                - master
+                - /^support.*/
+                - /^feature\/.*-i18n/
+                - /^release.*/
+      - unit_test:
```
oshi97, line 144: for my edification why does the unit step not need filters, but the deploy steps do?

oshi97: talked on hangouts about this!

cea2aj: The filters on the build step determine whether or not the workflow (build_and_deploy or build_and_deploy_i18n) will run. The unit test steps depends on the build steps, so only branches that pass the filters on the build steps will be run for the corresponding unit test steps. The deploy steps need filters because the deploy process is different for their associated branches. The deploy step filters are subsets of their corresponding build step filters

.circleci/config.yml
```diff
@@ -121,15 +134,62 @@ workflows:
...
+            branches:
+              ignore:
+                - master
+                - develop
+                - /^support.*/
+                - /^feature\/.*-i18n/
+                - /^release.*/
+          requires:
+            - unit_test
+  build_and_deploy_i18n:
```
oshi97, line 160: would it make sense to split this workflow into one for release tags + master, and one for support, feature-i18n, and release?

.circleci/config.yml
```diff
@@ -121,15 +134,62 @@ workflows:
...
+    jobs:
+      - build_i18n:
           filters:
             tags:
               only: /^v.*/
+            branches:
+              only:
+                - master
+                - /^support.*/
```
oshi97: nit but could we make it match `support\/.*` instead of just `support.*`? Just to not match a
branch like `supportive-boy` and only match things like `support/blah-blah`. same thing with release

.circleci/config.yml
cea2aj: Ya, we could do that

oshi97: that might simplify some of the filtering logic

.circleci/config.yml
cea2aj: sure!

oshi97 (approved):
this lgtm! I feel like the config is easier to read now, despite adding more functionality.

PR #1108, opened 2020-09-29
VerticalResults: revert hideResultsHeader config

oshi97:
This can already be specified with
showAppliedFilters: false (which you can already
specify two different ways) and showResultCount: false.

TEST=manual

Test that this config does not do anything anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -834,8 +834,6 @@ ANSWERS.addComponent('VerticalResults', {
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, whether to hide the default results header that VerticalResults provides. Defaults to false.
-  hideResultsHeader: false,
-  // Optional, the card used to display each individual result, see the Cards section for more details,
   card: {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -834,8 +834,6 @@ ANSWERS.addComponent('VerticalResults', {
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, whether to hide the default results header that VerticalResults provides. Defaults to false.
-  hideResultsHeader: false,
-  // Optional, the card used to display each individual result, see the Cards section for more details,
```
cea2aj: It looks like you deleted the comment below the option rather than the comment above the option

cea2aj:
In order to hide the default results header, what are the steps now? It seems like showAppliedFilters and showResultsCount are indirectly related to this functionality

oshi97:
> In order to hide the default results header, what are the steps now? It seems like showAppliedFilters and showResultsCount are indirectly related to this functionality

basically if you set both to false, you end up hiding the results header

README.md
oshi97: ugh my bad

PR #1109, opened 2020-09-29
Use lodash escapeRegExp on plural translation keys

cea2aj:
Use escapeRegExp rather than escapeInterpolationBrackets for plural translation key.

There was a bug where the following translation was not being translated properly:
msgid "([[resultsCount]] result)" msgid_plural "([[resultsCount]] results)" msgstr[0] "[[resultsCount]] resultado" msgstr[1] "[[resultsCount]] resultados"

The translation was failing due to the open and closing parentheses being interpreted as a regex capture group during the key lookup. This was because we manually check translation keys inside i18next with a regex test for plural translations. Use lodash.escapeRegExp to solve this issue and escape all other possible RegExp characters.

J=none
TEST=manual

Run an internationalized jambo build and observe the proper translation inside of fr-answerstemplates.compiled.min.js. View the working translation on the local test page.

### DIFF
--- a/conf/i18n/translator.js
+++ b/conf/i18n/translator.js
@@ -1,4 +1,5 @@
 const i18next = require('i18next');
+const escapeRegExp = require('lodash.escaperegexp');
 
 /**
  * This class wraps an instance of the i18next library and provides methods supporting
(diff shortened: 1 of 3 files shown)


oshi97 (approved):
looks great!

PR #1110, opened 2020-09-29
AppliedFilters: listen to new FACETS_LOADED key

oshi97:
This commit adds a FACETS_LOADED global storage key, which
sends an 'update' event when facets are rendered from the response,
and updates the AppliedFilters component to listen to this key
in addition to VERTICAL_RESULTS.

Prior to this change, AppliedFilters had to have its addComponent AFTER the Facets
component. This is because the AppliedFilters
component and Facets component were both listening to the same
StorageKey (DYNAMIC_FILTERS), so the order in which their
moduleId listeners ran depended on the addComponent order.
The Facets component has to run first, though, because in addition
to rendering the facets, it is also responsible for registering
each new vertical search's new set of facets into FilterNodes stored
in FilterRegistry.

For some context, each vertical search response is capable of
returning a set of facet filters, which we want displayed both
in the AppliedFilters component, as well as the Facets component.
The Facets component is ALSO responsible for loading these
facets into the SDK's FilterRegistry (it should not be).

Created SLAP-721 for correctly dealing with this. It would
be a pretty large code change, so for now we're adding this
short-term fix.

J=SLAP-712
TEST=manual

test that I can search "Consulting", which will return a
search response with the "Consulting" facet selected,
and the AppliedFilters component will show a removable
"Consulting" filter, regardless of whether I add AppliedFilters
before or after Facets inside the onReady()

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -233,6 +233,7 @@ export default class Core {
         if (typeof analyticsEvent === 'object') {
           this._analyticsReporter.report(AnalyticsEvent.fromData(analyticsEvent));
         }
+        this.globalStorage.set(StorageKeys.VERTICAL_SEARCH_LOADED, true);
         window.performance.mark('yext.answers.verticalQueryResponseRendered');
       });
   }
--- a/src/ui/components/results/appliedfilterscomponent.js
+++ b/src/ui/components/results/appliedfilterscomponent.js
@@ -27,7 +27,7 @@ export default class AppliedFiltersComponent extends Component {
     this._verticalKey = this._config.verticalKey ||
       this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
 
-    this.moduleId = StorageKeys.DYNAMIC_FILTERS;
+    this.moduleId = StorageKeys.VERTICAL_SEARCH_LOADED;
   }
 
   static areDuplicateNamesAllowed () {
--- a/src/core/storage/storagekeys.js
+++ b/src/core/storage/storagekeys.js
@@ -36,5 +36,6 @@ export default {
   RESULTS_HEADER: 'results-header',
   API_CONTEXT: 'context',
   REFERRER_PAGE_URL: 'referrerPageUrl',
-  QUERY_TRIGGER: 'queryTrigger'
+  QUERY_TRIGGER: 'queryTrigger',
+  VERTICAL_SEARCH_LOADED: 'vertical-search-loaded'
 };


cea2aj (approved):
LGTM!

src/core/core.js
```diff
@@ -233,6 +233,7 @@ export default class Core {
         if (typeof analyticsEvent === 'object') {
           this._analyticsReporter.report(AnalyticsEvent.fromData(analyticsEvent));
         }
+        this.globalStorage.set(StorageKeys.VERTICAL_SEARCH_LOADED, true);
```
tmeyer2115: I might be off-base here, but this seems a bit roundabout to me. We need the `AppliedFiltersComponent` to `setState` only after the `FilterOptionsComponent` has set the new Facets in the registry. Should we set a key in `globalStorage` in that component after it makes the call? It feels like we're assuming that `FilterOptions` will make the update by this point in the code. 

I'm a bit wary of the key name too. It sounds similar to something we have already. `VerticalResults` has a `searchState` that can be `SEARCH_LOADING`. I wouldn't want people to get confused about two similarly named things.

Lastly, do the `updateListeners` get invoked whenever a `globalStorage.set` is called on their `StorageKey`? I'm assuming we don't do a comparison first to see if the value's changed before invoking the listeners. If we do, I guess we'd need to reset this value.

oshi97: We do a shallow equality comparison, so we don't need to reset this value. I think a new key in the component makes more sense for the specific bug we're trying to fix!

src/ui/components/filters/facetscomponent.js
```diff
@@ -275,5 +275,6 @@ export default class FacetsComponent extends Component {
     );
 
     this._filterbox.mount();
+    this.core.globalStorage.set(StorageKeys.FACETS_LOADED, true);
```
tmeyer2115, line 278: Is this the only place we need to set this? If I execute a new query and it returns new Facets, do we go through this code pathway again?

src/ui/components/results/appliedfilterscomponent.js
```diff
@@ -27,7 +28,13 @@ export default class AppliedFiltersComponent extends Component {
     this._verticalKey = this._config.verticalKey ||
       this.core.globalStorage.getState(StorageKeys.SEARCH_CONFIG).verticalKey;
 
-    this.moduleId = StorageKeys.DYNAMIC_FILTERS;
+    this.moduleId = StorageKeys.FACETS_LOADED;
+
+    this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, results => {
+      if (results.searchState === SearchStates.SEARCH_COMPLETE) {
```
tmeyer2115, line 34: For my own edification, we reach the `SEARCH_COMPLETE` state only after new Facets have been added to the registry?

src/core/core.js
oshi97: I updated things so that Facets component sets FACETS_LOADED at the end of it's onMount. The problem with this is that the AppliedFilters component would then have to listen to both FACETS_LOADED and VERTICAL_RESULTS, otherwise it wouldn't work properly without a Facets component on the page. This results in 2 setStates per query if you're using Facets (verified with console logs). Visually I couldn't see any flickering behavior at all, it feels a little silly, but I guess it's not the worst. 

Another alternative would be flipping the order VERTICAL_RESULTS and DYNAMIC_FILTERS are updated. I'm pretty sure this would not have any side effects (did some quick testing and looks like it doesn't), but it's a little scary.

Anyways, I feel like either we would want to find a better name for another StorageKey is updated after DYNAMIC_FILTERS during a VerticalSearch, or live with the double setState

src/ui/components/results/appliedfilterscomponent.js
oshi97: This is to handle cases where we don't have a Facets component on the page. I wrote a small explanation and some possible other options above.

oshi97: SEARCH_COMPLETE happens before Facets are updated in the registry

src/ui/components/filters/facetscomponent.js
oshi97: Yup. every time we execute a new query, DYNAMIC_FILTERS is updated, which calls the Facets component setState, and then its onMount

src/core/core.js
tmeyer2115: Could you explain a bit more how `AppliedFilters` will work if the `FacetsComponent` is not present? I got the impression that the `FacetsComponent` was needed for `AppliedFilters` to work, since it added facets to the registry. If the `AppliedFilters` component works fine on it's own (when `FacetsComponent` is not around), I don't totally follow why having them together is an issue.

oshi97: `AppliedFilters` just pulls from all the filters in FilterRegistry, it doesn't care about the difference between facet filters vs static filters. The issue is, currently the SDK relies on `Facets` to register a new search's facet filters into FilterRegistry. If the `AppliedFilters` setState happens before `Facets` setState, `AppliedFilters` will pull all FilterNodes from `FilterRegistry` before `Facets` has a chance to register them, which ends up pulling all the facet FilterNodes sent in the request, not the ones parsed from the response. This leads to issues when you run a search like "consulting", which does not send any facets in the request, but will return a response with the "consulting" facet selected.

oshi97: If `Facets` are not present, `AppliedFilters` will still pull all filter nodes from `FilterRegistry` as normal, which handles static filters properly since static filters don't change when they come back.

tmeyer2115: Because we rely on the `Facets` component to register the new dynamic filters, can `AppliedFilters` show these filters if `Facets` has not been added to the page?

tmeyer2115: But it can't show any of the dynamic filters that were applied without `Facets`? If so, that seems like an issue and I bet Product expects that we can show dynamic applied filters, even if `Facets` hasn't been added.

oshi97: Nope, without a `Facets` component no Facets will be added to the FilterRegistry. If we move the creation of FilterNodes for facets outside of the `Facets` component they would be, though.

tmeyer2115: We should tell Rose about this. She should hear it from us and not a TechOps.

tmeyer2115: If we can quickly come up with a sane way to avoid the double `setState`, I say let's do it. If this seems like it will be difficult or time consuming to avoid, I'm okay leaving it for now. We already know that the SDK's filters are in terrible shape and are planning an imminent refactor. We don't want to hold up 1.6 over this.

PR #1111, opened 2020-09-29
Readme: add performance marks section

oshi97:
TEST=m a n u a l

Double checked in chrome dev tools that things were in the right order
checked that links and stuff worked on our github readme for the branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Outline:
    - [Conversion Tracking](#conversion-tracking)
    - [On-Search Analytics](#on-search-analytics)
 8. [Rich Text Formatting](#rich-text-formatting)
+9. [Performance Metrics](#performance-metrics)
 
 # Install and Setup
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2260,3 +2261,40 @@ ANSWERS.ponyfillCssVariables({
...
 ```
+
+# Performance Metrics
+
+The SDK uses the Performance API, via `window.performance.mark()`, to create performance metrics regarding ANSWERS.init(), vertical search, and universal search. These marks can be viewed through browser developer tools, or programmatically through `window.performance.getEntries()`.
+
+## ANSWERS.init()
+
+1. `'yext.answers.initStart'` called when ANSWERS.init beings
+
```
cea2aj, line 2272: Are the numbers here important? The 1, 2, 3, etc. Can we remove them?

oshi97: those are the order the the marks are executed, feel like I’d rather have them but dont feel too strong either way. is there a reason you dislike them?

cea2aj: Oh, I'd keep them then

cea2aj (approved):

PR #1112, opened 2020-09-29
Readme: deprecate the ResultsHeader component built into VerticalResults

oshi97:
Update the readme to guide people to the new VerticalResultsCount
and AppliedFilters components instead of the ResultsHeader component
tacked on top of VerticalResults

TEST=n0ne

### DIFF
--- a/README.md
+++ b/README.md
@@ -827,10 +827,6 @@ ANSWERS.addComponent('VerticalResults', {
   itemTemplate: `<div> Custom template </div>`,
   // Optional, set a max number of columns to display at the widest breakpoint. Possible values are 1, 2, 3 or 4, defaults to 1
   maxNumberOfColumns: 1,
-  // Optional, whether to display the total number of results, default true
-  showResultCount: true,
-  // Optional, a custom template for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount.
-  resultsCountTemplate: '<div>{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}</div>',
   // Optional, a modifier that will be appended to a class on the results list like this `yxt-Results--{modifier}`
   modifier: '',
   // Optional, whether to hide the default results header that VerticalResults provides. Defaults to false.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -851,7 +847,14 @@ ANSWERS.addComponent('VerticalResults', {
...
+  // DEPRECATED: we recommend setting hideResultsHeader to true, and using the VerticalResultsCount and AppliedFilters components instead.
+  showResultCount: true,
+  // Optional, a custom template for the results count. You can specify the variables resultsCountStart, resultsCountEnd, and resultsCount.
+  // DEPRECATED: we recommend setting hideResultsHeader to true, and using the VerticalResultsCount and AppliedFilters components instead.
+  resultsCountTemplate: '<div>{{resultsCountStart}} - {{resultsCountEnd}} of {{resultsCount}}</div>',
   // Configuration for the applied filters bar in the header.
+  // DEPRECATED: we recommend setting hideResultsHeader to true, and using the VerticalResultsCount and AppliedFilters components instead.
   appliedFilters: {
     // If true, show any applied filters that were applied to the vertical search. Defaults to true
     show: true,
```
cea2aj: This is a lot of text in the readme dedicated to deprecated options. What are you thoughts on putting a single message such as // DEPRECATION NOTICE: The options showResultsCount, resultsCountTemplate, and appliedFilters are deprecated. We recommend setting hideResultsHeader to true, and using the VerticalResultsCount and AppliedFilters components instead.

This would let us remove a lot of text including the appliedFilters options. If people need to see the options they can look at an older release

oshi97: yeah I think something like that would make a lot of sense!

README.md
```diff
@@ -851,6 +847,17 @@ ANSWERS.addComponent('VerticalResults', {
     // Optional, whether to display all results in the vertical when no results are found. Defaults to false, in which case only the no results card will be shown.
     displayAllResults: false
   },
+
+  /**
+   * DEPRECATED: The config below is DEPRECATED.
```
cea2aj: Thoughts on "The config options below are DEPRECATED" to make it clearer that it is multiple options rather than a single one?

cea2aj (approved):
🔥

PR #1114, opened 2020-09-30
Fix sending object instead of boolean for sessionTrackingEnabled

oshi97:
This change was probably lost in a merge, the original PR is #992.

TEST=manual

see that on a vertical test page I no longer see [Object object]
being sent in the query for the sesionTrackingEnabled param


cea2aj:
Huh, that's really odd that this change was lost. The PR got into release/v1.6, but git blame doesn't show at what point it was deleted

cea2aj (approved):

PR #1116, opened 2020-10-01
AppliedFilters/VerticalResultsCount: remove extra margin

oshi97:
When these components were factored out of the ResultsHeader
component, they retained margin styling that was originally
there to space them correctly on top of VerticalResults.
This has been removed, since they are no longer tied down
to vertical results.

TEST=manual

Test that AppliedFilters, removable: true and false, and
showFieldNames: true and false, does not have extra margin

test that VerticalResultsCount also does not have extra margin


cea2aj (approved):
LG!

PR #1118, opened 2020-10-05
AppliedFilters: only render when has filters

oshi97:
This commit guards the rendering of the applied filters component.
Now it will only render when there are
filters to be displayed.

TEST=manual

linked local sdk to a theme page, checked that applied filters dont render when there are no filters

also checked on a raw sdk site


cea2aj (approved):
LG!

PR #1122, opened 2020-10-06
Add browserstack + testcafe integration with circleci

oshi97:
This commit updates our circle ci config to run our testcafe
acceptance tests on browserstack. Videos can be viewed under
the browserstack automate section of browserstack.

TEST=auto

See that in circle ci, and in browserstack automate, our acceptance
tests are being run

see that in browserstack automate, our tests have the Answers SDK project,
each branch gets its own build tag, and each commit in a branch gets a test
session under the build tab

test that acceptance tests are run for build_and_deploy_i18n as well


tests/acceptance/acceptancesuite.js
```diff
@@ -29,7 +29,7 @@ test('Basic universal flow', async t => {
   await t.expect(sections.length).eql(2);
 
   const faqsSectionTitle = await sections[1].getTitle();
-  await t.expect(faqsSectionTitle).contains('FAQ');
+  await t.expect(faqsSectionTitle.toUpperCase()).contains('FAQ');
```
oshi97, line 32: fix for ie11, innerText doesn't respect text-transform: uppercase in ie11

.circleci/run_acceptance.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+export BROWSERSTACK_USE_AUTOMATE="1"
+export BROWSERSTACK_PROJECT_NAME="Answers SDK"
+export BROWSERSTACK_BUILD_ID=$CIRCLE_BRANCH
+COMMIT_MSG_TITLE=$(git log -n 1 --pretty=format:%s)
```
cea2aj: Should we set the `BROWSERSTACK_BUILD_ID` to something that is unique? I think there will be collisions if we use `CIRCLE_BRANCH`. Would `CIRCLE_BUILD_NUM` be a better choice here because those numbers are unique for each job?

cea2aj:
Can we split up test into unit_test and acceptance_test? That would make them run in parallel and would give us more visibility into what failed.

oshi97:
> Can we split up test into unit_test and acceptance_test? That would make them run in parallel and would give us more visibility into what failed.

hmm splitting them into different jobs would still have them run in serial, but I can look into whether we can do some parallel stuff here

.circleci/run_acceptance.sh
oshi97: I was trying to group them by branch name, so the automate ui would look like
image
but I guess if somebody were to use the same branch name for two different prs that could be annoying

cea2aj:
> > Can we split up test into unit_test and acceptance_test? That would make them run in parallel and would give us more visibility into what failed.
> 
> hmm splitting them into different jobs would still have them run in serial, but I can look into whether we can do some parallel stuff here

If they both depend on the build step, shouldn't they run in parallel?

.circleci/run_acceptance.sh
cea2aj: It looks like the `BROWSERSTACK_BUILD_ID` doesn't mind having multiple entries of the same value, so I think this works.

oshi97:
> > > Can we split up test into unit_test and acceptance_test? That would make them run in parallel and would give us more visibility into what failed.
> > 
> > 
> > hmm splitting them into different jobs would still have them run in serial, but I can look into whether we can do some parallel stuff here
> 
> If they both depend on the build step, they shouldn't they run in parallel?

you're right! we get the parallelism for free. I remembered reading about specifying a specific job's parallelism level, but two separate jobs can run in parallel without any of that!

.circleci/run_acceptance.sh
oshi97: I still think what you said makes sense though, updated it to use CIRCLE_BUILD_NUM.

cea2aj: Alternatively, we could append `CIRCLE_BUILD_NUM` to `CIRCLE_BRANCH`

cea2aj (changes requested):
Do we want to run acceptance tests for the build_and_deploy_i18n and the build_and_deploy_hold workflows? Other than that, LGTM!

cea2aj (approved):

PR #1124, opened 2020-10-08
Verify all translations are present in CircleCI

cea2aj:
Add a translation_test job to Circle which tests that all the PO files contain all the translations defined in the POT file.

Use the utility "msgcmp" to compare the files

J=SLAP-724
TEST=manual

Observe CircleCI pass when all translations are present. Push a test commit where a translation is missing and observe it fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,21 @@ jobs:
       - attach_workspace:
           at: ~/answers
       - run: ./.circleci/run_acceptance.sh
-        
+  # verify all translations are present
+  translation_test:
+    docker:
+      - image: circleci/node.14.5
+    working_directory: ~/answers
+    steps:
+      - setup-workspace
+      - attach_workspace:
+          at: ~/answers
+      - run: 
+        name: Verify translation files
+        command: |
+          npm run extract-translations
+          cd conf/i18n/translations
+          find . -type f -name "*.po" | xargs -n1 -I {{po_file}} msgcmp {{po_file}} messages.pot
   # deploy to S3 in the root folder, overwriting the existing latest version
   deploy_latest:
     docker:
@@ -158,6 +172,9 @@ workflows:
       - acceptance_test:
           requires:
             - build
+      - translation_test:
+          requires:
+            - build
       - deploy_canary:
           filters:
             branches:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -92,7 +92,22 @@ jobs:
...
+    working_directory: ~/answers
+    steps:
+      - setup-workspace
+      - attach_workspace:
+          at: ~/answers
+      - run: 
+          name: Verify translation files
+          command: |
+            npm run extract-translations
+            sudo apt-get install gettext
```
oshi97: maybe we could add a quiet flag (looks like this would be -qq) to make the output slightly quieter? it's not a big deal but it would make the output easier to read when there's an error

.circleci/config.yml
```diff
@@ -92,7 +92,22 @@ jobs:
...
+            cd conf/i18n/translations
+            find . -type f -name "*.po" | xargs -n1 -I {{po_file}} msgcmp {{po_file}} messages.pot
```
oshi97: would it be a little clearer what's happening if we just did a for loop through the files in the translations folder? something like 
```sh
FILES=/path/to/*
for f in $FILES
```
happy with what we have right now as well it's very cool guy so up to you

.circleci/config.yml
cea2aj: sure

.circleci/config.yml
cea2aj: Updated to make this clearer!

.circleci/config.yml
```diff
@@ -92,7 +92,24 @@ jobs:
...
+          at: ~/answers
+      - run: 
+          name: Verify translation files
+          command: |
+            npm run extract-translations
+            sudo apt-get install -qq gettext
+            cd conf/i18n/translations
+            for po_file in *.po
+              do msgcmp $po_file messages.pot
+            done
```
alexisgrow: Would it make sense move this into its own file similar to `./.circleci/run_acceptance.sh`? I'm not familiar with CircleCI config so definitely your call

alexisgrow (commented):
One comment, but going to defer to Oliver on the +2!

.circleci/config.yml
cea2aj: ya, I can do that

alexisgrow (approved):
This LGTM but I'd feel more comfortable if Tom or Oliver also approved since I haven't spent much time configuring our CircleCI setup!

oshi97 (approved):
this lgtm!

PR #1125, opened 2020-10-09
DefaultTemplatesLoader: Fix async template loading race condition

oshi97:
If somebody was using the built-in template loader, instead of
manually specifying a script tag with the templates, we had a race condition
between the master switch api call and the default templates loader.
There was a chance components would be addComponent-ed before any
templates were registered.

This commit uses Promise.all() so we don't delay the masterswitch api +
ponyfill while waiting for the templates to load.
Without this update, waiting for the template loader, and after it's finished only
then running the ponyfill and checking the master switch api, caused an additional ~75-100ms
slowdown. Updating the code to use Promise.all() got us back the old performance, 
barring a few occasional spikes (which I'm assuming would have errored out without 
this change) his was measured by adding a performance mark at the start of the onReady().
Theoretically the page would also be slightly faster on ie11, since the master switch api no longer
blocks the ponyfill, but I haven't confirm this.

There was some consideration to use Promise.race() with the master switch api,
so that if the master switch was disabled then the site would fail immediately
instead of waiting for the other async deps, but it felt weird to optimize for
this and it made the code more complicated.

T=https://yextops.zendesk.com/agent/tickets/356651
TEST=manual

for these changes, I had to change the COMPILED_TEMPLATES_URL
to point to my localhost

test that if I force the template loader to wait an extra 2 seconds
before creating the templates script tag, the rest of the page will
wait for the templates to load (and check that without this code
change we would see the page fail to load without waiting)

test that if that master switch is disabled, the experience will
still not load

test that if the master switch takes longer than the timeout,
the page will load (tested by changing the timeout to 0ms)

test that I can chain more .then()s after the ANSWERS.init()
promise (was worried that using finally() would prevent this)

test speed as noted above

test that ponyfill still works on ie11 with --yxt-filter-options-options-max-height

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -5,7 +5,7 @@ module.exports = [
   },
   {
     path: 'dist/answers-modern.min.js',
-    limit: '105kb'
+    limit: '110kb'
   },
   {
     path: 'dist/answerstemplates.compiled.min.js',
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -294,19 +294,6 @@ class Answers {
     return Promise.all([loadTemplates, ponyfillCssVariables, masterSwitch]);
   }
 
-  _checkMasterSwitch () {
-    window.performance.mark('yext.answers.statusStart');
-    const handleFulfilledMasterSwitch = (isDisabled) => {
-      this._disabledByMasterSwitch = isDisabled;
-    };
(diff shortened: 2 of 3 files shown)


cea2aj (approved):
Looks good! This is a lot cleaner

PR #1128, opened 2020-10-12
Add redirectUrlTarget config option to SearchBar component

alexisgrow:
Add a config option that allows a user to specify the target of the redirect url, if provided.

TEST=manual

Specify "redirectUrlTarget: "_blank"" with a redirectUrl, see it open in new tab. Also tested "_self", "_parent" and "_top".

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,7 +431,7 @@ ANSWERS.addComponent('SearchBar', {
   clearButton: true,
   // Optional, redirect search query to url
   redirectUrl: 'path/to/url',
-  // Optional, target frame for the redirect url, defaults to current frame
+  // Optional, target frame for the redirect url, defaults to current frame. Expects a valid target: "_blank", "_self", "_parent", "_top" or the name of a frame
   redirectUrlTarget: '_self',
   // Optional, defaults to native form node within container
   formSelector: 'form',
--- a/README.md
+++ b/README.md
@@ -431,6 +431,8 @@ ANSWERS.addComponent('SearchBar', {
   clearButton: true,
   // Optional, redirect search query to url
   redirectUrl: 'path/to/url',
+  // Optional, target frame for the redirect url, defaults to current frame
+  redirectUrlTarget: '_self',
   // Optional, defaults to native form node within container
   formSelector: 'form',
   // Optional, defaults to true. When true, a form is used as the query submission context.
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -137,8 +137,6 @@ export default class SearchComponent extends Component {
      * redirectUrl is also supplied.
      * Optional, defaults to current frame.
      *
-     * If no redirectUrlTarget provided, we keep the page as a single page app.
-     *
      * @type {string}
      */
     this.redirectUrlTarget = config.redirectUrlTarget || '_self';
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -133,7 +133,7 @@ export default class SearchComponent extends Component {
     this.redirectUrl = config.redirectUrl || null;
 
     /**
-     * redirectUrlTarget will force the search query submission to open in the frame specified, if
+     * redirectUrlTarget will force the search query submission to open in the frame specified if
      * redirectUrl is also supplied.
      * Optional, defaults to current frame.
      *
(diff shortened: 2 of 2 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -487,7 +498,12 @@ export default class SearchComponent extends Component {
     // serialized and submitted.
     if (typeof this.redirectUrl === 'string') {
       if (this._allowEmptySearch || query) {
-        window.location.href = this.redirectUrl + '?' + params.toString();
+        const newUrl = this.redirectUrl + '?' + params.toString();
+        if (this.redirectUrlTarget) {
+          window.open(newUrl, this.redirectUrlTarget) || (window.location.href = newUrl);
```
oshi97: would it be possible to default redirectUrlTarget to the current window, and not have this if check?

README.md
```diff
@@ -431,6 +431,8 @@ ANSWERS.addComponent('SearchBar', {
   clearButton: true,
   // Optional, redirect search query to url
   redirectUrl: 'path/to/url',
+  // Optional, target frame for the redirect url, defaults to current frame
+  redirectUrlTarget: '_self',
```
tmeyer2115, line 435: Did Rose say anything about grouping these two config options? Something like:

```
redirect: {
  url: '...',
  target: '...'
}
```
Personally, I prefer related options grouped like this. It's not a big deal, but I think it allows people to compartmentalize. It would mean that the `redirectUrl` is now deprecated, which might not be worth it.

src/ui/components/search/searchcomponent.js
```diff
@@ -487,7 +498,8 @@ export default class SearchComponent extends Component {
     // serialized and submitted.
     if (typeof this.redirectUrl === 'string') {
       if (this._allowEmptySearch || query) {
-        window.location.href = this.redirectUrl + '?' + params.toString();
+        const newUrl = this.redirectUrl + '?' + params.toString();
+        window.open(newUrl, this.redirectUrlTarget) || (window.location.href = newUrl);
```
tmeyer2115, line 500: For my own edification, what's the `|| (window.location.href = newUrl)` do? If the `target` is `_self` does `window.open` return false?

src/ui/components/search/searchcomponent.js
```diff
@@ -128,10 +128,21 @@ export default class SearchComponent extends Component {
...
      * If no redirectUrl provided, we keep the page as a single page app.
      *
-     * @type {boolean}
+     * @type {string}
      */
     this.redirectUrl = config.redirectUrl || null;
 
+    /**
+     * redirectUrlTarget will force the search query submission to open in the frame specified, if
+     * redirectUrl is also supplied.
```
cea2aj: nit: I don't think you need a comma here

README.md
cea2aj: Do you think it would be helpful to list possible options for the target such as "_parent", "_blank" or "_top"?

alexisgrow: Yeah, that was actually an alternative considered in the item! This is what that portion of the description said:

> Alternatives considered:
> 
> Nest it a new object, but this just adds more mess for backwards compatibility
> 
> redirect {
>    url: test.answers.com
>    target: "top"
> }
> 
>

src/ui/components/search/searchcomponent.js
alexisgrow: Users can configure their browser settings so that window.open won't work well. For example, if someone has an aggressive pop-up blocking setup, this would fall back to opening the redirect URL in the current frame using the `window.location.href = newUrl`. To my knowledge, _self should work OK with default browser settings.

README.md
tmeyer2115: Okay! Sounds like we decided it wasn't worth the compatibility headache.

src/ui/components/search/searchcomponent.js
```diff
@@ -128,10 +128,21 @@ export default class SearchComponent extends Component {
...
+     * @type {string}
      */
     this.redirectUrl = config.redirectUrl || null;
 
+    /**
+     * redirectUrlTarget will force the search query submission to open in the frame specified if
+     * redirectUrl is also supplied.
+     * Optional, defaults to current frame.
+     *
+     * If no redirectUrlTarget provided, we keep the page as a single page app.
```
tmeyer2115: Minor nit, but having this sentence and "Optional, defaults to current frame" is a bit redundant.

tmeyer2115 (commented):
One small thing, other than that lgtm!

src/ui/components/search/searchcomponent.js
alexisgrow: Oh hahah totally makes sense, updating! I copy-pasted the JSDoc from the redirectUrl as a template and should have read it more critically! 🙈

tmeyer2115 (approved):

PR #1129, opened 2020-10-12
Add unit tests to the i18n build process

cea2aj:
Add tests to the code responsible for transpiling the javascript and hbs helpers into processTranslation calls.

Test coverage includes `TranslationResolver`, `TranslateCallParser`, `TranslateHelperVisitor`, `Translator`, `fromMustacheStatementNode()`, and `generateProcessTranslationJsCall()`.

This PR also adds tests for the translation extractor and adds pluralWithContext support to the translation resolver. The tests for the translator are from the Jambo translator test suite. I tweaked the `ALL_LANGUAGES` to make it work with the tests.

J=SLAP-700
TEST=auto

Run unit tests and confirm they pass

### DIFF
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -2,7 +2,7 @@ exports.TRANSLATION_FLAGGER_REGEX = /TranslationFlagger.flag\((\s)*\{[^;]+?\}(\s
 
 exports.DEFAULT_LOCALE = 'en';
 
-exports.LANGUAGES_TO_LOCALES = {
+const LANGUAGES_TO_LOCALES = {
   'en': [
     'en_AE',
     'en_AI',
@@ -104,5 +104,6 @@ exports.LANGUAGES_TO_LOCALES = {
     'ja_JP'
   ]
 };
+exports.LANGUAGES_TO_LOCALES = LANGUAGES_TO_LOCALES;
 
-exports.ALL_LANGUAGES = ['en', 'es', 'fr', 'it', 'de', 'ja'];
+exports.ALL_LANGUAGES = Object.keys(LANGUAGES_TO_LOCALES);
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -105,4 +105,4 @@ exports.LANGUAGES_TO_LOCALES = {
   ]
 };
 
-exports.ALL_LANGUAGES = Object.keys('en', 'es', 'fr', 'it', 'de', 'ja');
+exports.ALL_LANGUAGES = ['en', 'es', 'fr', 'it', 'de', 'ja'];
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -105,4 +105,4 @@ exports.LANGUAGES_TO_LOCALES = {
   ]
 };
 
-exports.ALL_LANGUAGES = Object.keys(this.LANGUAGES_TO_LOCALES);
+exports.ALL_LANGUAGES = Object.keys('en', 'es', 'fr', 'it', 'de', 'ja');
--- a/tests/conf/i18n/translatehelpervisitor.js
+++ b/tests/conf/i18n/translatehelpervisitor.js
@@ -13,10 +13,14 @@ async function createTranslator () {
 }
 
 describe('TranslateHelperVisitor translates visited nodes', () => {
-  it('static translation', async () => {
+  let translator;
+  beforeEach(async () => {
+    translator = await createTranslator();
+  });
+
+  it('static translation', () => {
     const template = `{{ translate phrase='Hello' }}`;
     const ast = Handlebars.parse(template);
-    const translator = await createTranslator();
 
     new TranslateHelperVisitor(translator).accept(ast);
     const indexOfFirstStatement = 0;
@@ -25,10 +29,9 @@ describe('TranslateHelperVisitor translates visited nodes', () => {
     expect(translatedValue).toEqual('Bonjour');
   });
 
-  it('plural translation', async () => {
+  it('plural translation', () => {
     const template = `{{ translate phrase='result' pluralForm='results' count=resultCount }}`;
     const ast = Handlebars.parse(template);
-    const translator = await createTranslator();
 
     new TranslateHelperVisitor(translator).accept(ast);
     const indexOfFirstStatement = 0;
@@ -50,11 +53,10 @@ describe('TranslateHelperVisitor translates visited nodes', () => {
     expect(hashPairs).toMatchObject([expectedCount, expectedPluralForm0, expectedPluralForm1]);
   });
 
-  it('a non-translate helper should not be modified', async () => {
+  it('a non-translate helper should not be modified', () => {
     const template = `{{ yext phrase='Hello' }}`;
     const ast = Handlebars.parse(template);
     const originalAst = _.cloneDeep(ast);
-    const translator = await createTranslator();
 
     new TranslateHelperVisitor(translator).accept(ast);
 
--- a/tests/conf/i18n/translatehelpervisitor.js
+++ b/tests/conf/i18n/translatehelpervisitor.js
@@ -26,7 +26,7 @@ describe('TranslateHelperVisitor usage', () => {
   });
 
   it('plural translation', async () => {
-    const template = `{{ translate phrase='result' pluralForm='results' count='resultCount' }}`;
+    const template = `{{ translate phrase='result' pluralForm='results' count=resultCount }}`;
     const ast = Handlebars.parse(template);
     const translator = await createTranslator();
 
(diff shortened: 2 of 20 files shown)


conf/i18n/constants.js
```diff
@@ -105,4 +105,4 @@ exports.LANGUAGES_TO_LOCALES = {
   ]
 };
 
-exports.ALL_LANGUAGES = Object.keys(this.LANGUAGES_TO_LOCALES);
```
oshi97, line 108: how come this doesn't work with tests?

conf/i18n/constants.js
```diff
@@ -105,4 +105,4 @@ exports.LANGUAGES_TO_LOCALES = {
...
+exports.ALL_LANGUAGES = ['en', 'es', 'fr', 'it', 'de', 'ja'];
```
alexisgrow: I saw in the PR comment that the reason for this change was to make it work with the tests and to simplify. What would happen if `ALL_LANGUAGES` and `LANGUAGES_TO_LOCALES` were to get out of sync? It seems like it could be an opportunity for a bug in the future if someone didn't realize they needed to update both constants

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,65 @@
...
+
+    new TranslateHelperVisitor(translator).accept(ast);
+    const indexOfFirstStatement = 0;
+    const translatedValue = ast.body[indexOfFirstStatement].value;
+
+    expect(translatedValue).toEqual('Bonjour');
+  });
+
+  it('plural translation', async () => {
+    const template = `{{ translate phrase='result' pluralForm='results' count='resultCount' }}`;
```
alexisgrow: Should count for plurization be a number?

tests/conf/i18n/translator.js
```diff
@@ -0,0 +1,257 @@
...
+      2: 'Nepavyksta nusiųsti apžvalgų el. paštu'
+    };
+
+    expect(translation).toEqual(expectedResult);
+  });
+
+  it('pluralization with interpolation works as expected', () => {
+    const translation = translator.translatePlural(
+      '1 location selected',
+      '[[count]] locations selected]'
```
alexisgrow: Is the `]` at the end intentional?

alexisgrow (commented):
These tests are great 🔥🔥

tests/conf/i18n/translator.js
cea2aj: nope, removing

tests/conf/i18n/translatehelpervisitor.js
cea2aj: It should be a variable because we don't know the value of count until runtime. I realized that I have it as a string and not a variable here , so I just pushed a commit which fixes that

tests/conf/i18n/translationresolver.js
```diff
@@ -0,0 +1,77 @@
+const TranslationResolver = require('../../../conf/i18n/translationresolver');
```
oshi97, line 1: definitely out of scope for your pr, but (cc: @tmeyer2115) do we still have plans to unify the SDK and Jambo translation classes? it feels a little weird to have double coverage for things

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,71 @@
...
+async function createTranslator () {
+  const locale = 'fr';
+  return Translator.create(locale, [], { [locale]: { translation: {
+    'Hello': 'Bonjour',
+    'result': 'résultat',
+    'result_plural': 'résultats'
+  } } });
+}
+
+describe('TranslateHelperVisitor usage', () => {
```
oshi97: can we update this describe to be more descriptive

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,71 @@
...
+    const expectedCount = {
+      key: 'count',
+      value: { type: 'PathExpression', parts: ['resultCount'] }
+    };
+
+    const actualPluralForm0 = hashPairs.find(hashPair => hashPair.key === 'pluralForm0');
+    const actualPluralForm1 = hashPairs.find(hashPair => hashPair.key === 'pluralForm1');
+    const actualCount = hashPairs.find(hashPair => hashPair.key === 'count');
+
+    expect(actualPluralForm0).toMatchObject(expectedPluralForm0);
```
oshi97: Could you try passing a the AST to Handlebars.compile and testing that way? I think we could simplify these tests a lot that way

tests/conf/i18n/translationplaceholderutils.js
```diff
@@ -0,0 +1,32 @@
...
+      phrase='result' 
+      pluralForm='results' 
+      count='resultCount' 
+      param1='param1'
+      param2='param2'
+      context='testing' 
+    }}`;
+
+    const indexOfFirstMustacheStatement = 0;
+    const mustacheStatement = Handlebars.parse(template).body[indexOfFirstMustacheStatement];
```
oshi97, line 17: Could we just have the mustache statement object hard coded instead of using Handlebars.parse to get it? I think it might simplify this test a little, and also make it more clear what is being taken in by `fromMustacheNode`.

tests/conf/i18n/translationresolver.js
```diff
@@ -0,0 +1,77 @@
...
+    'Hello': 'Bonjour',
+    'result': 'résultat',
+    'result_plural': 'résultats',
+    'object_noun': 'objet',
+    'object_noun_plural': 'objets',
+    'mail_noun': 'courrier'
+  } } });
+}
+
+describe('TranslationResolver usage', () => {
```
oshi97: can we replace "usage" with something more descriptive (just slightly more descriptive don't have to go overboard) also for the other describe blocks

tests/conf/i18n/translationresolver.js
oshi97: On nevermind, SLAP-701 still exists!

conf/i18n/constants.js
cea2aj: If they got out of sync, builds would not run. I'm reverting this change though, as I got it to work with the tests

conf/i18n/constants.js
cea2aj: 'this' was undefined, but I got it working now

tests/conf/i18n/translationresolver.js
cea2aj: sure

tests/conf/i18n/translationplaceholderutils.js
cea2aj: I could, but it would add at least 180 lines of code since the mustache statement object is so large. Is that worth it?

tests/conf/i18n/translatehelpervisitor.js
cea2aj: Can you explain how that would work?

oshi97: my bad I thought there was an easy way to convert the AST back into the template. I tried some googling and it doesn't look like there are any libraries to do this right now. But this test still feels like it's a little long, maybe split the 3 expects into three different `it` blocks nested inside the plural one?

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,71 @@
...
+    'result': 'résultat',
+    'result_plural': 'résultats'
+  } } });
+}
+
+describe('TranslateHelperVisitor usage', () => {
+  it('static translation', async () => {
+    const template = `{{ translate phrase='Hello' }}`;
+    const ast = Handlebars.parse(template);
+    const translator = await createTranslator();
```
oshi97: do we need to create the translator twice, or could this be done inside the describe block?

tests/conf/i18n/translationplaceholderutils.js
oshi97: hmm okay it's probably fine then!

oshi97 (commented):
overall looks great!

tests/conf/i18n/translatehelpervisitor.js
cea2aj: I managed to reduce this quite a bit by passing an array to toMatchObject()

tests/conf/i18n/translatehelpervisitor.js
cea2aj: I moved this to a beforeEach in the describe block

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -12,11 +12,15 @@ async function createTranslator () {
   } } });
 }
 
-describe('TranslateHelperVisitor usage', () => {
-  it('static translation', async () => {
+describe('TranslateHelperVisitor translates visited nodes', () => {
+  let translator;
+  beforeEach(async () => {
```
oshi97: I guess I was thinking we don't need to make a new translator before each test, the translator should be able to handle translating multiple templates without reinstantiating itself. We should be able to have one translator used between all the tests

cea2aj: Ok, just switched it to beforeAll instead

tests/conf/i18n/translatehelpervisitor.js
```diff
@@ -0,0 +1,66 @@
...
+  return Translator.create(locale, [], { [locale]: { translation: {
+    'Hello': 'Bonjour',
+    'result': 'résultat',
+    'result_plural': 'résultats'
+  } } });
+}
+
+describe('TranslateHelperVisitor translates visited nodes', () => {
+  let translator;
+  /* global beforeAll */
```
oshi97: lets add this to the globals in the package.json (beforeEach is there as well)

cea2aj: Oh, that's a much better spot!

oshi97 (approved):
looks great! thanks for dealing with my annoying feedback

PR #1130, opened 2020-10-13
Expose an ANSWERS.search function

alexisgrow:
This can be used to conduct a search from outside the library's SearchBar component.

TEST=manual

With an existing search experience set up with a UniversalResults component and a SearchBar component and see the query appear into the SearchBar and a search is conducted.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -397,6 +397,16 @@ class Answers {
     return this.components.create('Component', opts).mount();
   }
 
+  /**
+   * Conducts a search in the Answers experience
+   *
+   * @param {string} query
+   */
+  search (query) {
+    this.core.setQuery(query, { setQueryParams: true });
+    this.core.persistentStorage.set(StorageKeys.QUERY, query);
+  }
+
   registerHelper (name, cb) {
     this.renderer.registerHelper(name, cb);
     return this;


src/answers-umd.js
```diff
@@ -397,6 +397,16 @@ class Answers {
     return this.components.create('Component', opts).mount();
   }
 
+  /**
+   * Conducts a search in the Answers experience
+   *
+   * @param {string} query
+   */
+  search (query) {
+    this.core.setQuery(query, { setQueryParams: true });
```
oshi97, line 424: it's a little weird that this method is called search but to what it does is update the query, is there some other way to do this? Alternatively maybe we just call this setQuery?

oshi97: if it's not too hard, would it be possible to refactor core.setQuery to also update persistent storage? I imagine every time core.setQuery is called persistent storage is already being updated, just outside of setQuery (hopefully anyways)

alexisgrow: Yeah that's a good point. I've been testing with a SearchBar component on the page, which listens to updates on `StorageKeys.QUERY` and conducts a search. However, if someone didn't have a SearchBar on their page I'm not sure this would work. I'll look into this!

For updating other usages of setQuery, 3/6 of them don't update persistent storage, do you know if that is intentional? I'm not familiar enough with persistent storage to say for sure. (Examples that do not set it: SearchComponent's onCreate ([link](https://github.com/yext/answers/blob/develop/src/ui/components/search/searchcomponent.js#L293)), SearchComponent's listener on `StorageKeys.QUERY` ([link](https://github.com/yext/answers/blob/develop/src/ui/components/search/searchcomponent.js#L183)), answers_umd.js _setDefaultInitialSearch ([link](https://github.com/yext/answers/blob/develop/src/answers-umd.js#L467)))

oshi97: ah I tried to take a look and it's really hard to unravel what's going on, setting persistent storage here should be fine!

oshi97 (approved):

PR #1131, opened 2020-10-13
Add autocomplete onOpen and onClose hooks

alexisgrow:
TEST=manual

Test with the theme, specifying both onOpen and onClose with simple `console.log` statements. See the "open" print statement print once when the autocomplete opens, and "close" print once when autocomplete closes.

### DIFF
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -273,6 +273,12 @@ export default class SearchComponent extends Component {
      * @type {string}
      */
     this.inputIdName = `yxt-SearchBar-input--${this.name}`;
+
+    /**
+     * Options to pass to the autocomplete component
+     * @type {Object}
+     */
+    this._autocompleteConfig = config.autocomplete || {};
   }
 
   static get type () {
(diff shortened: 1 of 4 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -273,6 +273,12 @@ export default class SearchComponent extends Component {
      * @type {string}
      */
     this.inputIdName = `yxt-SearchBar-input--${this.name}`;
+
+    /**
+     * Options to pass to the autocomplete component
+     * @type {Object}
+     */
+    this._autocompleteConfig = config.autocomplete || {};
```
oshi97: what kind of config is being passed? I think if possible we want to avoid passing config directly

oshi97: took a look at the theme prs and I'm back! would it make sense to just expose the config needed for the overlay? we would probably want to add it to the readme as well

alexisgrow: Ah thanks!! I just updated to pass only the specific new config: onOpen and onClose hooks!

oshi97 (approved):
lgtm!

PR #1132, opened 2020-10-13
Add onSubmit function to SearchBar component

alexisgrow:
Add an onSubmit hook to the SearchBar component. This `onSubmit` function is necessary for the Overlay and should happen as (or before) the search network call is conducted.

This is different than the onUniversalSearch and onVerticalSearch because those functions are meant to be used for analytics. Those functions happen after the search is conducted and provide result information to a user; they also are expected to return an analytics event. 

TEST=manual

Add a simple onSubmit function with a console.log statement. See output printed to console when search is conducted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,7 +464,7 @@ ANSWERS.addComponent('SearchBar', {
     // Optional, a function invoked when a search is submitted via this component. The users search
     // is passed in as a string
     onSubmit: function(searchTerms) {}
-  }
+  },
   // Optional, options to pass to the autocomplete component
   autocomplete: {
     // Optional, boolean used to hide the autocomplete when the search input is empty (even if the
--- a/README.md
+++ b/README.md
@@ -460,8 +460,11 @@ ANSWERS.addComponent('SearchBar', {
   // Optional, functions invoked when certain events occur
   customHooks: {
     // Optional, a callback invoked when the clear search button is clicked
-    onClearSearch: function() {}
-  },
+    onClearSearch: function() {},
+    // Optional, a function invoked when a search is submitted via this component. The users search
+    // is passed in as a string
+    onSubmit: function(searchTerms) {}
+  }
   // Optional, options to pass to the autocomplete component
   autocomplete: {
     // Optional, boolean used to hide the autocomplete when the search input is empty (even if the
@@ -471,12 +474,6 @@ ANSWERS.addComponent('SearchBar', {
     onClose: function() {},
     // Optional, callback invoked when the autocomplete component changes from closed to open.
     onOpen: function() {},
-  },
-  // Optional, functions invoked when certain events occur
-  customHooks: {
-    // Optional, a function invoked when a search is submitted via this component. The users search
-    // is passed in as a string
-    onSubmit: function(searchTerms) {}
   }
 })
 ```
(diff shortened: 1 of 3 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -626,6 +631,7 @@ export default class SearchComponent extends Component {
    * @returns {Promise} A promise that will perform the query and update globalStorage accordingly.
    */
   search (query, searchOptions) {
+    this.onSubmit(query);
```
oshi97: I think this method is also called for searches on load, is that okay? Or just any searches that are triggered by the StorageKeys.QUERY global storage listener

README.md
```diff
@@ -456,7 +456,10 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
-  }
+  },
+  // Optional, a callback invoked when a search is submitted via this component. The users search
+  //           is passed in as a string
```
tmeyer2115: Minor nit: "Optional" and "is passed in" should be the same level of indentation.

src/ui/components/search/searchcomponent.js
tmeyer2115: The component also has an `onQuerySubmit` function. I believe that's fired only when the component's `form` is submitted, the search button is clicked, or a user hits return in the input. In short, I think it's only ever fired for searches caused by explicit user interaction. If you don't want `onSubmit` fired for searches on load,`onQuerySubmit` may be the place to put it.  

The `onQuerySubmit` function isn't exposed publicly. In the code, though, it may cause a bit of confusion to have that method and something called `onSubmit`. What do you think? One idea: we could call it `onSubmit` publicly, but name it something else in the component itself (i.e. `this.blah = config.onSubmit`)

alexisgrow: Yeah that totally makes sense to me! I combined this with the other feedback about nesting these commits under a "customHooks" top-level config option. In the code now, it's `this.customHooks.onSubmit()` which helps with the ambiguity. I also moved it to the `onQuerySubmit` function

README.md
```diff
@@ -456,6 +456,12 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
+  },
+  // Optional, functions invoked when certain events occur
+  customHooks: {
+    // Optional, a function invoked when a search is submitted via this component. The users search
+    // is passed in as a string
+    onSubmit: function(searchTerms) {}
```
tmeyer2115, line 466: This hook is only fired when `onQuerySubmit` is fired. Can a query be submitted in this component without going through `onQuerySubmit`? If so, we should make explicit the cases that `onSubmit` will be fired.

alexisgrow: In a case where the global storage's "StorageKeys.QUERY" is set outside of this function, a search is conducted and the `onSubmit` wouldn't be fired.

However, since that case involves a search as a side effect from a separate action, I think the concept of "that search was not submitted via this component" makes sense. The argument the other way is: technically the "debouncedSearch" function lives in this component/file; however, that seems like an implementation detail. What do you think?

cea2aj: I think it makes sense to only fire the hook when a search happens from within this component

tmeyer2115: Sure, the distinction makes sense to me!

tmeyer2115 (approved):

PR #1133, opened 2020-10-13
Add onClearSearch hook to the SearchBar component

alexisgrow:
TEST=manual

Add a simple console.log statement in the `onClearSearch` config option. Serve site locally, see expected output printed when I clicked the clear button.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,7 +456,9 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
-  }
+  },
+  // Optional, a callback invoked when the clear search button is clicked
+  onClearSearch: function() {}
 })
 ```
 
--- a/README.md
+++ b/README.md
@@ -457,8 +457,11 @@ ANSWERS.addComponent('SearchBar', {
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
   },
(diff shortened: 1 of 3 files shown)


oshi97 (approved):
lgtm!

README.md
```diff
@@ -456,7 +456,9 @@ ANSWERS.addComponent('SearchBar', {
     enabled: false,
     // Optional, the message in the alert. Defaults to the below
     message: "We are unable to determine your location"
-  }
+  },
+  // Optional, a callback invoked when the clear search button is clicked
+  onClearSearch: function() {}
```
tmeyer2115: This component has many config options and they're all on the top-level. There's no config hierarchy, save for the `geolocation` stuff. What would you think about adding a `customHooks` config option (or something like that) that's an `Object`? The `Object` would have an `onClearSearch` entry, `onSubmit`, etc. I really think that grouping like config options together helps with progressive disclosure of complexity and compartmentalization.

tmeyer2115 (commented):
It looks like the build is failing because the size of one of our assets has increased a bit. You should be able to get the build passing by upping the size limit of `answers-modern.min.js` to 110KB (technically, 106KB would work, but we round to the nearest 5KB).

README.md
alexisgrow: Yes totally agree! Updated to have a `customHooks` option.

oshi97 (approved):
lgtm!

PR #1134, opened 2020-10-15
Revert "Verify all translations are present in CircleCI"

cea2aj:
Reverts yext/answers#1124

This went in as a regular merge and it should have been a squash and merge


oshi97:
Reverting won't clean up the git history, if that's the goal. You'll probably want to manually squash all the commits (I would just reset --soft to right before the pr then make a new commit) and force push it

PR #1142, opened 2020-10-20
VerticalResultsCount: add noResults.visible config option

oshi97:
By default, the VerticalResultsCount should be hidden during
NoResults.

J=SLAP-767
TEST=manual

Test that when unspecified, will hide the component during no results
when set to false, will hide the component during no results
when set to true, will display the component during no results

### DIFF
--- a/src/ui/components/results/verticalresultscountcomponent.js
+++ b/src/ui/components/results/verticalresultscountcomponent.js
@@ -1,6 +1,7 @@
 import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
 import SearchStates from '../../../core/storage/searchstates';
+import ResultsContext from '../../../core/storage/resultscontext';
 
 export default class VerticalResultsCountComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
@@ -10,6 +11,8 @@ export default class VerticalResultsCountComponent extends Component {
         this.setState(verticalResults);
       }
     });
+
+    this._visibleForNoResults = (config.noResults || {}).visible;
   }
 
   static areDuplicateNamesAllowed () {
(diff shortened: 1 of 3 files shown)


src/ui/components/results/verticalresultscountcomponent.js
```diff
@@ -10,6 +11,8 @@ export default class VerticalResultsCountComponent extends Component {
         this.setState(verticalResults);
       }
     });
+
+    this._visibleForNoResults = (config.noResults || {}).visible;
```
tmeyer2115: Minor nit, but let's make sure this is an actual boolean using '!!' or something like that. I think relying on truthy/falsy can get confusing, especially when a variable's name suggests it is a boolean.

src/ui/templates/results/verticalresultscount.hbs
```diff
@@ -1,19 +1,21 @@
 {{#unless (eq total 0)}}
-  <div class="yxt-VerticalResultsCount"
-    aria-label="{{translate phrase='[[start]] through [[end]] of [[resultsCount]]' start=pageStart end=pageEnd resultsCount=total }}">
-    <div aria-hidden="true">
-      {{translate
-        phrase=
+  {{#unless (and isNoResults hiddenForNoResults)}}
```
tmeyer2115: Can we combine the two `unless` statements ?
```
{{#unless (or (eq total 0) (and isNoResults hiddenForNoResults)) }}
```
Alternatively, in the JS, you could just make a variable called `shouldDisplayCount` that encapsulates this compound check.

oshi97: sure

src/ui/components/results/verticalresultscountcomponent.js
oshi97: I'm still 50/50 about which I prefer so don't mind siding this way

tmeyer2115 (approved):

PR #1143, opened 2020-10-20
Pagination: remove extra DOM nodes

oshi97:
This commit removes the vast majority of extra DOM nodes
in the pagination component that are being display: none'd,
using bhaines impl at https://github.com/yext-pages/answers-mercy.pagescdn.com/pull/92
it works by not rendering pagination nodes that are being display: none'd
on both desktop and mobile

J=SLAP-698
TEST=manual

test that in the DOM explorer, we aren't rendering extra page numbers
test that I can still click through the page numbers, and run new searches,
  and the page numbers will still be there


alexisgrow (approved):
this LGTM!

PR #1144, opened 2020-10-21
highlightedvalue: add transform function parameter

creotutar:
Previously, in the SpellCheck component, if the suggested text contained
an escaped character (for example `'`), then we would escape the string
before applying our highlighting transformations (in this case, adding
`<strong>` markup in between the highlighted values).

This caused the indices to be misaligned (as, for example, ' would
become `&#x27;`). Thus the `<strong>` tags would interrupt the escaped
strings and they would not longer be escaped.

To fix this, we accept a transformFunction in the highlightedValue get
functions. This is a generalization of the escapeExpression function, so
any transformations can be used. We change the highlightValue hbs
partial to call the highlighted values with an escape function as the
transformFunction.

J=SLAP-696
TEST=manual

Test on a local HH Theme Jambo site using a local SDK, with a SpellCheck
component.

Try searching for "my locaton's", this should give you a string
Did you mean: my location's
And not
Did you mean: my location&#x27;s

You should also see the entire word "location's" be highlighted, not a
portion of the string.

Try searching for "my locaton&", this should give you a SpellCheck
string
Did you mean: my location&
Where location is highlighted, not the &

Tested on an npm run test, added a test case for the transformFn

### DIFF
--- a/src/core/models/highlightedvalue.js
+++ b/src/core/models/highlightedvalue.js
@@ -19,6 +19,16 @@ export default class HighlightedValue {
   }
 
   /**
+   * get highlighted value string
+   * @param {Function} transformFunction takes a string and returns the transformed string
+   * @returns {string} The value interpolated with highlighting markup and transformed in between
+   */
+  getWithTransformFunction (transformFunction) {
+    this._sortMatchedSubstrings();
+    return this.buildHighlightedValue(this.value, this.matchedSubstrings, transformFunction);
+  }
+
+  /**
    * get inverted highlighted value string
    * @returns {string}
    */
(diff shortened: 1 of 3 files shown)


src/core/models/highlightedvalue.js
```diff
@@ -18,6 +18,16 @@ export default class HighlightedValue {
     return this.buildHighlightedValue(this.value, this.matchedSubstrings);
   }
 
+  /**
+   * get highlighted value string
+   * @param {Function} transformFunction takes a string and returns the transformed string
+   * @returns {string} The value interpolated with highlighting markup and transformed in between
+   */
+  getWithTransformFunction (transformFunction) {
```
oshi97, line 26: do we still need the original get()? same thing with getInverted()

oshi97 (approved):
lgtm!

cea2aj:
In the description, is "my location's" supposed to be bold? I don't think the formatting got through:

> Try searching for "my locaton's", this should give you a string
> Did you mean: my location's
> And not
> Did you mean: my location's

cea2aj (approved):

PR #1148, opened 2020-10-21
Add Percy Integration

oshi97:
This commit adds a percy_snapshot github action,
that uploads snapshots of our acceptance test html fixtures
to percy to check for visual changes. A PERCY_TOKEN secret under
the yext/answers repo settings will be necessary for this to work
(tested by just hardcoding the env var). 

To explain why github actions were used instead of adding to our circle config,
our git strategy is more complicated than what Percy can handle without PR integration.
We have multiple "master" percy branches (support, release, develop,
and master), but percy only lets you set one "master" branch. Because of this, in
order to use Percy properly, we have to use it with PR integration turned in.
In order for a percy build to have PR integration, a PR for that commit has to
exist before the percy stuff is run. However, circleci does not have an "on
pull request hook", and builds are run immediately after pushing to the remote.
[apparently other people want this too](https://ideas.circleci.com/cloud-feature-requests/p/trigger-new-build-when-a-pull-request-is-opened)
This creates an unfortunate real life race condition, where if you don't create a PR before
the percy step is run, you will not get PR integration with your percy build. We could
add a hold step to our circleci builds, to give people time to create a PR, or maybe do
something with a webhook, but it seemed simpler to just create a github action for this instead.

J=SLAP-291
TEST=manual

tested that percy builds are run when a pull request is made, for any branch
tested that, for branches that don't match `dev/*`,  percy builds are automatically
run and approved. This is needed so that PRs will have base builds to compare to.
tested that merging in a `dev/*` branch into a non `dev/*` branch (like `develop` or `master`) triggers a 
percy build, which can form the base build for future PRs


cea2aj:
We are requiring percy_snapshots everywhere except on standard branches?

tests/acceptance/percysnapshot.js
```diff
@@ -0,0 +1,63 @@
+import { setupServer, shutdownServer } from './server';
```
tmeyer2115: Is this file a copy of the existing acceptancesuite? Was the reason for that so people don't generate snapshots if they run the tests locally? I figured we'd have one set of tests and we could pass a flag or something to generate snapshots (or not).

oshi97:
> We are requiring percy_snapshots everywhere except on standard branches?

I think I added the snapshots to all workflows, did I miss one of them?

tests/acceptance/percysnapshot.js
```diff
@@ -0,0 +1,63 @@
...
+  'Universal page',
+  async t => {
+    await percySnapshot(t, `universal page pre search`);
+
+    const searchComponent = UniversalPage.getSearchComponent();
+    await searchComponent.enterQuery('Tom');
+    await searchComponent.submitQuery();
+    await t.expect(Selector('.yxt-Results').exists).ok();
+
+    await percySnapshot(t, `universal page post search`);
```
tmeyer2115: Does this PR allow for different snapshots per minor? If not, I think we need to figure that bit out. Otherwise the tests are likely to get turned off or ignored. Minors may have different functionality. A later minor may have UI/UX that an earlier doesn't. If we tried to hold both minors to the same snapshot, we could run into trouble.

cea2aj:
> > We are requiring percy_snapshots everywhere except on standard branches?
> 
> I think I added the snapshots to all workflows, did I miss one of them?

The first deploy_branch (after deploy_canary) doesn't depend on percy_snapshots

tests/acceptance/percysnapshot.js
oshi97: originally, I assumed Percy was "dumb" and all it did was look at the name of the snapshot, and compare it to the most recent snapshot with the same name. In reality Percy claims it has smart logic for choosing the base branch to compare to, and will try to pick the most recent commit on the branch that had percy snapshots (I did some light testing and this seems to be the case?) So new prs on a support/v1.7 will be compared to the last commit on support/v1.7, and won't be compared to say the develop branch. Here's a link to the article I read https://docs.percy.io/docs/baseline-picking-logic

tests/acceptance/percysnapshot.js
oshi97: Snapshots aren't generated unless you run testcafe with a percy --exec. The issue was that the acceptance tests are run 4 times total, once per browser we care about. On the other hand, Percy only wants you to take snapshots once per build, so the percy_snapshots circleci job had to be separate from the regular acceptance_tests job. At that point it felt a little weird to run the entire acceptance suite again, when all that step cared about was taking snapshots, but I can update this PR to put the snapshots into acceptancesuite.js again.

oshi97: Actually I like being able to put snapshots in the main tests more, even if it means running them an extra time. The browserstack automate video is pretty garbage so this can also give some insight into failing tests, even if it makes the acceptance suite have a little more responsibility

oshi97: *Sorry I tried again and running acceptance tests that contain percySnapshot() calls even without `percy exec`ing them is fine EXCEPT in browserstack ie11, so it might be necessary to have the snapshots be separate

tests/acceptance/percysnapshot.js
oshi97: update: percy is not that smart, and looks like it has to be used with github PR integration for more our git flow. circleci doesn't let you run a build only when a pull request has been made, but github actions DOES let you do that, so I've updated this pr to use github actions for percy, instead.

oshi97:
could somebody try approving the percy build? Tom said he can't access the percy project, despite being in the browserstack org, so wanted to see whether this was a specific issue with Tom's account (maybe because he made his percy account before percy.io was acquired by browserstack?)

alexisgrow:
> could somebody try approving the percy build? Tom said he can't access the percy project, despite being in the browserstack org, so wanted to see whether this was a specific issue with Tom's account (maybe because he made his percy account before percy.io was acquired by browserstack?)

Approved!

.github/workflows/percy_snapshots.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [15.x]
```
tmeyer2115, line 15: Should we use node 12? I think that's what most people on the team are using.

.github/workflows/percy_snapshots.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build
+    - name: Percy Snapshots
+      uses: percy/exec-action@v0.3.1
+      with:
+        command: npx testcafe "chrome:headless" tests/acceptance/percysnapshots.js
```
tmeyer2115, line 28: Different browsers have different user agent stylings. Is it possible to generate snapshots for a build in additional browsers?

tests/acceptance/percysnapshots.js
```diff
@@ -0,0 +1,17 @@
+import { setupServer, shutdownServer } from './server';
+import percySnapshot from '@percy/testcafe';
+
+/**
+ * This file contains testcafe tests that take percy snapshots.
+ * These have to be separated out because taking percy snapshots
+ * on browserstack:ie, even without percy exec, will error.
+ */
+fixture`Universal search page works as expected`
```
tmeyer2115: Update fixture description

.github/workflows/percy_snapshots.yml
oshi97: percy takes care of that automatically, though only firefox and chrome are supported. you only need to run testcafe in one browser. running in multiple will actually have percy complain about "multiple snapshots with the same name in the same build"

.github/workflows/percy_snapshots.yml
oshi97: percy snapshots aren't really something we want to take locally, so I don't think there's any need to use an older version?

oshi97: tbh I just I want to use the newest, shiniest, fastest toys possible, and have an ingrained fear of being left behind. I don't mind downgrading to 14 or 12 (which are both still LTS)

.github/workflows/percy_snapshots.yml
oshi97: here's an example of percy complaining when you exec it on multiple browsers https://app.circleci.com/pipelines/github/yext/answers/4773/workflows/1278c236-94c8-49d6-bbb7-3ce0453d6ce0/jobs/13226

tmeyer2115 (approved):

cea2aj (approved):

PR #1162, opened 2020-10-27
SearchBar: Update to show clear button on external query

alexisgrow:
Since this PR (https://github.com/yext/answers/pull/1130), searches can be triggered from outside of the SearchBar component using a function `ANSWERS.search`. However, the SearchBar component was previously not adding the "clear button" icon when this happened. Updated the SearchBar component to show the clear button for external queries as well as internal ones.

TEST=manual,unit

Test with the Overlay prototype which uses the ANSWERS.search function. Also tested manually typing in the search input and see clear button appear.

### DIFF
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -184,6 +184,8 @@ export default class SearchComponent extends Component {
         }
         return;
       }
+      const clearButton = DOM.query(this._container, '.js-yxt-SearchBar-clear');
+      this._updateClearButton(clearButton, q);
 
       const queryTrigger = this.core.globalStorage.getState(StorageKeys.QUERY_TRIGGER);
       const resetPagination = this._verticalKey &&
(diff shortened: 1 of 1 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -759,4 +755,24 @@ export default class SearchComponent extends Component {
...
     DOM.query(this._container, this._inputEl).focus();
   }
+
+  /**
+   * Updates the Search inputs clear button based on the current input value
+   *
+   * @param {Element} button
+   * @param {string} input
+   */
+  _updateClearButton (button, input) {
```
oshi97: would it make sense to add a this.getClearButton() instead of passing in the button? It looks like we have DOM.query(this._container, '.js-yxt-SearchBar-clear'); in two different places already so maybe we could consolidate those?

cea2aj: What are your thoughts on `_updateClearButtonVisibility`?

oshi97 (approved):
lgtm!

PR #1163, opened 2020-10-29
Analytics: Add query source config option

alexisgrow:
Add an analytics query source config option to the ANSWERS.init function; this is passed in the universal and vertical search requests as "source". The analytics "source" key will be used to differentiate analytics data between integration types. Current types accepted by the back end are "OVERLAY" and "STANDARD"

J=SLAP-768
TEST=manual

Specify this option and conduct searches for both vertical and universal. Confirm with the back end analytics teams that they are seeing the source param come through for the request queryIds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ function initAnswers() {
     onUniversalSearch: function() {},
     // Optional, opt-out of automatic css variable resolution on init for legacy browsers
     disableCssVariablesPonyfill: false,
+    // Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD' or 'OVERLAY', defaults to 'STANDARD'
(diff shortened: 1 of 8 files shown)


tmeyer2115 (approved):
lgtm! More of a Theme thing, but what if one Jambo repo is used to host two different integrations of the same experience? Conceivably both may require different `querySources`

alexisgrow:
> lgtm! More of a Theme thing, but what if one Jambo repo is used to host two different integrations of the same experience? Conceivably both may require different `querySources`

Yes good question! From within the theme, we know whether we're inside the Overlay on page load so we can do this: https://github.com/yext/answers-hitchhiker-theme/pull/445

tmeyer2115 (approved):

PR #1165, opened 2020-10-29
Sort options focus state

cea2aj:
Add focus state to the sort options

This PR allows users to tab into the SortOptions component and navigate through the options with the arrow keys (The W3C spec). In order to make this possible, the SortOptions component had to be modified so that it did not re-render when a different sort option was selected.

Note: To get this feature to work on Safari, you must go to "Preferences -> Accessibility" and select "Press Tab to highlight each item on a webpage"

J=SLAP-727
TEST=manual, acceptance

Test that you can tab into the sort options component and navigate the options with the arrow keys on Chrome, Firefox, Safari, and IE11. Test that this change is compatible with the searchOnChange and the showReset options. Test this feature on an Answers site using the Hitchhiker's Theme and Collapsible Filters. Add an acceptance test that confirms that focus works on options.


.circleci/run_acceptance.sh
```diff
@@ -6,5 +6,7 @@ COMMIT_MSG_TITLE=$(git log -n 1 --pretty=format:%s)
 export BROWSERSTACK_TEST_RUN_NAME=$COMMIT_MSG_TITLE
 
 # Browserstack only allows for 2 threads to run at once, so we have to break up the browsers
-npx testcafe "browserstack:firefox,browserstack:chrome" tests/acceptance/acceptancesuite.js
-npx testcafe "browserstack:safari,browserstack:ie@11.0" tests/acceptance/acceptancesuite.js
\ No newline at end of file
+npx testcafe "browserstack:firefox" tests/acceptance/acceptancesuite.js
```
oshi97: oh did we run into clobbering issues?

src/ui/sass/modules/_SortOptions.scss
```diff
@@ -45,6 +45,18 @@
     margin-left: 5px;
   }
 
+  &-optionSelector {
+    &:focus {
+      outline: .0625rem auto black;
```
oshi97, line 50: could you double check whether the other focus styling uses 1px or .0625 rem? (I haven't checked myself)

tests/ui/components/filters/sortoptionscomponent.js
```diff
@@ -92,23 +91,25 @@ describe('sort options component', () => {
   it('renders correctly for default values', () => {
     const component = COMPONENT_MANAGER.create('SortOptions', defaultConfig);
     expect(component.selectedOptionIndex).toEqual(0);
+    component.handleSearchResults(); // The component will only render if there are search results
```
oshi97: I think this is a breaking change, previously sort options would render even before search results were loaded. We also want to avoid directly calling methods on the component if we can help it

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -104,8 +140,35 @@ export default class SortOptionsComponent extends Component {
...
+      if (this.selectedOptionIndex === optionIndex) {
+        optionEl && optionEl.setAttribute('checked', '');
+      } else {
+        optionEl && optionEl.removeAttribute('checked', '');
+      }
+    });
+  }
+
+  /**
+   * Set the visibility of the reset button based on this.showReset
```
oshi97: this is technically not setting the visibility, since visibility is a different css property than display

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,57 @@ export default class SortOptionsComponent extends Component {
     this.selectedOptionIndex = parseInt(this.core.globalStorage.getState(this.name)) || 0;
     this.options[this.selectedOptionIndex].isSelected = true;
     this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex < this._config.showMoreLimit;
-    this.showReset = this._config.showReset && this.selectedOptionIndex !== 0;
+    this.searchOnChangeIsEnabled = this._config.searchOnChange;
+    this.showResetIsEnabled = this._config.showReset;
+    this.showReset = this.showResetIsEnabled && this.selectedOptionIndex !== 0;
+    this.isSearchResults = false;
```
oshi97: could we call this isNoResults instead?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,57 @@ export default class SortOptionsComponent extends Component {
...
+          this.handleSearchResults();
+        } else {
+          this.handleNoSearchResults();
+        }
       }
     });
   }
 
-  setState (data = {}) {
+  handleSearchResults () {
```
oshi97: if possible, could we combine these two handler methods + the vertical results handler into a single one, maybe handleVerticalResultsUpdate?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,57 @@ export default class SortOptionsComponent extends Component {
     this.selectedOptionIndex = parseInt(this.core.globalStorage.getState(this.name)) || 0;
     this.options[this.selectedOptionIndex].isSelected = true;
     this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex < this._config.showMoreLimit;
-    this.showReset = this._config.showReset && this.selectedOptionIndex !== 0;
+    this.searchOnChangeIsEnabled = this._config.searchOnChange;
+    this.showResetIsEnabled = this._config.showReset;
+    this.showReset = this.showResetIsEnabled && this.selectedOptionIndex !== 0;
```
oshi97, line 28: would it make sense to put these statelike things (showReset, isSearchResults) into the setState, and reference them with this.getState('showReset')?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,57 @@ export default class SortOptionsComponent extends Component {
...
-     * This component listens to updates to vertical results, and sets its state to it when
-     * an update occurs.
-     * @type {string}
+     * This component should only render if there are search results, so it should listen
+     * to updates to vertical results and handle them accordingly.
      */
     this.core.globalStorage.on('update', StorageKeys.VERTICAL_RESULTS, verticalResults => {
-      if (verticalResults.searchState === SearchStates.SEARCH_COMPLETE) {
-        this.setState(verticalResults);
+      const isSearchComplete = verticalResults.searchState === SearchStates.SEARCH_COMPLETE;
```
oshi97, line 36: did the focus/blur listeners we touched on end up not working?

.circleci/run_acceptance.sh
oshi97: if you're having trouble with browserstack, you can also run tests on browserstack through your local terminal, without committing (will probably be much faster)

cea2aj: I was just seeing if it was safari or ie which was failing the tests. I'll test it locally and revert this change

tests/ui/components/filters/sortoptionscomponent.js
cea2aj: I was thinking that not rendering before a search (when there are no results) might better align with the spec, but if that's not what we want I can revert

src/ui/sass/modules/_SortOptions.scss
cea2aj: The ask was .0625rem, but the other focus styling uses 1px. I can double check with rose

src/ui/components/filters/sortoptionscomponent.js
cea2aj: the issue is that it would trigger a re-render when we don't want it to

src/ui/components/filters/sortoptionscomponent.js
cea2aj: The problem was that the blur listeners would execute sooner than expected which caused the component to lose the knowledge of which option was supposed to be focused during the next render. Reflecting back on this now, I think we might be able to store focused option before it is lost by using the component onUpdate() function. I can try that out if you'd prefer that approach

tests/ui/components/filters/sortoptionscomponent.js
oshi97: discussed offline, the "No Results" state doesn't include before search results have loaded, "No Results" is definitely not the best name

src/ui/components/filters/sortoptionscomponent.js
oshi97: oh, I was thinking that every time showReset or isSearchResults change we do a setState anyway (I think?). And if that's the case, you could just add them as properties to the state instead. Then any time you want to guard a setState you could reference `this.getState('showReset')` or `this.getState('isSearchResults')`. let me know if I'm completely off

(`this.getState('showReset')` won't trigger a re-render but I don't think that's what you meant)

src/ui/components/filters/sortoptionscomponent.js
oshi97: ah okay, I think I prefer what we have right now a little more assuming they both work equally well

src/ui/components/filters/sortoptionscomponent.js
cea2aj: We don't always want to call `setState()` every time `showReset` changes because then the component would re-render and the focus would be lost as the user navigates between the default and the second non-default option with the keyboard. It could work for changes to isNoResults, but I don't think it's worth refactoring just for that.

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,53 @@ export default class SortOptionsComponent extends Component {
...
+
+    if (isNoResults) {
+      this.setState(); // Call setState to trigger a re-render
+      return;
+    }
+
+    // If previously there weren't any search results, but now there are, re-render by
+    // calling setState. Otherwise, don't re-render in order to maintain focus on the
+    // selected sort option.
+    if (!isPreviousSearchResults) {
```
oshi97: does isPreviousSearchResults and this if block do anything? if isNoResults is true it won't reach this line, and if it's false the if statement will be skipped over

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,53 @@ export default class SortOptionsComponent extends Component {
...
       }
     });
   }
 
-  setState (data = {}) {
+  handleVerticalResultsUpdate (isNoResults) {
+    const isPreviousSearchResults = !this.isNoResults;
+    this.isNoResults = isNoResults;
+
+    if (isNoResults) {
```
oshi97: it looks like this handler will only be triggered if you do a search that goes to No Results, and not if you go from No Results to a regular search? if you search for "Frodo Client Delivery" on a rose's "people" vertical, which goes to No Results, then do a search for "Virginia", does the sort component become hidden for the first search and visible again for the second?

src/ui/components/filters/sortoptionscomponent.js
cea2aj: If `isNoResults` is false, this if statement isn't skipped over. The purpose of this block is to trigger a re-render when previously there were no results, and now there are. If there previously were results, and there still are results, then we don't want to re-render.

src/ui/components/filters/sortoptionscomponent.js
cea2aj: Just tested this. It hides when there are no results and it re-appears when there are results. I can modify this a little to try and make it clearer what is happening

src/ui/components/filters/sortoptionscomponent.js
oshi97: oh sorry I read !this.isNoResults as just !isNoResults, lgtm!

oshi97: would `wasNoResults` be clearer than `isPreviousSearchResults`?

oshi97: would it be the same behavior if we just checked whether the isNoResults state changed? instead of 2 if statements

cea2aj: ya, that works!

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -23,31 +23,53 @@ export default class SortOptionsComponent extends Component {
...
+
+    // If previously there weren't any search results, but now there are, re-render by
+    // calling setState. Otherwise, don't re-render in order to maintain focus on the
+    // selected sort option.
+    if (!isPreviousSearchResults) {
+      this.setState();
+    }
+  }
+
+  setState () {
```
oshi97: I think we should keep the data param, just to be consistent with every other component.

cea2aj: ah, I wasn't aware that was the convention. I'll add it back

oshi97 (approved):

PR #1168, opened 2020-10-30
Search: update search hook placement

alexisgrow:
Move custom hook from `onQuerySubmit` to `search`. This means that the hook is invoked whenever a search is conducted. Since the hook has not been in any releases, we're free to change its name without any backwards compatibility issues. Changed from `onSubmit` to `onConductSearch`.

TEST=manual,unit

Test with Overlay locally. Confirm that this hook is invoked when searches are conducted via typing in the search bar as well as with the ANSWERS.search function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,9 +461,8 @@ ANSWERS.addComponent('SearchBar', {
   customHooks: {
     // Optional, a callback invoked when the clear search button is clicked
     onClearSearch: function() {},
-    // Optional, a function invoked when a search is submitted via this component. The users search
-    // is passed in as a string
-    onSubmit: function(searchTerms) {}
+    // Optional, a function invoked when a search is conducted. The search terms are passed in as a string
+    onConductSearch: function(searchTerms) {}
   },
   // Optional, options to pass to the autocomplete component
   autocomplete: {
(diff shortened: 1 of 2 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -602,6 +600,7 @@ export default class SearchComponent extends Component {
     this._throttled = true;
     setTimeout(() => { this._throttled = false; }, this._searchCooldown);
 
+    this.customHooks.onConductSearch(query);
```
creotutar: Why here and not in the `search` function?

alexisgrow: Ah, good point 👀

nanhu95 (approved):
LGTM!

README.md
```diff
@@ -461,9 +461,8 @@ ANSWERS.addComponent('SearchBar', {
   customHooks: {
     // Optional, a callback invoked when the clear search button is clicked
     onClearSearch: function() {},
-    // Optional, a function invoked when a search is submitted via this component. The users search
-    // is passed in as a string
-    onSubmit: function(searchTerms) {}
+    // Optional, a function invoked when a search is conducted. The search terms are passed in as a string
+    onConductSearch: function(searchTerms) {}
```
creotutar, line 467: This is probably why the wording was changed, but it's OK that this is called when we do a defaultInitialSearch in `ANSWERS.init` right? It's wanted behavior that this custom hook in the SearchBar component is called when that happens?

alexisgrow: Yes exactly! It should be called whenever a search is conducted - default initial search, ANSWERS.search or a search from typing in the search input!

creotutar (approved):
✨ ✨ ✨ LGTM

tmeyer2115 (approved):

PR #1171, opened 2020-10-30
Run firefox and chrome acceptance tests in headless mode

oshi97:
Currently we're capped at 2 parallel threads for browserstack automate.
This makes it really annoying when multiple people are trying to debug
acceptance tests.

If we run the chrome and firefox acceptance tests in headless mode in
the circleci container, we can speed up our builds a lot. Testcafe
only supports firefox and chrome in headless mode. This will also
cut our browserstack usage in half. On the flip side we wont get
the browserstack automate video for chrome/firefox

TEST=auto

see tests run properly in the circleci build


tmeyer2115 (approved):
lgtm. I think we should petition to upgrade our account so that we have more than 2 parallel threads. Even with this change, seems like we may run into some issues when we have 5 people hammering away at the SDK. Additionally, I think having the automate videos in every browser is helpful. They provide a good starting point for debugging. But, speeding up our builds to better enable development is the more important thing now. cc: @bryanreed

PR #1176, opened 2020-11-05
Update the acceptance tests to use the new Slapshot account.

tmeyer2115:
Previously, all of our acceptance tests were using Rose's test
account. This isn't ideal since Rose frequently makes changes
to her account. These changes have the potential to break our tests.
To mitigate this, we have created the Slapshot Test account. We
control all aspects of the account, so data won't be changed out
from under us unexpectedly, breaking our tests. This PR points the
tests to the new account.

J=SLAP-348
TEST=manual

Made sure all the acceptance tests passed.

### DIFF
--- a/tests/acceptance/fixtures/html/facets.html
+++ b/tests/acceptance/fixtures/html/facets.html
@@ -21,9 +21,9 @@
 
   <script>
     ANSWERS.init({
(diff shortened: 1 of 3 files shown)


alexisgrow (approved):
🎊🎊

nanhu95 (approved):
LGTM!

PR #1178, opened 2020-11-06
gulp: Add external sourcemaps for minified content

creotutar:
We add sourcemaps to the minified content for all library assets that
are minified (Legacy IIFE and UMD, Modern). This is a response to
feedback that it is hard to debug code that is using answers.min.js when
an error occurs.

As a result, we now get errors in a Jambo site that say the line number
in answers.min.js with the error, instead of all on line 1.

The dist folder now has three new files:
        dist/answers-modern.min.js.map
        dist/answers-umd.min.js.map
        dist/answers.min.js.map

We store the sourcemaps in a separate file instead of inline so that the
sourcemaps are only loaded when dev tools are open.

J=SLAP-825
TEST=manual

In a Jambo HH Site, try purposefully creating errors. Change `profile`
to `profil` in the job card and build. The error in the browser dev
tools should show a trace with line numbers other than 1.

### DIFF
--- a/conf/gulp-tasks/bundle/minifytaskfactory.js
+++ b/conf/gulp-tasks/bundle/minifytaskfactory.js
@@ -3,6 +3,7 @@ const getBundleName = require('../utils/bundlename');
 const { src, dest } = require('gulp');
 const rename = require('gulp-rename');
 const uglify = require('gulp-uglify-es').default;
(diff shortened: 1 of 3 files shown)


nanhu95 (approved):
LGTM!

tmeyer2115 (approved):

alexisgrow (approved):

PR #1180, opened 2020-11-06
Add README for ANSWERS.registerTemplate

creotutar:
J=SLAP-649
TEST=auto

npm test

### DIFF
--- a/README.md
+++ b/README.md
@@ -2096,8 +2096,8 @@ ANSWERS.addComponent('SearchBar', {
 })
 ```
 
-The SDK also offers an `ANSWERS.registerTemplate` function. This is to register
-a template string to any given template name in our template bundle.
+The SDK also offers an `ANSWERS.registerTemplate` function. This will map
+a template string to an entry in the Answers handlebars renderer.
 
 ```js
   /**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2096,6 +2096,37 @@ ANSWERS.addComponent('SearchBar', {
...
+   * Compile and add a template
+   * @param {string} templateName The unique name for the template
+   * @param {string} template The handlebars template string
+   */
+  registerTemplate (templateName, template)
+```
+
+If the template name does not exist, a new template name is created. If
+it does exist, it overrides the current template for the template name. This
+allows you override default template names.
```
alexisgrow, line 2115: nit: What do you think about describing these as entries in a registry/dictionary keyed off of template name? I think it's already currently implied, but unsure if we should explicitly say it

README.md
```diff
@@ -2096,6 +2096,37 @@ ANSWERS.addComponent('SearchBar', {
 })
 ```
 
+The SDK also offers an `ANSWERS.registerTemplate` function. This is to register
+a template string to any given template name in our template bundle.
```
alexisgrow: Could "in our template bundle" be swapped out for phrasing that describes registering the template with the Handlebars compiler rather than the SDK's bundle itself?

alexisgrow (commented):
😍😍 read me updates, a few comments!

alexisgrow (approved):
this LGTM!

PR #1183, opened 2020-11-12
package.json: Add prepublishOnly and postpublish scripts for README

creotutar:
We want to publish to NPM for the first time. There is a problem because
NPM uses the root-level README.md as its source for the package README.
Our current root-level README has langauge that is specific to the
GitHub repository. Thus, we want to provide a shorter version of the
README to NPM that will link to the current GitHub README.

To do this, before any time we do an `npm publish`, we want to switch
out the root-level README with an NPM README, and then postpublish,
switch things back to the way they are currently.

J=SLAP-812
TEST=manual

Tested on a test NPM package and successfully pushed a changed README.md
to the NPM registry and saw it reflected there.


conf/npm/README.md
```diff
@@ -0,0 +1,4 @@
+This is a javascript library that helps you build search experiences on top of
```
oshi97, line 1: could you clarify what you mean by "language specific to github" nothing really stood out to me but I'm also illiterate

package.json
```diff
@@ -81,7 +81,9 @@
     "acceptance": "testcafe safari,chrome tests/acceptance/acceptancesuite.js",
     "size": "size-limit",
     "fix": "semistandard --fix",
-    "extract-translations": "gulp extractTranslations"
+    "extract-translations": "gulp extractTranslations",
```
oshi97, line 84: are these all the changes we need to publish to npm? I thought we might need to update `main` in the package.json as well, and maybe rebuild the sdk before publish?

creotutar: yeee, coming to you hot in another `dev` branch
for your latter thought, was thinking of a spruce-tools script, wdyt?

conf/npm/README.md
creotutar: sure,
from @tmeyer2115 :
```
The Answers Javascript API Library does not need to be installed locally. Instead, it can be used with script tags on a webpage. The instructions below explain how to do this; they will walk you through adding the Answers stylesheet, JS library, and an intialization script to an HTML page. After doing this, you can view your page in the browser.
```

where it mentions not being installed locally, though I could see the case where we are saying "it does not _need_ to be installed locally, you can also do this instead"

also got feedback that the README as it exists currently may be too large for NPM

package.json
oshi97: I think the scripts in spruce-tools are for things we use manually, would we need to do anything besides an npm run build?

conf/npm/prepublishOnly.sh
```diff
@@ -0,0 +1,6 @@
+# This is run before a publish using `npm publish`
+# We want to switch out the current (GitHub) README
+# with the short README we have for NPM.
+
+mv ./README.md ./README.tmp
+cp ./ci/npm/README.md ./README.md
```
oshi97: I don't see this ci folder anywhere, am I missing something?

oshi97 (commented):
+1 from me! will ask tom to take a look too

conf/npm/postpublish.sh
```diff
@@ -0,0 +1,4 @@
+# This is after before a publish using `npm publish`
```
tmeyer2115: "This is after a publish using..."

package.json
tmeyer2115: Supposedly, we want this item to take care of https://yexttest.atlassian.net/browse/CPB-247. We don't want Consulting on dist tags any more. I'm not sure I follow the comment of CPB-247. Rose wants the package to do an `npm run build` after it's installed. That doesn't make a whole lot of sense to me. It seems like we should just publish what's in the `dist` directory to npm. We should follow up with her on that.

creotutar: `dist` is being added

creotutar: (in another PR)

creotutar: and there will be a new manual step to increment the `package.json` version and commit before our tagging process

tmeyer2115 (approved):

PR #1184, opened 2020-11-12
package.json: Changes for public npm release

creotutar:
Change the license to match the LICENSE file.
Change the main file to the dist umd file, as that is the file most
likely to be imported when including this package.

Add the repository, bugs, and homepage fields to point to the current
GitHub repository.

Add a .npmignore for the registry that is exactly like the .gitignore
except it allows the dist folder. We allow the dist on publish so that
we do not need to build on postinstall. The dist files are available as
soon as installation is done.

J=SLAP-812
TEST=manual

Tested on a test NPM package
Saw all changes reflected on the test package in the registry after
pushing.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,2 @@
 node_modules/
-app/
 **/.DS_Store
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+node_modules/
+app/
+**/.DS_Store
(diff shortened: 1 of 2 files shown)


alexisgrow (approved):
LGTM!

.npmignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
+app/
```
oshi97: I don't think we need app/ here (we should probably also remove it from the gitignore since it doesn't exist anymore?)

oshi97 (approved):
lgtm!

PR #1185, opened 2020-11-13
Create the new CoreAdapter wrapper.

tmeyer2115:
This PR creates the `CoreAdapater`, which will use the new Core library to
perform all network requests. The adapter is essentially the SDK's old
`Core` class, with the addition of an `init` method. This method calls the new
`provideCore` function to get an instance of the library. Note that the Core
library is not added to the SDK's package.json. This is because it has not yet
been published to npm and the Github repo does not contain a dist build.

You can point the SDK at your local version of answer-core by adding a line
like this to the package.json dependencies:

`"@yext/answers-core": "file:../answers-core"`

This will enable you to test your answers-core changes against the SDK. Once
answers-core has been published to npm, we will commit the dependency to the
package.json. We will also be able to get rid of the answers-core specific
things added to bundle.js

I initially considered having answers-umd invoke `provideCore` and then
instantiate `CoreAdapter` after `_loadAsyncDependencies`. This would not have
worked, however. It would mean that `ANSWERS` instance methods, such as
`setSessionsOptIn` could not be called until `onReady`. This would be a
breaking change.

J=SLAP-836
TEST=manual

Pointed the SDK to my local answers-core build. Created a bundle of the SDK
and made sure that `CoreAdapter` was initialized with a Core library instance.

### DIFF
--- a/src/core/core-adapter.js
+++ b/src/core/core-adapter.js
@@ -130,18 +130,18 @@ export default class CoreAdapter {
   /**
    * Initializes the {@link CoreAdapter} by providing it with an instance of the Core library.
    */
-  init() {
+  init () {
     const params = { apiKey: this._apiKey, experienceKey: this._experienceKey };
     return provideCore(params).then(coreLibrary => {
       this._coreLibrary = coreLibrary;
-    })
+    });
   }
 
   /**
    * @returns {boolean} A boolean indicating if the {@link CoreAdapter} has been
    *                    initailized.
    */
-  isInitialized() {
+  isInitialized () {
     return !!this._coreLibrary;
   }
 
--- a/src/core/core-adapter.js
+++ b/src/core/core-adapter.js
@@ -1,4 +1,5 @@
 /** @module Core */
+import provideCore from '@yext/answers-core';
 
 import SearchDataTransformer from './search/searchdatatransformer';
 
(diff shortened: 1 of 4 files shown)


tmeyer2115:
Ignore the broken build here. That is expected because no dependency was added to the `package.json`. That will be done by you during your local testing.

src/core/core-adapter.js
```diff
@@ -125,6 +127,24 @@ export default class Core {
     this.onVerticalSearch = config.onVerticalSearch || function () {};
   }
 
+  /**
+   * Initializes the {@link CoreAdapter} by providing it with an instance of the Core library.
+   */
+  init () {
+    const params = { apiKey: this._apiKey, experienceKey: this._experienceKey };
```
cea2aj, line 134: Do we want to add the locale and possibly the configurationLabel ( 'STAGING' | 'PRODUCTION' ) here as well?

tmeyer2115: I was going off what is in master. In master, it looks like `provideCore` only takes in `apiKey` and `experienceKey`. Is that the incorrect branch of `answers-core`?

cea2aj: oh, I was thinking this was pointed at feature/search-service. This is code is correct for master, and I think it's fine to leave this pointed at master for now

cea2aj (approved):

PR #1191, opened 2020-11-18
Adapt universal search

cea2aj:
Add adapters that convert the UniversalSearchResponse from answers-core into an SDK `UniversalResults` model

The commented-out lines in the core-adapter represent work that will be completed in future PRs. Future PRs will also add tests for these adapters.

In order to use this, "@yext/answers-core": "file:../answers-core", needs to be added to the package.json and answers-core needs to be in the same directory as answers. The tests are failing in percy and circle because they do not have access to the core library.

The truncate function is the same as the truncate function in `ResultsFactory`. In future PRs, I expect to remove `ResultsFactory` completely

J=SLAP-878
TEST=manual

Build a sample site and perform universal searchers. Compare the results to the same site on the master branch and confirm the results are the same. Also manually inspect the resulting objects to verify correctness.


src/core/adapters/adapthighlightinfoarray.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * {
+ *   name: '<strong>Yext</strong>',
+ *   description: {
+ *     featured: '<strong>Yext</strong> is the offical answers company'
+ *   }
+ * }
+ */
+export default function adaptHighlightInfoArray (highlightInfoArray) {
+  if (highlightInfoArray === undefined || highlightInfoArray.length === 0) {
```
alexisgrow: Do we want to check if this is falsy since L24 assumes it's truthy?

src/core/adapters/adaptsimplefilter.js
```diff
@@ -0,0 +1,15 @@
+import Filter from '../models/filter';
+
+/**
+ * Constructs an SDK Filter model from an answers-core SimpleFilter model
+ *
+ * @param {SimpleFilter} filter from answers-core
+ * @returns {@link Filter}
+ */
+export default function adaptSimpleFilter (filter) {
+  return new Filter({
```
oshi97: Would it make sense to have a separate item to completely rip out the current sdk filter model? Ideally we could do it here but that also would bleed over into vertical search. It feels a little weird to convert from the raw filter shape to the new core's shape and then back to the original shape here

src/core/adapters/adapthighlightinfoarray.js
```diff
@@ -0,0 +1,62 @@
...
+    return {};
+  }
+
+  return highlightInfoArray.reduce((highlightedData, highlightInfo) => {
```
alexisgrow: With highlightedData, highlightInfo and highlightedValue - its a little tricky to tell what should be stored in each one and how they differ. Here and in the `adaptResult` function, would it be possible to rename?

src/core/adapters/adaptresults.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Constructs an SDK Result from an answers-core Result
+ *
+ * @param {Result} result from answers-core
+ * @returns {@link Result}
+ */
+export default function adaptResult (result) {
+  const highlightedData = adaptHighlightInfoArray(result.highlightedValues);
+  const details = highlightedData.description || result.description;
+  const truncatedDetails = truncate(details);
```
oshi97: what's the truncation for?

src/core/core-adapter.js
```diff
@@ -312,35 +313,36 @@ export default class CoreAdapter {
...
+        query: queryString,
+        coordinates: this.globalStorage.getState(StorageKeys.GEOLOCATION),
         skipSpellCheck: this.globalStorage.getState('skipSpellCheck'),
         queryTrigger: queryTrigger,
         sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN).value,
         context: context,
-        referrerPageUrl: referrerPageUrl,
-        querySource: this.globalStorage.getState(StorageKeys.QUERY_SOURCE)
+        referrerPageUrl: referrerPageUrl
+        // querySource: this.globalStorage.getState(StorageKeys.QUERY_SOURCE)
```
alexisgrow, line 322: This is used for analytics to track whether or not queries come from the Overlay - so we'll want to add this back in eventually but I'm thinking it's not in the answers-core lib yet?

alexisgrow (commented):
this is looking great so far - some minor comments

src/core/adapters/adapthighlightinfoarray.js
```diff
@@ -0,0 +1,62 @@
...
+    const highlightedValue = new HighlightedValue()
+      .buildHighlightedValue(highlightInfo.value, highlightInfo.matchedSubstrings);
```
oshi97: are we intentionally not passing things into the constructor and directly calling buildHiglightedValue()?

src/core/adapters/adaptverticalresults.js
```diff
@@ -0,0 +1,22 @@
...
+import adaptResult from './adaptresults';
+
+/**
+ * Constructs an SDK Section model from an answers-core VerticalResult array
+ *
+ * @param {VerticalResults[]} verticalResults
+ * @param {Object<string, string>} urls keyed by vertical key
+ * @returns {@link Section}
+ */
+export default function adaptVerticalResults (verticalResults, urls) {
```
alexisgrow: To make it easier to use the "adaptVerticalResults" for both universal and vertical responses, what do you think about changing this to take in a single VerticalResults rather than an array? I think the mental model could/should be that universal responses have a list of VerticalResults, so it would make sense that they'd need to call the adapt function on each one

src/core/adapters/adapthighlightinfoarray.js
cea2aj: Sure. I'm going to go through these adapters and make my falsy checks a little more robust

src/core/adapters/adapthighlightinfoarray.js
cea2aj: Maybe` highlightedValue` becomes `highlightedString`, and `highlightedData` becomes `highlightedFieldMap`?

src/core/core-adapter.js
cea2aj: It's not in the answers-core yet but I can put up a PR

src/core/adapters/adaptresults.js
cea2aj: The SDK currently truncates the descriptions of results so that they aren't super long

src/core/adapters/adapthighlightinfoarray.js
cea2aj: ya, the `HighlightedValue` class is a little wonky. I wasn't sure if refactoring it was outside the scope of this PR

src/core/adapters/adaptverticalresults.js
cea2aj: ya, that works!

src/core/adapters/adaptsimplefilter.js
cea2aj: Completely redoing the filters in the SDK is a larger change that makes sense to do in a separate item. I think we will handle that work when we rewrite the components in preact. For now, converting the data twice will let us get this out faster

src/core/utils/strings.js
```diff
@@ -0,0 +1,31 @@
+/**
```
oshi97, line 1: could we update resultfactory.js to use this truncate method as well?

cea2aj: That all is going to get deleted in a future PR, so no need

src/core/models/result.js
```diff
@@ -101,4 +104,28 @@ export default class Result {
...
+  /**
+   * Constructs an SDK Result from an answers-core Result
+   *
+   * @param {Result} result from answers-core
+   * @returns {@link Result}
+   */
+  static fromCore (result) {
+    const highlightedData = HighlightedFieldMap.fromCore(result.highlightedValues);
+    const details = highlightedData.description || result.description;
+    const truncatedDetails = truncate(details);
```
alexisgrow, line 117: Could truncating the highlighted data result in potentially invalid HTML if one of the `<strong>` tags is split?

cea2aj: I can truncate before the highlighting is applied just in case

cea2aj: After testing further, I learned that this works as-is because the truncate method splits on spaces, and the highlighted data seems to highlight each word rather than highlighted sequences of words which may contain spaces. The SDK currently truncates after applying highlighting, so it seems safe to do for now. The risk is if the backend changes and highlights a value which contains a space rather than highlighting each word, it may break

src/core/models/appliedqueryfilter.js
```diff
@@ -0,0 +1,29 @@
+import Filter from './filter';
+
+export default class AppliedQueryFilter {
```
tmeyer2115, line 6: Add some JSDoc

src/core/models/highlightedfieldmap.js
```diff
@@ -0,0 +1,76 @@
...
+   * value strings.
+   *
+   * @param {Object<string, string|object>} data Keyed by fieldName. It may consist of nested fields
+   *
+   * Example object:
+   *
+   * {
+   *   name: '<strong>Yext</strong>',
+   *   description: {
+   *     featured: '<strong>Yext</strong> is the offical answers company'
```
tmeyer2115, line 18: Is this the same structure as the object output by `ResultFactory.computeHighlightedData`?

src/core/models/result.js
```diff
@@ -101,4 +104,28 @@ export default class Result {
...
+
+    return new Result({
```
tmeyer2115, line 119: Does the Core library's search response have a notion of `source`? The `source` could be KG, Google CSE, Algolia, etc. The `source` may impact how the SDK's `Result` is constructed. I notice I left it out of the tech design 👎 , but I think it's something we need to account for.

cea2aj: The `answers-core` handles the transformations from the various sources. I think it may be worth adding `source` as a property to the core however since the properties on the rawData field will depend on the source. That change wouldn't impact the SDK.

src/core/models/highlightedfieldmap.js
cea2aj: yes

tmeyer2115 (approved):

Issue #570, opened 2020-03-16
400 Response in Q&A Endpoint Should Display Error


rosiegrant:
**Current Behavior:** When the q&a endpoint returns an error (400), we still display a successful response on the front end.

**Expected Behavior:** If an error is returned from the API, we should display our error state (which already exists here, https://github.com/yext/answers/blob/8af0620afc6a8fab42b0e113414c73fe546e3b60/src/ui/templates/questions/questionsubmission.hbs#L44)

**To Replicate:** 
- In your test account, add an incorrect `entityId` to the QASubmission config
- Submit a question using the frontend component
- You'll see a "thanks for your question", despite the fact that the request 400's and no question was submitted. 

<img width="779" alt="Screen Shot 2020-03-16 at 6 49 46 PM" src="https://user-images.githubusercontent.com/22160788/76806118-f7f5a080-67b6-11ea-8c10-a021cd428d26.png">

cea2aj:
Solved by https://github.com/yext/answers-search-ui/pull/1505

PR #1192, opened 2020-11-18
Use extracted core question submission service

oshi97:
This commit removes the previous question submission
service code, and uses the new extracted core one.

The qa submission component does not use the data
returned from the endpoint, and the only data returned
is the uuid of the request and any errors.
Currently there is an issue where the component will
ignore any errors in the response json, as long
as no error is thrown within the javascript.

https://github.com/yext/answers/issues/570

This will be addressed in a later pr, once we decide
how to do errors in the new extracted core.

J=SLAP-838
TEST=manual

tested that I can use the qa submission component
added a console.log to the extracted core for sanity's sake


cea2aj (approved):
lg

PR #1193, opened 2020-11-20
Use CoreAdapter and answers-core library for vertical searches

alexisgrow:
Update the SDK to use the answers-core library as the network layer. Update the SearchDataTransformer and its test cases. Removed old "from" static methods from the models if they were unused as a result of this change.

TEST=manual,unit

Create a simple page with SearchBar and VerticalResults components. Log the output in the console and see the models are the expected shape. Also see the results appear properly in the VerticalResults DOM. Also ran the unit tests with `npm run test` and confirmed all tests pass.


alexisgrow:
The build failures on this PR are expected since CircleCI doesn't have the `answers-core` library. I ran the unit tests and manually tested locally.

oshi97 (approved):
lgtm!

src/core/models/alternativeverticals.js
```diff
@@ -13,15 +13,15 @@ export default class AlternativeVerticals {
...
-  static from (response, formatters) {
-    const alternativeVerticals = response.alternativeVerticals;
-    if (!alternativeVerticals || !alternativeVerticals.modules) {
+  static from (alternativeVerticals) {
+    if (!alternativeVerticals || alternativeVerticals.length < 0) {
       return new AlternativeVerticals();
     }
 
-    return new AlternativeVerticals(Section.from(alternativeVerticals.modules, {}, formatters));
+    return new AlternativeVerticals(alternativeVerticals.map(alternativeVertical => {
```
cea2aj, line 23: Is it possible to replace this arrow function with just `VerticalResults.fromCore`?

alexisgrow: Not really, it is functionally different since `VerticalResults.fromCore` has more than one argument and we'd like the defaults applied for every arg except the first

tests/fixtures/responseWithNoResults.json
```diff
@@ -0,0 +1,2228 @@
...
+    "results": [],
+    "resultsCount": 0,
+    "source": "KNOWLEDGE_MANAGER"
+  },
+  "queryId": "94c572c7-8fa5-4435-a956-7c13667af173",
+  "searchIntents": [],
+  "facets": [],
+  "locationBias": {
+    "latitude": 43.6349,
+    "longitude": -116.2023,
```
cea2aj: I would maybe fudge these numbers a little bit if you haven't already

cea2aj:
For the testing fixtures, it is possible to remove some results so that they are more manageable?

src/core/models/alternativeverticals.js
```diff
@@ -13,15 +13,15 @@ export default class AlternativeVerticals {
...
    * Create alternative verticals from server data
-   * @param {Object} response The server response
-   * @param {Object.<string, function>} formatters The field formatters to use
+   * @param {Object[]} alternativeVerticals
    */
-  static from (response, formatters) {
-    const alternativeVerticals = response.alternativeVerticals;
-    if (!alternativeVerticals || !alternativeVerticals.modules) {
+  static from (alternativeVerticals) {
+    if (!alternativeVerticals || alternativeVerticals.length < 0) {
```
tmeyer2115: Should this be `!alternativeVerticals || alternativeVerticals.length === 0`?

src/core/search/searchdatatransformer.js
```diff
@@ -30,42 +30,46 @@ export default class SearchDataTransformer {
     };
   }
 
-  static transformVertical (data, formatters, verticalKey) {
```
tmeyer2115, line 33: Was the `formatters` logic moved elsewhere? While we don't want any concept of `formatters` in the Core library, this will still need to work in the SDK.

tests/fixtures/responseWithNoResults.json
alexisgrow: hahah true, i think this is in Boise somewhere! I'll do a passthrough and replace things with cleaned data

src/core/search/searchdatatransformer.js
```diff
@@ -30,42 +30,46 @@ export default class SearchDataTransformer {
...
-      [StorageKeys.INTENTS]: SearchIntents.from(response.searchIntents),
+      [StorageKeys.VERTICAL_RESULTS]: VerticalResults.fromCore(
+        response.verticalResults, {}, resultsContext, verticalKey),
+      [StorageKeys.DYNAMIC_FILTERS]: DynamicFilters.from(response.facets, resultsContext),
+      [StorageKeys.INTENTS]: SearchIntents.fromCore(response.searchIntents),
       [StorageKeys.SPELL_CHECK]: SpellCheck.from(response.spellCheck),
-      [StorageKeys.ALTERNATIVE_VERTICALS]: AlternativeVerticals.from(response, formatters),
-      [StorageKeys.LOCATION_BIAS]: LocationBias.from(response.locationBias)
+      [StorageKeys.ALTERNATIVE_VERTICALS]: AlternativeVerticals.from(response.alternativeVerticals),
+      [StorageKeys.LOCATION_BIAS]: LocationBias.fromCore(response.locationBias)
```
tmeyer2115, line 54: I think we should be consistent with the naming of the `from/fromCore` methods. Either they are all called `fromCore` or all called `from` (and it's implicitly understood that we're translating from Core library models).

src/core/models/alternativeverticals.js
alexisgrow: Oh, yes 🙈

src/core/search/searchdatatransformer.js
cea2aj: That's on me! I suspect Alexis was following what I did in the universal adapter which was to leave them out. I planned on getting them in a future PR, but that got put on the backburner because of the theme release. At this point, we can get the formatters in with this PR (and update the universal search as well), or we can do it in separate PR

alexisgrow: Exactly, since the "ResultsFactory.from" was removed in the feature branch, the formatters arg was unused. I can add it back here, but we'll likely need a new PR to actually apply the formatters to the results

src/core/search/searchdatatransformer.js
alexisgrow: Yeah that sounds good to me! At this point, it looks like there is less code change if I rename all the "from"'s to "fromCore". Updating!

src/core/search/searchdatatransformer.js
alexisgrow: Added it back with a "TODO" in the Section's constructor!

cea2aj (approved):
🔥

PR #1194, opened 2020-11-20
Additonal universal search adapters

cea2aj:
This PR creates the models that need to be created during a universal search that do not include the universal search results themselves

These models include: navigation, direct answer, search intents, spell check, and location bias.

A future PR will include tests

J=SLAP-878
TEST=manual

Perform searches on an answers site build with master and one built with this PR. Observe the created models and confirm that they are the same shape and include the same data


src/core/adapters/adaptlocationbias.js
```diff
@@ -0,0 +1,20 @@
...
+      accuracy: 'UNKNOWN'
+    });
+  }
+  return new LocationBias({
+    accuracy: locationBias.method,
+    latitude: locationBias.latitude,
+    longitude: locationBias.longitude,
+    locationDisplayName: locationBias.displayName
+  });
+}
```
alexisgrow: Would it make sense to use the existing LocationBias.from() as a substitute for this adapter?

src/core/core-adapter.js
```diff
@@ -325,24 +328,32 @@ export default class CoreAdapter {
         // querySource: this.globalStorage.getState(StorageKeys.QUERY_SOURCE)
       })
       .then(data => {
+        const universalResults = adaptUniversalSearchResponse(data, urls);
+        const navigation = Navigation.fromVerticalResults(data.verticalResults);
+        const directAnswer = adaptDirectAnswer(data.directAnswer);
+        const searchIntents = SearchIntents.from(data.searchIntents);
+        const spellCheck = SpellCheck.from(data.spellCheck);
+        const locationBias = adaptLocationBias(data.locationBias);
```
alexisgrow: This is a stylistic call so up to you, what are your thoughts on the pattern where we have an adapter like the current `.then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))` that turns an adapted object with this data vs declaring these constants?

src/core/core-adapter.js
```diff
@@ -325,24 +328,32 @@ export default class CoreAdapter {
...
 
         this.globalStorage.delete('skipSpellCheck');
         this.globalStorage.delete(StorageKeys.QUERY_TRIGGER);
-        /*
+
         const exposedParams = this._getOnUniversalSearchParams(
-          data[StorageKeys.UNIVERSAL_RESULTS].sections,
+          data.verticalResults,
           queryString);
+        console.log(exposedParams);
```
alexisgrow: Could this be removed?

src/core/models/navigation.js
```diff
@@ -16,4 +16,14 @@ export default class Navigation {
     }
     return new Navigation(nav);
   }
+
+  /**
+   * Constructs a Navigation component from an answers-core VerticalResults array
```
alexisgrow: nit: This constructs the model, not component, right?

src/core/adapters/adaptdirectanswer.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns {@link DirectAnswer}
+ */
+export default function adaptDirectAnswer (directAnswer) {
+  if (!directAnswer) {
+    return new DirectAnswer();
+  }
+
+  const relatedResult = directAnswer.relatedResult || {};
+
+  return new DirectAnswer({
```
oshi97: would it be possible to update direct answer to just use the new core's model? I'm pretty sure direct answer is universal results only. Also down for making it a separate pr!

src/core/adapters/adaptlocationbias.js
cea2aj: They are a little different since the method and display name fields are different, but It might make sense to just replace the current 'from' methods with all the adapter code

src/core/core-adapter.js
cea2aj: ya, that makes sense

src/core/models/navigation.js
cea2aj: ya, i'll update

src/core/adapters/adaptdirectanswer.js
cea2aj: ya, that actually may be simpler in this case. I can handle it in this PR

edit: Updating the component involves more complexity, so I'm going to stick with the adapter for now

src/core/core-adapter.js
```diff
@@ -268,7 +267,7 @@ export default class CoreAdapter {
     this.globalStorage.set(StorageKeys.INTENTS, new SearchIntents({}));
     this.globalStorage.set(StorageKeys.NAVIGATION, new Navigation());
     this.globalStorage.set(StorageKeys.ALTERNATIVE_VERTICALS, new AlternativeVerticals({}));
-    this.globalStorage.set(StorageKeys.DIRECT_ANSWER, new DirectAnswer({}));
+    this.globalStorage.set(StorageKeys.DIRECT_ANSWER, {});
```
oshi97: ah does the direct answer still reset with backnav whether it's `new DirectAnswer({})` or `{}` here? just want to be extra careful with back nav behavior

src/core/models/directanswer.js
```diff
@@ -7,22 +7,35 @@ export default class DirectAnswer {
...
-    const data = { ...response };
-    const { answer, relatedItem } = data;
-
-    if (answer && formatters[answer.fieldApiName]) {
-      answer.value = formatters[answer.fieldApiName](
-        answer.value,
-        relatedItem.data.fieldValues,
-        relatedItem.verticalConfigId,
-        true);
+  static fromCore (directAnswer) {
```
oshi97, line 15: I like this pattern much more! should be easier to prune old code this way

oshi97 (commented):
seems close!

src/core/core-adapter.js
cea2aj: I was experimenting with something and then I reverted, so this should actually just be `new DirectAnswer({})`

src/core/models/directanswer.js
tmeyer2115, line 15: Do we need to apply `formatters` here as well?

src/core/search/searchdatatransformer.js
```diff
@@ -18,16 +18,15 @@ import ResultsContext from '../storage/resultscontext';
...
-      [StorageKeys.INTENTS]: SearchIntents.from(response.searchIntents),
-      [StorageKeys.SPELL_CHECK]: SpellCheck.from(response.spellCheck),
-      [StorageKeys.LOCATION_BIAS]: LocationBias.from(response.locationBias)
+      queryId: data.queryId,
+      navigation: Navigation.fromCore(data.verticalResults),
+      directAnswer: DirectAnswer.fromCore(data.directAnswer),
+      universalResults: UniversalResults.fromCore(data, urls),
+      searchIntents: SearchIntents.fromCore(data.searchIntents),
+      spellCheck: SpellCheck.fromCore(data.spellCheck),
+      locationBias: LocationBias.fromCore(data.locationBias)
```
tmeyer2115: Small thing, but I think we should continue to use `[StorageKeys.blah]` as the keys here. It indicates that an entry is going to be put into storage and where. While it's unlikely that we will actually change any of the `StorageKey` values, we might. If we did, we don't want to have to update the value in multiple places.

cea2aj: Sure, i can revert

src/core/models/directanswer.js
cea2aj: ya, i'll just go ahead and include the formatters in this PR. Edit: After taking another look, I decided that I will include the formatter stuff in a future PR

oshi97 (approved):
this lgtm!

PR #1195, opened 2020-11-24
Location Bias Fix

cea2aj:
Fix a bug where the location bias component would load improperly

The bug occurred when navigating to a blank vertical, performing a search, and hitting the back button. The location bias component would then have the properties for its state set to null. This caused the component to load since the component was only checking that the accuracy was not undefined. This commit makes sure the state is not null before loading the component.

J=SLAP-574
Test=manual

Verify that the bug described above no longer occurs. Test other searches like 'locations near me' and check that the location bias component still works as expected


creotutar:
This fixes https://yexttest.atlassian.net/browse/SLAP-574

I think generally we want the `clearResults` to set the value to the correct value (ideally in a system where a `resetComponent` exists)

But I like this solution now

creotutar (approved):
🍇 🍇 🍇 LGTM

PR #1205, opened 2020-12-02
Add facets to vertical search request using answers-core

alexisgrow:
TEST=manual,unit

Run unit tests locally, confirm they pass. Locally build and serve site, confirm facets are sent in the request and returned properly with results updating. Test selecting three facets on the same field - see they are sent in the request and received in the response; see results update accordingly. Test selecting three facets on one field and two facets on another, confirm all five are sent in request and received in response with results updating.


src/core/models/verticalresults.js
```diff
@@ -42,21 +42,22 @@ export default class VerticalResults {
...
-  static fromCore (verticalResults, urls = {}, resultsContext = ResultsContext.NORMAL) {
+  static fromCore (verticalResults, urls = {}, resultsContext = ResultsContext.NORMAL, verticalKey) {
     if (!verticalResults) {
       return new Section();
     }
 
     return new Section(
       {
-        verticalConfigId: verticalResults.verticalKey,
+        verticalConfigId: verticalResults.verticalKey || verticalKey,
```
tmeyer2115: Is `verticalResults.verticalKey` sometimes not present?

src/core/filters/filterregistry.js
```diff
@@ -73,19 +72,14 @@ export default class FilterRegistry {
...
-  }
-
-  _getFacetFilterPayload () {
-    const getFilters = fn => fn.getChildren().length
-      ? fn.getChildren().flatMap(getFilters)
-      : fn.getFilter();
-    const filters = this.getFacetFilterNodes().flatMap(getFilters);
-    return Facet.fromFilters(this.availableFieldIds, ...filters);
+    return this.getFacetFilterNodes()
+      .flatMap(filterNode => filterNode.children || filterNode)
```
tmeyer2115: Can we use the `getChildren()` getter here? If possible, we should always use the getter instead of the property directly. That's safer.

src/core/models/verticalresults.js
```diff
@@ -42,21 +42,22 @@ export default class VerticalResults {
...
         resultsCount: verticalResults.resultsCount,
         appliedQueryFilters: verticalResults.appliedQueryFilters.map(AppliedQueryFilter.fromCore),
         results: verticalResults.results.map(Result.fromCore)
       },
-      urls[verticalResults.verticalKey],
+      urls[verticalResults.verticalKey || verticalKey],
```
tmeyer2115: Can we make a local variable called `verticalKey` (or something like that) so we only have to do this defaulting logic once?

src/core/search/searchdatatransformer.js
```diff
@@ -46,7 +46,7 @@ export default class SearchDataTransformer {
       [StorageKeys.QUERY_ID]: response.queryId,
       [StorageKeys.NAVIGATION]: new Navigation(), // Vertical doesn't respond with ordering, so use empty nav.
       [StorageKeys.VERTICAL_RESULTS]: VerticalResults.fromCore(
-        response.verticalResults, {}, resultsContext),
+        response.verticalResults, {}, resultsContext, verticalKey),
       [StorageKeys.DYNAMIC_FILTERS]: DynamicFilters.from(response.facets, resultsContext),
```
tmeyer2115: Do we need to do any kind of transformation for the SDK to properly parse the facets returned by the Core library?

src/core/models/verticalresults.js
alexisgrow: Right! It looks like it's not present in vertical search responses (though we supply it in the request), but it is present in `alternativeVerticals` and universal result's `verticalResults`. This is something that existed in the non-answers-core branches of the SDK, it was removed in this PR https://github.com/yext/answers/pull/1193. I can move the change over to that PR so it doesn't seem facets-specific.

src/core/filters/filterregistry.js
alexisgrow: I think I remember trying that and I got different results, but totally agree, going to test again

src/core/search/searchdatatransformer.js
alexisgrow: Yes, we do! This PR is only really for the facets request, the response parsing changes are in this PR: https://github.com/yext/answers/pull/1193

src/core/filters/filterregistry.js
```diff
@@ -73,19 +72,14 @@ export default class FilterRegistry {
   }
 
   /**
-   * Gets the facet filter string to send in a search query.
-   * @returns {string}
+   * Gets the facet filters as an array of SimpleFilters to send to the answers-core.
+   *
+   * @returns {SimpleFilter[]}
```
cea2aj: This is an answers-core `SimpleFilter`, correct? If so, can we indicate that with a comment such as 'from answers-core'?

src/core/filters/filterregistry.js
alexisgrow: Ah, it was a combination of the "refreshing the page doesn't actually apply facets to the search" and that `getChildren()` is not falsy when there are no children (empty array). Resolved and updating to use `getChildren()` and `getFilters()`!

tmeyer2115 (approved):

PR #1206, opened 2020-12-02
Add static filters to vertical search request to answers-core

alexisgrow:
Adds support for using static filters with the answers-core library.

TEST=manual,unit

Run the unit tests and ensure they pass. Build and serve a page locally with the answers-core library and a FilterBox component with three filters. "Check" a combination of filters from all three filters; test just two filters, test selecting options from only one filter and check selecting only a single option. Also test with a DateRangeFilter, confirm that the start date is being sent in the request.


src/core/filters/filterregistry.js
```diff
@@ -57,18 +56,65 @@ export default class FilterRegistry {
...
-    const totalNode = FilterNodeFactory.and(...filterNodes);
-    return totalNode.getFilter();
+  _transformFilter (filterNode) {
+    if (filterNode.children) {
+      return this._recurseCombinedFilterNodes(filterNode.children, filterNode.combinator);
+    }
+
+    // All static filters come through this code path, even if the filters have not been
+    // interacted with. We should ignore filters that effectively don't filter anything.
+    if (!filterNode.filter || Object.entries(filterNode.filter).length === 0) {
```
tmeyer2115: Instead of having this check here, should `getStaticFilterNodes` simply not return such `filterNode`s?

src/core/filters/filterregistry.js
```diff
@@ -57,18 +56,65 @@ export default class FilterRegistry {
...
+  _recurseCombinedFilterNodes (filterNodes, combinator) {
+    return {
+      filters: filterNodes.flatMap(filterNode => {
+        return this._transformFilter(filterNode);
+      }).filter(filter => filter),
+      combinator: combinator
+    };
+  }
+
+  _transformSimpleFilter (filterNode) {
```
tmeyer2115: I think we should rename this method. The name suggests that we're transforming a `SimpleFilter` into something else, when we're really doing the opposite.

src/core/filters/filterregistry.js
```diff
@@ -57,18 +56,65 @@ export default class FilterRegistry {
...
+      return this._transformFilter(filterNode);
+    }).filter(filter => filter);
+
+    if (transformedFilterNodes.length > 1) {
+      return {
+        combinator: '$and',
+        filters: transformedFilterNodes
+      };
+    }
+    return transformedFilterNodes[0];
```
tmeyer2115: Do we need to consider the case of `filterNodes.length === 0`?

src/core/filters/filterregistry.js
```diff
@@ -57,18 +56,65 @@ export default class FilterRegistry {
...
   }
 
-  _getStaticFilterPayload () {
-    const filterNodes = this.getStaticFilterNodes();
-    const totalNode = FilterNodeFactory.and(...filterNodes);
-    return totalNode.getFilter();
+  _transformFilter (filterNode) {
```
tmeyer2115: Can we come up with a more descriptive name or add some JSDocs? It's not clear from the method signature what the result of the transform is. Potentially, we could name this `transformFilterNode` (as that's what is being transformed) and add some JSDoc about how it is parsed into either a `CombinedFilter` or `SimpleFilter`

src/core/filters/filterregistry.js
alexisgrow: Looking into this! `getStaticFilterNodes` returns what's in global storage, which appears to be one node per static filter on the page. If static filter nodes were only simple filters, I'd say definitely yes, but since combined filters can be nested, it makes it a little trickier to know if they contain filtering data

src/core/filters/filterregistry.js
alexisgrow: In that case, the `transformedFilterNodes = []` so this function would return undefined. Is that expected, or should we add in alternate fallback behavior?

src/core/filters/filterregistry.js
alexisgrow: Update here so far - I'm a little hesitant to change the behavior of `getStaticFilterNodes` since it's used in the ResultHeader and AppliedQueryFilters. Going to try to make the change and test things out to see what happens!

tmeyer2115: @oshi97, do you foresee any problems updating `getStaticFilterNodes` in this way?

src/core/filters/filterregistry.js
tmeyer2115: This is ultimately passed to the Core library, right? I'm assuming the library would interpret `undefined` as no static filters and so none would be included in the request.

alexisgrow: Yes exactly!

tmeyer2115: I think we should at least make it clear in the JSDoc that this method may return `undefined`. Someone reading through this code may be a bit confused to see an `[0]` accessor, without a check that `length === 1`.

src/core/filters/filterregistry.js
tmeyer2115: Instead of updating `getStaticFilterNodes`, could we put this filtering logic in `getStaticFilterPayload`? 
```
const transformedFilterNodes = filterNodes
    .filter(filterNode => ...)
    .map(this.transformFilter);
```
If we need this filtering in `_recurseCombinedFilterNodes`, maybe we could make a shared validation function. It just feels a bit odd to put the validation in `_transformFilter`, I think it should be up to the caller to pass it a proper `filterNode`. I think we should avoid returning `undefined` from a method when we can. When other functions begin to interpret an `undefined` output as meaning something, that can get a bit brittle.

alexisgrow: Oh I like that! Yes! I'll try that out now

oshi97: I think CombinedFilterNodes will never have any empty filter nodes because in FilterNodeFactory any SimpleFilterNodes that are empty are filtered out. I... think this is something that's safe to assume though we might want some kind of refactor later since you ARE able to break this rule if you don't use FilterNodeFactory and call the constructors directly.

I can't remember exactly why we let you store empty filter nodes, but I think having a validation step here makes sense!

src/core/filters/filterregistry.js
```diff
@@ -1,5 +1,6 @@
 /** @module FilterRegistry */
 
+import { FilterCombinator } from '@yext/answers-core/dist/models/searchservice/request/CombinedFilter';
```
tmeyer2115: Can we add a `TODO (update)` here? This is alright for now, but I want to make sure things are exported in `answers-core` in such a way that people don't have to reach deep into `dist` to get what they need.

src/core/filters/filterregistry.js
```diff
@@ -56,56 +57,48 @@ export default class FilterRegistry {
   }
 
   /**
-   * Gets the static filters as a {@link CombinedFilter} to send to the answers-core
+   * Gets the static filters as a {@link SimpleFilter|CombinedFilter} to send to the answers-core
+   * @returns {CombinedFilter|SimpleFilter|undefined} Returns undefined if no filters with
```
tmeyer2115: Newline before `@returns`

src/core/filters/filterregistry.js
oshi97: you could probably just do `filterNodes.filter(fn => fn.getFilter().getFilterKey());` (ripped straight from FilterNodeFactory)

src/core/filters/filterregistry.js
```diff
@@ -56,56 +57,48 @@ export default class FilterRegistry {
...
+        return this._transformFilterNodes(filterNode.children, filterNode.combinator);
+      }
 
-    // All static filters come through this code path, even if the filters have not been
-    // interacted with. We should ignore filters that effectively don't filter anything.
-    if (!filterNode.filter || Object.entries(filterNode.filter).length === 0) {
-      return;
-    }
+      return !this._isEmpty(filterNode) && this._transformSimpleFilterNode(filterNode);
+    }).filter(filter => filter);
```
tmeyer2115: Since we have the `!this._isEmpty` check do we need this `filter` any more?

src/core/filters/filterregistry.js
```diff
@@ -56,56 +57,48 @@ export default class FilterRegistry {
...
-    return transformedFilterNodes[0];
+    return this._transformFilterNodes(filterNodes, FilterCombinator.AND);
   }
 
-  _transformFilter (filterNode) {
-    if (filterNode.children) {
-      return this._recurseCombinedFilterNodes(filterNode.children, filterNode.combinator);
-    }
+  /**
+   * Transforms a filter node {@link CombinedFilterNode} or {@link SimpleFilterNode} to
```
tmeyer2115: Maybe "Transforms a list of filter nodes ..." since this operates on an `Array` of filter node.

src/core/filters/filterregistry.js
alexisgrow: Yeah totally! I can just use the SDK's built-in `FilterCombinators` for now (they are the same) and we can update all at once?

tmeyer2115: Sure, that works too!

src/core/filters/filterregistry.js
alexisgrow: Ooh, per Oliver's comment I think we can move the `isEmpty` check up to the `getStaticFilterPayload` and then yes we won't need this! Removing!

src/core/filters/filterregistry.js
```diff
@@ -58,38 +58,43 @@ export default class FilterRegistry {
...
-      return !this._isEmpty(filterNode) && this._transformSimpleFilterNode(filterNode);
-    }).filter(filter => filter);
+      return this._transformSimpleFilterNode(filterNode);
+    });
 
-    return filters.length > 1
-      ? {
+    return filters.length === 1
+      ? filters[0]
+      : {
```
tmeyer2115, line 94: If the length was zero, would we want to return what's below?

alexisgrow: I'm not sure! Technically, something like `{"$or":[]}` is valid to the API (it's the same as no filters), and we have a check in the caller, but I could move the check from `getStaticFilterPayload()`:
```
filterNodes.length > 0
      ? this._transformFilterNodes(filterNodes, FilterCombinators.AND)
      : null;
```
into this function to be safer!

tmeyer2115: Okay! If the API knows how to interpret this correctly when it's `[]`, then I'm fine either way!

tmeyer2115 (approved):

oshi97 (approved):
lgtm!

PR #1207, opened 2020-12-02
Update size limit 0.21.1 -> 4.9.0

oshi97:
This commit updates size-limit. It also installs the
@size-limit/file plugin, which is now separate from the
main size-limit package, and is what we were using to
check our bundle sizes.

TEST=manual

see that size-limit still works after building the sdk, and will freak out if the bundle
is bigger than the limit


creotutar (approved):
🏔️ 🏔️ 🏔️ LGTM

PR #1208, opened 2020-12-04
globalstorage: 2.0 scaffold + tests

creotutar:
Add the scaffolded global storage to the SDK (for storage 2.0). Add
tests for the new functions assuming new functionality. Note: The unit
tests are not expected to pass at this time as the global storage has
not been implemented yet. The linter is supposed to pass, though.

All tests from globalstorage are added except for the setAll and the
previous implementation of getAll (which was really a `getAllWithPrefix`).

J=SLAP-867
TEST=unit

### DIFF
--- a/tests/core/storage/storage.js
+++ b/tests/core/storage/storage.js
@@ -20,18 +20,18 @@ it('calls update and reset listeners onpopstate', () => {
 
 describe('init', () => {
   it('should be initialized with empty map w/o init', () => {
-    expect(storage.getAll()).toEqual({});
+    expect(storage.getAll()).toEqual(new Map());
   });
 
   it('should be initialized with empty map with empty string', () => {
     storage.init('');
-    expect(storage.getAll()).toEqual({});
+    expect(storage.getAll()).toEqual(new Map());
   });
 
-  const expectedResult = {
-    key1: 'val1',
-    key2: 'val2'
-  };
+  const expectedResult = new Map([
+    ['key1', 'val1'],
+    ['key2', 'val2']
+  ]);
 
   it('should be initialized with an absolute url', () => {
     storage = new GlobalStorage();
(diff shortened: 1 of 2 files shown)


tests/core/storage/storage.js
```diff
@@ -0,0 +1,339 @@
...
+it('calls update and reset listeners onpopstate', () => {
+  storage = new GlobalStorage(stateUpdateListener, stateResetListener);
+  window.onpopstate();
+  expect(stateUpdateListener).toBeCalled();
+  expect(stateResetListener).toBeCalled();
+});
+
+describe('init', () => {
+  it('should be initialized with the correct formats', () => {
+    expect(storage.getAll()).toEqual({});
```
oshi97: it might be a little cleaner to have one it() per expect

tests/core/storage/storage.js
```diff
@@ -0,0 +1,339 @@
...
+
+    storage = new GlobalStorage();
+    storage.init('key1=val1&key2=val2');
+    expect(storage.getAll()).toEqual(expectedResult);
+  });
+});
+
+describe('set', () => {
+  describe('vanilla set', () => {
+    it('correctly stores primitives', () => {
```
oshi97: I think I would break up these tests into one expect per each too, instead of `correctly stores primitives` something like `correctly stores number/array/etc`

tests/core/storage/storage.js
```diff
@@ -0,0 +1,339 @@
...
+
+      storage.set(StorageKeys.QUERY, undefined);
+      expect(storage.get(StorageKeys.QUERY)).toEqual(undefined);
+
+      storage.set(StorageKeys.QUERY, '');
+      expect(storage.get(StorageKeys.QUERY)).toEqual('');
+    });
+
+    it('stores data by the provided key', () => {
+      storage.set(StorageKeys.AUTOCOMPLETE, { test: 'test autocomplete data' });
```
oshi97, line 111: is this test redundant with the "correctly stores structures" one?

tests/core/storage/storage.js
```diff
@@ -0,0 +1,339 @@
...
+
+        storage.setWithPersist(StorageKeys.QUERY, undefined);
+        expect(storage.get(StorageKeys.QUERY)).toEqual(undefined);
+
+        storage.setWithPersist(StorageKeys.QUERY, '');
+        expect(storage.get(StorageKeys.QUERY)).toEqual('');
+      });
+
+      it('stores data by the provided key', () => {
+        storage.setWithPersist(StorageKeys.AUTOCOMPLETE, { test: 'test autocomplete data' });
```
oshi97: same question about redundancy

oshi97 (commented):
looking good! just a few small things

oshi97 (approved):
lgtm!

PR #1209, opened 2020-12-04
Add custom formatter support to core adapters

cea2aj:
Add support for a custom formatting object which formats result fields

J=SLAP-918
Test=manual

Test the formatters on with universal search, vertical search, and direct answers. View the formatted results on a test site. Use toUpperCase() as the test formatter. Also confirm that the formatters have access to highlighted fields.

### DIFF
--- a/src/core/models/alternativeverticals.js
+++ b/src/core/models/alternativeverticals.js
@@ -15,7 +15,7 @@ export default class AlternativeVerticals {
    * Create alternative verticals from server data
    *
    * @param {Object[]} alternativeVerticals
-   * @param {Object.<string, function>} formatters to apply to the result fields
+   * @param {Object<string, function>} formatters applied to the result fields
    */
   static fromCore (alternativeVerticals, formatters) {
     if (!alternativeVerticals || alternativeVerticals.length === 0) {
--- a/src/core/models/alternativeverticals.js
+++ b/src/core/models/alternativeverticals.js
@@ -13,15 +13,17 @@ export default class AlternativeVerticals {
 
   /**
    * Create alternative verticals from server data
+   *
    * @param {Object[]} alternativeVerticals
+   * @param {Object.<string, function>} formatters to apply to the result fields
    */
-  static fromCore (alternativeVerticals) {
+  static fromCore (alternativeVerticals, formatters) {
     if (!alternativeVerticals || alternativeVerticals.length === 0) {
       return new AlternativeVerticals();
     }
 
     return new AlternativeVerticals(alternativeVerticals.map(alternativeVertical => {
-      return VerticalResults.fromCore(alternativeVertical);
+      return VerticalResults.fromCore(alternativeVertical, {}, formatters);
     }));
   }
 }
--- a/src/core/models/directanswer.js
+++ b/src/core/models/directanswer.js
@@ -45,7 +45,7 @@ export default class DirectAnswer {
    * @param {Object.<string, function>} formatters
    */
   format (formatters) {
-    if (!this.answer || !this.relateditem) {
+    if (!this.answer || !this.relatedItem) {
       return this;
     }
 
--- a/src/core/models/directanswer.js
+++ b/src/core/models/directanswer.js
@@ -7,19 +7,21 @@ export default class DirectAnswer {
   }
 
   /**
-   * Constructs an SDK DirectAnswer from an answers-core DirectAnswer
+   * Constructs an SDK DirectAnswer from an answers-core DirectAnswer and applies formatting
    *
    * @param {DirectAnswer} directAnswer from answers-core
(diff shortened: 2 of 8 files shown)


src/core/models/alternativeverticals.js
```diff
@@ -13,15 +13,17 @@ export default class AlternativeVerticals {
 
   /**
    * Create alternative verticals from server data
+   *
    * @param {Object[]} alternativeVerticals
+   * @param {Object.<string, function>} formatters to apply to the result fields
```
tmeyer2115: `Object<string, function>`

src/core/models/directanswer.js
```diff
@@ -38,4 +38,28 @@ export default class DirectAnswer {
       }
     });
   }
+
+  /**
+   * Applies a formatter to the direct answer if a formatter exists for the direct answer field
+   *
+   * @param {Object.<string, function>} formatters
```
tmeyer2115: `Object<string, function>`

src/core/models/directanswer.js
```diff
@@ -38,4 +38,28 @@ export default class DirectAnswer {
...
+   */
+  format (formatters) {
```
oshi97: would it make sense to call these format() methods within fromCore() instead of making them public?

src/core/models/directanswer.js
```diff
@@ -38,4 +38,28 @@ export default class DirectAnswer {
...
+
+    if (formatters[answer.fieldApiName]) {
+      answer.value = formatters[answer.fieldApiName](
+        answer.value,
+        relatedItem.data.fieldValues,
+        relatedItem.verticalConfigId,
+        true);
+    }
+
+    return this;
```
tmeyer2115: This method mutates the original `DirectAnswer`. Instead of that, we could add `formatters` to the `fromCore` method. Aside from avoiding mutation, it may simplify things if this is handled by `fromCore`. Otherwise, we will need to remember to always call `format` on the result of `fromCore`. What do you think?

src/core/models/result.js
```diff
@@ -128,4 +129,57 @@ export default class Result {
...
     });
   }
+
+  /**
+   * Applies formatters to the result
+   *
+   * @param {Object.<string, function>} formatters to apply to the result fields
+   * @param {string} verticalKey
+   */
+  format (formatters, verticalKey) {
```
tmeyer2115: Same thing here about combining `formatters` with `fromCore`.

src/core/models/directanswer.js
cea2aj: That makes sense. I'll update

src/core/models/directanswer.js
```diff
@@ -36,6 +38,33 @@ export default class DirectAnswer {
...
+
+  /**
+   * Applies a formatter to a direct answer value
+   *
+   * @param {Object} data directAnswerData
+   * @param {function} formatter a field formatter to apply to the answer value field
+   */
+  static _getFormattedValue (data, formatter) {
+    if (!data.answer || !data.relatedItem) {
+      return null;
```
tmeyer2115, line 63: If we can't apply formatters to the value, should we return `data` here?

src/core/models/result.js
```diff
@@ -126,6 +129,63 @@ export default class Result {
...
+      resultData.details = formattedData.description;
+    }
+
+    return new Result(resultData);
+  }
+
+  /**
+   * Returns an object which contains formatted fields
+   *
+   * @param {Object<string, function>} formatters to apply to the result fields
```
tmeyer2115, line 152: Missing one `@param`

src/core/models/result.js
```diff
@@ -126,6 +129,63 @@ export default class Result {
...
+   * @param {string} verticalKey the verticalKey associated with the result
+   * @returns {Object<string, string>} keys are field names and values are the formatted data
+   */
+  static _getFormattedData (resultData, formatters, verticalKey) {
```
tmeyer2115, line 156: Looking through the SDK code, it seems like we only formatted data if it came from the KnowledgeManager back-end. Do we need to have such a safeguard here?

src/core/models/directanswer.js
cea2aj: This function is supposed to return a string (i'll update the jsdoc), and I thought it made more sense to return a string or null if the formatter can't be applied. Returning a string or an object if the formatter can't be applied seems like a strange interface to me

src/core/models/result.js
cea2aj: That's a good catch. Since we recently added a Source param to the result, I can add a check to make sure that the result is from KnowledgeManager before applying formatting

oshi97 (commented):
+1 from me!

tmeyer2115 (approved):

PR #1210, opened 2020-12-08
globalstorage 2.0: Initial implementation

creotutar:
We implement the global storage 2.0 class. There are a few notable
differences between this class and the old class.

1) There is a persistent storage instance variable and a buffer variable
for the persistent storage. This will allow us to do delayed
persistence like context/referrerPageUrl.

2) The update listeners for persistent storage set/delete are handled
here.

3) There is a getUrlWithCurrentState function that will return url query
params of the current persistent storage buffer.

4) We serialize the data in setWithPersist (instead of relying on
serialization in the persistent storage).

5) getAll() no longer gives data with a certain key prefix. instead it
returns all entries in the global storage

6) setAll was removed

7) getState was renamed to get

We also fix our tests (some to accomodate changes in the
implementation).

J=SLAP-866
TEST=auto

npx jest --clearCache && npm run test -- tests/core/storage/storage.js

### DIFF
--- a/src/core/storage/listener.js
+++ b/src/core/storage/listener.js
@@ -1,0 +1,15 @@
+/**
+ * The storage listener is a listener on changes to the SDK global storage
+ */
+export default class StorageListener {
+  /**
+   * @param {string} eventType The type of event to listen for e.g. 'update'
+   * @param {string} storageKey The key to listen for e.g. 'Pagination'
+   * @param {Function} callback The callback to call when the event is emitted on the storage key
+   */
+  constructor (eventType, storageKey, callback) {
+    this.eventType = eventType;
+    this.storageKey = storageKey;
+    this.callback = callback || function () {};
+  }
+}
--- a/src/core/storage/listener.js
+++ b/src/core/storage/listener.js
@@ -10,6 +10,6 @@ export default class StorageListener {
   constructor (eventType, storageKey, callback) {
     this.eventType = eventType;
     this.storageKey = storageKey;
-    this.callback = callback || function () {};
+    this.callback = callback;
   }
 }
--- a/src/core/storage/storage.js
+++ b/src/core/storage/storage.js
@@ -1,6 +1,5 @@
 import DefaultPersistentStorage from '../../../node_modules/@yext/answers-storage/dist/index';
 import { AnswersStorageError } from '../errors/errors';
-import ModuleData from './moduledata';
 
 /**
  * GlobalStorage is a container around application state.  It
@@ -39,9 +38,9 @@ export default class GlobalStorage {
     /**
      * The core data for the global storage
      *
-     * @type {Map<string, ModuleData}
+     * @type {Map<string, *}
      */
-    this.moduleDataContainer = new Map();
+    this.storage = new Map();
 
     /**
      * The persistent storage implementation to store state
@@ -60,11 +59,11 @@ export default class GlobalStorage {
     this.persistentStorageBuffer = new DefaultPersistentStorage(this.popListener);
 
     /**
-     * The listeners to apply to uninitialized storage modules
+     * The listeners to apply on changes to global storage
      *
-     * @type {Map<string, Object>}
+     * @type {Listener[]}
      */
-    this.futureListeners = new Map();
+    this.listeners = [];
   }
 
   /**
@@ -97,11 +96,8 @@ export default class GlobalStorage {
       throw new AnswersStorageError('Invalid storage key provided', key, data);
     }
 
-    if (this.moduleDataContainer.get(key) === undefined) {
-      this.moduleDataContainer.set(key, new ModuleData(key));
-      this._applyFutureListeners(key);
-    }
-    this.moduleDataContainer.get(key).set(data);
+    this.storage.set(key, data);
+    this._callListeners('update', key);
   }
 
   /**
@@ -150,23 +146,16 @@ export default class GlobalStorage {
    * @return {*} The state for the provided key, undefined if key doesn't exist
    */
   get (key) {
-    if (!this.moduleDataContainer.get(key)) {
-      return this.moduleDataContainer.get(key);
-    }
-    return this.moduleDataContainer.get(key).raw();
+    return this.storage.get(key);
   }
 
   /**
    * Get the current state for all key/value pairs in storage
    *
-   * @return {Map<string, ModuleData>} mapping from key to value representing the current state
+   * @return {Map<string, *>} mapping from key to value representing the current state
    */
   getAll () {
-    const entries = new Map();
-    this.moduleDataContainer.forEach((value, key) => {
-      entries.set(key, this.get(key));
-    });
-    return entries;
+    return this.storage;
   }
 
   /**
@@ -175,7 +164,7 @@ export default class GlobalStorage {
    * @param {string} key The storage key to delete
    */
   delete (key) {
-    this.moduleDataContainer.delete(key);
+    this.storage.delete(key);
     this.persistentStorageBuffer.delete(key);
   }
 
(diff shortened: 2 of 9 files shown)


creotutar:
For unit tests, I may need to publish @yext/storage to the registry with a beta flag, because jest works locally but apparently cannot find the node_modules path in circleci.

src/core/storage/storage.js
```diff
@@ -1,3 +1,7 @@
+import DefaultPersistentStorage from '../../../node_modules/@yext/answers-storage/dist/index';
```
oshi97: would you be able to import from `@yext/answers-storage` instead?

src/core/storage/storage.js
```diff
@@ -32,65 +88,162 @@ export default class GlobalStorage {
...
    * Get the current state for the provided key
    *
    * @param {string} key The storage key to get
-   * @return {*} The state for the provided key
+   * @return {*} The state for the provided key, undefined if key doesn't exist
    */
-  get (key) {}
+  get (key) {
+    if (!this.moduleDataContainer.get(key)) {
+      return this.moduleDataContainer.get(key);
```
oshi97: would this.moduleDataContainer.has(key) be better (also for other similar usages like in set, on, and off)? could we also just return undefined instead of doing another get?

src/core/storage/storage.js
```diff
@@ -32,65 +88,162 @@ export default class GlobalStorage {
...
    * Get the current state for all key/value pairs in storage
    *
-   * @return {Object} mapping from key to value representing the
-   *                  current state
+   * @return {Map<string, ModuleData>} mapping from key to value representing the current state
    */
-  getAll () {}
+  getAll () {
+    const entries = new Map();
+    this.moduleDataContainer.forEach((value, key) => {
```
oshi97: I think you can also just do  `new Map(this.moduleDataContainer);` to make a copy?

src/core/storage/storage.js
```diff
@@ -32,65 +88,162 @@ export default class GlobalStorage {
...
+  off (event, moduleId, callback) {
+    const moduleData = this.moduleDataContainer.get(moduleId);
+    if (moduleData === undefined) {
+      if (this.futureListeners.get(moduleId) !== undefined) {
+        this.futureListeners.get(moduleId).pop(); // Is this right?
+      }
+      return;
+    }
+
+    moduleData.off(event, callback);
```
oshi97: I think this removes all callbacks for the moduleData, since this ends up calling event emitter's off(), which doesn't even take a callback arg? (ran into this a while back)

src/core/storage/storage.js
```diff
@@ -32,65 +88,162 @@ export default class GlobalStorage {
...
+   *
+   * @param {string} event The event being listened to, e.g. ‘update’
+   * @param {string} moduleId The module key being listened for e.g. Pagination
+   * @param {Function} callback The callback to be removed
+   */
+  off (event, moduleId, callback) {
+    const moduleData = this.moduleDataContainer.get(moduleId);
+    if (moduleData === undefined) {
+      if (this.futureListeners.get(moduleId) !== undefined) {
+        this.futureListeners.get(moduleId).pop(); // Is this right?
```
oshi97: wouldn't this only remove the most recent registered listener, not the specified callback? I think we would need to do something like loop through all listeners and check if the callback arg points to the same thing as any of these listeners? or maybe a different system for tracking callbacks like setTimeout does

oshi97 (commented):
just some small things (and also ModuleData.off() never working properly in the sdk)

src/core/storage/storage.js
creotutar: Yee, this was copied from old implementation, looking into ways of removing this way of storing listeners

src/core/storage/storage.js
creotutar: Ahh, good catch. Used to have `this.get(key).raw()` but kept it as the ModuleData. Was really unsure there but will change b/c currently it is not using raw

src/core/storage/storage.js
```diff
@@ -8,12 +11,59 @@
...
+     */
+    this.popListener = (queryParamsObject, queryParamsString) => {
+      this.stateUpdateListener(queryParamsObject, queryParamsString);
+      this.stateResetListener(queryParamsObject, queryParamsString);
+    };
+
+    /**
+     * The core data for the global storage
+     *
+     * @type {Map<string, *}
```
oshi97: closing angle bracket

src/core/storage/storage.js
creotutar: Moved away from module data! Returning the value directly now.

src/core/storage/storage.js
creotutar: Module data is out, using a different removeListener system!

src/core/storage/storage.js
```diff
@@ -32,65 +87,126 @@ export default class GlobalStorage {
...
+  }
 
   /**
    * Adds a listener to the given module for a given event
    *
-   * @param {string} event The event to listen to, e.g. ‘update’
-   * @param {string} module The module key to listen for e.g. Pagination
-   * @param {Function} callback The callback when an event is called,
-   *                   function is given the data if relevant
+   * @param {Listener} the listener to add
```
oshi97: should this be StorageListener?

src/core/storage/storage.js
```diff
@@ -32,65 +87,126 @@ export default class GlobalStorage {
...
+   */
+  addListener (listener) {
+    if (!listener.eventType || !listener.storageKey || !listener.callback) {
+      throw new AnswersStorageError(`Invalid listener applied in storage: ${listener}`);
+    }
+    this.listeners.push(listener);
+  }
+
+  /**
+   * Removes a listener for the given moduleId for a given event
```
oshi97: should `for the given moduleId for a given event` be removed?

src/core/storage/listener.js
```diff
@@ -0,0 +1,15 @@
...
+export default class StorageListener {
+  /**
+   * @param {string} eventType The type of event to listen for e.g. 'update'
+   * @param {string} storageKey The key to listen for e.g. 'Pagination'
+   * @param {Function} callback The callback to call when the event is emitted on the storage key
+   */
+  constructor (eventType, storageKey, callback) {
+    this.eventType = eventType;
+    this.storageKey = storageKey;
+    this.callback = callback || function () {};
```
oshi97: I like this model! maybe we shouldn't default callback to something? if somebody didn't pass a callback to a listener that's probably a bug right? maybe the validation that is done in storage.addListener() should just be done here?

src/core/storage/storage.js
```diff
@@ -32,65 +87,126 @@ export default class GlobalStorage {
...
   /**
    * Adds a listener to the given module for a given event
    *
-   * @param {string} event The event to listen to, e.g. ‘update’
-   * @param {string} module The module key to listen for e.g. Pagination
-   * @param {Function} callback The callback when an event is called,
-   *                   function is given the data if relevant
+   * @param {Listener} the listener to add
+   */
+  addListener (listener) {
```
oshi97: maybe registerListener? I'm fine with this name though!

src/core/storage/storage.js
```diff
@@ -8,12 +11,59 @@
  */
 export default class GlobalStorage {
   constructor (stateUpdateListener, stateResetListener) {
+    /**
+     * The update listener for changes in state (persistent storage changes)
+     *
+     * @type {Function}
+     */
+    this.stateUpdateListener = stateUpdateListener || function () {};
```
oshi97: state is a bit of a loaded term, though it makes sense here, maybe something like `persistedStateUpdateListener`? my suggestion is a little wordy so open to ideas! maybe a `persistedStateListeners` object instead of 2 params?

src/core/storage/storage.js
```diff
@@ -32,65 +87,126 @@ export default class GlobalStorage {
...
    * to the persistent storage.
    */
-  flushPersist () {}
+  flushPersist () {
+    this.persistentStorage = this.persistentStorageBuffer;
+
+    this.persistentStorageBuffer = new DefaultPersistentStorage(this.popListener);
+    this.persistentStorage.getAll().forEach((value, key) => {
+      this.persistentStorageBuffer.set(key, value);
+    });
```
cea2aj: When I think of flushing a buffer, I assume that the buffer is emptied. However that's not the case here, is it? Since the persistentStorageBuffer is immediately updated with all the values from the new persistentStorage?

creotutar: Persistent storage buffer is under construction, updates pending. I agree that it feels like flush should empty the buffer.

oshi97 (approved):
this lgtm!

oshi97 (approved):
lgtm!

src/core/storage/storage.js
```diff
@@ -32,65 +73,119 @@ export default class GlobalStorage {
    * @param {string} key The storage key to set
    * @param {*} data The data to set
    */
-  set (key, data) {}
+  set (key, data) {
+    if (key === undefined || key === null || typeof key !== 'string') {
+      throw new AnswersStorageError('Invalid storage key provided', key, data);
```
cea2aj: nit: Since key must be a string, what do you think of the error message: 'Storage key must be of type string'?

src/core/storage/storage.js
```diff
@@ -32,65 +73,119 @@ export default class GlobalStorage {
...
+    }
+
+    if (typeof data === 'undefined') {
+      throw new AnswersStorageError('Invalid storage key provided', key, data);
```
cea2aj: I believe this message should refer to data. How about 'Data must not be undefined'?

cea2aj (approved):
lgtm!

PR #1212, opened 2020-12-08
core-adapter: add autocomplete core adapter

nanhu95:
Updates core-adapter.js to use the core library's
autocomplete functions.

Adds an `AutoCompleteResponseTransformer` class to
transform responses from the core library into a format
that the SDK understands and uses.

J=884
TEST=manual
Set up a test site to use my local version of the SDK
with the core adapter and saw that the autocomplete data
that is being generated via the core library and transformer
is equivalent to that being generated via just the SDK.

### DIFF
--- a/src/core/search/autocompleteresponsetransformer.js
+++ b/src/core/search/autocompleteresponsetransformer.js
@@ -13,6 +13,7 @@ export default class AutoCompleteResponseTransformer {
    *
    * @param {import('@yext/answers-core').AutoCompleteResponse} response
    *  the response passed from the core library
+   * @return {AutoCompleteData}
    */
   static transformAutoCompleteResponse (response) {
     const sections = [{
@@ -32,6 +33,7 @@ export default class AutoCompleteResponseTransformer {
    *
    * @param {import('@yext/answers-core').FilterAutoCompleteResponse} response
    *  the response passed from the core library
+   * @return {AutoCompleteData}
    */
   static transformFilterAutoCompleteResponse (response) {
     if (response.sectioned && response.sections) {
--- a/src/core/search/autocompleteresponsetransformer.js
+++ b/src/core/search/autocompleteresponsetransformer.js
@@ -13,7 +13,7 @@ export default class AutoCompleteResponseTransformer {
    *
    * @param {import('@yext/answers-core').AutoCompleteResponse} response
    *  the response passed from the core library
-   * @return {AutoCompleteData}
+   * @returns {AutoCompleteData}
    */
   static transformAutoCompleteResponse (response) {
     const sections = [{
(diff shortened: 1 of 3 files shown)


src/core/search/autocompleteresponsetransformer.js
```diff
@@ -0,0 +1,37 @@
+import AutoCompleteData, { AutoCompleteResult } from '../models/autocompletedata';
+
+/**
+ * A data transformer that takes the response object from an
+ * AutoComplete request and transforms it into a front-end oriented
+ * data structure that our component library and core storage understand.
+ */
+export default class AutoCompleteResponseTransformer {
```
tmeyer2115, line 8: Let's add some unit tests for this class.

tmeyer2115 (commented):
Were you able to test the three different kinds of autocomplete?

src/core/search/autocompleteresponsetransformer.js
oshi97: +2 for tests!

src/core/search/autocompleteresponsetransformer.js
```diff
@@ -0,0 +1,37 @@
...
+      resultsCount: response.results.length
+    }];
+    return new AutoCompleteData({
+      sections: sections,
+      queryId: response.queryId,
+      inputIntents: response.inputIntents
+    });
+  }
+
+  static transformFilterAutoCompleteResponse (response) {
```
oshi97, line 37: Out of curiosity are we able to reference the types exported from the core in jsdoc? maybe with command + click working in vscode? https://github.com/yext/answers-core/pull/24

nanhu95: Yes, it looks like we are able to. Are you saying we should try to use those types here in the code?

oshi97: yes we can reference them in jsdoc! imo it's just a nice quality of life thing

oshi97: something like `@param {AutoCompleteResponse}` or maybe `@param {import('@yext/answers-core').AutoCompleteResponse}`

nanhu95: Ah gotcha! Sorry for some reason I was thinking you meant referencing them in the JS code (like `(response: AutoCompleteResponse)`) in the method signature, and I got confused. I'll add some JSdoc and push them up!

src/core/search/autocompleteresponsetransformer.js
```diff
@@ -6,6 +6,14 @@ import AutoCompleteData, { AutoCompleteResult } from '../models/autocompletedata
...
  */
 export default class AutoCompleteResponseTransformer {
+
+  /**
+   * Converts a universal or vertical autocomplete response from the
+   * core library into an object that the SDK understands.
+   *
+   * @param {import('@yext/answers-core').AutoCompleteResponse} response
+   *  the response passed from the core library
+   */
```
tmeyer2115, line 16: Here and below add lets add a `@returns` to indicate the type of what's returned.

tmeyer2115 (approved):

cea2aj (approved):
LGTM!

cea2aj (approved):
LG!

PR #1214, opened 2020-12-10
core-adapter: don't replace autocomplete result filter

nanhu95:
Attempting to assign result.filter another value in
the autocompleteresponsetransformer results in an
error because result.filter is readonly.

This makes the change so that in the response transformer,
we pass data from the result into the constructor for
AutoCompleteResult in a new object that contains
transformed filters and the rest of the data from the result, instead of attempting to reassigned result.filter.

Also passes the sessionTrackingEnabled field to the
autocomplete request.

TEST=auto, manual
Ran automated tests and saw them pass.
On a local test site, verified that the readonly error
is gone and that the results have been transformed
corrected. Also using the same site, verified that
sessionTrackingEnabled is being passed as a request
paramter to the backend.

### DIFF
--- a/src/core/core-adapter.js
+++ b/src/core/core-adapter.js
@@ -445,7 +445,6 @@ export default class CoreAdapter {
       })
       .then(response => AutoCompleteResponseTransformer.transformFilterAutoCompleteResponse(response))
       .then(data => {
-        console.log(data);
         this.globalStorage.set(`${StorageKeys.AUTOCOMPLETE}.${config.namespace}`, data);
       });
   }
--- a/src/core/core-adapter.js
+++ b/src/core/core-adapter.js
@@ -385,7 +385,8 @@ export default class CoreAdapter {
   autoCompleteUniversal (input, namespace) {
     return this._coreLibrary
       .universalAutoComplete({
-        input: input
+        input: input,
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN).value
       })
       .then(response => AutoCompleteResponseTransformer.transformAutoCompleteResponse(response))
       .then(data => {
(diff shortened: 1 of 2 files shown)


src/core/search/autocompleteresponsetransformer.js
```diff
@@ -52,10 +52,17 @@ export default class AutoCompleteResponseTransformer {
   }
 
   static _transformAutoCompleteResult (result) {
+    let transformedFilter = {};
     if (result.filter) {
-      result.filter = this._transformFilter(result.filter);
```
nanhu95, line 56: Attempting to reassign `result.filter` results in a read only error. The fix here is to pass data into the constructor for `AutoCompleteResult` using a new object that contains the transformed filter. Open to suggestions if there is a more graceful way of mitigating the issue.

src/core/search/autocompleteresponsetransformer.js
```diff
@@ -52,10 +52,17 @@ export default class AutoCompleteResponseTransformer {
   }
 
   static _transformAutoCompleteResult (result) {
+    let transformedFilter = {};
```
oshi97: would a ternary operator work here instead?

src/core/core-adapter.js
```diff
@@ -438,10 +440,12 @@ export default class CoreAdapter {
       .filterAutoComplete({
         input: input,
         verticalKey: config.verticalKey,
-        searchParameters: searchParams
+        searchParameters: searchParams,
+        sessionTrackingEnabled: this.globalStorage.getState(StorageKeys.SESSIONS_OPT_IN).value
       })
       .then(response => AutoCompleteResponseTransformer.transformFilterAutoCompleteResponse(response))
       .then(data => {
+        console.log(data);
```
oshi97: console log

cea2aj (approved):
lgtm

PR #1218, opened 2020-12-18
core-adapter: more tests for autocompleteresponsetransformer

nanhu95:
Adds more test cases for autocompleteresponsetransformer.js.

Makes a change in autocompleteresponsetransformer.js to not
pass in `result.shortValue` to the AutoCompleteResult constructor
in `_transformAutoCompleteResult` since it is no longer part
of the response from the core library.

J=SLAP-941
TEST=auto
Ran tests with `npm run test` and saw that all of them passed.

### DIFF
--- a/src/core/search/autocompleteresponsetransformer.js
+++ b/src/core/search/autocompleteresponsetransformer.js
@@ -57,8 +57,7 @@ export default class AutoCompleteResponseTransformer {
       filter: transformedFilter,
       key: result.key,
       matchedSubstrings: result.matchedSubstrings,
-      value: result.value,
-      shortValue: result.shortValue
+      value: result.value
     });
   }
 
--- a/tests/core/search/autocompleteresponsetransformer.js
+++ b/tests/core/search/autocompleteresponsetransformer.js
@@ -1,7 +1,7 @@
 const { default: AutoCompleteResponseTransformer } = require('../../../src/core/search/autocompleteresponsetransformer');
 
 describe('transform autocomplete response', () => {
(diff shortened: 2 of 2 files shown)


src/core/search/autocompleteresponsetransformer.js
```diff
@@ -57,8 +57,7 @@ export default class AutoCompleteResponseTransformer {
       filter: transformedFilter,
       key: result.key,
       matchedSubstrings: result.matchedSubstrings,
-      value: result.value,
-      shortValue: result.shortValue
```
nanhu95, line 61: Removing this since the core's response no longer contains a `shortValue` property.

tests/core/search/autocompleteresponsetransformer.js
```diff
@@ -107,4 +107,143 @@ describe('transform autocomplete response', () => {
...
+      },
+      intents: [],
+      key: 'name',
+      matchedSubstrings: [
+        {
+          length: 8,
+          offset: 0
+        }
+      ],
+      shortValue: 'Virginia Beach',
```
cea2aj, line 214: This should be removed since we are removing shortValue, right?

nanhu95: Ah, not quite. The `AutoCompleteResult` of the SDK itself actually has `shortValue` as a property, and it is used elsewhere in the SDK, so for the transformed result, we should be expecting `shortValue`.

cea2aj: ah, got it

cea2aj (approved):

PR #1223, opened 2021-01-06
Storage Acceptance Tests

cea2aj:
Add various acceptance tests related to storage

Components tested include Navigation, Spellcheck, Pagination, Facets, and SortOptions. The basic usage of storage for the SearchBar component is covered by existing acceptance tests, so I didn't add any.

These tests cover the persistence bugs that Christian addressed in #1060 , and they more broadly cover usages of persistent storage set, get, and delete functions. Global storage is tested indirectly through the test suite, so I didn't add tests targeted at global storage. Examples of global storage being tested include the SearchComponent setting query on search, the VerticalResultsCount component reading the search offset from the pagination component, and filters being applied during the use of the facets.

I also added some helper functions in a utils file.

J=SLAP-868
TEST=auto

Run the acceptance tests and confirm that they pass

### DIFF
--- a/tests/acceptance/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuite.js
@@ -3,7 +3,7 @@ import VerticalPage from './pageobjects/verticalpage';
 import { setupServer, shutdownServer } from './server';
 import FacetsPage from './pageobjects/facetspage';
 import { Selector } from 'testcafe';
-import { getURLSearchParams, hitBrowserBackButton, reloadPage, waitForResultsToRender } from './utils';
+import { getURLSearchParams, browserBackButton, browserRefreshPage } from './utils';
 
 const UNIVERSAL_PAGE = 'http://localhost:9999/tests/acceptance/fixtures/html/universal';
 const VERTICAL_PAGE = 'http://localhost:9999/tests/acceptance/fixtures/html/vertical';
@@ -15,6 +15,7 @@ const FACETS_PAGE = 'http://localhost:9999/tests/acceptance/fixtures/html/facets
  * up to serve the search page and the dist directory of Answers.
  * This server is closed once all tests have completed.
  */
+
 fixture`Universal search page works as expected`
   .before(setupServer)
   .after(shutdownServer)
@@ -56,7 +57,7 @@ test('navigating and refreshing mantains that page number', async t => {
   await t.navigateTo(`${VERTICAL_PAGE}?query=Virginia`);
   const paginationComponent = VerticalPage.getPaginationComponent();
   await paginationComponent.clickNextButton();
-  await reloadPage();
+  await browserRefreshPage();
   const pageNum = await paginationComponent.getActivePageLabelAndNumber();
   await t.expect(pageNum).eql('Page 2');
 });
@@ -65,8 +66,7 @@ test('navigating pages and hitting the browser back button lands you on the righ
   const paginationComponent = VerticalPage.getPaginationComponent();
   await paginationComponent.clickNextButton();
   await paginationComponent.clickNextButton();
-  await t.eval(() => window.history.back());
-  await t.wait(1000);
+  await browserBackButton();
   const pageNum = await paginationComponent.getActivePageLabelAndNumber();
   await t.expect(pageNum).eql('Page 2');
 });
--- a/tests/acceptance/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuite.js
@@ -3,6 +3,11 @@ import VerticalPage from './pageobjects/verticalpage';
 import { setupServer, shutdownServer } from './server';
 import FacetsPage from './pageobjects/facetspage';
 import { Selector } from 'testcafe';
+import { getURLSearchParams, hitBrowserBackButton, reloadPage, waitForResultsToRender } from './utils';
+
+const UNIVERSAL_PAGE = 'http://localhost:9999/tests/acceptance/fixtures/html/universal';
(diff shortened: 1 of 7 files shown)


tests/acceptance/acceptancesuite.js
```diff
@@ -3,17 +3,23 @@ import VerticalPage from './pageobjects/verticalpage';
 import { setupServer, shutdownServer } from './server';
```
oshi97, line 3: does running `npx testcafe chrome:headless,firefox:headless tests/acceptance/acceptancesuite.js` work locally?

cea2aj: ya, i'm looking into why it failed on circle

cea2aj: Re-running it fixed it, which means it's likely an intermittent error out of scope of this PR (the failure occurred in a previous acceptance test)

tests/acceptance/utils.js
```diff
@@ -0,0 +1,23 @@
+import { ClientFunction } from 'testcafe';
+
+/* global location */
+
+const getPageUrl = ClientFunction(() => window.location.href);
+
+exports.getURLSearchParams = async () => {
```
oshi97: would es6 style exports work here?

tests/acceptance/blocks/navigationcomponent.js
```diff
@@ -0,0 +1,47 @@
...
+   */
+  async getTabLabelsLowerCase () {
+    const tabList = this._navList.child();
+    const tabListCount = await tabList.count;
+    const tabLabels = [];
+
+    for (let i = 0; i < tabListCount; i++) {
+      const nthTab = tabList.nth(i);
+      const innerText = await nthTab.find('a').innerText;
+      tabLabels[i] = innerText.toLowerCase().trim();
```
oshi97, line 24: do we have to change these to lowercase?

oshi97 (commented):
these look great!

tests/acceptance/utils.js
cea2aj: I can try it

tests/acceptance/blocks/navigationcomponent.js
cea2aj: I added this because locally I was getting all uppercase, but in CI, I was getting mixed case which caused the test to fail. Do you have any idea why that might be?

oshi97: Ah, I think I remember ie11 not respecting text-transform for this? Iirc can see this behavior in dev tools too if you try to console log the inner text, lowercase sgtm!

tests/acceptance/utils.js
cea2aj: yep, it works

tests/acceptance/acceptancesuite.js
```diff
@@ -3,17 +3,23 @@ import VerticalPage from './pageobjects/verticalpage';
...
 import FacetsPage from './pageobjects/facetspage';
```
tmeyer2115, line 4: This is likely out of scope for your item, but I think our acceptance tests need some TLC. At the very least, we should break them out into multiple files (maybe one for vertical, universal, facets, and storage). This file is getting pretty bulky. But, we can make a separate item for that.

tests/acceptance/acceptancesuite.js
```diff
@@ -181,10 +208,86 @@ test('Facets, pagination, and filters do not persist accross experience links',
   await verifyCleanLink(viewAllLink);
 });
 
+fixture`Persistent Storage behavior for various components`
+  .before(setupServer)
+  .after(shutdownServer)
+  .page`${UNIVERSAL_PAGE}`;
+
+test('Navigation tab order respects persistent storage tab order', async t => {
```
tmeyer2115, line 216: @oshi97 do we have the capacity to write Jest tests for specific components? From what I can recall, some components have an associated Jest suite, but it's clunky. I'm wondering if some of these tests can be written as Jest integration tests. That is, we create a fake instance of the component with a mocked persistent storage and try some of these things. Acceptance tests are usually meant to verify an actual user flow. They are slower to run. 

Integration tests, on the other hand, ensure that different pieces of the system work together. They are a step above unit tests. Typically, they are fairly quick to run. I have a suspicion that some of the existing acceptance tests we have may be better suited as Jest integration tests. Likely, we made them acceptance tests because it was easier to do so. If our inability to effectively mock components and their dependencies is causing us to rely too much on acceptance tests, that's an issue. In the extreme, when you have tons of acceptance tests, nobody runs them locally because they take too long. 

Verifying that the Nav component respects the tab order in persistent storage feels more like an integration test to me. There's not really a user flow that's being tested. Same thing with ensuring that referrerPageUrl is included. Conversely, I think ensuring that you stay on the same page of results after a refresh or that pagination works with the back button is an acceptance test. There's a legitimate user flow being validated. It's kinda a fine line. 

If we don't have a straightforward path for making good integration tests, that's okay. We can keep the tests I referenced acceptance tests. But, we should make an item to make it easier to create integration tests.

oshi97: I think the tests you mention could be written as jest tests, and maybe also the 'selecting a sort option updates persistent storage' and the 'the spell check component deletes skipSpellCheck'? I don't think it would be the most tedious (though probably not as straightforward as it should be). We would probably have to mock out persistent storage in the tests, which would have to be updated when we swap the storages, but we'll have to do that for the other jest tests anyways.

cea2aj: The test that make sense as acceptance tests have been moved to #1224. I will make an item for adding an integration test suite.

PR #1224, opened 2021-01-07
Add persistence acceptance tests

cea2aj:
Add acceptance tests which cover uses of the browser back button and refresh page button

These tests cover the persistence bugs that Christian addressed in #1060. My initial attempt was #1223, however we decided that several of those tests should be integration tests rather than acceptance tests. This PR includes only the tests from #1223 which make sense as acceptance tests.

This PR includes utility functions for refreshing the page and using the back button.

The navigation tab links were incorrectly pointing to './links' rather than './facets' for the Facets page, so I fixed that here as well.

J=SLAP-868
TEST=auto


oshi97 (approved):
these lgtm! I like the renames + refactors

PR #1229, opened 2021-01-08
Update autocompleteOnLoad README

cea2aj:
Update the README from 'autoCompleteOnLoad' to 'autocompleteOnLoad'

J=SLAP-980
TEST=manaual

Build a site and confirm that the option works as 'autocompleteOnLoad'. Also confirm that the option does not work unless autoFocus = true

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ ANSWERS.addComponent('SearchBar', {
   // Optional, auto focuses the search bar. Defaults to false
   autoFocus: false,
   // Optional, opens the autocomplete suggestions on page load. Defaults to false. Requires autoFocus to be set to true
-  autoCompleteOnLoad: false,
+  autocompleteOnLoad: false,
   // Optional, on vertical search, allow a user to conduct an empty search. Should be set to true if the defaultInitialSearch is "".
   allowEmptySearch: false,
   // Optional, defaults to 300ms (0.3 seconds)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -420,9 +420,9 @@ ANSWERS.addComponent('SearchBar', {
   promptHeader: 'Header',
   // Optional, no default
   placeholderText: 'Start typing...',
-  // Optional, defaults to false
-  autoFocus: true,
-  // Optional, when auto focus on load,  open the autocomplete
+  // Optional, auto focuses the search bar. Defaults to false
+  autoFocus: false,
+  // Optional, opens the autocomplete suggestions on page load. Defaults to false. Requires autoFocus to be set to true
   autoCompleteOnLoad: false,
```
tmeyer2115: Should this be called `autocompleteOnLoad`?

tmeyer2115 (commented):
If we can call the configuration option `autocompleteOnLoad`, then we can get away with just the README update. @rosiegrant is that okay or should we fix the Component such that the option is called `autoCompleteOnLoad`? It looks like we do refer to it as `autocomplete`, not `autoComplete`, elsewhere in the README and configs.

README.md
cea2aj: ya, that's the whole purpose of this PR. I don't understand how I missed that change 😂

rosiegrant:
the read me update is sufficient, thanks!

tmeyer2115 (approved):

PR #1230, opened 2021-01-11
pass queryTrigger: suggest when clicking spell check

oshi97:
This ensures that we pass the queryTrigger: suggest param
after clicking a spell check link. Previously, this would
be overriden to query-parameter

This also fixes a bug where clicking a spell check link would
not reset the pagination from say page 2 to page 1. So if I were
to search for "varginia", go to page 2, then click the spell check,
prior to this I would be on page 2 instead of page 1 (though this
is probably exceptionally rare).

J=SLAP-979
TEST=manual, auto

test that clicking an autocomplete link will send queryTrigger: suggest

added acceptance test for spell check flow on a vertical page

### DIFF
--- a/tests/acceptance/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuite.js
@@ -83,6 +83,7 @@ test.requestHooks(spellCheckLogger)('spell check flow', async t => {
   await t.expect(pageNum).eql('Page 1');
 
   // Check that clicking spell check sends a queryTrigger=suggest url param
+  // TODO(oshi) investigate making this an integration test
   const requestUrl = spellCheckLogger.requests[spellCheckLogger.requests.length - 1].request.url;
   const queryTriggerParam = new URLSearchParams(requestUrl).get('queryTrigger');
   await t.expect(queryTriggerParam).eql('suggest');
--- a/tests/acceptance/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuite.js
@@ -66,29 +66,27 @@ test('navigating and refreshing mantains that page number', async t => {
 const spellCheckLogger = RequestLogger({
   url: /v2\/accounts\/me\/answers\/vertical\/query/
 });
-test
-  .requestHooks(spellCheckLogger)
-  ('spell check flow', async t => {
-    const searchComponent = VerticalPage.getSearchComponent();
-    await searchComponent.enterQuery('varginia');
-    await searchComponent.submitQuery();
+test.requestHooks(spellCheckLogger)('spell check flow', async t => {
+  const searchComponent = VerticalPage.getSearchComponent();
+  await searchComponent.enterQuery('varginia');
+  await searchComponent.submitQuery();
 
-    // Check that clicking spell check resets pagination to page 1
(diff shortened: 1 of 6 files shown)


tests/acceptance/acceptancesuite.js
```diff
@@ -63,6 +63,31 @@ test('navigating and refreshing mantains that page number', async t => {
...
+  await paginationComponent.clickNextButton();
+  let pageNum = await paginationComponent.getActivePageLabelAndNumber();
+  await t.expect(pageNum).eql('Page 2');
+
+  const spellCheckComponent = VerticalPage.getSpellCheckComponent();
+  await spellCheckComponent.clickLink();
+  pageNum = await paginationComponent.getActivePageLabelAndNumber();
+  await t.expect(pageNum).eql('Page 1');
+
+  // Check that clicking spell check sends a queryTrigger=suggest url param
```
cea2aj, line 85: This part of the test feels like it should be an integration test rather than an acceptance test because it is not directly related to user flow. What are your thoughts?

tests/acceptance/acceptancesuite.js
```diff
@@ -63,6 +63,31 @@ test('navigating and refreshing mantains that page number', async t => {
...
 
+const spellCheckLogger = RequestLogger({
+  url: /v2\/accounts\/me\/answers\/vertical\/query/
+});
+test.requestHooks(spellCheckLogger)('spell check flow', async t => {
+  const searchComponent = VerticalPage.getSearchComponent();
+  await searchComponent.enterQuery('varginia');
+  await searchComponent.submitQuery();
+
+  // Check that clicking spell check resets pagination to page 1
```
cea2aj, line 74: What do you think about moving this comment down to the next block of code? (above line 80)

tests/acceptance/acceptancesuite.js
tmeyer2115: As with our other item this Sprint, if we can't easily express this as an integration test (or unit test), I think we should keep it in the acceptancesuite for now. Make a TODO though indicating the test should be migrated once we improve our unit/integration testing abilities. This might be hard to do as anything but an acceptance test. I think we'd need to be able to mock the Core dependency given to the SpellCheck component.

tests/acceptance/acceptancesuite.js
oshi97: I kind of like it up here, since the only reason I'm changing the page from 1 to 2 is to test this, but don't really have strong feelings

tests/acceptance/acceptancesuite.js
oshi97: I can add a todo! I tried to think of ways to test this in jest, and I think it would be kind of difficult, since clicking the spell check suggestion (like varginia -> virginia) does a full page refresh and reloads all of the components. Probably still something we can set up in jest but requires a little more thought I think.

tmeyer2115 (approved):
lgtm, pending Connor taking a look at your responses!

tests/acceptance/acceptancesuite.js
cea2aj: ah, got it

cea2aj (approved):

PR #1231, opened 2021-01-12
Pagination storage tests

cea2aj:
Test the pagination component's interaction with global and persistent storage

J=SLAP-1005
TEST=auto

### DIFF
--- a/tests/ui/components/results/paginationcomponent.js
+++ b/tests/ui/components/results/paginationcomponent.js
@@ -161,12 +161,31 @@ describe('properly interacts with storage', () => {
     };
   });
 
-  it('the limit property is set to the global storage search-config.limit after being created', () => {
+  it('the global storage search-config.limit and results.resultsCount are used to calculate the total page count', () => {
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.SEARCH_CONFIG, { limit: 5 });
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.VERTICAL_RESULTS, { searchState: 'search-complete', resultsCount: 20 });
+
     const component = COMPONENT_MANAGER.create('Pagination', defaultConfig);
 
-    const limit = component._limit;
+    const wrapper = mount(component);
 
-    expect(limit).toEqual(5);
+    const paginationLink = wrapper.find('.yxt-Pagination-link').first();
+    const paginationInfo = JSON.parse(paginationLink.prop('data-eventoptions'));
+
+    expect(paginationInfo.totalPageCount).toEqual(4);
+  });
+
+  it('the global storage search-config.limit and search-offset are used to calculate the current page number', () => {
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, 10);
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.SEARCH_CONFIG, { limit: 5 });
+
+    const component = COMPONENT_MANAGER.create('Pagination', defaultConfig);
+
+    const wrapper = mount(component);
+
+    const activePageSpan = wrapper.find('#active-page');
+
+    expect(activePageSpan.text()).toEqual('Page 3');
   });
 
   it('updating the page sets global storage searchOffset', () => {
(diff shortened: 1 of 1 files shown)


tests/ui/components/results/paginationcomponent.js
```diff
@@ -144,3 +147,48 @@ describe('rendering the page numbers', () => {
     expect(paginate).toHaveBeenCalledWith(2, 1, 5);
   });
 });
+
+describe('properly interacts with storage', () => {
```
tmeyer2115, line 151: Since we're not using an actual instance of `PersistentStorage` or `GlobalStorage`, I think these would qualify as unit tests. As we discussed last night, though, that's fine! We should just make that clear in the commit message.

tests/ui/components/results/paginationcomponent.js
```diff
@@ -144,3 +147,48 @@ describe('rendering the page numbers', () => {
...
+
+    expect(limit).toEqual(5);
+  });
+
+  it('updating the page sets global storage searchOffset', () => {
+    const component = COMPONENT_MANAGER.create('Pagination', {
+      ...defaultConfig
+    });
+
+    component.updatePage(5);
```
oshi97: if possible can we do this through the component ui with simulated clicks

tests/ui/components/results/paginationcomponent.js
```diff
@@ -144,3 +147,48 @@ describe('rendering the page numbers', () => {
...
+    const searchOffset = component.core.globalStorage.getState(StorageKeys.SEARCH_OFFSET);
+
+    expect(searchOffset).toEqual(5);
+  });
+
+  it('updating the page sets persistent storage searchOffset', () => {
+    const component = COMPONENT_MANAGER.create('Pagination', {
+      ...defaultConfig
```
oshi97: nit but don't need the spread here (also above)

tests/ui/components/results/paginationcomponent.js
```diff
@@ -144,3 +147,48 @@ describe('rendering the page numbers', () => {
...
+      verticalKey: 'verticalKey'
+    };
+  });
+
+  it('the limit property is set to the global storage search-config.limit after being created', () => {
+    const component = COMPONENT_MANAGER.create('Pagination', {
+      ...defaultConfig
+    });
+
+    const limit = component._limit;
```
oshi97: could we test this through the ui instead of the variable value?

tests/ui/components/results/paginationcomponent.js
cea2aj: sure

tests/ui/components/results/paginationcomponent.js
cea2aj: This value doesn't show up in the UI so i'm not sure if testing it in the UI makes sense

oshi97: would it be possible to test that this is working by mocking out the results count and search offset?

cea2aj: I could test it indirectly by mocking resultsCount and searchOffset and checking if the nextPageButton is shown. Is that what you mean? Whether or not nextPageButton is shown depends on resultsCount, searchOffset, and limit

oshi97: it looks like limit is used to calculate stuff like the current and max page number, so maybe testing for those values in the ui? I think it would be just a little more robust

cea2aj: sure

tests/ui/components/results/paginationcomponent.js
```diff
@@ -144,3 +147,69 @@ describe('rendering the page numbers', () => {
...
+
+    defaultConfig = {
+      container: '#test-component',
+      verticalKey: 'verticalKey'
+    };
+  });
+
+  it('the global storage search-config.limit and results.resultsCount are used to calculate the total page count', () => {
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.SEARCH_CONFIG, { limit: 5 });
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.VERTICAL_RESULTS, { searchState: 'search-complete', resultsCount: 20 });
```
oshi97: nitty but line length here

oshi97 (commented):
besides the one nit I like these a lot!

tmeyer2115 (approved):

PR #1232, opened 2021-01-12
Vertical results storage integration tests

cea2aj:
Test the verticalresults component's interaction with global storage

J=SLAP-1005
TEST=auto

### DIFF
--- a/tests/ui/components/results/verticalresultscomponent.js
+++ b/tests/ui/components/results/verticalresultscomponent.js
@@ -4,43 +4,19 @@ import mockManager from '../../../setup/managermocker';
 import StorageKeys from '../../../../src/core/storage/storagekeys';
 import SearchStates from '../../../../src/core/storage/searchstates';
 import VerticalResultsComponent from '../../../../src/ui/components/results/verticalresultscomponent';
-
-const updateEvents = {};
+import GlobalStorage from '../../../../src/core/storage/globalstorage';
 
 const mockCore = {
-  globalStorage: {
-    set: (storageKey, value) => {
-      if (storageKey in updateEvents) {
-        updateEvents[storageKey](value);
-      }
-    },
-    on: (event, storageKey, cb) => {
-      if (event === 'update') {
-        updateEvents[storageKey] = cb;
-      }
-    },
-    getState: (storageKey) => {
-      if (storageKey === StorageKeys.VERTICAL_PAGES_CONFIG) {
-        return { get: () => { return []; } };
-      } else if (storageKey === StorageKeys.NO_RESULTS_CONFIG) {
-        return {};
-      } else if (storageKey === StorageKeys.API_CONTEXT) {
-        return undefined;
-      } else if (storageKey === StorageKeys.REFERRER_PAGE_URL) {
-        return '';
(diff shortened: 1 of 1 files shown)


tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -48,6 +61,29 @@ describe('vertical results component', () => {
     };
   });
 
+  it('the query propery is based on global storage when the component is created', () => {
```
tmeyer2115: query property?

tmeyer2115:
Again, I think these are more-so unit tests than integration tests. We should make that clear in the commit message.

tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -2,12 +2,23 @@ import DOM from '../../../../src/ui/dom/dom';
...
+import SearchStates from '../../../../src/core/storage/searchstates';
 import VerticalResultsComponent from '../../../../src/ui/components/results/verticalresultscomponent';
 
+const updateEvents = {};
+
 const mockCore = {
   globalStorage: {
-    set: () => {},
-    on: () => {},
+    set: (storageKey, value) => {
```
oshi97: at this point would it be better to use the real global storage? or were there issues with that

tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -48,6 +61,29 @@ describe('vertical results component', () => {
...
+      searchState: SearchStates.SEARCH_COMPLETE,
+      resultsCount: 15,
+      results: 'yext search results',
+      verticalConfigId: 'verticalKey',
+      resultsContext: {}
+    };
+
+    component.core.globalStorage.set(StorageKeys.VERTICAL_RESULTS, verticalResults);
+
+    expect(component.results).toEqual('yext search results');
```
oshi97: would it be possible to update this test to expect the component html or state to be a certain way, instead of an internal variable?

cea2aj: yes

tests/ui/components/results/verticalresultscomponent.js
cea2aj: The component has a query property which it gets from global storage. If we don't think the test is that useful I can remove it

tests/ui/components/results/verticalresultscomponent.js
cea2aj: that actually worked okay so i'll swap over

tests/ui/components/results/verticalresultscomponent.js
cea2aj: I'll update this test to the hopefully more useful: 'getVerticalUrl encodes the global storage query when the component is created'

tests/ui/components/results/verticalresultscomponent.js
cea2aj: updating to get get this value from the state

cea2aj:
> Again, I think these are more-so unit tests than integration tests. We should make that clear in the commit message.

Per Oliver's suggestion, I swapped to the real GlobalStorage so that these are integration tests

tmeyer2115 (approved):

PR #1234, opened 2021-01-12
Spellcheck link tests

cea2aj:
Add tests which verify the query params on spellcheck links

J=SLAP-1005

TEST=auto

### DIFF
--- a/tests/ui/components/search/spellcheckcomponent.js
+++ b/tests/ui/components/search/spellcheckcomponent.js
@@ -2,16 +2,10 @@ import DOM from '../../../../src/ui/dom/dom';
 import mockManager from '../../../setup/managermocker';
 import StorageKeys from '../../../../src/core/storage/storagekeys';
 import SpellCheckComponent from '../../../../src/ui/components/search/spellcheckcomponent';
-
-const globalStorage = {};
+import GlobalStorage from '../../../../src/core/storage/globalstorage';
 
 const mockCore = {
-  globalStorage: {
-    getState: key => globalStorage[key] || undefined,
-    set: (key, value) => {
-      globalStorage[key] = value;
-    }
-  }
+  globalStorage: new GlobalStorage()
 };
 
 const COMPONENT_MANAGER = mockManager(mockCore);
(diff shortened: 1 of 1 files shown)


tests/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+};
+
+const COMPONENT_MANAGER = mockManager(mockCore);
+
+describe('spellcheck redirect links', () => {
+  let linkUrlParams;
+
+  beforeEach(() => {
```
oshi97: Is it necessary to run this before each test? I kind of like the thought process but not sure if it's necessary for this specific case?

cea2aj: ah, ya I can probably remove the beforeEach and just put this in the describe block

tests/ui/components/search/spellcheckcomponent.js
```diff
@@ -0,0 +1,58 @@
...
+import SpellCheckComponent from '../../../../src/ui/components/search/spellcheckcomponent';
+import GlobalStorage from '../../../../src/core/storage/globalstorage';
+
+const mockCore = {
+  globalStorage: new GlobalStorage()
+};
+
+const COMPONENT_MANAGER = mockManager(mockCore);
+
+describe('spellcheck redirect links', () => {
```
tmeyer2115, line 13: While this test coverage is definitely good to have, is it really related to Storage? If it is, the relationship feels tangential. Again, more test coverage is always better. This item is fairly large in terms of scope already. So, I think we need to limit ourselves to tests that verify interactions with storage.

tmeyer2115 (approved):

tests/ui/components/search/spellcheckcomponent.js
cea2aj: That's correct, these tests cover the encoding of persistent storage rather than a direct interaction with persistent storage. It was one of the use cases that Christian pointed out so I added some coverage, however I will stick to more direct usage of storage in the future

oshi97 (approved):

PR #1235, opened 2021-01-13
Cut over SORT_BYS to new global storage

oshi97:
Cuts over SORT_BYS to the new global storage
(chosen because it's the simplest storage key)
and adds the new storage to core.

J=SLAP-1019
TEST=manual

chrome + ie11: test that the sort component still sorts, and passes the correct
sort_bys to the liveapi query

### DIFF
--- a/src/core/storage/storagekeys.js
+++ b/src/core/storage/storagekeys.js
@@ -30,7 +30,7 @@ export default {
   SESSIONS_OPT_IN: 'sessions-opt-in',
   VERTICAL_PAGES_CONFIG: 'vertical-pages-config',
   LOCALE: 'locale',
-  SORT_BYS: 'sort-bys', // Cut over to new global storage
+  SORT_BYS: 'sort-bys', // Has been cut over to new global storage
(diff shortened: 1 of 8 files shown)


src/core/storage/storagekeys.js
```diff
@@ -30,7 +30,7 @@ export default {
   SESSIONS_OPT_IN: 'sessions-opt-in',
   VERTICAL_PAGES_CONFIG: 'vertical-pages-config',
   LOCALE: 'locale',
-  SORT_BYS: 'sort-bys',
+  SORT_BYS: 'sort-bys', // Cut over to new global storage
```
cea2aj: nit: this almost reads like a todo when really it's just a note. How about '// Has been cut over to the new global storage'?

cea2aj:
Does it make sense to push an npm build of answers-storage so that these tests can pass? Is there another way?

src/core/storage/storagekeys.js
oshi97: I like that, will update in the next commit!

tmeyer2115 (approved):

PR #1236, opened 2021-01-13
Navigation integration tests

cea2aj:
Ensure that tab links encode tabOrder, referrerPageUrl, and context from global storage

Update the mockmanager to call component setState() on updates to keys matching a component's moduleId

J=SLAP-1005
TEST=auto

### DIFF
--- a/tests/setup/mockcomponentmanager.js
+++ b/tests/setup/mockcomponentmanager.js
@@ -42,11 +42,11 @@ export default class MockComponentManager {
 
     const component = new COMPONENT_REGISTRY[componentType](componentConfig, systemConfig).init(componentConfig);
 
-    if (component.moduleId === undefined || component.moduleId === null) {
-      return component;
-    }
-
     if (this.core && this.core.globalStorage) {
+      if (component.moduleId === undefined || component.moduleId === null) {
+        return component;
+      }
+
       this.core.globalStorage
         .on('update', component.moduleId, (data) => {
           component.setState(data);
--- a/tests/setup/mockcomponentmanager.js
+++ b/tests/setup/mockcomponentmanager.js
@@ -41,6 +41,18 @@ export default class MockComponentManager {
     };
 
     const component = new COMPONENT_REGISTRY[componentType](componentConfig, systemConfig).init(componentConfig);
+
+    if (component.moduleId === undefined || component.moduleId === null) {
+      return component;
+    }
+
+    if (this.core && this.core.globalStorage) {
+      this.core.globalStorage
+        .on('update', component.moduleId, (data) => {
+          component.setState(data);
+        });
+    }
+
     return component;
   }
 
--- a/tests/ui/components/navigation/navigationcomponent.js
+++ b/tests/ui/components/navigation/navigationcomponent.js
@@ -158,6 +158,21 @@ describe('navigation tab links are correct', () => {
     expect(tabLink).toContain('context=some+context');
   });
 
+  it('updating the context in global storage updates the context in tab links', () => {
+    const component = COMPONENT_MANAGER.create('Navigation', defaultConfig);
+    const wrapper = mount(component);
+
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.API_CONTEXT, 'new context');
+
+    // Re-render because the component state changed
+    wrapper.update();
+
+    const navItem = wrapper.find('.js-yxt-navItem').first();
+    const tabLink = navItem.prop('href');
+
+    expect(tabLink).toContain('context=new+context');
+  });
+
   it('tab links default to the tab order from global storage', () => {
     const verticalPagesConfig = new VerticalPagesConfig([
       { label: 'Home', url: './index.html' },
--- a/tests/ui/components/navigation/navigationcomponent.js
+++ b/tests/ui/components/navigation/navigationcomponent.js
@@ -1,5 +1,11 @@
 import DOM from '../../../../src/ui/dom/dom';
 import { Tab } from '../../../../src/ui/components/navigation/navigationcomponent';
+import mockManager from '../../../setup/managermocker';
+import { mount } from 'enzyme';
+import GlobalStorage from '../../../../src/core/storage/globalstorage';
+import PersistentStorage from '../../../../src/ui/storage/persistentstorage';
+import StorageKeys from '../../../../src/core/storage/storagekeys';
+import VerticalPagesConfig from '../../../../src/core/models/verticalpagesconfig';
 
 // The DOM doesn't exist within components in the JEST environment,
 // so we have to provide it to our DOM API properly.
@@ -14,7 +20,11 @@ beforeEach(() => {
   DOM.empty(bodyEl);
 
   // Create the container that our component will be injected into
-  DOM.append(bodyEl, DOM.createEl('div', { class: 'test-component' }));
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+
+  // Mock DOM.on and DOM.off because they will not work in this test environment
+  DOM.on = jest.fn();
+  DOM.off = jest.fn();
 });
 
 describe('navigation component configuration', () => {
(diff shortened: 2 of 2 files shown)


tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -102,3 +112,70 @@ describe('navigation component configuration', () => {
...
+    ]);
+
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.VERTICAL_PAGES_CONFIG, verticalPagesConfig);
+
+    const component = COMPONENT_MANAGER.create('Navigation', {
+      container: '#test-component',
+      verticalKey: 'verticalKey',
+      verticalPages: null
+    });
+    const wrapper = mount(component);
```
tmeyer2115, line 189: How slow is the component creation and `mount`? If performance isn't great, we can consolidate things by setting `globalStorage` and creating the component once. Then, we'd have three tests, one to verify each property was present in the URL. If performance isn't a problem, no worries.

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -102,3 +112,70 @@ describe('navigation component configuration', () => {
...
+    const component = COMPONENT_MANAGER.create('Navigation', defaultConfig);
+    const wrapper = mount(component);
+
+    const navItem = wrapper.find('.js-yxt-navItem').first();
+    const tabLink = navItem.prop('href');
+
+    expect(tabLink).toContain('referrerPageUrl=yext.com');
+  });
+
+  it('tab links contain the context from global storage', () => {
```
tmeyer2115, line 149: Could we also make a test that handles `setContext` being called after the component is initially mounted? In that case, the links should be re-rendered to include the new context.

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -102,3 +112,70 @@ describe('navigation component configuration', () => {
...
+    const component = COMPONENT_MANAGER.create('Navigation', defaultConfig);
+    const wrapper = mount(component);
+
+    const navItem = wrapper.find('.js-yxt-navItem').first();
+    const tabLink = navItem.prop('href');
+
+    expect(tabLink).toContain('context=some+context');
+  });
+
+  it('tab links default to the tab order from global storage', () => {
```
tmeyer2115, line 176: The back-end can return a new tab order as part of a vertical search response. Could we have a test that verifies the component re-renders properly when a new tab order comes in after it's mounted?

tests/ui/components/navigation/navigationcomponent.js
cea2aj: Sure. I don't think the tests have access to ANSWERS.setContext(), however can try it with globalStorage.set(StorageKeys.API_CONTEXT, context).

tests/ui/components/navigation/navigationcomponent.js
cea2aj: Yes, I was thinking of making that an acceptance test. Also, do you have an example query that re-orders the tabs on the slapshot test account? I wasn't able to find one

tests/ui/components/navigation/navigationcomponent.js
cea2aj: The time spent on component creation and mount seems negligible. We do that in lots of test (about 60), and our tests are relatively quick (about 6-7 seconds for the whole test suite)

tests/ui/components/navigation/navigationcomponent.js
tmeyer2115: I don't have one at hand. @creotutar might be more familiar. I believe the response just sends verticals in a particular order. We then set that order as the `tabOrder`. I think the relevant code is here: https://github.com/yext/answers/blob/master/src/core/models/navigation.js. The storage key is `StorageKeys.NAVIGATION`.

tmeyer2115: For now, it may be sufficient to just write a Jest test that verifies that behavior. Mount the component, update `StorageKeys.NAVIGATION`, and verify the links re-render properly.

cea2aj: I added a tab order test

tests/setup/mockcomponentmanager.js
```diff
@@ -41,6 +41,18 @@ export default class MockComponentManager {
     };
 
     const component = new COMPONENT_REGISTRY[componentType](componentConfig, systemConfig).init(componentConfig);
+
+    if (component.moduleId === undefined || component.moduleId === null) {
+      return component;
+    }
+
+    if (this.core && this.core.globalStorage) {
```
oshi97, line 45: would it be a little cleaner to  move the moduleId checks into this second if statement? and are there any tests that don't give components a this.core.globalStorage? we might be able to get rid of the check for this.core.globalStorage (maybe we even want it to bug out)

oshi97: actually if you're just trying to match the current logic this makes sense!

oshi97: this ~~could be slightly~~ is probably out of scope, but maybe we want to use the actual ComponentManager in MockComponentManager? At a quick glance it looks like the main differences are adding the 'enzyme-component' css class and setting the handlebars renderer?

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -102,3 +112,131 @@ describe('navigation component configuration', () => {
...
+  });
+
+  it('updating the context in global storage updates the context in tab links', () => {
+    const component = COMPONENT_MANAGER.create('Navigation', defaultConfig);
+    const wrapper = mount(component);
+
+    COMPONENT_MANAGER.core.globalStorage.set(StorageKeys.API_CONTEXT, 'new context');
+
+    // Re-render because the component state changed
+    wrapper.update();
```
oshi97, line 168: shouldn't the component rerender on it's own when API_CONTEXT is updated?

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -102,3 +112,131 @@ describe('navigation component configuration', () => {
...
+  it('tab order changes in response to the server', () => {
+    const component = COMPONENT_MANAGER.create('Navigation', defaultConfig);
+    const wrapper = mount(component);
+
+    const navStateWithReversedTabs = {
+      tabOrder: ['third', 'second', 'first']
+    };
+
+    component.core.globalStorage.set(StorageKeys.NAVIGATION, navStateWithReversedTabs);
+    wrapper.update();
```
oshi97, line 227: same question here

tests/ui/components/navigation/navigationcomponent.js
```diff
@@ -14,7 +20,11 @@ beforeEach(() => {
   DOM.empty(bodyEl);
 
   // Create the container that our component will be injected into
-  DOM.append(bodyEl, DOM.createEl('div', { class: 'test-component' }));
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+
+  // Mock DOM.on and DOM.off because they will not work in this test environment
+  DOM.on = jest.fn();
+  DOM.off = jest.fn();
```
oshi97, line 27: how come DOM.on and off don't work here? I thought we use this in other tests, but haven't looked at this in a while

tests/ui/components/navigation/navigationcomponent.js
oshi97: *I looked a little more into this (trying to write some unit tests of my own) and I think in the past we never tried to test how the component reacts to state changes in global storage (since we were using a mocked out one) and would set the state prior to mounting. A few tests do simulate clicks, and I'm not quite sure how those are working, but I think it could be related to AnswersAdapter (enzymeadapter.js)

tests/ui/components/navigation/navigationcomponent.js
oshi97: So I ALSO ran into this issue (when trying to write DirectAnswer tests), and my guess is that the jest DOM is not being kept in sync with how a real DOM would behave, and causes these event listener registrations to blow up because they're expecting an element to be there.

tests/setup/mockcomponentmanager.js
cea2aj: I'll update this to match your first suggestion, as that matches the current logic in the real ComponentManager. As for using the real ComponentManager, I don't think there's an immediate need but we can always do that if needed

tests/ui/components/navigation/navigationcomponent.js
cea2aj: Based on the jsdoc for `update()`, it seems like this is an appropriate use case for it. Here's what the jsdoc says: 'Forces a re-render. Useful to run before checking the render output if something external may be updating the state of the component somewhere.'

I also did some manual testing and it doesn't cause setState() to be called additional times which suggests that the answers component re-render lifecycle isn't occurring. So I believe it's just the ReactWrapper component updating rather than the answers component updating

tests/ui/components/navigation/navigationcomponent.js
cea2aj: see previous comment: https://github.com/yext/answers/pull/1236#discussion_r558638899

tests/ui/components/navigation/navigationcomponent.js
cea2aj: That would make sense as the error I'm getting when I don't disable DOM.on and DOM.off is: 'Error initializing component (Navigation) Caused By: '[object Window]' is not a valid selector (unknown)'

tests/ui/components/navigation/navigationcomponent.js
oshi97: Yeah, I think this is functionality our testing framework is lacking is lacking

oshi97 (approved):
lgtm!

PR #1238, opened 2021-01-14
cut over NAVIGATION to new global storage

oshi97:
Also adds support for component.moduleId in ComponentManager using
the new storage, and adds support for new core in managermocker.js
for our jest tests.

J=SLAP-1019
TEST=manual

test that searching "amani farooque phone number" will reorder
the navigation component tabs to have the people vertical first
(this is with the rosetest experience)

### DIFF
--- a/src/ui/components/componentmanager.js
+++ b/src/ui/components/componentmanager.js
@@ -4,6 +4,8 @@ import { AnswersComponentError } from '../../core/errors/errors';
 import DOM from '../dom/dom';
 import { COMPONENT_REGISTRY } from './registry';
 
+/** @typedef {import('../../core/core').default} Core */
+
 /**
  * ComponentManager is a Singletone that contains both an internal registry of
  * eligible components to be created, as well as keeps track of the current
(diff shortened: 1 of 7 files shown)


src/ui/components/componentmanager.js
```diff
@@ -152,6 +154,16 @@ export default class ComponentManager {
         return component;
       }
 
+      this._core.storage.registerListener({
+        eventType: 'update',
+        storageKey: component.moduleId,
+        callback: (data) => {
+          component.setState(data);
```
tmeyer2115, line 161: Are we going to have two, duplicate `setState` calls on certain updates? The duplicate calls would happen only for those `StorageKeys` that have been cut-over I think. Once the full cut-over is complete, the we'd just have the one listener per key?

tmeyer2115 (approved):

src/ui/components/componentmanager.js
oshi97: That's a valid concern! The plan is to fully cut over each storage key. As long as a storage key like NAVIGATION is fully cut over in a single PR, there's no chance for duplicate updates, since all older setState() calls will be updated to the new set() calls

tmeyer2115: Got it! Sounds good!

PR #1239, opened 2021-01-14
cut over VERTICAL_RESULTS to new global storage

oshi97:
Cuts over StorageKeys.VERTICAL_RESULTS to the new global storage.
Also adds some jest tests to make sure components are still
listening to updates to VERTICAL_RESULTS.

J=SLAP-1019
TEST=maunal,auto

add unit tests
test that the AppliedFilters, VerticalResultsCount, VerticalResults,
QASubmission, Pagination, and SortOptions component still rerender on vertical
results change. For SortOptions checked that it still is hidden
for the no-results state.


src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -32,12 +32,16 @@ export default class SortOptionsComponent extends Component {
...
+    this.core.storage.registerListener({
+      eventType: 'update',
+      storageKey: StorageKeys.VERTICAL_RESULTS,
+      callback: verticalResults => {
+        const isSearchComplete = verticalResults.searchState === SearchStates.SEARCH_COMPLETE;
+
+        if (isSearchComplete) {
+          const isNoResults = verticalResults.resultsContext === ResultsContext.NO_RESULTS;
+          this.handleVerticalResultsUpdate(isNoResults);
+        }
```
cea2aj, line 44: i love the new interface

tests/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -0,0 +1,36 @@
...
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+
+  defaultConfig = {
+    container: '#test-component',
+    entityId: 1234
+  };
+});
+
+describe('QASubmission component', () => {
+  it('listens to updates to VERTICAL_RESULTS in global storage', () => {
```
cea2aj, line 23: This is testing that the component appears only when there is at least one result, correct?

tests/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -0,0 +1,36 @@
...
+    const storage = COMPONENT_MANAGER.core.storage;
+    const component = COMPONENT_MANAGER.create(QuestionSubmissionComponent.type, defaultConfig);
+    const wrapper = mount(component);
+    expect(wrapper.find('.yxt-QuestionSubmission')).toHaveLength(0);
```
cea2aj: nit: Assuming these tests are testing whether or not the component is visible, what are your thoughts on an explanatory variable so that the code reads: `expect(componentIsVisible).toBeFalsy()` and `expect(componentIsVisible).toBeTruthy()`? The variable would be `const componentIsVisible = wrapper.find('.yxt-QuestionSubmission').length >= 1`

tests/ui/components/questions/questionsubmissioncomponent.js
oshi97: oh, results count doesn't matter here! I'll remove results and reusltscount from the data to make that clear

tests/ui/components/questions/questionsubmissioncomponent.js
oshi97: we can use .exists() here, would that be better? (updated the code with this)

cea2aj: Ya, that looks good!

cea2aj (approved):

PR #1243, opened 2021-01-15
Fix search bar buttons on very small screens

cea2aj:
Fix the search bar styling so that the buttons don't disappear on small screens

On screens with very small widths (<350px width), the search bar clear and search buttons shrink and eventually disappear as the screen width decreases. This PR prevents the buttons from shrinking and disappearing on very small screens.

J=SLAP-999
TEST=visual

Build a site and view it at various sizes between 300px and 400px wide and confirm that the search bar buttons no longer shrink or disappear. Test on a browswerstack iphone 8 with safari and on a windows laptop on IE11.

### DIFF
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -87,6 +87,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
     background-color: transparent;
     flex-grow: 1;
+    width: 100%;
     padding-top: calc(var(--yxt-base-spacing) / 2);
     padding-bottom: calc(var(--yxt-base-spacing) / 2);
     padding-left: var(--yxt-base-spacing);
@@ -121,6 +122,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-clear
   {
     display: flex;
+    flex: 0 0 auto;
     align-items: center;
     font: inherit;
     padding-top: calc(var(--yxt-base-spacing) / 2);
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -86,7 +86,6 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     border: none;
 
     background-color: transparent;
-    flex-grow: 1;
     width: 100%;
     padding-top: calc(var(--yxt-base-spacing) / 2);
     padding-bottom: calc(var(--yxt-base-spacing) / 2);
@@ -122,7 +121,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-clear
   {
     display: flex;
-    flex: 0 0 auto;
+    flex-shrink: 0;
     align-items: center;
     font: inherit;
     padding-top: calc(var(--yxt-base-spacing) / 2);


src/ui/sass/modules/_SearchBar.scss
```diff
@@ -87,6 +87,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
     background-color: transparent;
     flex-grow: 1;
```
oshi97: do we still need flex-grow: 1 if we have width: 100% now?

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -121,6 +122,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-clear
   {
     display: flex;
+    flex: 0 0 auto;
```
oshi97: would just setting flex-shrink: 0 here work? (did a quick search and the default is 0 1 auto) for my edification why do we need flex-shrink: 0 here?

oshi97 (commented):
could you double check that this works on ie11 as well, if you haven't already? specifically with a ui refresh site

src/ui/sass/modules/_SearchBar.scss
cea2aj: no, it's actually not needed

src/ui/sass/modules/_SearchBar.scss
cea2aj: You right, all we need is flex-shrink: 0. It prevents the buttons from shrinking. I used the full flex syntax because a  guide recommended using that syntax, however I can switch

cea2aj:
> could you double check that this works on ie11 as well, if you haven't already? specifically with a ui refresh site

Good call. I just tested it and it didn't work with ie11 so I'm going to look into an alternative solution. The UI refresh without this change does work on IE11

src/ui/sass/modules/_SearchBar.scss
cea2aj: updated to just have the width: 100%

cea2aj:
> > could you double check that this works on ie11 as well, if you haven't already? specifically with a ui refresh site
> 
> Good call. I just tested it and it didn't work with ie11 so I'm going to look into an alternative solution. The UI refresh without this change does work on IE11

Update: Discussed offline. The change does in fact work in IE11. The issue I was seeing was a separate CORS issue that occurred because how how I was serving the SDK locally. (SLAP-1024)

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -121,6 +121,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-clear
   {
     display: flex;
+    flex-shrink: 0;
```
tmeyer2115, line 124: Since we removed the `flex-grow` from `&-input` is this still necessary?

cea2aj: I tested this PR without flex-shrink it and the buttons still shrink at low resolutions. The buttons and the search bar elements are all flex items because their parent (the search bar form) has `display: flex;` set. By default the flex items have a flex-shrink of 1, which makes them susceptible to shrinking. Setting flex-shrink to 0 prevents that.

oshi97 (approved):
lgtm pending tom!

tmeyer2115 (approved):

PR #1245, opened 2021-01-15
cut over AUTOCOMPLETE to new global storage

oshi97:
J=SLAP-1019
TEST=manual

test that autocomplete still works for universal and vertical search

test that I can search for a filter in filtersearch, apply it, and
that the network tab shows a filter was applied to the query

test that I can search for a filter in the geolocation filter,
apply it, and that the network tab shows a filter was applied to the query


cea2aj (approved):
🔥

PR #1246, opened 2021-01-15
cut over FILTER to new global storage

oshi97:
J=SLAP-1019
TEST=manual

Test that I can use filtersearch to select a filter,
then if I refresh the page, the filter will still be applied
to the page, and also be displayed in the applied filters bar


cea2aj (approved):
lgtm

PR #1249, opened 2021-01-19
Cut over UNIVERSAL_RESULTS to new global storage

oshi97:
J=SLAP-1019
TEST=manual,auto

test that a universal page can still run searches, and that
I can search for "virginia", search for "office space", and then
hitting the back button will take me back to virginia

check that the qa submission component rerenders with universal
results, and is hidden while the search is loading

add tests for UniversalResults and QASubmission for listening
to UNIVERSAL_RESULTS in global storage


tests/ui/components/results/universalresultscomponent.js
```diff
@@ -0,0 +1,33 @@
...
+  DOM.empty(bodyEl);
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+
+  defaultConfig = {
+    container: '#test-component'
+  };
+});
+
+describe('UniversalResults component', () => {
+  it('listens to updates to UNIVERSAL_RESULTS in global storage', () => {
```
tmeyer2115: Do we already have a test for the `Component` base class that verifies it re-renders when the `storage.[moduleId]` is updated? If not, I think that would be useful. It would save us needing to write these kinds of tests for every component.

oshi97: no, but that makes sense, I can add it instead of this!

tmeyer2115 (approved):

PR #1250, opened 2021-01-19
FilterOptions integration test

cea2aj:
Test the FilterOptions component's usage of globalStorage.delete()

J=SLAP-1005
TEST=auto

### DIFF
--- a/tests/ui/components/filters/filteroptionscomponent.js
+++ b/tests/ui/components/filters/filteroptionscomponent.js
@@ -7,6 +7,7 @@ import FilterMetadata from 'src/core/filters/filtermetadata';
 import FilterType from '../../../../src/core/filters/filtertype';
 import StorageKeys from '../../../../src/core/storage/storagekeys';
 import PersistentStorage from 'src/ui/storage/persistentstorage';
+import GlobalStorage from '../../../../src/core/storage/globalstorage';
 
 describe('filter options component', () => {
   DOM.setup(document, new DOMParser());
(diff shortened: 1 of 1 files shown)


tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -657,4 +658,42 @@ describe('filter options component', () => {
...
+        container: '#test-component',
+        control: 'singleoption',
+        options: options,
+        label: filterOptionsLabel
+      };
+
+      globalStorage = COMPONENT_MANAGER.core.globalStorage;
+    });
+
+    it('deletes the global storage FilterOptions key when constructed', () => {
```
oshi97, line 689: could we update this to check it deletes the storage key equal to the component name you give it through config?
would it be okay to get rid of the extra line breaks too?

tmeyer2115, line 689: This test should help Oliver as he continues on in his refactor. But, I'm not sure how useful the test is once the refactor is complete. We're testing essentially one line of source code here. That line is a fairly low-level implementation detail. Ideally, we would test some higher-level behavior of the component and in so doing, verify a particular storage interaction. For example, Oliver wrote a test to make sure the `QuestionSubmissionComponent` is hidden while a search is loading. The test has the side effect of verifying that the component correctly receives `UNIVERSAL_RESULTS` storage updates. Is there some higher-level behavior we could test here that would have the side effect of verifying the `delete`?

cea2aj: I can test that the clear button of the search component deletes the search offset from global storage, so I'll close this PR

PR #1251, opened 2021-01-19
Search bar clear button integration tests

cea2aj:
Add tests that verify that the search bar clear button properly interacts with global and persistent storage

J=SLAP-1005
TEST=auto

### DIFF
--- a/tests/ui/components/search/searchcomponent.js
+++ b/tests/ui/components/search/searchcomponent.js
@@ -16,11 +16,14 @@ describe('SearchBar component', () => {
   const COMPONENT_MANAGER = mockManager({
     globalStorage: new GlobalStorage(),
     persistentStorage: new PersistentStorage(),
-    setQuery: jest.fn(),
-    autoCompleteVertical: jest.fn()
+    setQuery: function (query) {
+      this.globalStorage.set(StorageKeys.QUERY, query);
+    },
+    autoCompleteVertical: jest.fn(() => Promise.resolve())
   });
 
   const globalStorage = COMPONENT_MANAGER.core.globalStorage;
+  const persistentStorage = COMPONENT_MANAGER.core.persistentStorage;
 
   beforeEach(() => {
     let bodyEl = DOM.query('body');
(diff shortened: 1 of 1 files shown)


tmeyer2115:
> search barf

Connor, the SDK's not that bad, come on.

tests/ui/components/search/searchcomponent.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const globalStorage = COMPONENT_MANAGER.core.globalStorage;
+
+  beforeEach(() => {
+    let bodyEl = DOM.query('body');
+    DOM.empty(bodyEl);
+    DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+  });
+
+  it('clicking the clear button deletes SEARCH_OFFSET from global storage', () => {
```
tmeyer2115: I think the clear button does a couple of other things too. Since we're here already, could we make a `clicking the clear button updates storage correctly` test or something like that? It would make sure `SEARCH_OFFSET` is removed from global and persistent storage, `QUERY` is set to `''`, etc.

cea2aj: sure, what do you think about making that a describe block and having multiple tests for each of those storage checks. That way we can keep the tests clean with one or two expects per test

tmeyer2115: Sure, that sounds good!

tests/ui/components/search/searchcomponent.js
```diff
@@ -0,0 +1,87 @@
...
+      persistentStorage.set(StorageKeys.SEARCH_OFFSET, 15);
+      expect(persistentStorage.get(StorageKeys.SEARCH_OFFSET)).toBeTruthy();
+
+      const clearButton = wrapper.find('.js-yxt-SearchBar-clear');
+      clearButton.simulate('click');
+
+      expect(persistentStorage.get(StorageKeys.SEARCH_OFFSET)).toBeFalsy();
+    });
+
+    it('clear button deletes QUERY from global storage', () => {
```
tmeyer2115: The `QUERY` chunk of storage isn't exactly deleted. The chunk is kept around, but it's set to the empty string. Off the top of my head, I don't know if that distinction really matters. But, if it is important that it be set to the empty string rather than wholly deleted, I think we should make sure the tests explicitly verify that (and are named accordingly).

cea2aj: updated!

tmeyer2115 (approved):

tests/ui/components/search/searchcomponent.js
```diff
@@ -0,0 +1,87 @@
...
+      globalStorage.set(StorageKeys.QUERY, 'what does yext do?');
+      expect(globalStorage.getState(StorageKeys.QUERY)).toBeTruthy();
+
+      const clearButton = wrapper.find('.js-yxt-SearchBar-clear');
+      clearButton.simulate('click');
+
+      expect(globalStorage.getState(StorageKeys.QUERY)).toEqual('');
+    });
+
+    it('clear button sets QUERY to the empty string in persistent storage', () => {
```
oshi97, line 74: could you combine this test with the one above it?

cea2aj: sure. I'll combine both sets of tests

oshi97: ~~actually eh separate is fine too~~ ignore me

oshi97 (approved):

PR #1252, opened 2021-01-19
storage: use new setEntries for resetListener and post-search

oshi97:
This commit updates the current persistent storage resetListener
to also reseed the new storage in addition with the old storage.

To do this a new setEntries() method was added to the new storage,
which will set all entries given BEFORE calling any listeners, so
we don't need to worry about the order of storage listeners are called
anymore (see logic around StorageKeys.QUERY in  current
GlobalStorage's setAll()).
This was also added to core.search() and core.verticalSearch().

It also refactors the new storage listeners to be a map of
storage key to listener, so that setEntries() doesn't do a
double for loop when calling listeners.

J=SLAP-869
TEST=manual,auto

test that after searching for "virginia", then searching for "technology",
then hitting backspace, that ANSWERS.core.storage.get('query') returns
"virginia"

unit test that makes sure setEntries() will only call listeners after all data has been updated

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -182,6 +182,7 @@ class Answers {
           this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, 0);
         }
         globalStorage.setAll(data);
+        this.core.storage.setEntries(new Map(Object.entries(data)));
       }
     });
     globalStorage.setAll(persistentStorage.getAll());
--- a/src/core/storage/storage.js
+++ b/src/core/storage/storage.js
@@ -47,7 +47,7 @@ export default class GlobalStorage {
     this.persistentStorage = new DefaultPersistentStorage(this.popListener);
 
     /**
-     * A map of storage key to StorageListener, which apply on
+     * A map of storage key to StorageListener[], which apply on
      * changes to global storage.
      *
      * @type {Map<string, StorageListener[]>}
--- a/src/core/storage/storage.js
+++ b/src/core/storage/storage.js
@@ -50,7 +50,7 @@ export default class GlobalStorage {
      * A map of storage key to StorageListener, which apply on
      * changes to global storage.
      *
-     * @type {Map<string, Array<StorageListener[]>}
+     * @type {Map<string, StorageListener[]>}
      */
     this.listeners = new Map();
   }
--- a/src/core/storage/storage.js
+++ b/src/core/storage/storage.js
@@ -47,11 +47,12 @@ export default class GlobalStorage {
     this.persistentStorage = new DefaultPersistentStorage(this.popListener);
 
     /**
-     * The listeners to apply on changes to global storage
+     * A map of storage key to StorageListener, which apply on
+     * changes to global storage.
      *
-     * @type {StorageListener[]}
+     * @type {Map<string, Array<StorageListener[]?>}
      */
-    this.listeners = [];
+    this.listeners = new Map();
   }
 
   /**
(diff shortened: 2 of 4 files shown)


src/answers-umd.js
```diff
@@ -182,6 +182,7 @@ class Answers {
           this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, 0);
         }
         globalStorage.setAll(data);
+        this.core.storage.setEntries(new Map(Object.entries(data)));
```
creotutar, line 185: Definitely an aside from this PR's purpose: I was kind of imagining storage being initiated here in answers-umd and passed to core (that way we could operate directly on storage instead of using a Core function).

Is there a reason we moved it to be made in core?

src/core/storage/storage.js
```diff
@@ -90,6 +91,20 @@ export default class GlobalStorage {
     this._callListeners('update', key);
   }
 
+  /**
+   * Sets all of the given entries before calling 'update' event listeners.
+   *
+   * @param {Map} entries the data to set
+   */
+  setEntries (entries) {
```
creotutar, line 99: I really like this solution! I'm not sure how this solves the problem of the `query`. Couldn't it still be the case that `query` is the first updated key in the set and then its the first listener called and the remaining aren't called because a search was conducted?

src/answers-umd.js
oshi97: it's still being init-ed in answers-umd, this is just for the resetListener

src/core/storage/storage.js
creotutar: I guess searches don't do a page re-load and instead just update the current page

oshi97: nope! since we wait until everything is updated before calling the update listeners

src/answers-umd.js
creotutar: Ohhhhh I see! Do you have any preference to keeping it initiated close to where it was initiated before? (we may need to for the other storage setters below). Definitely outside of this PR

cea2aj:
+1 from me

src/core/storage/storage.js
creotutar: So say we call the `StorageKeys.QUERY` update listener first, and then it starts calling the other listeners with new data (since the StorageKeys.QUERY update added new information to the storage from the search).

Then, because we updated the StorageKeys from the search, we call the other listeners a second time. My worry is that there is a situation here where you wouldn't want this listener to be called twice with the same data (a flash on the page or it causes behavior you wouldn't expect, since the previous listener would be expected to be called with old data).

Am I wrong in this worry?

src/answers-umd.js
oshi97: yeah I can move it up!

src/core/storage/storage.js
oshi97: I think your worry is justified but is also not what this is trying to solve. The problem this pr is trying to solve making sure that, when you update multiple storage keys at once, that all of the storage updates run before the listeners do. 
The problem you're describing sounds more like "how do we handle people messing with storage inside their listeners?" I think adding setEntries() will help this second problem a little, but you're right it won't solve it. If you have a listener that updates storage you will still have to be careful that it doesn't do something weird, or god forbid infinite loop.
We could make sure that all of the listeners are called with the original data, even if some other listener updates storage, if that's something that makes sense? I don't think that's currently something the SDK needs, but could come up in the future. Currently the closest thing is the QUERY listener, but that fires off a search, which only updates storage after the api request comes back, and because of that all of the listeners will run before that happens. I could look into it though (or make an item) if that's something we care about?

oshi97: leaning towards shelving this for now after talking with christian a little more

oshi97:
closing for now, this is out of scope for SLAP-869 (storage cutover). we probably need to think a little bit more about how to handle this and what our opinions are!

PR #1253, opened 2021-01-20
cut over QUERY to new global storage

oshi97:
This commit cuts over StorageKeys.QUERY, and also
adds a reseed of the new storage to the old persistent
storage's resetListener, so that the back button works.

I found a preexisting bug while testing this, currently,
if you search for say "office space", go to page 4,
then search for "virginia" (which only displays 2 pages
on my setup), the SDK will still use the page 4 offset,
even though there aren't 4 pages of results so you just
get a blank results page and pagination only displays
page 1 and 2 but shifted to the left.

J=SLAP-1019
TEST=manual,auto

added question submission and alternative verticals component tests
test that I can have a "" search on the page, search for something,
hit the back button and go back to the "" search
test that qasubmission will use your current query in the question box
test that no results will display your current query, and also use query
properly in its suggestion links for other verticals and universal
test that vertical and universal searches work
test that location bias searches work on vertical and universal


src/ui/components/filters/geolocationcomponent.js
```diff
@@ -119,10 +119,14 @@ export default class GeoLocationComponent extends Component {
      * The query string to use for the input box, provided to template for rendering.
      * @type {string}
      */
-    this.query = this.core.globalStorage.getState(`${StorageKeys.QUERY}.${this.name}`) || '';
-    this.core.globalStorage.on('update', `${StorageKeys.QUERY}.${this.name}`, q => {
-      this.query = q;
-      this.setState();
+    this.query = this.core.storage.get(`${StorageKeys.QUERY}.${this.name}`) || '';
```
tmeyer2115, line 122: This might be an existing SDK idiosyncrasy, but is the `.${this.name} ` suffix here ever actually present? A cursory glance through the rest of this PR makes it seem like we're never setting `${StorageKeys.QUERY}.${this.name}`, only `${StorageKeys.QUERY}`.

tests/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,28 @@
...
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+
+  defaultConfig = {
+    container: '#test-component',
+    verticalsConfig: []
+  };
+});
+
+describe('AlternativeVerticals component', () => {
+  it('passes query to state correctly', () => {
```
tmeyer2115: Could we test something a bit higher-level here? This test is verifying a lower-level implementation detail of the component: that query makes it into the state. These kinds of details can be subject to change. What's more isolated from change is the higher-level behavior of the component. Specifically, that the query appears in the alternative verticals text. Could we change the test to verify that the text is computed properly? It will have the side benefit of ensuring that the component somehow parses the query from globalStorage.

src/ui/components/filters/geolocationcomponent.js
oshi97: I think these come from persistentStorage, from when we seed globalStorage with persistentStorage on page load

oshi97: the new storage also has this seeding behavior it its init()!

tests/ui/components/results/alternativeverticalscomponent.js
oshi97: I had a lot of trouble getting the alternativeVerticals component to render properly, so resorted to this, but can take another look!

oshi97: updated the test!

tests/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,68 @@
...
+      alternativeVerticals: [{
+        resultsCount: 2,
+        verticalConfigId: 'KM'
+      }]
+    }
+  };
+});
+
+describe('AlternativeVerticals component', () => {
+  it('uses query from the url in suggestLinks', () => {
```
tmeyer2115: This test does a bit more than make sure that the query is in the `suggestLink`. It basically makes sure that the entire `suggestLink` is created properly. Could we rename the test `constructs vertical suggest link properly`?

tests/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,68 @@
...
+    const component = COMPONENT_MANAGER.create(AlternativeVerticalsComponent.type, defaultConfig);
+    const wrapper = mount(component);
+    const suggestionHref = wrapper.find('.yxt-AlternativeVerticals-suggestionLink').prop('href');
+    expect(suggestionHref).toEqual('locations.html?query=office+space');
+  });
+
+  it('uses query from storage in universal links', () => {
```
tmeyer2115: Same question here about potentially renaming the test.

tests/ui/components/results/alternativeverticalscomponent.js
```diff
@@ -0,0 +1,68 @@
...
+        label: 'Home',
+        url: 'index.html'
+      },
+      {
+        verticalKey: 'KM',
+        label: 'Locations',
+        url: 'locations.html'
+      }
+    ],
+    baseUniversalUrl: './index.html?query=office+space',
```
tmeyer2115: Should `baseUniversalUrl` have the `query` param?

src/ui/components/filters/geolocationcomponent.js
tmeyer2115: Ah, okay!

tests/ui/components/results/alternativeverticalscomponent.js
oshi97: currently VerticalResults DOES pass query params down, but the AlternativeVerticals component doesn't require you do this so I removed the `query` param from the test

tests/ui/components/results/alternativeverticalscomponent.js
oshi97: renamed all 3 tests!

tmeyer2115 (approved):

PR #1254, opened 2021-01-20
cut over DYNAMIC_FILTERS and RESULTS_HEADER to new storage

oshi97:
This commit cuts over DYNAMIC_FILTERS and RESULTS_HEADER storage keys
to the new storage. Since the new storage doesn't have the same bugs
as the old one, we can remove the usage of RESULTS_HEADER for just
DYNAMIC_FILTERS and finally remove the TODO. We should still
send updates to the RESULTS_HEADER storage key for backwards
compatibility purposes, in case somebody registered listeners to it
or overrode the results header component.

This commit also adds removal of moduleId listeners when the component
is removed during componentManager.remove(), otherwise, as the results
header component is unmounted and remounted with every rerender of
vertical results, a new module id listener is added without removing
the old one

J=SLAP-1019
TEST=manual,auto

test that running additional searches, and thereby removing and
remounting the results header component,  will not add extra unneeded
listeners to storage, by logging the below onStateChange
ANSWERS.core.storage.listeners.filter(a => a.storageKey === 'dynamic-filters'))

test that facets and filters work as expected, refreshing the page
after selecting facets does not have any regressions (still buggy
but not more than before), and you can remove filters through the
results header or applied filters components


oshi97:
I rebased to get connor's test and borked everything one second

tmeyer2115 (approved):

PR #1255, opened 2021-01-20
cut over DIRECT_ANSWER to new storage

oshi97:
TEST=manual,auto

test that I can search "amani farooque phone number" and get her
direct answer

test that it disappears on search, and only reappears after
the search has loaded and if is directly answering the query


tests/ui/components/results/directanswercomponent.js
```diff
@@ -0,0 +1,52 @@
...
+        type: 'ce_person',
+        website: 'http://www.test.com'
+      }
+    },
+    type: 'FIELD_VALUE'
+  };
+});
+
+describe('DirectAnswer component', () => {
+  it('displays the answer correctly', () => {
```
tmeyer2115, line 48: Can we add a TODO to make this test more thorough? I think what you have is fine for now (especially since there were no tests to start with). Ultimately, we want to verify specific things about the display, such as the breadcrumb view at the top, that the displayed value is correct, etc.

oshi97: sure I can make an item!

src/core/core.js
```diff
@@ -310,8 +310,9 @@ export default class Core {
       .then(data => {
         this.globalStorage.set(StorageKeys.QUERY_ID, data[StorageKeys.QUERY_ID]);
         this.storage.set(StorageKeys.NAVIGATION, data[StorageKeys.NAVIGATION]);
-        this.globalStorage.set(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
+        this.storage.set(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
         this.storage.set(StorageKeys.UNIVERSAL_RESULTS, data[StorageKeys.UNIVERSAL_RESULTS], urls);
+        this.globalStorage.set(StorageKeys.DIRECT_ANSWER, data[StorageKeys.DIRECT_ANSWER]);
```
tmeyer2115: Do we need to get rid of this?

oshi97: sorry yeah that's an extra from the merge, removed it!

tmeyer2115 (approved):

PR #1260, opened 2021-01-20
cut over SEARCH_OFFSET to new global storage

oshi97:
J=SLAP-1019
TEST=manual

test that I can use the pagination button, and if I refresh
the page I'm on the same page

test that searching with sort options with reset the search_offset

test that searching with the searchbar will clear the search_offset

test that VerticalResults and VerticalResultsCount components
display the correct page counts


src/ui/components/results/paginationcomponent.js
```diff
@@ -113,13 +113,17 @@ export default class PaginationComponent extends Component {
...
-    this.core.globalStorage.set(StorageKeys.SEARCH_OFFSET, Number(offset));
-    this.core.globalStorage.on('update', StorageKeys.SEARCH_OFFSET, offset => {
-      if (typeof offset === 'number') {
-        return;
+    const offset = this.core.storage.get(StorageKeys.SEARCH_OFFSET) || 0;
+    this.core.storage.set(StorageKeys.SEARCH_OFFSET, Number(offset));
+    this.core.storage.registerListener({
+      eventType: 'update',
+      storageKey: StorageKeys.SEARCH_OFFSET,
+      callback: offset => {
```
creotutar, line 121: This feels like it's not pagination component's job (also not part of this PR, just commenting)

oshi97: 🍝 🍝 🍝

tests/ui/components/results/verticalresultscountcomponent.js
```diff
@@ -42,17 +42,8 @@ describe('results count component', () => {
...
-        getState: key => {
-          if (key === StorageKeys.SEARCH_OFFSET) {
-            return 40;
-          }
-          return null;
-        }
-      }
-    });
+    const COMPONENT_MANAGER = mockManager();
+    COMPONENT_MANAGER.core.storage.set(StorageKeys.SEARCH_OFFSET, 40);
```
creotutar, line 46: We return `undefined` instead of `null` (as was done previously) in the storage if the key doesn't exist. I don't think this matters since tests are passing, just pointing it out

creotutar (approved):
👍 👍 👍 LGTM

PR #1263, opened 2021-01-20
cut over SESSION_OPT_IN to new global storage

oshi97:
J=SLAP-1019
TEST=manual

test that when I run a search, it still passes my sessionTrackingEnabled
from my config as sessionTrackingEnabled

test that session-opt-in=true in the url will take priority over
sessionTrackingEnabled in config

test that calling setSessionsOptIn in the console will cause universal
results to flash. most easily seen in either the maps portion of universal,
or by manually deleting stuff from the dom, then seeing it reappear when calling
setSessionsOptIn(). saw this with universalresults, navigation, and alternativeverticals
checked that the api call sent had the correct value


oshi97:
I rebased to get connor's test and borked everything one second

cea2aj (approved):
lgtm

PR #1265, opened 2021-01-20
cut over QUERY_SOURCE to the new storage

oshi97:
TEST=manual

see that still defaults to STANDARD and is sent in query
as source: STANDARD

see that I can specify a querySource in either my config,
and it will be sent as the source param in my query


src/answers-umd.js
```diff
@@ -168,6 +168,7 @@ class Answers {
     parsedConfig.search = new SearchConfig(parsedConfig.search);
     parsedConfig.verticalPages = new VerticalPagesConfig(parsedConfig.verticalPages);
 
+    const storage = new Storage().init(window.location.search);
```
creotutar, line 171: love it

creotutar (approved):
👍 👍 👍 LGTM

PR #1266, opened 2021-01-20
cut over NO_RESULTS_CONFIG to new storage

oshi97:
TEST=manual

see that vertical results will use the global no results config's template
setting, and also displayAllResults will control whether to view all possible
results

see that pagination will use the global no results config's visible
setting

### DIFF
--- a/tests/ui/components/results/paginationcomponent.js
+++ b/tests/ui/components/results/paginationcomponent.js
@@ -17,6 +17,7 @@ const createCore = () => {
   const persistentStorage = {};
   const storage = new Storage().init();
   storage.set(StorageKeys.VERTICAL_RESULTS, { searchState: SearchStates.SEARCH_COMPLETE, resultsCount: 21 });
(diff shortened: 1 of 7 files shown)


tests/ui/components/results/paginationcomponent.js
```diff
@@ -17,6 +17,7 @@ const createCore = () => {
   const persistentStorage = {};
   const storage = new Storage().init();
   storage.set(StorageKeys.VERTICAL_RESULTS, { searchState: SearchStates.SEARCH_COMPLETE, resultsCount: 21 });
+  storage.set(StorageKeys.NO_RESULTS_CONFIG, { displayAllResults: true });
```
creotutar, line 19: are we going to remove this in the globalStorage call above at a later time (e.g. when we do SEARCH_OFFSET AND SEARCH_CONFIG?)

oshi97: nope we can remove this right now!

creotutar (approved):
👍 👍 👍 LGTM

PR #1267, opened 2021-01-20
remove SearchIntents

oshi97:
This is not used anywhere anymore. It seems highly unlikely anybody is relying
on this model, they would have to be looking into global storage manually.
This was originally added so that the searchBar could listen
to changes to searchIntents, which it no longer does.
See https://github.com/yext/answers/commit/b036a5f473ecb510cab69fc9c3ac97bc6c65fd31

TEST=manual

test that I can set my ANSWERS.setGeolocation, can do a near me search,
and the location bias component will update to the location I set


tmeyer2115 (commented):
I think you're right, this isn't used anywhere. The `SearchBar` component does not use any stored `INTENTS`, it re-computes them in each call to `debouncedSearch`. Could we just get rid of this chunk of `storage` all together?

oshi97:
> I think you're right, this isn't used anywhere. The `SearchBar` component does not use any stored `INTENTS`, it re-computes them in each call to `debouncedSearch`. Could we just get rid of this chunk of `storage` all together?

yup will kill it! 🔥 🔥 🔥

tmeyer2115 (approved):
LGTM!

PR #1268, opened 2021-01-20
cut over VERTICAL_PAGES_CONFIG to new storage

oshi97:
TEST=manual

check that navigation will use my verticalPages config
check that no results will use the universal page specified in verticalPages


cea2aj (approved):
LGTM

PR #1269, opened 2021-01-20
cut over API_CONTEXT to new global storage

oshi97:
TEST=manual,auto

test that I can load up noresults, see that the
universal link does nto have a context param, enter
ANSWERS.setContext({'state': 'hi'}) into the console,
then see that the universal link DOES have a context param


cea2aj (approved):
lg

PR #1270, opened 2021-01-20
cut over REFERRER_PAGE_URL to new global storage

oshi97:
TEST=manual

check that referrerPageUrl is preserved across
navigation tabs, view all button on universal, and spellcheck

check that I clicking a link to an answers experience will
set referrerPageUrl to the page with the link


creotutar (approved):
👍 👍 👍 LGTM

PR #1271, opened 2021-01-20
cut over QUERY_TRIGGER to new global storage

oshi97:
TEST=manual

check that queryTrigger is set to "initialize" if a
defaultInitialSearch is set and there is no preexisting query url param

check that clicking spellcheck will send queryTrigger=suggest in the search request, and further searches
will not send queryTrigger=suggest


cea2aj (approved):
LGTM

PR #1272, opened 2021-01-20
cut over FACETS_LOADED to new storage

oshi97:
TEST=manual

checked that searching "technology" will create a removable
filter in the AppliedFilters component, due to the "technology"
querying causing the backend to apply a facet to the search.


cea2aj (approved):
nice and simple

PR #1273, opened 2021-01-21
cut over non StorageKey usages of storage

oshi97:
Cuts over usages of storages that aren't
necessarily listed in storagekeys.js,
i.e. usages of this.name where this is the
component itself.

Also fix a typo from #916 (many moons ago) where
ALTERNATIVE_VERTICALS was mispelled ALERNATIVE_VERTICALS.
It looks like the AlternativeVertical component's setState() doesn't
depend on the data you pass it, so probably nothing was broken.

Since all the StorageKeys are cut over, we can
remove the cut over comment markers.

J=SLAP-1039
TEST=manual

checked that sortoptions, range, date range filters are
seeded on page load but don’t affect the initial search
checked that this is consistent with 1.7.1
check that facets still do the wonderful thing where they pretend to apply on load
also checked that 1.7.1 still does this

### DIFF
--- a/tests/ui/components/filters/sortoptionscomponent.js
+++ b/tests/ui/components/filters/sortoptionscomponent.js
@@ -12,17 +12,6 @@ const mockedCore = () => {
     },
     clearSortBys: () => {},
     verticalSearch: () => {},
-    globalStorage: {
-      on: () => {},
-      getState: storageKey => {
-        expect(['SortOptions', StorageKeys.QUERY]).toContain(storageKey);
-        return null;
-      },
-      getAll: storageKey => {
-        expect([StorageKeys.FACET_FILTER_NODE, StorageKeys.STATIC_FILTER_NODE]).toContain(storageKey);
-        return [];
-      }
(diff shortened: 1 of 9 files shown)


creotutar (approved):
👍 👍 👍 LGTM

tests/ui/components/filters/sortoptionscomponent.js
```diff
@@ -12,17 +12,6 @@ const mockedCore = () => {
     },
     clearSortBys: () => {},
     verticalSearch: () => {},
-    globalStorage: {
```
tmeyer2115, line 15: Is `globalStorage` (or `storage`) not needed for these tests?

tmeyer2115 (approved):

tests/ui/components/filters/sortoptionscomponent.js
oshi97: storage is still needed, but it's being automatically provided by mockManager() !

PR #1276, opened 2021-01-27
replace PersistentStorage with Storage

oshi97:
This commit removes the old PersistentStorage in favor of our new Storage.
Additionally, it makes FilterSearch and static FilterBox work with backnavigation
and page refresh. It also fixes a bug where hitting backnav with FilterBox (that was
configured with searchOnChange=true) would trigger one extra liveapi verticalsearch
per FilterOptions inside the FilterBox.

To replace PersistentStorage, a few steps were taken.
1. replace all usages of persistentStorage.set() and delete() with storage.setWithPersist() and delete()
2. replace usages of window.location.search.substr(1) with storage.getCurrentStateUrlMerged() where needed
3. call storage.pushStateToHistory() in core after a search is completed. core.shouldPushState() was added 
    to prevent an extra state from being pushed on page load, from say a defaultInitialSearch or `query`
    url parameter.

There were also a few cases where the SDK was relying on globalStorage and persistentStorage
being separate and desynced, which had to be fixed
1. The SpellCheck component was deleting SKIP_SPELL_CHECK and QUERY_TRIGGER from persistentStorage
    in its onCreate(), while still expecting these values to be in globalStorage when the api call was made.
    This was removed. Now, these are deleted from storage at the end of a core.search() or core.verticalSearch()
2. FilterOptions had a listener to `this.name` in globalStorage, and in this listener, update persistentStorage. 
    Unifying the storages caused this to infinite loop in certain scenarios. Because the original purpose of the
    `this.name` listener was for back navigation, and new HISTORY_POP_STATE key + listener was added, which
    triggers when back or forward navigation occurs. 
3. FilterOptions, in its constructor, was pulling `this.name` from globalStorage in order to get the initial state of
    its selected options. However, FilterOptions, when different filter options were selected, it would only update
    persistentStorage but not globalStorage. This never worked properly with Facets, where it could cause a Facet
    checkbox to be displayed as selected even when it did not impact the search, or vice versa. This happens
    much more frequently after unifying the storages, because previously, the value of `this.name` in globalStorage
    would always just be whatever the initial state of the page was, typically null, and if it was null would not do anything.
    By unifying the storages, the value of `this.name` in globalStorage is now kept in sync in persistentStorage, which
    WAS being updated, causing the bug to happen much more frequently.
    To fix this, now, if a FilterOptions is part of a Facets component, it doesn't try to set anything in storage.
4. In a similar vein, FilterSearch had a listener for a storageKey it used to only be setting in persistentStorage, which would run a search. This listener was updated to use HISTORY_POP_STATE. It was a small change to update this listener to work for back navigation, so I included it in this PR.

As part of the HISTORY_POP_STATE listener (number 3 directly above), this commit also fixes a preexisting bug where, for a static filterbox with searchOnChange = true, back navigation would trigger additional searches and cause the page to flash.
This was fixed by changing updateListener() to updateListener(false, true), in the HISTORY_POP_STATE listener,
to block the extra searches. 

J=SLAP-1027
TEST=manual, auto

- added an acceptance test for static filters, including back navigation for filterbox (which was fixed this pr,
for “free”)
- added an acceptance test for filtersearch back navigation (also fixed this pr)

- added a jest test to make sure we don’t run extra queries on on back nav when a static filterbox
with searchOnChange is on the page
added an acceptance test for filtersearch back navigation
- ie11 does a very nasty thing of aggressively caching all ajax requests by default. These cached requests end up ignored by testcafe's RequestLogger. Added IE11NoCacheHook to deal with this, using a workaround found on the testcafe github (for a slightly different issue). Also added network logs to our browserstack automate

- check that only one state is pushed per search on a page with facets
- checked that in 1.7.1, when you have a static filterbox, it's expected that selecting a filter,
then hitting the back button causes multiple searches to be called
- check that navigation and no results links are updated with the current query correctly
- check that in 1.7.1 range and daterange filters dont work with backnavigation in any way
and also do not apply the filter properly on load
- check that in 1.7.1 sort options will appear visually to apply the state from the url, but not
actually impact the search
it also does not work with back nav in any way
- check that in 1.7.1, the filtersearch half of the geolocation filter (where you search for a filter using 
its searchbar) does not work with either back nav or on page load
on page load it will persist the query inside its searchbar but not actually apply the filter
- check that in 1.7.1, filtersearch will work on load (it will persist both the filter and the query
in its personal searchbar) but does not work with back navigation
a new bug was introduced, where if you selected a filtersearch filter say filter APPLE,
selected a different filtersearch filter, say filter BANANA, hit backnav, which would not do anything
to filtersearch (filter BANANA) would still be displayed on the page, and then refreshed the page,
filter APPLE would be applied when previously filter BANANA would be applied. Fixing this was
a relatively small change, so it was included in this PR.
- check that in 1.7.1, on universal, if you land on a page with no initial search, run a search, then hit backnav, it’s expected to see QA submission on the page
check that sessionTrackingEnabled is passed correctly in requests
- test that getCurrentStateUrlMerged() will include query params that are in the current url but not in persistentStorage. Tested by typing "window.history.replaceState(null, '', '?a=b')" into the console, then loading up a no results page and seeing that the links had a=b as an additional param


src/answers-umd.js
```diff
@@ -167,32 +166,45 @@ class Answers {
...
-      resetListener: data => {
-        if (!data[StorageKeys.QUERY]) {
+    const storage = new Storage({
+      update: (data, url) => {
+        if (parsedConfig.onStateChange) {
+          parsedConfig.onStateChange(Object.fromEntries(data), url);
+        }
+      },
+      /**
+       * Called when the 'popstate' event is fired, i.e. window.history
```
creotutar: Love the jsdoc, how do we feel about putting it in update as well?

src/core/core.js
```diff
@@ -546,6 +548,25 @@ export default class Core {
...
+   * If queryTrigger is QUERY_PARAMETER, this only occurs either on page load
+   * or when hitting the history back and forward buttons. For both cases we
+   * don't push an extra state.
+   * If queryTrigger is SUGGEST, this only occurs when clicking a spellcheck
+   * link, which reloads the page, so don't push an extra state.
+   *
+   * @param {QueryTriggers} queryTrigger SDK query trigger
+   * @returns {boolean}
+   */
+  shouldPushState (queryTrigger) {
```
creotutar, line 564: LG, I just want to make sure I understand this. Previously, we would always set the persistentStorage for things that were waiting in global storage (`context`, `referrerPageUrl`). Here, because we have a new push state to history paradigm (all or nothing), AND because we need context and referrerPageUrl to show up, we have to be picky about when we push a state. 

1. We don't push history on INITIALIZE otherwise we would never be able to navigate backwards in history on page load.
2. We don't push history on QUERY_PARAMETER we don't push for the same reason.
3. This case is very similar to the QUERY_PARAMETER case, it's just a special subcase

src/answers-umd.js
```diff
@@ -167,32 +166,45 @@ class Answers {
...
+       * back or forward navigation.
+       *
+       * @param {Map} data
```
cea2aj: What data is this listener called with?

src/answers-umd.js
```diff
@@ -282,7 +292,9 @@ class Answers {
     });
 
     if (parsedConfig.onStateChange && typeof parsedConfig.onStateChange === 'function') {
-      parsedConfig.onStateChange(persistentStorage.getAll(), window.location.search.substr(1));
+      parsedConfig.onStateChange(
+        Object.fromEntries(storage.getAll()),
```
cea2aj, line 290: nit: This is out of scope of this PR, but what are your thoughts on changing storage.getAll() to return the object so that we don't have to call Object.fromEntries? We could also have a storage.getAllAsMap() if we want the Map object specifically

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -306,18 +306,18 @@ export default class FilterOptionsComponent extends Component {
       // this.name otherwise, instead opt for this.core.setStaticFilterNodes()
```
cea2aj: I believe this comment needs to be updated to say StorageKeys.HISTORY_POP_STATE rather than this.name

src/answers-umd.js
oshi97: hmm it seems kind of weird to put it twice, I'll move it into the storage.js file

src/answers-umd.js
oshi97: I'll move this jsdoc into the storage.js file

src/core/core.js
oshi97: yup! 3 is a subcase of 2.

src/answers-umd.js
oshi97: hmm I'm not really sure on the pros or cons, Object.fromEntries is being called here purely for backwards compatibility reasons, otherwise maybe we could pass onStateChange a map instead ... ? I'm a little more comfortable using a regular Object instead of a Map but maybe using Map is more big brained (i.e. more ES6 less ES5)?

src/ui/components/filters/filteroptionscomponent.js
oshi97: ah good catch I'll just remove this comment, it doesnt look like it'll age very well anyways

src/core/storage/storage.js
```diff
@@ -24,7 +24,9 @@ export default class Storage {
     /**
      * The listener for window.pop in the persistent storage
      *
-     * @type {Function}
+     * @param {Map<string, string>} queryParamsObject  A Map containing the persisted state
+     * @param {string} queryParamsString the url params of the new state
+     *                                   e.g. "?query=virginia&otherParam=yext"
      */
     this.popListener = (queryParamsObject, queryParamsString) => {
```
tmeyer2115: Going off the JSDoc, I'm not sure what the difference is between these two params. Is `queryParamsObject` just a parsed `queryParamsString`?

oshi97: yeah, that's the only difference. I changed `queryParamsObject` to `queryParamsMap` and gave an example, if that makes the jsdoc clearer?

tests/acceptance/acceptancesuite.js
```diff
@@ -160,6 +164,132 @@ test(`Facets load on the page, and can affect the search`, async t => {
   await t.expect(actualResultsCount).eql(initialResultsCount);
 });
 
+test(`static filterboxes`, async t => {
```
cea2aj: I'm having a hard time understanding which pieces of functionality these tests are testing. Is it possible to split these up into multiple tests where each test is testing one thing and the functionality to be tested is included in the test title?

tests/acceptance/acceptancesuite.js
```diff
@@ -160,6 +164,132 @@ test(`Facets load on the page, and can affect the search`, async t => {
...
+  await t.expect(filterTagText).eql('Virginia, United States');
+  await expectResultsCountToEql(4);
+
+  // Choose the 'New York City, New York, United States' filter option
+  await filterSearch.enterQuery('new york');
+  await filterSearch.selectFilter('New York City, New York, United States');
+  await t.expect(filterTags.count).eql(1);
+  filterTagText = await filterTags.nth(0).find('.yxt-ResultsHeader-removableFilterValue').innerText;
+  await t.expect(filterTagText).eql('New York City, New York, United States');
+  await expectResultsCountToEql(1033);
```
cea2aj: Is it possible to make these tests less dependent on the exact number of results? I know that we control the testing KG, but if we can test this without depending on the exact number I think that would be ideal. Similarly to my other comment, it is not very clear to me how some of these expect statements relate to the functionality that is being tested. Would explanatory boolean variables possibly work here and have the expects operate on those? (For example, why are we expecting results to equal 1033? Were we just expecting the number to respond to the filter? If so, we could have an resultsCountUpdated variable and make our expect something like `t.expect(resultsCountUpdated).eql(true)`

src/core/storage/storage.js
tmeyer2115: Do we need both pieces of data passed in? It seems like one can be constructed from the other.

tests/acceptance/acceptancesuite.js
oshi97: originally I used a testcafe requestLogger to track that we were sending the correct filter in the request. BUT for some reason requestLogger wouldn't log the request from a page refresh, and only in browserstack (not locally or with the headless acceptance tests). I even tried upgrading our browserstack-testcafe-provider and testcafe npm package versions. I'll try to play around with it again in another branch because it definitely would be better to not care about the specific response

tests/acceptance/acceptancesuite.js
oshi97: I can use more words in the description, these tests can't be split up because they're testing back/forward navigation

oshi97: actually, I think I can split up the radius_filter and static_filter testing! it's unlikely these two will affect each other, though they still both need to be tested despite being on the same component

tests/acceptance/acceptancesuite.js
oshi97: Updated the tests to use RequestLogger! IE11 caches all ajax requests, which causes them to not be tracked by testcafe's RequestLogger (this was really annoying to uncover...), so I added a new IE11NoCacheHook which adds Cache-Control: no-store to the requests you tell the hook to care about

src/core/storage/storage.js
oshi97: Yeah it seems a little redundant, we can also get queryParamsString from getUrlWithCurrentState(). This listener's implementation is inside the bespoke answers-storage library, though. Is it alright if we make a separate item to get rid of queryParamsString (or look into getting rid of it?)

tmeyer2115: Yeah, no worries. We can have a separate item.

tests/acceptance/acceptancesuite.js
```diff
@@ -84,8 +97,8 @@ test.requestHooks(spellCheckLogger)('spell check flow', async t => {
 
   // Check that clicking spell check sends a queryTrigger=suggest url param
   // TODO(oshi) investigate making this an integration test
-  const requestUrl = spellCheckLogger.requests[spellCheckLogger.requests.length - 1].request.url;
-  const queryTriggerParam = new URLSearchParams(requestUrl).get('queryTrigger');
+  const queryParams = await getMostRecentQueryParamsFromLogger(spellCheckLogger);
```
tmeyer2115, line 96: This doesn't feel much like an acceptance test. Are we essentially verifying that the `SpellCheck` component constructs the suggested query link properly? If so, I think we should aim to make this a unit or integration test.

tests/acceptance/acceptancesuite.js
```diff
@@ -160,6 +173,213 @@ test(`Facets load on the page, and can affect the search`, async t => {
...
+  await expectRequestLocationRadiusToEql(radiusFilterLogger, 40233.6);
+  radiusFilterLogger.clear();
+
+  // Refresh the page
+  await browserRefreshPage();
+  await t.expect(filterTags.count).eql(1);
+  await expectRequestLocationRadiusToEql(radiusFilterLogger, 40233.6);
+});
+
+test(`static filter filterbox works with back/forward navigation and page refresh`, async t => {
```
tmeyer2115: Does the SDK treat location radius different than other static filters? If not, I think we'd be fine with just this test here.

tests/acceptance/acceptancesuite.js
```diff
@@ -160,6 +173,213 @@ test(`Facets load on the page, and can affect the search`, async t => {
...
+  filterSearchLogger.clear();
+
+  // Hit the forward button, expect to see the 'New York' filter applied
+  await browserForwardButton();
+  await expectOnlyFilterTagToEql('New York City, New York, United States');
+  await expectRequestFiltersToEql(filterSearchLogger, newYorkFilter);
+  filterSearchLogger.clear();
+});
+
+test(`pagination works with page navigation after selecting a filtersearch filter`, async t => {
```
tmeyer2115: This file is starting to get too big. Can we break some of these tests out into their own acceptance suite? Like one for static filters, one for filter search, etc. Don't worry about the existing tests in the suite. We can take care of those later.

tests/acceptance/blocks/filtersearchcomponent.js
```diff
@@ -0,0 +1,27 @@
+import { Selector, t } from 'testcafe';
+
+/**
+ * This class models user interactions with the {@link FilterSearchComponent}.
+ */
+export default class FilterOptionsComponentBlock {
```
tmeyer2115: Should this be called `FilterSearchComponentBlock`?

tests/acceptance/ie11nocachehook.js
```diff
@@ -0,0 +1,31 @@
+import { RequestHook } from 'testcafe';
+
+/**
+ * IE11NoCacheHook solves the problem of IE11 caching all ajax requests
+ * by default by adding a 'no-store' header to requests in ie11.
+ *
+ * Without this hook, ajax requests cached by IE11 will be ignored
+ * testcafe's RequestLogger. This is only a problem in IE11.
```
tmeyer2115, line 8: Just to make sure I understand, IE11 was caching the Ajax requests. This meant, on a page refresh, that a cached request was used. This wouldn't be caught by TestCafe's logger. Is that right?

tests/acceptance/utils.js
```diff
@@ -1,13 +1,34 @@
-import { ClientFunction } from 'testcafe';
+import { ClientFunction, t } from 'testcafe';
+import IE11NoCacheHook from './ie11nocachehook';
 
 /* global location */
 
 export async function browserRefreshPage () {
   await ClientFunction(() => location.reload())();
-  await new Promise(resolve => setTimeout(resolve, 2500));
+  return t.wait(2500);
```
tmeyer2115, line 8: For the refresh, back, and forward actions, could we use `t.navigateTo`? The `navigateTo` has a built-in waiting mechanism. From my experience with Selenium, it is better to rely on the built-in mechanisms than to add explicit `wait`s. Adding a `wait` like this can lead to spurious failures.

tests/acceptance/acceptancesuite.js
oshi97: yup it does! locationRadius is a different query param than filters. That being said FilterBox probably doesn't treat them different enough to warrant an additional acceptance test. Originally I added this test because I had issues with backnav with only a RADIUS_FILTER filterbox, but that was due to one particular issue I think I can add a jest test for.

tests/acceptance/ie11nocachehook.js
oshi97: Yup! Also on back nav/forward nav. You can see this in browserstack in the ie11 network tab (how I originally found this) it'll say "from cache" or something like that.

tests/acceptance/utils.js
oshi97: I tried it and I think navigateTo is functionally different than a refresh. Is it alright if I keep the waits for now, since they existed previously? I tried removing them and it looks like that makes things fail more often than not. There's probably some way to ask testcafe to smartly wait for things but it seems a little out of scope

tests/acceptance/acceptancesuite.js
oshi97: Actually, is it alright if I just rephrase it into one test for singleoption and one test for multioption? That's probably a more
important distinction, especially since there's two separate pathways within the component for seeding the component state depending on that. RADIUS_FILTER vs STATIC_FILTER just change the filternode that is created, and which core method we call to store the filternode.

tests/acceptance/acceptancesuite.js
oshi97: Yup! There's already a todo, I gave it an item (SLAP-1062) so it's a little easier to track

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -290,8 +290,9 @@ export default class VerticalResultsComponent extends Component {
 
     const navigationData = this.core.storage.get(StorageKeys.NAVIGATION);
     const dataTabOrder = navigationData ? navigationData.tabOrder : [];
-    const tabOrder = getTabOrder(this._verticalsConfig, dataTabOrder);
-    const params = new SearchParams(window.location.search.substring(1));
+    const tabOrder = getTabOrder(
+      this._verticalsConfig, dataTabOrder, this.core.storage.getUrlWithCurrentState());
+    const params = new SearchParams(this.core.storage.getUrlWithCurrentState());
```
creotutar: This is interesting. Whereas before we would take whatever is in the window.location, now we are explicitly using things that are in persistentStorage only. This is a change I would feel like may take people by surprise, if they were adding their own query parameters after page load without knowing they needed to add it to persistent storage as well. What do you think?

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -138,11 +145,11 @@ export default class FilterSearchComponent extends Component {
     return 'search/filtersearch';
   }
 
-  // TODO(oshi): SPR-1925 check that it is safe to remove this, it runs an extra search
-  // For no obvious reasons
+  // This is needed for filtersearch only pages, however it will run an duplicate search
```
creotutar: nit: a duplicate

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -223,14 +235,14 @@ export default class FilterSearchComponent extends Component {
    * optionally redirecting based on config. Uses window.setTimeout to allow
    * other filters to finish rendering before searching.
    */
-  search () {
+  search (resetPagination = true) {
```
creotutar, line 238: Did we need to add this flag for some bug unrelated to storage? What is a case we'd search without resetting pagination?

oshi97: on page reload or backnav, we don't want to resetPagination if a search-offset param exists

src/ui/components/results/verticalresultscomponent.js
oshi97: talked offline, added getCurrentStateUrlMerged() to storage. Tested by entering `window.history.replaceState(null, '', '?a=b')` into console, loading up noresults, and seeing an additional a=b param in the links

src/ui/components/search/spellcheckcomponent.js
```diff
@@ -33,11 +33,6 @@ export default class SpellCheckComponent extends Component {
     return 'search/spellcheck';
   }
 
-  onCreate () {
-    this.core.persistentStorage.delete(StorageKeys.SKIP_SPELL_CHECK, true);
```
creotutar, line 37: Does this mean these parameters will stay in the URL when you click on the spellcheck?

oshi97: from over gchat: prior to this we were deleting them from globalStorage after the search in core.js and deleting them from persistenStorage on page load with the lines you commented on
now they're both deleted after a search.

src/core/storage/storage.js
```diff
@@ -152,6 +156,20 @@ export default class Storage {
...
   }
 
+  /**
+   * Returns the url representing the current persisted state, merged
+   * with any additional query params currently in the url.
+   *
+   * @returns {string}
+   */
+  getCurrentStateUrlMerged () {
+    const searchParams = new SearchParams(window.location.search.substring(1));
```
creotutar, line 166: I'm worried about having this in the function instead of passed into the function. What if we get into the situation where someone is changing the URL in between calls and we get two different Navigation tab URLs? COuld we move it to a parameter? I feel like it would make the test easier too

creotutar: As we talked about online, this is the behavior of the SDK currently. This is something we should address but not necessarily in this PR as it is not a breaking change.

creotutar (approved):
🏖️ 🏖️ 🏖️ LGTM

tmeyer2115 (approved):

PR #1285, opened 2021-02-09
Add answers-core to dependencies

cea2aj:
Add @yext/answers-core to the dev dependencies

Also update the interface because provide core changed from async to sync.

Set browser: true for the resolve plugin so that rollup uses the browser ponyfill for cross-fetch rather than the node ponyfill.

Use the commonjs build to avoid the 'this is undefined' warning which appears because the core library outputs an awaiter helper which references `this` for the async answers core functions (universalSearch, verticalSearch, etc). The helpers are created because the core esm bundle targets an older version of JS than es2017.

Tests fail because I still need to update the core adapter to use the latest core api. That will be in a future PR.

J=SLAP-839
TEST=manual

Load a site with the SDK and confirm that Answers initializes


oshi97:
oh in the commit message it says dev dependencies but it should just say dependencies

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -88,7 +94,13 @@ exports.legacyBundle = function (callback, outputConfig, bundleName, locale, lib
           '@babel/plugin-transform-object-assign'
         ]
       })
-    ]
+    ],
+    onwarn: function(warning) {
```
oshi97, line 98: I think I would rather use the commonjs bundle than have this suppression, if the bundle size is basically the same

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -29,16 +29,22 @@ exports.modernBundle = function (callback, outputConfig, bundleName, locale, lib
...
       }),
       babel({
         babelrc: false,
-        exclude: [/node_modules/, /answers-core/],
+        exclude: [/node_modules/],
         presets: ['@babel/env']
       })
-    ]
+    ],
+    onwarn: function(warning) {
```
oshi97: I think I'd rather use the commonjs bundle than have this onwarn, assuming the bundle sizes are basically the same. not sure what other people think

cea2aj: ya, they're nearly the same. I'll update.

Leaving it as it is now would suppress the warning for other dependencies, and we probably don't want that

oshi97 (approved):

PR #1288, opened 2021-02-10
Update location adapter

cea2aj:
Update the location adapter for vertical and universal searches

Update the param name and add a helper function which performs a transformation from storage to the model needed by core

J=SLAP-839
TEST=manual

Perform a search and confirm that no location is sent by default. Use ANSWERS.setGeolocation() and then perform another search. Confirm that location is properly sent. Test both for both universal and vertical.

### DIFF
--- a/src/core/core-adapter.js
+++ b/src/core/core-adapter.js
@@ -196,7 +196,7 @@ export default class CoreAdapter {
       .verticalSearch({
         verticalKey: verticalKey || searchConfig.verticalKey,
         limit: this.storage.get(StorageKeys.SEARCH_CONFIG).limit,
-        coordinates: this.storage.get(StorageKeys.GEOLOCATION),
+        location: this._getLocationPayload(),
         query: parsedQuery.input,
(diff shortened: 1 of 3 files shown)


src/core/core-adapter.js
```diff
@@ -587,6 +587,19 @@ export default class CoreAdapter {
...
+  /**
+   * Gets the location object needed for answers-core
+   * 
+   * @returns {LatLong|undefined} from answers-core
+   */
+  _getLocationPayload () {
+    const geolocation = this.storage.get(StorageKeys.GEOLOCATION);
+    return geolocation && {
+      latitude: geolocation.lat,
+      longitude: geolocation.lng,
```
creotutar: nit: I don't think we do trailing commas in the SDK (though I am not opposed to that in the future)

cea2aj: ah, I had the linter disabled temporarily and missed this

oshi97 (approved):

PR #1290, opened 2021-02-10
Core adapter filter naming updates

cea2aj:
Update the core adapter and some tests to match an updated core interface

- comparator -> matcher
- comparedValue -> value

Also update filterAutoComplete to filterSearch in the test

J=SLAP-839
TEST=none

Update tests and confirm they pass


creotutar (approved):
👍 👍 👍 LGTM

PR #1292, opened 2021-02-11
Update filterbox acceptance test

cea2aj:
Update the multioption filterbox test to support the core library.

When no static filters are supplied, the core library doesn't sent the query param. The SDK previously would send an empty object. Update the test to match the way that the core library sends it (as null)

J=SLAP-839
TEST=none

Confirm that all acceptance tests now pass

### DIFF
--- a/tests/acceptance/acceptancesuites/filterboxsuite.js
+++ b/tests/acceptance/acceptancesuites/filterboxsuite.js
@@ -104,7 +104,7 @@ test(`multioption filterbox works with back/forward navigation and page refresh`
 
   // Hit the back button, see the 'Marty' filter disappear
   await browserBackButton();
-  await expectRequestFiltersToEql(filterBoxLogger, null);
+  await expectRequestDoesNotContainParam(filterBoxLogger, 'filters');
   await t.expect(filterTags.count).eql(0);
 
   // Hit the forward button, see the 'Marty' filter applied again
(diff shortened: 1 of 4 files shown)


creotutar:
It looks like acceptance tests are failing but unrelated to your change, is this intentional?

cea2aj:
> It looks like acceptance tests are failing but unrelated to your change, is this intentional?

ya, these tests depend on the latest version of core which hasn't been pushed to npm yet

creotutar (approved):
💯 💯 💯 LGTM

tests/acceptance/acceptancesuites/filterboxsuite.js
```diff
@@ -104,7 +104,7 @@ test(`multioption filterbox works with back/forward navigation and page refresh`
 
   // Hit the back button, see the 'Marty' filter disappear
   await browserBackButton();
-  await expectRequestFiltersToEql(filterBoxLogger, {});
+  await expectRequestFiltersToEql(filterBoxLogger, null);
```
oshi97: instead of expecting filters to equal null, can we add a new helper that expects a param to not be in the request?

cea2aj: sure

cea2aj: that helper is already in there, so I'll use that

oshi97 (approved):

PR #1296, opened 2021-02-12
Additional functionality for the ANSWERS instance

cea2aj:
Expose additional functionality on the ANSWERS instance

The methods include:
- verticalSearch()
- resetAllFilters()
- setStorage()
- getFromStorage()
- deleteFromStorage()
- registerListener()

J=SLAP-1097
TEST=manual

Use the corresponding theme branch and confirm that the test site now works with cfilters and this version of the SDK. Test the 'clear search' button, the 'sorts and filter' page, and the 'reset filters' button on a theme locations page with cfilters. Manually call setStorage, getFromStorage, deleteFromStorage, registerListener in the browser devtools and observe that they are working properly.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -418,6 +418,52 @@ class Answers {
     this.coreAdapter.storage.setWithPersist(StorageKeys.QUERY, query);
   }
 
+  /**
+   * Conduct a vertical search with the provided options.
+   * @param {Object} options
+   * @param {string} options.verticalKey the vertical ID for the search
+   * @param {string} options.query the query string of the search
+   * @param {boolean} options.useFacets enables facets if true
+   * @param {boolean} options.resetPagination resets the search offset if true
+   * @param {string} options.queryId the queryId associated with the search
+   * @param {boolean} options.appendQuery if true, adds the results of this query to the current results
+   */
+  verticalSearch (options) {
+    const verticalKey = options.verticalKey;
+    const verticalOptions = {
+      ...options.useFacets && { useFacets: options.useFacets },
+      ...options.resetPagination && { resetPagination: options.resetPagination }
+    };
+    const queryOptions = {
+      ...options.query && { input: options.query },
+      ...options.queryId && { id: options.queryId },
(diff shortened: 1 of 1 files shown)


src/answers-umd.js
```diff
@@ -418,6 +418,61 @@ class Answers {
...
+    const queryOptions = {
+      ...options.query && { input: options.query },
+      ...options.queryId && { id: options.queryId },
+      ...options.appendQuery && { append: options.appendQuery }
+    };
+    this.coreAdapter.verticalSearch(verticalKey, verticalOptions, queryOptions);
+  }
+
+  /**
+   * Rests all filters in the filter registry.
```
creotutar: nit: Resets

src/answers-umd.js
```diff
@@ -418,6 +418,61 @@ class Answers {
...
+      node.remove();
+    });
+  }
+
+  /**
+   * Retrieve data from storage.
+   *
+   * @param {string} storageKey
+   */
+  getFromStorage (storageKey) {
```
creotutar: This feels like we should support all CRUD operations. I have `delete` and `set` in the InteractiveMap operations. Is this something we're thinking about adding?

cea2aj: ya, if we have a use case for it I think we should add it here. What do you think of the names 'deleteFromStorage' and 'setStorage'?

src/answers-umd.js
creotutar: I am setting filters in the filter registry for my use case as well. Did we want to support CRUD for these filters as well? This is almost feeling like a wrapper around core 🤔

cea2aj: I don't think a wrapper is necessarily a bad thing because it lets us swap out the internals without changing the public interface. But you're right, we could just rename coreAdapter to core and access everything we need with ANSWERS.core or ANSWERS.core.storage.

cea2aj: @tmeyer2115 What are your thoughts on this?

cea2aj:
Closing this because we are moving this functionality to the theme. See: https://github.com/yext/answers-hitchhiker-theme/pull/607

PR #1297, opened 2021-02-12
Rename CoreAdapter to Core

cea2aj:
Rename CoreAdapter to Core

The suffix is a bit unnecessary. The theme will need to reference ANSWERS.core, and having the name be 'core' rather than 'coreAdapter' makes more sense.

J=SLAP-1097
TEST=manual

Smoke test on a test site

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -282,8 +282,6 @@ class Answers {
 
     this._setDefaultInitialSearch(parsedConfig.search);
 
-    this.core.init();
-
     this._onReady = parsedConfig.onReady || function () {};
 
     const asyncDeps = this._loadAsyncDependencies(parsedConfig);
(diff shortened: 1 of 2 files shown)


src/answers-umd.js
```diff
@@ -273,20 +273,22 @@ class Answers {
...
     }
 
     this.components
-      .setCore(this.coreAdapter)
+      .setCore(this.core)
       .setRenderer(this.renderer);
 
     this._setDefaultInitialSearch(parsedConfig.search);
 
+    this.core.init();
```
creotutar: Why did we move this out?

cea2aj: Core is no longer an async dependency. I can revert this though. This is better put in a different PR

creotutar (approved):
🌽 🌽 🌽 LGTM

PR #1298, opened 2021-02-12
Add back MasterSwitch

cea2aj:
Add back masterswitch because the core library no longer performs the check

Make the init to core synchronous because it is no longer async.

Also update the HttpRequester to use cross-fetch. This prevents us from needing to include both whatwg-fetch and cross-fetch and saves us  a few KBs on the bundle size.

J=SLAP-1097
TEST=manual

Smoke test a site and confirm that the page loads. Manually test setting the masterswitch to disabled and confirm that the proper error is thrown.


oshi97 (approved):

oshi97:
it looks like acceptance tests are unhappy in ie11? probably not from this change though

oshi97 (approved):

PR #1299, opened 2021-02-17
update static FilterOptions and FilterBox to use persisted filter params 

oshi97:
This commit updates static FilterOptions (and by extension FilterBox)
to use the new `filters` and `locationRadius` url params for persisting filters
in the url. Facets will use a separate param, `facetFilters`, and be in a separate PR.

On back navigation these storage keys are deleted, because navigating
to a page with no url params does not automatically delete them from
storage, despite them no longer being in the url. 
This is also behavior experienced by other storage keys.

This commit also adds the afterAnswersOnReady component method, 
which will be called after the ANSWERS.onReady() call is finished.
This fixes a preexisting bug where filters marked as selected in the config, e.g. something like
```js
  {
    label: "50 miles",
    value: 80467.2,
    selected: true
  },
```
would not be applied on page load, if promptForLocation in the SearchBar's config was set to
false. We were previously relying on the Promise [here](https://github.com/yext/answers/blob/20928fb208cd7d1aa07a60ebcd67229de81b5480/src/ui/components/search/searchcomponent.js#L608) to execute the FilterBox component's onMount BEFORE the SearchBar
would run a search. Turning off promptForLocation would make it so that the FilterBox's onMount
would happen after the query params for the search were determined.
Adding afterAnswersOnReady ensures that the SearchBar's search is always run only AFTER all
components have finished mounting.

This commit also fixes an old bug, where back navigation would not work properly for
when the query in the url was equal to blank string. Previously, no search would be run
on back navigation, even though blank string is a valid query.
Also, the FilterBox component has been updated to search using setQuery and setting
the QUERY_TRIGGER, relying on the SearchBar's QUERY listener instead of manually calling
a search. This is for an edge case where a user has no defaultInitialSearch, and runs a search
with a FilterBox component. In those cases, the query for the api call would be blank string,
but QUERY would not be set in storage & persistent storage. All other filters will be updated
to run searches this way in following PRs.

J=SLAP-1087
TEST=manual, auto

tested that both RADIUS_FILTER and STATIC_FILTER FilterOptions, as well
as a FilterBox that has a FilterOptions for both, works properly with page reload
and back nav. there's a number of acceptance tests that already exist for testing
this behavior

added new acceptance test page filterbox.html, to test for a bug I made during this PR
where if a FilterBox contained a RADIUS_FILTER FilterOptions, that was not the first
index of the FilterBox's config filters, then it would not work properly with back navigation

Testing running a search after page load with no defaultInitialSearch,
using a filter component, and back nav/refreshing

tested running a search after page load WITH a defaultInitialSearch, using
a filter component, and seeing it still use the defaultInitialSearch in the search
done by the filter


oshi97:
~~noticed some weirdness when working on facets, will mark this as ready when it's fixed~~
fixed and updated acceptance tests to check for this backnav edge case

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -295,31 +320,18 @@ export default class FilterOptionsComponent extends Component {
...
-        }
-      });
-    }
+  _registerBackNavigationListener () {
+    this.core.storage.registerListener({
+      eventType: 'update',
+      storageKey: StorageKeys.HISTORY_POP_STATE,
+      callback: () => {
+        this._initVariables(this._config);
+        this.updateListeners(true, true);
```
oshi97, line 334: the first param, `alwaysSaveFilterNodes`, was changed from false to true so that back navigation would work for searchOnChange = false FilterBoxes. When searchOnChange = true the behavior is the same

src/core/filters/filterregistry.js
```diff
@@ -82,6 +83,16 @@ export default class FilterRegistry {
       : null;
   }
 
+  /**
+   * Computes the filter that will be persisted in the URL in the next
+   * history state.
+   */
+  createPersistedFilter () {
```
tmeyer2115: Rename this to indicate it just returns the combination of the static filters

src/core/storage/storagekeys.js
```diff
@@ -40,3 +42,4 @@ export default {
   QUERY_SOURCE: 'query-source',
   HISTORY_POP_STATE: 'history-pop-state'
 };
+export default StorageKeys;
```
oshi97, line 45: this was so my intellisense could autoimport StorageKeys correctly,
autoimport without this
`import storagekeys from '../storagekeys'`
with this
`import StorageKeys from '../storagekeys'`
not sure if other people have this issue

src/ui/tools/filterutils.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param {Filter} filter
+ * @param {Filter} persistedFilter
+ * @returns {boolean}
+ */
+export function filterIsPersisted (filter, persistedFilter) {
+  const childFilters =
+    persistedFilter[FilterCombinators.AND] || persistedFilter[FilterCombinators.OR];
+  if (childFilters) {
+    return !!childFilters.find(childFilter => filterIsPersisted(filter, Filter.from(childFilter)));
```
creotutar, line 18: If we're assuming there are no child filters (re: jsdoc), should we be doing this recursion?

oshi97: the jsdoc is for the filter to be tested not the persisted filter, I can update the jsdoc

src/answers-umd.js
```diff
@@ -588,6 +595,25 @@ class Answers {
...
     return this.core.storage.get(StorageKeys.LOCALE);
   }
+
+  /**
+   * Parses a value from persistent storage, which stores strings,
+   * into the shape the SDK expects.
+   *
+   * @param {string} key
+   * @param {string} value
+   * @returns {string|number|Filter|number}
```
creotutar: nit: number twice

oshi97: 👀 👀

tests/acceptance/acceptancesuites/filterboxsuite.js
```diff
@@ -84,23 +84,21 @@ test(`multioption filterbox works with back/forward navigation and page refresh`
   await searchComponent.submitQuery();
 
   // Click the 'Marty' checkbox
-  // 2 filter tags should be rendered, 1 from the backend facet and 1 from the static filter
   await staticFilter.toggleOption('Marty');
   await expectRequestFiltersToEql(filterBoxLogger, martyFilter);
-  await t.expect(filterTags.count).eql(2);
+  await t.expect(filterTags.count).eql(1);
```
cea2aj, line 89: for my own edification, why is this changing? (edit: nvm: it's beause these tests are on a new page which don't include facets)

src/core/storage/storagekeys.js
```diff
@@ -6,18 +6,21 @@
...
   VERTICAL_RESULTS: 'vertical-results',
   ALTERNATIVE_VERTICALS: 'alternative-verticals',
   AUTOCOMPLETE: 'autocomplete',
   DIRECT_ANSWER: 'direct-answer',
   FILTER: 'filter', // DEPRECATED
-  STATIC_FILTER_NODE: 'static-filter-node',
+  PERSISTED_FILTER: 'filters',
+  STATIC_FILTER_NODES: 'static-filter-nodes',
+  LOCATION_RADIUS_FILTER_NODE: 'location-radius-filter-node',
+  PERSISTED_LOCATION_RADIUS: 'locationRadius',
```
creotutar, line 20: it's a bit weird to use `locationRadius` when we have `location-radius-filter-node`. is this because we want it to match some LiveAPI format?

cea2aj:
+1 from me (pending Christian's comments)

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -634,9 +646,6 @@ export default class FilterOptionsComponent extends Component {
       default:
         throw new AnswersComponentError(`Unknown optionType ${this.config.optionType}`, 'FilterOptions');
     }
-    if (!this.config.isDynamic) {
-      this.saveSelectedWithPersist();
```
creotutar, line 638: Confirming: this is moved to core now, right? Would someone still see filters in their URL if they did not have a filter options component? I don't think this is a case we need to worry about, just bringing it up. It seems like if someone were adding their filters without the component, they should be responsible for, during any upgrade, to make sure it's not in the URL after a search.

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -32,7 +33,7 @@ const OptionTypes = {
 };
 
 class FilterOptionsConfig {
-  constructor (config) {
+  constructor (config, persistedFilter, persistedLocationRadius) {
```
creotutar: Nit: Only one or the other of these `persistedX` variables should be non-null at any time right? It makes me feel like this can be factored out into a `persistedNode` as one parameter. Hyf?

tests/acceptance/acceptancesuites/filterboxsuite.js
oshi97: yup! took the facets off to make the tests a little more focused

src/core/storage/storagekeys.js
oshi97: yup, locationRadius is just a number like 1000.5. location-radius-filter-node is the whole node with unneeded metadata

src/ui/components/filters/filteroptionscomponent.js
oshi97: ah, yeah even if they don't have any filter components they will see `filters={}` in the url. With the current implementation he param won't do anything without corrresponding static filter components, though. I can ask Rose about this when she gets back next week

src/ui/components/filters/filteroptionscomponent.js
oshi97: They can both be non-null at the same time, though each FilterOptions will only use one or the other depending on whether it's a RADIUS_FILTER or STATIC_FILTER. I can combine them into a single object

oshi97: Okay, updated!

oshi97:
@creotutar @tmeyer2115 @cea2aj Added an `afterAnswersOnReady` method to the base component class, to fix old weird behavior with the SearchBar that was being very annoying

src/core/core.js
```diff
@@ -189,8 +189,10 @@ export default class Core {
         verticalKey: verticalKey
       });
     }
+    const locationRadius = this._getLocationRadius();
+    // The backend throws an error when locationRadius is 0
```
cea2aj: What is the reason for supporting a locationRadius of 0 if the backend doesn't support it? Is it because they are going to support it in the future?

oshi97: A locationRadius of 0 in the SDK means "Don't set a locationRadius". 
(This is the previous behavior as well, with the old logic living in the FilterOptions component instead of core. Previously, we would set a FilterNode with an empty value if the locationRadius was 0. Now, in order to persist the actual value in core, we need to pass this 0 value through.)
In a distance filter, having a radio button selected that has a locationRadius of 0 is slightly different than not having a radio button selected at all. The query param sent to the backend is the same (locationRadius will be unset) but for the former the radio button will be checked, and we will also display it as an applied filter in the AppliedFilters/ResultsHeader components. We don't do this for the latter.

I think my comment is unclear, though, that is it because the backend does not support locationRadius of 0, the SDK has decided to make that mean "unset my locationRadius"

cea2aj: ah, got it!

cea2aj (approved):
lgtm

src/answers-umd.js
```diff
@@ -293,9 +301,19 @@ class Answers {
         throw new Error('MasterSwitchApi determined the front-end should be disabled');
       }
       this._onReady();
+      this._searchOnLoad();
     });
   }
 
+  /**
+   * Calls each SearchBar's searchAfterAnswersOnReady().
```
tmeyer2115: Give a little more detail here. I think we need to explain why we're calling things in this order.

src/answers-umd.js
```diff
@@ -588,6 +606,25 @@ class Answers {
   _getInitLocale () {
     return this.core.storage.get(StorageKeys.LOCALE);
   }
+
+  /**
+   * Parses a value from persistent storage, which stores strings,
```
tmeyer2115, line 621: What do you think about having this be its own class? I was envisioning we'd have a class some day whose responsibility is to properly seed `storage` with whatever is in the URL. Out of scope for this PR, but certain components have parsing logic like this. With such a class, we could move that parsing out of the components.

src/core/core.js
```diff
@@ -189,8 +189,12 @@ export default class Core {
...
+    // Partially because of this, the SDK has decided to make locationRadius of 0
+    // mean "unset my locationRadius"
+    const nonZeroLocationRadius = locationRadius === 0 ? undefined : locationRadius;
```
tmeyer2115: Having a `locationRadius` and a `nonZeroLocationRadius` is kind of confusing.

src/core/filters/filterregistry.js
oshi97: this has already been renamed!

src/answers-umd.js
oshi97: Yeah I think that makes sense!  is it okay if I leave it here for this pr?

src/ui/components/search/searchcomponent.js
```diff
@@ -314,7 +314,7 @@ export default class SearchComponent extends Component {
     return 'search/search';
   }
 
-  onCreate () {
+  afterAnswersOnReady () {
```
creotutar: This is 🥇 . The change I never knew I needed

src/core/core.js
oshi97: ah yeah it is, nonZeroLocationRadius is gone now!

src/ui/components/search/searchcomponent.js
oshi97: talked with Tom and we ended up killing it :( The idea is that only the searchbar (and filtersearch ~kind of) need this functionality, so for now it's just a thing that the searchbar can do

oshi97: we can probably refactor these two components to not be responsible for searching on page load, though it would probably be difficult, but then we wouldn't need this lifecycle hook

src/answers-umd.js
```diff
@@ -155,36 +157,45 @@ class Answers {
...
+      resetListener: data => {
+        let query = data.get(StorageKeys.QUERY);
+        if (!query && data.has(StorageKeys.PERSISTED_FILTER)) {
+          query = '';
+        }
+        this.core.storage.delete(StorageKeys.PERSISTED_LOCATION_RADIUS);
+        this.core.storage.delete(StorageKeys.PERSISTED_FILTER);
+        this.core.filterRegistry.clearAllFilterNodes();
+
+        if (!query && query !== '') {
```
oshi97, line 174: this is to handle when query is equal to blank string

src/ui/components/filters/filterboxcomponent.js
```diff
@@ -308,10 +310,8 @@ export default class FilterBoxComponent extends Component {
    * Trigger a search with all filters in storage
    */
   _search () {
-    this.core.verticalSearch(this._config.verticalKey, {
-      setQueryParams: true,
-      resetPagination: true,
-      useFacets: true
-    });
+    const query = this.core.storage.get(StorageKeys.QUERY) || '';
```
oshi97, line 313: Now, FilterBox searches by setting a query trigger and query, instead of directly calling the verticalSearch method.
This is so, if somebody lands on a page with no search triggered on load, then triggers a search using a filter component, the query will be updated to "", which is what a search triggered by a static filter will use if there is no query set in storage

oshi97: This was an edge case with back navigation

src/answers-umd.js
tmeyer2115: Sure, but let's make a TODO.

oshi97:
headless acceptance tests derping out (but not browserstack) investigating why
it looks like the tests pass locally in live mode (-L command line param) but fail if run normally...
edit: fixed now, looks like it had something to do with the logger not waiting properly

src/answers-umd.js
oshi97: added a todo!

src/answers-umd.js
oshi97: Updated the comment, how does it look?

tmeyer2115 (approved):

PR #1301, opened 2021-02-22
provide helpful error messages for missing apiKey or experienceKey

bjulian5:
The SDK does not provide a helpful error message when either apiKey or
experienceKey are missing. Instead, it produces a cruptic error that is
hard to diagnose.

This commit adds an AnswersConfig class that parses a raw answers
configuration object and validates it to ensure things like the apiKey
and experienceKey are properly set. If they are not set, it provides a
helpful error messsage. Optional configuration options not supplied by
the user are given default values.

J=SLAP-1016
TEST=manual

used dev.html to run an  answers configuration without an apiKey or
experienceKey and confirmed that helpful messages appeared in the console.

Ran all acceptance tests and unit tests and confirmed that all we're
still passing.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -19,7 +19,6 @@ import { AnswersComponentError } from './core/errors/errors';
 import AnalyticsEvent from './core/analytics/analyticsevent';
 import StorageKeys from './core/storage/storagekeys';
 import QueryTriggers from './core/models/querytriggers';
-import SearchConfig from './core/models/searchconfig';
 import AutoCompleteApi from './core/search/autocompleteapi';
 import MockAutoCompleteService from './core/search/mockautocompleteservice';
 import QuestionAnswerApi from './core/search/questionanswerapi';
@@ -27,14 +26,13 @@ import MockQuestionAnswerService from './core/search/mockquestionanswerservice';
 import SearchApi from './core/search/searchapi';
 import MockSearchService from './core/search/mocksearchservice';
 import ComponentManager from './ui/components/componentmanager';
-import VerticalPagesConfig from './core/models/verticalpagesconfig';
-import { SANDBOX, PRODUCTION, LOCALE, QUERY_SOURCE } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
 import RichTextFormatter from './core/utils/richtextformatter';
 import { isValidContext } from './core/utils/apicontext';
 import FilterNodeFactory from './core/filters/filternodefactory';
 import { urlWithoutQueryParamsAndHash } from './core/utils/urlutils';
 import TranslationProcessor from './core/i18n/translationprocessor';
+import AnswersConfig from './core/models/answersconfig';
 
 /** @typedef {import('./core/services/searchservice').default} SearchService */
 /** @typedef {import('./core/services/autocompleteservice').default} AutoCompleteService */
@@ -49,11 +47,6 @@ import TranslationProcessor from './core/i18n/translationprocessor';
  * @property {QuestionAnswerService} questionAnswerService
  * @property {ErrorReporterService} errorReporterService
  */
-
-const DEFAULTS = {
-  locale: LOCALE,
-  querySource: QUERY_SOURCE
-};
 
 /**
  * The main Answers interface
@@ -160,17 +153,10 @@ class Answers {
    */
   init (config, statusPage) {
     window.performance.mark('yext.answers.initStart');
-    const parsedConfig = this.parseConfig(config);
-    this.validateConfig(parsedConfig);
-
-    parsedConfig.search = new SearchConfig(parsedConfig.search);
-    parsedConfig.verticalPages = new VerticalPagesConfig(parsedConfig.verticalPages);
-
+    const answersConfig = new AnswersConfig(config);
     const storage = new Storage({
       update: (data, url) => {
-        if (parsedConfig.onStateChange) {
-          parsedConfig.onStateChange(Object.fromEntries(data), url);
-        }
+        answersConfig.onStateChange(Object.fromEntries(data), url);
       },
       reset: data => {
         if (!data.get(StorageKeys.QUERY)) {
@@ -199,10 +185,10 @@ class Answers {
       }
     });
     storage.init(window.location.search);
-    storage.set(StorageKeys.SEARCH_CONFIG, parsedConfig.search);
-    storage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
-    storage.set(StorageKeys.LOCALE, parsedConfig.locale);
-    storage.set(StorageKeys.QUERY_SOURCE, parsedConfig.querySource);
+    storage.set(StorageKeys.SEARCH_CONFIG, answersConfig.search);
+    storage.set(StorageKeys.VERTICAL_PAGES_CONFIG, answersConfig.verticalPages);
+    storage.set(StorageKeys.LOCALE, answersConfig.locale);
+    storage.set(StorageKeys.QUERY_SOURCE, answersConfig.querySource);
 
     // Check if sessionsOptIn data is stored in the URL. If it is, prefer that over
     // what is in parsedConfig.
@@ -210,7 +196,7 @@ class Answers {
     if (!sessionOptIn) {
       storage.set(
         StorageKeys.SESSIONS_OPT_IN,
-        { value: parsedConfig.sessionTrackingEnabled, setDynamically: false });
+        { value: answersConfig.sessionTrackingEnabled, setDynamically: false });
     } else {
       // If sessionsOptIn was stored in the URL, it was stored only as a string.
       // Parse this value and add it back to storage.
@@ -219,7 +205,7 @@ class Answers {
         { value: (/^true$/i).test(sessionOptIn), setDynamically: true });
     }
 
-    parsedConfig.noResults && storage.set(StorageKeys.NO_RESULTS_CONFIG, parsedConfig.noResults);
+    answersConfig.noResults && storage.set(StorageKeys.NO_RESULTS_CONFIG, answersConfig.noResults);
     const isSuggestQueryTrigger =
       storage.get(StorageKeys.QUERY_TRIGGER) === QueryTriggers.SUGGEST;
     if (storage.get(StorageKeys.QUERY) && !isSuggestQueryTrigger) {
@@ -240,24 +226,24 @@ class Answers {
     }
 
     this._masterSwitchApi = statusPage
-      ? new MasterSwitchApi({ apiKey: parsedConfig.apiKey, ...statusPage }, storage)
-      : MasterSwitchApi.from(parsedConfig.apiKey, parsedConfig.experienceKey, storage);
+      ? new MasterSwitchApi({ apiKey: answersConfig.apiKey, ...statusPage }, storage)
+      : MasterSwitchApi.from(answersConfig.apiKey, answersConfig.experienceKey, storage);
 
-    this._services = parsedConfig.mock
+    this._services = answersConfig.mock
       ? getMockServices()
-      : getServices(parsedConfig, storage);
+      : getServices(answersConfig, storage);
 
-    this._eligibleForAnalytics = parsedConfig.businessId != null;
+    this._eligibleForAnalytics = answersConfig.businessId != null;
     // TODO(amullings): Initialize with other services
-    if (this._eligibleForAnalytics && parsedConfig.mock) {
+    if (this._eligibleForAnalytics && answersConfig.mock) {
       this._analyticsReporterService = new NoopAnalyticsReporter();
     } else if (this._eligibleForAnalytics) {
       this._analyticsReporterService = new AnalyticsReporter(
-        parsedConfig.experienceKey,
-        parsedConfig.experienceVersion,
-        parsedConfig.businessId,
-        parsedConfig.analyticsOptions,
-        parsedConfig.environment);
+        answersConfig.experienceKey,
+        answersConfig.experienceVersion,
+        answersConfig.businessId,
+        answersConfig.analyticsOptions,
+        answersConfig.environment);
 
       // listen to query id updates
       storage.registerListener({
@@ -271,35 +257,31 @@ class Answers {
     }
 
     this.core = new Core({
-      apiKey: parsedConfig.apiKey,
+      apiKey: answersConfig.apiKey,
       storage: storage,
-      experienceKey: parsedConfig.experienceKey,
-      fieldFormatters: parsedConfig.fieldFormatters,
-      experienceVersion: parsedConfig.experienceVersion,
-      locale: parsedConfig.locale,
+      experienceKey: answersConfig.experienceKey,
+      fieldFormatters: answersConfig.fieldFormatters,
+      experienceVersion: answersConfig.experienceVersion,
+      locale: answersConfig.locale,
       searchService: this._services.searchService,
       autoCompleteService: this._services.autoCompleteService,
       questionAnswerService: this._services.questionAnswerService,
       analyticsReporter: this._analyticsReporterService,
-      onVerticalSearch: parsedConfig.onVerticalSearch,
-      onUniversalSearch: parsedConfig.onUniversalSearch
+      onVerticalSearch: answersConfig.onVerticalSearch,
+      onUniversalSearch: answersConfig.onUniversalSearch
     });
 
-    if (parsedConfig.onStateChange && typeof parsedConfig.onStateChange === 'function') {
-      parsedConfig.onStateChange(
-        Object.fromEntries(storage.getAll()),
-        this.core.storage.getCurrentStateUrlMerged());
-    }
+    answersConfig.onStateChange(Object.fromEntries(storage.getAll()), this.core.storage.getCurrentStateUrlMerged());
 
     this.components
       .setCore(this.core)
       .setRenderer(this.renderer);
 
-    this._setDefaultInitialSearch(parsedConfig.search);
+    this._setDefaultInitialSearch(answersConfig.search);
 
-    this._onReady = parsedConfig.onReady || function () {};
+    this._onReady = answersConfig.onReady;
 
-    const asyncDeps = this._loadAsyncDependencies(parsedConfig);
+    const asyncDeps = this._loadAsyncDependencies(answersConfig);
     return asyncDeps.finally(() => {
       if (this._disabledByMasterSwitch) {
         throw new Error('MasterSwitchApi determined the front-end should be disabled');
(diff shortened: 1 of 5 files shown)


src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
+/** @module AnswersConfig */
+
+import { LOCALE, PRODUCTION, QUERY_SOURCE, SANDBOX } from '../constants';
+import SearchConfig from '../models/searchconfig';
+import VerticalPagesConfig from '../models/verticalpagesconfig';
+
+const sandboxPrefix = `${SANDBOX}-`;
+const DEFAULTS = {
```
bjulian5: @tmeyer2115 I added some more defaults here, not sure if we need all of these

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+     */
+    this.disableCssVariablesPonyfill = parsedConfig.disableCssVariablesPonyfill;
+
+    /**
+     * Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD' or 'OVERLAY', defaults to 'STANDARD'
+     * @type {string}
+     */
+    this.querySource = parsedConfig.querySource;
+
+    this.mock = parsedConfig.mock;
```
bjulian5: Should I add a jsdoc for this? Didn't see it in the Answers readme but it's used by answers-umd.js

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+     * @private
+     */
+    this._environment = parsedConfig.environment;
+
+    /**
+     * Required, the Yext Answers API Key
+     * @private
+     * @type {string}
+     */
+    this._apiKey = parsedConfig.apiKey;
```
bjulian5: Can I set these as non-private and still have the getters for these?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+    this.disableCssVariablesPonyfill = parsedConfig.disableCssVariablesPonyfill;
+
+    /**
+     * Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD' or 'OVERLAY', defaults to 'STANDARD'
+     * @type {string}
+     */
+    this.querySource = parsedConfig.querySource;
+
+    this.mock = parsedConfig.mock;
+    this.analyticsOptions = parsedConfig.analyticsOptions;
```
bjulian5: This is not in the Answers README but is referenced in answers-umd.js. Not sure what the jsdoc for this parameter should be.

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+    this.fieldFormatters = parsedConfig.fieldFormatters;
```
bjulian5: Same thing as above

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+    this.validate();
+    Object.freeze(this);
+  }
+
+  /**
+   * Validates the Answers config object to ensure things like api key and experience key are
+   * properly set.
+   */
+  validate () {
+    if (typeof this._apiKey !== 'string') {
```
bjulian5: Are there any other values we'd want to validate?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+   * user are given default values.
+   * @param {Object} config The user supplied config.
+   */
+  parseConfig (config) {
+    const parsedConfig = Object.assign({}, DEFAULTS, config);
+    let sessionTrackingEnabled = true;
+
+    if (typeof config.sessionTrackingEnabled === 'boolean') {
+      sessionTrackingEnabled = config.sessionTrackingEnabled;
+    }
```
bjulian5: this feels weird, maybe we should be catching this in the validation phase

src/answers-umd.js
```diff
@@ -19,22 +19,20 @@ import { AnswersComponentError } from './core/errors/errors';
...
 import ComponentManager from './ui/components/componentmanager';
-import VerticalPagesConfig from './core/models/verticalpagesconfig';
-import { SANDBOX, PRODUCTION, LOCALE, QUERY_SOURCE } from './core/constants';
 import MasterSwitchApi from './core/utils/masterswitchapi';
 import RichTextFormatter from './core/utils/richtextformatter';
 import { isValidContext } from './core/utils/apicontext';
 import FilterNodeFactory from './core/filters/filternodefactory';
 import { urlWithoutQueryParamsAndHash } from './core/utils/urlutils';
 import TranslationProcessor from './core/i18n/translationprocessor';
+import AnswersConfig from './core/models/answersconfig';
```
bjulian5: Should I add a typeDef here?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+  debug: false,
+  sessionTrackingEnabled: true,
+  onStateChange: function () {},
+  onVerticalSearch: function () {},
+  onUniversalSearch: function () {},
+  disableCssVariablesPonyfill: false,
+  querySource: QUERY_SOURCE
+};
+
+export default class AnswersConfig {
```
tmeyer2115, line 9: Let's add some JSDoc for this class.

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+  constructor (config) {
```
tmeyer2115: Small nit: could we call this `rawConfig` or something like that? It just feels a bit weird for a class called `AnswersConfig` to be initialized with a `config`.

src/core/models/answersconfig.js
tmeyer2115: The existing name we had in `answers-umd` is a bit misleading. This value tells us whether or not we should use the mocked versions of the various services. We use these mocks in testing. It may be clearer to call this `useMock` or something like that.

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+
+    this.validate();
+    Object.freeze(this);
+  }
+
+  /**
+   * Validates the Answers config object to ensure things like api key and experience key are
+   * properly set.
+   */
+  validate () {
```
tmeyer2115: We should add some unit tests for this class. We want to make sure the parsing happens properly and that the `validate` method throws in the expected situations.

src/answers-umd.js
cea2aj: I believe that @oshi97 likes to add them so that the IDE can read the jsdocs from the class. Are you able to see any help text when you click on the class name?

src/core/models/answersconfig.js
tmeyer2115, line 9: What would you think about having two classes: `AnswersConfig` and an `AnswersConfigBuilder`. The role of the latter would be to create the former. The validation logic would live in the `AnswersConfigBuilder#build` method. Here's an article about the builder pattern in JS: https://medium.com/better-programming/the-builder-pattern-in-javascript-6f3d85c3ae4a. It might be nice to separate out the actual `AnswersConfig` data model from how it is constructed. The call chain could be something like:

```
const answersConfig = new AnswersConfigBuilder()
.setRawConfig(config)
.build();
```

The `AnswersConfig` would basically be a wrapper around the object it's passed, nothing too fancy. In the future, we might do something more interesting with the `AnswersConfig` model. If we ever use TypeSciprt, we can make it a strongly typed concept.

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * The internal Yext environment
+     * @type {string}
+     * @private
+     */
+    this._environment = parsedConfig.environment;
+
+    /**
+     * Required, the Yext Answers API Key
+     * @private
```
cea2aj: nit (nitpick): Can you put `@private` below the type for consistency?

src/answers-umd.js
oshi97: Yeah I like my intellisense haha, it makes me feel like I'm doing things correctly. If your favorite code completion is working properly already there's no need.

src/core/models/answersconfig.js
cea2aj: You cannot have a property hold a value and also have a getter bound to it. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#description

src/answers-umd.js
oshi97: Since we're using the AnswersConfig constructor directly here I think we should be fine

src/core/models/answersconfig.js
cea2aj: I think you're good since this covers what was in the JIRA item

src/core/models/answersconfig.js
cea2aj: i agree, this makes more sense in the validation phase

src/core/models/answersconfig.js
cea2aj: Since this method is only used internally, we would typically prefix it with an underscore

src/core/models/answersconfig.js
cea2aj: I second this idea. Something else to keep in mind is that the validation is currently happening on the parsed config after the defaults have been applied. I think it makes more sense to perform the validation on the raw config, and Tom's suggested changes might make that easier to do

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+    if (typeof this.experienceKey !== 'string') {
+      throw new Error('Missing required `experienceKey`. Type must be {string}');
+    }
+
+    if (typeof this.onVerticalSearch !== 'function') {
+      throw new Error('onVerticalSearch must be a function. Current type is: ' + typeof this.onVerticalSearch);
+    }
+
+    if (typeof this.onUniversalSearch !== 'function') {
+      throw new Error('onUniversalSearch must be a function. Current type is: ' + typeof this.onUniversalSearch);
```
creotutar: do we want to throw an `AnswersConfigError` here?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,216 @@
...
+   * The Yext Answers API Key. If the apiKey contains the sandbox prefix, then it is removed leaving only the API key.
+   * @returns {string}
+   */
+  get apiKey () {
+    return this._apiKey.replace(sandboxPrefix, '');
+  }
+
+  /**
+   * The internal Yext Environment.
+   *  Defaults to Production
```
creotutar: nit: is this extra space here intentional?

bjulian5: Probably not, I think it's a go convention that I do out of habit

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,214 @@
...
+  onVerticalSearch: function () {},
+  onUniversalSearch: function () {},
+  disableCssVariablesPonyfill: false,
+  querySource: QUERY_SOURCE
+};
+
+/**
+ * AnswersConfig is a model that is used that is used to turn a raw configuration into a data model that also storesVerticalPagesConfig}. It is used to initialze the Answers SDK
+ */
+class AnswersConfig {
```
tmeyer2115: Let's break this file up into two classes: one for `AnswersConfig`, the other for `AnswersConfigBuilder`. It's a nicer separation of concerns: one class and one file for the config, another class and another file for the builder.

tests/core/models/answersconfig.js
```diff
@@ -0,0 +1,131 @@
...
+  it('throws an error when the api key is missing from the raw answers config', () => {
+    const config = {
+      experienceKey: key
+    };
+    const buildingAnswersConfig = () => new AnswersConfigBuilder().setRawConfig(config).build();
+    expect(buildingAnswersConfig).toThrow(AnswersConfigError);
+    expect(buildingAnswersConfig).toThrow('Missing required `apiKey`. Type must be {string}');
+  });
+
+  it('throws an error when the api key is missing from the raw answers config', () => {
```
tmeyer2115: when the experience key

src/answers-umd.js
```diff
@@ -271,35 +260,31 @@ class Answers {
...
+      onVerticalSearch: answersConfig.onVerticalSearch,
+      onUniversalSearch: answersConfig.onUniversalSearch
     });
 
-    if (parsedConfig.onStateChange && typeof parsedConfig.onStateChange === 'function') {
-      parsedConfig.onStateChange(
-        Object.fromEntries(storage.getAll()),
-        this.core.storage.getCurrentStateUrlMerged());
-    }
+    answersConfig.onStateChange(Object.fromEntries(storage.getAll()), this.core.storage.getCurrentStateUrlMerged());
```
creotutar: nit: long line, can we keep to 100 characters?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+    this.environment = rawConfig.environment;
+
+    /**
+     * Required, the key used for your Answers Experience
+     * @type {string}
+     */
+    this.experienceKey = rawConfig.experienceKey;
+
+    /**
+     *  Optional, initialize components here, invoked when the Answers component library is loaded/ready.
```
creotutar: nit: can we fix the spacing here b/t `*` and text?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+     *  If components are not added here, they can also be added when the init promise resolves
+     * @type {function()}
+     */
+    this.onReady = rawConfig.onReady;
+
+    /**
+     * Optional*, Yext businessId, *required to send analytics events
```
creotutar: nit: what does `Optional*` and `*required` mean?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+     * @type {number|string|*}
```
creotutar: why do we allow more than one type here? i would think that is a number based on its usage in AnalyticsReporter. although looking at our documentation and jambo `ANSWERS.init` call, it looks like we pass in a string.

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+     */
+    this.businessId = rawConfig.businessId;
+
+    /**
+     * Optional, if false, the library will not fetch pre-made templates. Only use change this to false if you provide a
```
creotutar: "Only use change this to false"

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+     * Optional, vertical no results settings. See Vertical No Results
+     * @type {Object}
+     */
+    this.noResults = rawConfig.noResults;
+
+    /**
+     * Optional, the locale will affect how queries are interpreted and the results returned. Defaults to 'en'
+     * @type {string}
+     */
+    this.locale = rawConfig.locale;
```
creotutar, line 78: is this defaulting to `en`?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+    this.onStateChange = rawConfig.onStateChange;
+
+    /**
+     * Optional, analytics callback after a vertical search, see onVerticalSearch Configuration for additional details
+     * @type {function()}
+     */
+    this.onVerticalSearch = rawConfig.onVerticalSearch;
+
+    /**
+     * Optional, analytics callback after a universal search, see onUniversalSearch Configuration for additional details
```
creotutar: does this "see onUniversalSearch" comment make sense here outside of the context of the README?

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Used to signify if mocked versions of services should be used
+     * @type{boolean}
+     */
+    this.useMock = rawConfig.mock;
+
+    /**
+     * Global options to include with every analytic event reported to the server
+     * @type {object}
```
creotutar: nit: should this be `Object`?

src/core/models/answersconfig.js
bjulian5: I believe that was my auto-jsdoc. There is a string usage of it in tests/acceptance/fixtures/html/vertical.html where as other places have it being an integer

src/core/models/answersconfig.js
bjulian5: I copied the description from the README so I'm not sure myself haha

src/core/models/answersconfig.js
bjulian5: it defaults to
```
/** The current locale, injected by the build process */
export const LOCALE = '@@LOCALE';
```
which is located in src/core/constants.js

src/core/models/answersconfig.js
bjulian5: I don't think that needs to be there so removing

cea2aj:
At the top of PR descriptions, we usually give a one-line summary of the PR. It's usually pretty similar to the PR title

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
+/** @module AnswersConfig */
+import SearchConfig from '../models/searchconfig';
+import VerticalPagesConfig from '../models/verticalpagesconfig';
+
+/**
+ * AnswersConfig is a model that is used that is used to turn a raw configuration into a data model.
```
cea2aj: "that is used" appears twice in a row here

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+    this.locale = rawConfig.locale;
+
+    /**
+     * Optional, the Answers Experience version to use for api requests
+     * @type {string}
+     */
+    this.experienceVersion = rawConfig.experienceVersion;
+
+    /**
+     * Optional, prints full Answers error details when set to \`true\`. Defaults to false.
```
cea2aj: What do you think about leaving out the defaults from these jsdocs? By leaving them in here, it means that we have to remember to update this file when the defaults change in the config builder

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Required, the key used for your Answers Experience
+     * @type {string}
+     */
+    this.experienceKey = rawConfig.experienceKey;
+
+    /**
+     * Optional, initialize components here, invoked when the Answers component library is loaded/ready.
+     * If components are not added here, they can also be added when the init promise resolves
+     * @type {function()}
```
cea2aj: We usually write the type without the parentheses after the function like: `@type {function}`

tests/core/models/answersconfigbuilder.js
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it('has defaults properly applied', () => {
+    const config = {
+      apiKey: key,
+      experienceKey: key
+    };
+
+    const answersConfig = new AnswersConfigBuilder().setRawConfig(config).build();
+    expect(answersConfig.environment).toBe(PRODUCTION);
```
cea2aj, line 86: nit: We generally try to minimize the number of expects in each test. For example, you may be able to use `.toMatchObject` to do this in one expect. I don't think you need to change this as it's pretty clear what's going on, but it's something to keep in mind

src/core/models/answersconfig.js
```diff
@@ -0,0 +1,147 @@
...
+    this.onUniversalSearch = rawConfig.onUniversalSearch;
+
+    /**
+     * Optional, opt-out of automatic css variable resolution on init for legacy browsers
+     * @type {boolean}
+     */
+    this.disableCssVariablesPonyfill = rawConfig.disableCssVariablesPonyfill;
+
+    /**
+     * Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD' or 'OVERLAY', defaults to 'STANDARD'
```
creotutar: nit: can we restrict this to 100 characters too?

tmeyer2115 (approved):

tmeyer2115:
Since we don't currently have the `feat(..): ` convention, I'd suggest removing it from the commit title. It may be something we adopt in the future though!

PR #1304, opened 2021-02-24
Remove mock services and errorreporter

oshi97:
We actually had a discussion about whether anybody reads the sentry errors reported by the error reporter.
We don't, and neither do they. (We'll probably do this to main SDK branch at some point)
With the mock services, we don't even have tests that use these, but they take a little bit of bundle space.
I think this should be the last easy slimming of the mercy SDK fork. Additional improvements would probably
have to actually refactor code instead of just remove deprecated junk.

TEST=manual

test that mercy with the dev/mercy-remove-more-stuff-dist branch, see results load for each vertical


benmcginnis (approved):
deleting code is so cathartic

PR #1306, opened 2021-02-26
universalresults,verticalresults: remove default start icons

bjulian5:
Remove default star icons from SDK

If you don't specify an iconin the SDK in any of the following:
1. The top level verticals object
2. The component that is using it (vertical results for the no results
template, and universal results)

It should default to not showing an icon at all instead of a "star"

J=SLAP-1043
TEST=manual

- Performed universal search and confirmed that "star" icon no longer
  showed up in the DOM

### DIFF
--- a/src/ui/templates/results/verticalresults.hbs
+++ b/src/ui/templates/results/verticalresults.hbs
@@ -26,17 +26,17 @@
 {{#*inline "titleBar"}}
   {{#if _config.isUniversal}}
     <div class="yxt-Results-titleBar">
-    {{#if _config.icon}}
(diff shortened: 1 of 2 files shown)


src/ui/templates/results/verticalresults.hbs
```diff
@@ -26,6 +26,7 @@
 {{#*inline "titleBar"}}
   {{#if _config.isUniversal}}
     <div class="yxt-Results-titleBar">
+    {{#if _config.icon}}
```
creotutar: could we move this (and its children) one indent to the right?

tmeyer2115 (approved):

PR #1311, opened 2021-03-03
Add transformFacets options to the Facets component

cea2aj:
Add transformFacets options to the Facets component

Supports full customization of the facets data. One of the main use cases of this is to allow the displayNames to be customized for boolean facets.

This PR updates the DynamicFilters model to use the answers-core DisplayableFacet model internally. This allow us to apply `transformFacets` on the core data model and then finally convert it to the SDK model in the facet component's `setState()`

This PR deprecates the facets "fields" option

J=SLAP-1121
TEST=manual, unit

Use the config option to change the label name of boolean facets. Confirm that the fields option still works. Test that the field option is disabled if the transformFacets option is supplied

### DIFF
--- a/README.md
+++ b/README.md
@@ -1386,15 +1386,17 @@ Here's an example of using this option to customize a boolean facet.
 
 ```js
 transformFacets: facets => {
-  facets.forEach(facet => {
-    if (facet.fieldId === 'c_acceptingNewPatients') {
-      facet.options.forEach(option => {
-        if (option.value === false) { option.displayName = "Not Accepting Patients"; }
-        if (option.value === true) { option.displayName = "Accepting Patients"; }
-      });
-    }
+  return facets.map(facet => {
+    const options = facet.options.map(option => {
+      if (facet.fieldId === 'c_acceptingNewPatients') {
+        const displayName = option.value === true
+          ? "Accepting Patients"
+          : "Not Accepting Patients";
+      }
+      return Object.assign({}, option, { displayName });
+    });
+    return Object.assign({}, facet, { options });
   });
-  return facets;
 },
 ```
 
--- a/README.md
+++ b/README.md
@@ -1388,10 +1388,10 @@ Here's an example of using this option to customize a boolean facet.
 transformFacets: facets => {
   return facets.map(facet => {
     const options = facet.options.map(option => {
+      let displayName = option.displayName;
       if (facet.fieldId === 'c_acceptingNewPatients') {
-        const displayName = option.value === true
-          ? "Accepting Patients"
-          : "Not Accepting Patients";
+        if (option.value === false) { displayName = "Not Accepting Patients"; }
+        if (option.value === true) { displayName = "Accepting Patients"; }
       }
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1373,6 +1377,26 @@ ANSWERS.addComponent('Facets', {
...
 
+### Transforming Facets
+
+The `transformFacets` option of the Facets component allows facets data to be fully customized. The function takes in and returns an array of the answers-core DisplayableFacet which is described [here](https://github.com/yext/answers-core/blob/master/docs/answers-core.displayablefacet.md). The function also has access to the Facets config as the second parameter.
+
+Here's an example of using this option to customize a boolean facet:
+
+```js
+transformFacets: facets => {
+  facets.forEach(facet => {
```
creotutar: Do we want to be changing the input `facets` object? I think we would want them to construct a new facets object, right?

cea2aj: no, that's a good point. i'll update

cea2aj: I updated the example to use JSON for the deep copy. Do you think we might need something more robust such as lodash? JSON might not be sufficient if the facet has a date in it

cea2aj: Just updated to lodash

cea2aj: spoke to Tom offline. We decided to create a deep copy within the facets component so that someone can safely modify the facets object passed into this function

creotutar: Sounds good! I would think we want to give them example code without side effects too, right?

cea2aj: While the example modifies the param object, there aren't any side effects within the facets component since that object was created as a deep clone. I tested this out and confirmed that the internal state is not affected when the facets object is modified. We could give them an example with a lodash deep clone, but that seems unnecessary since that would mean the facets object would be deep cloned twice which just means extra computation for no additional benefit

src/ui/components/filters/facetscomponent.js
```diff
@@ -237,14 +260,17 @@ export default class FacetsComponent extends Component {
       this._filterbox.remove();
     }
 
-    let { filters, resultsContext } = this._state.get();
+    let filters = this._getFacets();
```
oshi97: could we do this transformation in setState()? I don't think there's a huge difference but it seems a little more thematic of the SDK, also a few less this._state.get() calls and we probably don't need to deepclone. It feels like transformFacets is a transformData but with a stricter, less leaky promise

cea2aj: Are you saying to call transformFacets during setState and store that in the state?

oshi97: yeah exactly that, instead of the original "data". we can also use the "isNoResults" state

README.md
cea2aj: The example now no longer has any side effects

src/ui/components/filters/facetscomponent.js
cea2aj: updated to do all the transforming in setState!

creotutar (approved):
👍 👍 👍 LGTM

PR #1312, opened 2021-03-08
Empty universal search

cea2aj:
Support empty universal searches

J=SLAP-1153
TEST=manual, auto

Refer to the minispec and test the 5 listed use cases which are combinations of allowEmptySearch (true or false), defaultInitialSearch (empty string, string, or not supplied). Also modify the data transformer to mock empty universal search results and make sure that the "no results" message shows up in accordance with the spec.

Add a unit test to simulate a default initial search. I added a `wasSearchRanPromise` because the search method is ran asynchronously and if we try to use a standard mock function with expect().toHaveBeenCalled(), the test fails because the expect runs before the search method

### DIFF
--- a/src/ui/components/results/universalresultscomponent.js
+++ b/src/ui/components/results/universalresultscomponent.js
@@ -59,7 +59,7 @@ export default class UniversalResultsComponent extends Component {
       isPreSearch: searchState === SearchStates.PRE_SEARCH,
       isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
       isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE,
-      showNoResults: sections.length === 0 && query,
+      showNoResults: sections.length === 0 && (query || query === ''),
       query: query,
       sections: sections
     }, val));
--- a/tests/ui/components/search/searchcomponent.js
+++ b/tests/ui/components/search/searchcomponent.js
@@ -97,6 +97,6 @@ describe('SearchBar component', () => {
       });
     });
 
-    return expect(wasSearchRanPromise).resolves.toBe(true);
+    return expect(wasSearchRanPromise).resolves.toBeTruthy();
   });
 });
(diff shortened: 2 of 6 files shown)


src/ui/components/results/universalresultscomponent.js
```diff
@@ -59,7 +59,7 @@ export default class UniversalResultsComponent extends Component {
       isPreSearch: searchState === SearchStates.PRE_SEARCH,
       isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
       isSearchComplete: searchState === SearchStates.SEARCH_COMPLETE,
-      showNoResults: sections.length === 0 && query,
+      showNoResults: sections.length === 0 && (query || query === ''),
```
tmeyer2115, line 62: Do we need to add `allowEmptySearch` in here at all? Like `(query || (query === ''" && allowEmptySearch)`. Or am I worried about an edge case that doesn't exist?

tests/ui/components/search/searchcomponent.js
```diff
@@ -79,4 +80,23 @@ describe('SearchBar component', () => {
...
+    storage.set(StorageKeys.QUERY, defaultInitialSearch);
+    storage.set(StorageKeys.QUERY_TRIGGER, QueryTriggers.INITIALIZE);
+
+    const component = COMPONENT_MANAGER.create('SearchBar', {
+      ...defaultConfig,
+      allowEmptySearch: true
+    });
+
+    const wasSearchRanPromise = new Promise(resolve => {
+      component.search = jest.fn(() => {
```
oshi97: would it make sense to mock core.search instead?

src/ui/components/results/universalresultscomponent.js
cea2aj: The search component prevents empty searches from occurring if `allowEmptySearch` is false, so extra logic in the universalresultscomponent isn't necessary

tests/ui/components/search/searchcomponent.js
cea2aj: I tried that, but I wasn't able to call the promise resolve function from within this test. Do you have other ideas on how to test that it was called in an asynchronous way?

oshi97: ah I see that IS tricky because there's no callback for the component's async stuff. I think we could still change the mock from component.search to component.core.search though

cea2aj: updated!

oshi97 (approved):
lgtm!

PR #1314, opened 2021-03-09
Set QUERY_PARAMETER query trigger even when query=""

oshi97:
This commit fixes a bug where the QUERY_PARAMETER query trigger
was not being set when query was a blank string. This was causing
pagination to not be persisted on page refresh for blank string queries.

J=SLAP-1089
TEST=manual,auto

test that I can run a blank query, update the page, then refresh
and the page will be persisted

double checked that undoing this fix will cause the newly added
acceptance test to fail


cea2aj (approved):
looks good

PR #1315, opened 2021-03-09
Add 'types' option to the direct answer component

cea2aj:
Add a types option to the DA component and deprecate cardOverrides

Add a validation step to the constructor to make sure that no keys are supplied which aren't supported

J=SLAP-1156
TEST=auto

Add tests

### DIFF
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -327,6 +327,8 @@ export default class DirectAnswerComponent extends Component {
 
   /**
    * Throws an error if any typeOptions includes unsupported keys
+   *
+   * @param {Objet} typeOptions
    * @throws if validation fails
    */
   _validateTypeOptions (typeOptions) {
@@ -335,11 +337,10 @@ export default class DirectAnswerComponent extends Component {
   }
 
   /**
-   * Validates the cardOverrides option within the types option.
-   * Throws an error if the cardOverrides contains the 'type' option.
+   * Throws an error if cardOverrides within the types option contain unsupported keys
    *
    * @param {Object[]} overrides
-   lc
+   * @throws if validation fails
    */
   _validateCardOverrides (overrides) {
     const supportedKeys = ['fieldName', 'entityType', 'fieldType', 'cardType'];
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -326,7 +326,7 @@ export default class DirectAnswerComponent extends Component {
   }
 
   /**
-   * Throws an error if any typeOptions include unsupported keys
+   * Throws an error if any typeOptions includes unsupported keys
    * @throws if validation fails
    */
   _validateTypeOptions (typeOptions) {
--- a/src/ui/components/results/directanswercomponent.js
+++ b/src/ui/components/results/directanswercomponent.js
@@ -104,6 +104,8 @@ export default class DirectAnswerComponent extends Component {
      * @type {Object}
      */
     this._types = config.types;
+
+    this._validateTypes();
   }
 
   static get type () {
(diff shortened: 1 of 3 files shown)


src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +219,112 @@ export default class DirectAnswerComponent extends Component {
...
+      if (directAnswerPropeties[propertyToMatch] !== propertyValue) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Validates the cardOverrides option within the types option.
+   * Throws an error if the cardOverrides contains the 'type' option.
```
oshi97: I'm probably missing context but why do we throw an error for this?

src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +219,112 @@ export default class DirectAnswerComponent extends Component {
...
+   * @param {Object} directAnswer The direct answer state
+   * @returns {string}
+   */
+  _getCardBasedOnTypes (directAnswer) {
+    return Object.entries(this._types).reduce((cardType, [directAnswerType, typeOptions]) => {
+      if (directAnswerType !== directAnswer.type) {
+        return cardType;
+      }
+
+      const cardFallback = typeOptions.cardType || this._defaultCard;
```
creotutar: Is it a valid `types` object if there's no `cardType`? Do we want to default for them here?

src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +219,112 @@ export default class DirectAnswerComponent extends Component {
...
+  }
+
+  /**
+   * Determines the card that should be used based on the types option
+   *
+   * @param {Object} directAnswer The direct answer state
+   * @returns {string}
+   */
+  _getCardBasedOnTypes (directAnswer) {
+    return Object.entries(this._types).reduce((cardType, [directAnswerType, typeOptions]) => {
```
oshi97: could we just directly access the correct directAnswerType in this._types instead of looping?

src/ui/components/results/directanswercomponent.js
cea2aj: Checking with product

src/ui/components/results/directanswercomponent.js
cea2aj: This was called out in the spec of something that isn't supported. The "types" object is already keyed by the type of the direct answer, so it doesn't really make sense if we let people put another type within the cardOverrides. I asked Rose what to do in this situation and she said to throw an error.

src/ui/components/results/directanswercomponent.js
cea2aj: ah, ya the reduce is probably overkill here

src/ui/components/results/directanswercomponent.js
oshi97: maybe it would make sense to just add a validation step for the config? On the off chance somebody has some a complicated setup, but maybe has a typo in their cardType we can still fail gracefully then

cea2aj: sure, that makes sense

src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +221,150 @@ export default class DirectAnswerComponent extends Component {
...
+
+  /**
+   * Throws an error if an object has keys other than the supported keys.
+   *
+   * @param {string[]} supportedKeys A list of supported keys
+   * @param {Object} object The object to check the keys of
+   * @throws if validation fails
+   */
+  _validateSupportedKeysOfObject (supportedKeys, object) {
+    const supportedKeysString = supportedKeys.join(' and ');
```
oshi97: nit: can we move supportedKeysString into the if statement

src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +221,150 @@ export default class DirectAnswerComponent extends Component {
...
+    });
+  }
+
+  /**
+   * Throws an error if any typeOptions includes unsupported keys
+   *
+   * @param {Objet} typeOptions
+   * @throws if validation fails
+   */
+  _validateTypeOptions (typeOptions) {
```
oshi97: nit: would it be a little more concise to include these 3 methods into _validateTypes() instead? _validateSupportedKeysOfObject could be an inline function if you don't want to repeat things

src/ui/components/results/directanswercomponent.js
```diff
@@ -237,6 +221,150 @@ export default class DirectAnswerComponent extends Component {
...
+    Object.entries(this._types).forEach(([directAnswerType, typeOptions]) => {
+      this._validateTypeOptions(typeOptions);
+      typeOptions.cardOverrides && this._validateCardOverrides(typeOptions.cardOverrides);
+    });
+  }
+
+  /**
+   * Throws an error if any typeOptions includes unsupported keys
+   *
+   * @param {Objet} typeOptions
```
oshi97: Objet

src/ui/components/results/directanswercomponent.js
cea2aj: i'll update

oshi97 (approved):

PR #1316, opened 2021-03-10
Add CSS classes to indicate state of search on results components

bjulian5:
Toggle CSS classes onto results containers to ensure we can target them
in Theme Styling

J=SLAP-957
TEST=manual
 - Checked loading states when making a search in both universal and
 vertical searches
 - Wrote a unit test for VerticalSearch that asserts that the correct
   class is rendered in all loading states


cea2aj:
Percy is showing a diff on the facets page. @oshi97 can you take a look at that diff? I'm not sure what the right behavior should be

bjulian5:
The diff is from filter search 
image

I branched off of the tip of develop, is that correct?

cea2aj:
> I branched off of the tip of develop, is that correct?

Ya, that's correct. It looks like the problem is that percy is diffing this PR against master rather than develop. The snapshots for this branch match the percy snapshots for develop, so I'll approve the percy build

oshi97:
> The diff is from filter search
> image
> 
> I branched off of the tip of develop, is that correct?

Yup this should be fine! it's weird that percy is comparing to master instead of the feature branch, there might be something weird with the our setup/PR integration and how it works with feature branches. it looks like the feature branch was auto approved before this pr was made, though, so not sure exactly what

src/ui/templates/results/universalresults.hbs
```diff
@@ -1,12 +1,15 @@
-{{#if isSearchComplete}}
-  {{#if showNoResults}}
-    {{> results/noresults}}
-  {{else}}
-    <div class="universal-result-list">
-      <div class="universal-result-sections"
-          data-component="VerticalResults"
-          data-prop="sections">
+<div class="yxt-ResultsContainer
```
oshi97: could we add this css class to `universal-result-list` instead? We have to be really careful about changing the DOM structure of SDK components.

src/ui/templates/results/universalresults.hbs
```diff
@@ -1,12 +1,15 @@
...
+  yxt-ResultsContainer--{{#if isPreSearch}}preSearch{{else if isSearchLoading}}loading{{else if isSearchComplete}}loaded{{/if}}">
```
oshi97: would it make sense to call this `yxt-UniversalResults` or something similar? the sdk kind of has this pattern of `yxt-{{component name}}` for css classes, though it's not followed 100% (as you probably already saw)

src/ui/templates/results/verticalresults.hbs
```diff
@@ -1,17 +1,20 @@
-{{#if isSearchComplete}}
-  {{#if showNoResults}}
-    {{> noResults}}
+<div class="yxt-ResultsContainer
```
oshi97: same questions about DOM structure and changing this to `yxt-Results` or `yxt-VerticalResults`

tests/ui/components/results/verticalresultscomponent.js
```diff
@@ -44,6 +45,20 @@ describe('vertical results component', () => {
     expect(component.getVerticalURL()).toContain('query=yext');
   });
 
+  it('sets correct loading state css class', () => {
```
oshi97, line 48: love the test!

src/ui/templates/results/universalresults.hbs
bjulian5: I had originally had it this way but I changed it to a wrapper because currently there are no dom objects if a search is not complete. I'm not sure if adding a wrapper or having `universal-result-list` permanently be in the dom was better. @tmeyer2115 should I be worried about leaving `universal-result-list` in the dom in all phases of the search? Also Should I be adding the class to `results/noresults` as well?

src/ui/components/component.js
```diff
@@ -342,7 +350,6 @@ export default class Component {
     }
 
     DOM.append(this._container, this.render(this._state.asJSON()));
-
```
creotutar: is this newline removal desired?

src/ui/dom/dom.js
```diff
@@ -144,6 +144,19 @@ export default class DOM {
     }
   }
 
+  static removeClass (node, className) {
```
creotutar, line 152: i know this is a copy of `addClass` but can we add jsdocs and turn our `let`s into `const`s when we can?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -51,11 +51,24 @@ export default class UniversalResultsComponent extends Component {
     return true;
   }
 
+  mount (container) {
+    const searchState = this.getState('searchState');
+    const camelCase = (str) => {
```
creotutar: can we make this a general util function and add jsdocs to it? it might be too specific if it does the `-` splitting but can take in an array of words to camel case. i believe we have a strings.js util file

src/ui/components/results/universalresultscomponent.js
```diff
@@ -51,11 +51,24 @@ export default class UniversalResultsComponent extends Component {
...
+      return str
+        .split('-')
+        .reduce((a, b) => a + b.charAt(0).toUpperCase() + b.slice(1));
+    };
+    Object.keys(SearchStates).forEach(state => super.removeContainerClass(`yxt-Results--${camelCase(state)}`));
+    super.addContainerClass(`yxt-Results--${camelCase(searchState)}`);
```
creotutar: is there a reason we're using `super` specifically here (and above) instead of `this`?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -254,9 +255,17 @@ export default class VerticalResultsComponent extends Component {
...
+  mount (container) {
     if (Object.keys(this.getState()).length > 0) {
-      super.mount();
+      const searchState = this.getState('searchState');
+      const camelCase = (str) => {
+        return str
+          .split('-')
+          .reduce((a, b) => a + b.charAt(0).toUpperCase() + b.slice(1));
+      };
+      Object.entries(SearchStates).forEach(([k, state]) => super.removeContainerClass(`yxt-Results--${camelCase(state)}`));
```
creotutar: is there a reason we use a different for loop b/t here and universalresults?

src/ui/components/results/universalresultscomponent.js
bjulian5: Oops, let me change that

src/ui/dom/dom.js
bjulian5: sure!

src/ui/components/results/verticalresultscomponent.js
bjulian5: Oops, universal results is wrong. I'll update

src/ui/dom/dom.js
```diff
@@ -144,6 +144,19 @@ export default class DOM {
...
+    if (!node) {
+      return;
+    }
+
+    let classes = className.split(',');
+    let len = classes.length;
+
+    for (let i = 0; i < len; i++) {
+      node.classList.remove(classes[i]);
+    }
```
cea2aj: We try to use functional js when possible, so something like this:
(class is a reserved keyword so I used clazz instead)

```suggestion

    classes.forEach(clazz => {
      node.classList.remove(clazz);
    };
```

src/ui/components/results/universalresultscomponent.js
bjulian5: We don't have a strings.js file but I can make one

src/ui/components/component.js
```diff
@@ -243,6 +243,22 @@ export default class Component {
...
+  /**
+   * Adds a class to the container of the component.
+   * @param {string} className - A comma separated value of classes
+   */
+  addContainerClass (className) {
+    DOM.addClass(this._container, className);
+  }
+
+  /**
+   * Removes the specified classes from the container of the compenent
```
cea2aj:
```suggestion
   * Removes the specified classes from the container of the component
```

src/ui/dom/dom.js
```diff
@@ -144,6 +144,20 @@ export default class DOM {
...
+   * Removes classes from a specified element.
+   * @param {HTMLElement} node - The html element to be acted upon
+   * @param {string} className - A comma separated list of classes to be removed
+   */
+  static removeClass (node, className) {
+    if (!node) {
+      return;
+    }
+
+    let classes = className.split(',');
```
cea2aj: nit: can this be a const?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -254,9 +256,13 @@ export default class VerticalResultsComponent extends Component {
     };
   }
 
-  mount () {
+  mount (container) {
```
oshi97: could this be done in an onMount instead? same question for universal results, it's just a little more common to do this kind of stuff in the onMount instead of overriding mount

src/ui/components/results/universalresultscomponent.js
```diff
@@ -51,11 +52,20 @@ export default class UniversalResultsComponent extends Component {
     return true;
   }
 
+  mount (container) {
```
oshi97: would it be possible to not pass in container here?

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -254,9 +256,13 @@ export default class VerticalResultsComponent extends Component {
...
     if (Object.keys(this.getState()).length > 0) {
-      super.mount();
+      const searchState = this.getState('searchState');
+      Object.entries(SearchStates).forEach(([k, state]) =>
+        this.removeContainerClass(`yxt-Results--${snakeCaseToCamelCase(state)}`));
```
oshi97: would it be clearer to just do a switch here? This is kind of using the implementation detail that we use snake case strings as enums, in order to create a specific css class. It's also slightly harder to search for where `yxt-Results--{state}` classes come from

bjulian5: I see you're point. Updating!

src/core/utils/stringutils.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Takes a snake-case string and converts it to camel-case
+ * @param {string} str - A string in snake-case
+ * @returns {string}
+ */
+export function snakeCaseToCamelCase (str) {
```
oshi97: this is unused I think

src/ui/components/results/abstractresultscomponent.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * AbstractResultsComponent is an abstraction for common functionality between {@link UniversalResultsComponent} and
+ * {@link VerticalResultsComponent}.
+ */
+export default class AbstractResultsComponent extends Component {
+  constructor (config = {}, systemConfig = {}) {
+    super(config, systemConfig);
+  }
+
+  onMount () {
```
oshi97: I think we try to avoid using inheritance for deduping code, because the inheritance chain can longterm make it more confusing to find out what's happening. We've kind of just been stuffing shared method into "utils" files

src/ui/components/results/verticalresultscomponent.js
```diff
@@ -215,7 +215,8 @@ export default class VerticalResultsComponent extends Component {
       eventType: 'update',
       storageKey: StorageKeys.VERTICAL_RESULTS,
       callback: results => {
-        if (results.searchState === SearchStates.SEARCH_COMPLETE) {
+        if (results.searchState === SearchStates.SEARCH_COMPLETE ||
```
oshi97, line 219: Just realizing this now, but would it be possible to call addContainerClass here, instead of in the onMount? that way we can avoid the extra render

src/ui/components/results/abstractresultscomponent.js
bjulian5: Should this live in `core/utils`?

oshi97: Yeah that would work! Either there or in src/ui/tools I think.

src/core/utils/resultsutils.js
```diff
@@ -0,0 +1,17 @@
...
+ * @returns {string}
+ */
+export function getContainerClass (searchState) {
+  switch (searchState) {
+    case SearchStates.PRE_SEARCH:
+      return 'yxt-Results--preSearch';
+    case SearchStates.SEARCH_LOADING:
+      return 'yxt-Results--searchLoading';
+    case SearchStates.SEARCH_COMPLETE:
+      return 'yxt-Results--searchComplete';
```
creotutar, line 15: looks like this can return `undefined`, can we get a default case?

src/core/utils/resultsutils.js
```diff
@@ -0,0 +1,17 @@
+import SearchStates from '../storage/searchstates';
+
+/**
+ * Returns a CSS class for the input searchState
+ * @param {string} searchState - A search state from {@link SearchStates}
```
creotutar: can this type be `SearchStates`

src/ui/components/component.js
```diff
@@ -243,6 +243,22 @@ export default class Component {
     return this;
   }
 
+  /**
+   * Adds a class to the container of the component.
+   * @param {string} className - A comma separated value of classes
```
creotutar: we normally don't have dashes in our jsdoc, can we remove in these cases?

src/core/utils/resultsutils.js
bjulian5: Isn't searchStates the entire object while the individual searchState is the string? I originally had it this way but my jsdoc yells at me it isn't a string.

src/ui/components/results/universalresultscomponent.js
```diff
@@ -55,6 +56,9 @@ export default class UniversalResultsComponent extends Component {
     const sections = data.sections || [];
     const query = this.core.storage.get(StorageKeys.QUERY);
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
+    ['preSearch', 'searchLoading', 'searchLoaded']
+      .forEach(state => this.removeContainerClass(`yxt-Results--${state}`));
```
creotutar: can we loop through the SearchStates and `getContainerClass` instead? that way the remove and add use the same class construction fn?

src/core/utils/resultsutils.js
bjulian5: Should we throw an error here or add an undefined searchState?

cea2aj:
+1 from me outside of Christian's comments

src/core/utils/resultsutils.js
```diff
@@ -0,0 +1,20 @@
...
+  switch (searchState) {
+    case SearchStates.PRE_SEARCH:
+      return 'yxt-Results--preSearch';
+    case SearchStates.SEARCH_LOADING:
+      return 'yxt-Results--searchLoading';
+    case SearchStates.SEARCH_COMPLETE:
+      return 'yxt-Results--searchComplete';
+    default:
+      console.trace('encountered an unknown search state');
+      return '';
```
cea2aj, line 18: Would returning `yxt-Results` be a little safer?

bjulian5: talked with @creotutar and we thought it would be better to return an empty string rather than an unintended class

cea2aj (approved):
LGTM

creotutar (approved):
💯 💯 💯 LGTM

PR #1320, opened 2021-03-11
Refactor SearchListener logic out of the SearchBar component

oshi97:
This commit adds a new SearchListener class to the SDK, which lives
inside core.js. This class is responsible for running a search when
a query is updated, which previously lived in the SearchBar component.
This refactor is necessary so that SearchBar-less pages can be made,
such as pages that use filtersearch as the main searchbar, or pages
that only have filters on them.

SearchBar specific details, i.e. fetching NEAR_ME intent
and the onConductSearch hook,  are now added to this SearchListener as
middleware. These middleware functions are run and awaited on before a search
is made.

The various filter components, as well as location bias and pagination,
have also been updated to use this new SearchListener.
This commit also fixes a bug introduced in the core-extraction branch,
where we no longer pass the queryId param to searches made by pagination.
The corresponding update has also been made to answers-core, and the SDK will need
to update it's version of answers-core once the update has been published.
Tests will still pass, because unit tests don't use answers-core directly,
and we don't have an acceptance test for queryId (the unit test coverage
should be fine as long as answers-core works as expected).

J=SLAP-1089
TEST=manual,auto

test that location bias can still run searches
test that pagination will send queryId in the request, if I plug in
my local answers-core with the update
test that spell check works, and will reset the pagination

test that a searchbar-less page can make searches, and change the page
test that defaultInitialSearch will be run on page load with no searchbar
test that defaultInitialSearch="" works on universal if allowEmptySearch is true
test that defaultInitialSearch="" will not init the query to ""  if allowEmptySearch is false
test that allowEmptySearch in the searchbar stops blank searches from firing, but still
allows filter components to fire blank searches

added acceptance test for blank defaultInitialSearch on universal
our slanswers experience key does not display any results on universal for blank
search, so used the old sdkacceptancetests experience key
unit tests for SearchListener
removed the request log checking in the filtersearch acceptance tests
for some reason the filters returned from the filtersearch request are
sometimes different when run through circleci. checking that the
filter tag pops up with the correct text should be enough to test things
test double searchbars still works (you still need different names for both components)


tests/acceptance/acceptancesuites/filtersearchsuite.js
```diff
@@ -32,63 +25,39 @@ test(`filtersearch works with back/forward navigation and page refresh`, async t
   };
   const filterTags = Selector('.yxt-ResultsHeader-removableFilterTag');
   const filterSearch = FacetsPage.getFilterSearch();
-  const virginiaFilter = {
```
oshi97: removed these because they were being inconsistent on circleci, I checked the browserstack video and it looked like the correct filtersearch option was being selected, but the actual filter being sent in the request was neither the virginia nor new york filter (and also this was passing locally) which makes me thinks circleci is getting a different filtersearch response
image

src/core/statelisteners/searchlistener.js
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    /**
+     * Middleware functions to be run before a search. Can be either async or sync.
+     */
+    this.middleware = [];
+  }
+
+  /**
+   * Sets the defaultInitialSearch, if no query is already set, and register's the
```
cea2aj: Since this function doesn't reference `defaultInitialSearch`, does it make sense to instead have this comment say "Sets the on load query and registers the QUERY listener"?

src/core/statelisteners/searchlistener.js
```diff
@@ -0,0 +1,177 @@
...
+    this.storage.registerListener({
+      storageKey: StorageKeys.QUERY,
+      eventType: 'update',
+      callback: query => this._handleQueryUpdate(query)
+    });
+    return this;
+  }
+
+  /**
+   * Sets a search query and corresponding QUERY_TRIGGER on initialization. A prepoluated query
```
cea2aj: prepoluated

src/core/statelisteners/searchlistener.js
```diff
@@ -0,0 +1,176 @@
...
+   * in the URL takes precendence over a defaultInitialSearch.
+   */
+  _setQueryOnLoad () {
+    const prepopulatedQuery = this.storage.get(StorageKeys.QUERY);
+    if (prepopulatedQuery != null) {
+      if (this.storage.get(StorageKeys.QUERY_TRIGGER) !== QueryTriggers.SUGGEST) {
+        this.storage.setWithPersist(StorageKeys.QUERY_TRIGGER, QueryTriggers.QUERY_PARAMETER);
+      }
+    } else if (this.config.defaultInitialSearch != null) {
+      if (!this.config.defaultInitialSearch && !this.config.allowEmptySearch) {
```
cea2aj: Some of these if statements are a little hard to follow. Is it possible to add some explanatory boolean values?

src/core/statelisteners/searchlistener.js
```diff
@@ -0,0 +1,176 @@
...
+        return;
+      }
+      this.storage.setWithPersist(StorageKeys.QUERY_TRIGGER, QueryTriggers.INITIALIZE);
+      this.storage.set(StorageKeys.QUERY, this.config.defaultInitialSearch);
+    }
+  }
+
+  _handleQueryUpdate (query) {
```
cea2aj: jsdoc?

src/core/statelisteners/searchlistener.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Returns the current search options state, based on the current QUERY_TRIGGER.
+   *
+   * @returns {Object}
+   */
+  _getSearchOptions () {
+    const queryTrigger = this.storage.get(StorageKeys.QUERY_TRIGGER);
+    if (queryTrigger === QueryTriggers.FILTER_COMPONENT) {
```
cea2aj: What do you think of having the pagination or filter component be responsible for the search options when triggering a search? It would let us avoid having to update search listener when a component changes its search options

oshi97 (commented):
github bugging out (keeps saying I have a review pending... I tried to respond to some comments during the github outage)

src/core/statelisteners/searchlistener.js
oshi97: yeah that's better!

src/core/statelisteners/searchlistener.js
oshi97: I think that makes sense, but I also kind of like how doing it by query trigger categorizes the different search scenarios. Like, "when a filter makes a search, this is what happens," and you can just look at one place and get an idea of what happens for filters, instead of having to check every filter component to see what kind of thing it's doing. I guess this kind of encapsulates the logic? As opposed to having it in lots of different places. Plus we already have this notion of query trigger, so we don't have to add something new like  a QUERY_OPTIONS storage key (though I don't think this would be a big deal). 

Maybe this comes down to what kind of granularity we need for search options. Are we going to be adding new types of "search options" (and we should probably stop adding new ones in either case), and new features that we'll have to add one new query trigger each for? I think... probably not? Even then I kind of like how it makes it easier to reuse the same logic, on the off chance we need to trigger that search in two different places it's a little easier to write. I could have just been looking at this for too long though

cea2aj: I'm definitely a fan of how clean the interface is for components triggering a search. I think where I'm coming from is that it maybe this mapping of keys search options shouldn't be the responsibility of the search listener class? Maybe we would have a better separation of concerns if we were to create a new class ( or a helper function ) whose sole responsibility is to determine the search options based on the QueryTrigger?

However, it doesn't seem terribly out of place here so I don't mind leaving it as is.

cea2aj (approved):

src/core/statelisteners/searchlistener.js
oshi97: Yeah I think we can pull it out to another class!

oshi97: actually I think there's a number of things that can be split out from this class

src/answers-umd.js
```diff
@@ -302,6 +306,10 @@ class Answers {
         throw new Error('MasterSwitchApi determined the front-end should be disabled');
       }
       this._onReady();
+      // initSearchListener happens after the onReady, so that we Wait for addComponent
```
cea2aj: I think this comment needs to be fixed a little bit

cea2aj (approved):
lgtm

tmeyer2115 (approved):

PR #1321, opened 2021-03-11
Add support for doc search and highlighted fields

cea2aj:
Add support for doc search and highlighted fields

Currently the feature/core-extraction branch Result contains a HighlightedFieldMap which contains highlighted fields with the highlighting applied. I renamed this model to `AppliedHighlightedFields` and introduced a new model `HighlightedFields` which is a similar model except it doesn't have highlighting applied. It matches the highlightedFields property from Live API.

J=SLAP-1125
TEST=manual, auto

Point develop of the theme to this PR, and point this PR to https://github.com/yext/answers-core/pull/69. Test document search and see the direct answer appear correctly. Print out the created result object and confirm the model is shaped properly


src/core/utils/objectutils.js
```diff
@@ -0,0 +1,20 @@
...
+ * the function will return the object:
+ *
+ * {
+ *   a: {
+ *     b: 'Hello, world!'
+ *   }
+ * }
+ *
+ * @param {*} value
+ * @param {string[]} keys
```
creotutar, line 14: This fn is really cool. Can we get a @returns

src/core/models/appliedhighlightedfields.js
```diff
@@ -0,0 +1,56 @@
...
+   *
+   * {
+   *   name: '<strong>Yext</strong>',
+   *   description: {
+   *     featured: '<strong>Yext</strong> is the offical answers company'
+   *   }
+   * }
+   */
+  constructor (data = {}) {
+    Object.assign(this, data);
```
creotutar, line 24: Do we not know the field names beforehand? It feels like it's hard to reason what this object looks like without actually having one to log.

cea2aj: That's correct, we don't know the field names beforehand since live API choses which fields to highlight based on the search query

creotutar (approved):
👍 👍 👍 LGTM

PR #1335, opened 2021-03-16
Default boolean facet formatting

cea2aj:
Format boolean facets as 'True' and 'False' by default

J=SLAP-1159
TEST=manual, auto

Open a test site with boolean facets and observe the correct display names for the options. Test that if `transformFacets` modifies the display name of a boolean facet, it takes precedence.

### DIFF
--- a/src/ui/components/filters/facetscomponent.js
+++ b/src/ui/components/filters/facetscomponent.js
@@ -252,10 +252,33 @@ export default class FacetsComponent extends Component {
       updatedFacets = previousFacets;
     }
 
+    updatedFacets = this._transformBooleanFacets(updatedFacets);
+
     return super.setState({
       ...data,
       filters: Facet.fromCore(updatedFacets),
       isNoResults: data.resultsContext === ResultsContext.NO_RESULTS
+    });
+  }
+
+  /**
+   * Applies default formatting to boolean facets
+   *
+   * @param {DisplayableFacet[]} facets from answers-core
+   */
+  _transformBooleanFacets (facets) {
+    return facets.map(facet => {
+      const options = facet.options.map(option => {
+        let displayName = option.displayName;
+        if (option.value === true && displayName === 'true') { 
+          displayName = TranslationFlagger.flag({ phrase: 'True' });
+        }
+        if (option.value === false && displayName === 'false') {
+          displayName = TranslationFlagger.flag({ phrase: 'False' });
+        }
+        return Object.assign({}, option, { displayName });
+      });
+      return Object.assign({}, facet, { options });
     });
   }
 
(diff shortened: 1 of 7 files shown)


src/ui/components/filters/facetscomponent.js
```diff
@@ -252,13 +252,36 @@ export default class FacetsComponent extends Component {
...
       ...data,
       filters: Facet.fromCore(updatedFacets),
       isNoResults: data.resultsContext === ResultsContext.NO_RESULTS
     });
   }
 
+  /**
+   * Applies default formatting to boolean facets
+   *
+   * @param {DisplayableFacet[]} facets from answers-core
```
creotutar, line 261: could we add a `@returns` here?

cea2aj: updated!

src/ui/components/filters/facetscomponent.js
```diff
@@ -252,13 +252,37 @@ export default class FacetsComponent extends Component {
       updatedFacets = previousFacets;
     }
 
+    updatedFacets = this._transformBooleanFacets(updatedFacets);
```
tmeyer2115: This is a minor thing, but it feels a bit weird that we pass all facets through a method called `_transformBooleanFacets`, even if they may not be boolean facets at all. What would you think about having a method called `processFacets` or something like that. The method could then have some logic like:
```
processFacets(facets) {
   // Find facets that are boolean facets, parse only those with _transformBooleanFacets.
   // If any other specific type of facet needs special processing in the future, this would be the
  // entry point for that tool.
}
```

cea2aj: I like that idea for the code organization. The one thing I want to be careful of though is not looping through the facets too many times, so I'm going to see if I can do it in one pass.

src/ui/components/filters/facetscomponent.js
```diff
@@ -240,25 +240,72 @@ export default class FacetsComponent extends Component {
...
+      const firstOption = (facet.options && facet.options[1]) || {};
+      return firstOption['value'] === true || firstOption['value'] === false;
+    };
+
+    facets.forEach(facet => {
+      if (isBooleanFacet(facet)) {
+        booleanFacets.push(facet);
+      } else {
+        processedFacets.push(facet);
+      }
```
oshi97: is it possible to avoid these pushes by changing transformBooleanFacets() to take in a single facet instead of an array, and using a map instead of forEach on line 273?

cea2aj: sure, I like that better

oshi97 (approved):
looks great!

PR #1336, opened 2021-03-17
Apply facetFilters from the url

oshi97:
This commit enables application of facets from the url.
This is done using a new "facetFilters" param that is persisted
in the url. These facets will be used on "non-new" queries,
i.e. queries supplied by the `query` url parameter, the defaultInitialSearch,
or if a search is made using a filter component.

J=SLAP-1089
TEST=manual,auto

test that on page reload, persisted facets are used on a
search on load (tested query parameter, default initial search,
and filtersearch)

test that if there is no search on load, but the user runs a search
using a filter component, the persisted facets are still used

test that if there is no search on load, and the user runs a search
with the searchbar, the facets are reset


cea2aj:
Per the description, why are we using the persisted facets if a user uses a filter component? What do you mean by that?

oshi97:
> Per the description, why are we using the persisted facets if a user uses a filter component? What do you mean by that?

I meant if you load up the page, with no defaultInitialSearch, and then click a static filter. Clicking a static filter will trigger a search. This search will still use the PERSISTED_FACETS. If instead you were to search with the searcbar, you would not use the PERSISTED_FACETS. I'll try to reword/remove it it's pretty unclear

creotutar (approved):
👍 👍 👍 LGTM

PR #1338, opened 2021-03-17
Allow range/daterange filters to set initialMin/initialMax to null

oshi97:
This PR allows somebody to set the initial values of a Range/DateRange filter
to null. Previously, the only way to do this was to set the initialMin/Max to blank string,
which felt unintuitive for a range filter specifically, since for a range filter initialMin/Max
is supposed to be a number.

J=SLAP-1088
TEST=manual

test that I can set the initial values in config to null, and that will
unset them. See that they still work with values persisted in the url.

### DIFF
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -74,8 +74,8 @@ export default class DateRangeFilterComponent extends Component {
      * @private
      */
     this._date = {
-      min: minDate || config.initialMin || todayString,
-      max: maxDate || config.initialMax || todayString
+      min: [minDate, config.initialMin, todayString].find(v => v !== undefined),
+      max: [maxDate, config.initialMax, todayString].find(v => v !== undefined)
     };
   }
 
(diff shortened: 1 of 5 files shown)


src/ui/components/filters/daterangefiltercomponent.js
```diff
@@ -74,8 +74,8 @@ export default class DateRangeFilterComponent extends Component {
      * @private
      */
     this._date = {
-      min: minDate || config.initialMin || todayString,
-      max: maxDate || config.initialMax || todayString
+      min: [minDate, config.initialMin, todayString].find(v => v !== undefined),
```
creotutar, line 77: I know this case is probably very very rare: Do we care that if `minDate` were null, we set the date to null? Whereas before it would go to the next fallback?

oshi97: yes that is intentional!

oshi97: minDate and maxDate will also be going away after this PR, since we're going to change how the min/max are persisted to use the PERSISTED_FILTER storage key

creotutar (approved):
🚐 🚐 🚐 LGTM

PR #1339, opened 2021-03-17
Transform filters with multiple matchers into combined filters

oshi97:
This commit updates FilterRegistry to transform simple filters with multiple matchers
into combined filters. Previously, we were assuming that all simple filters in the SDK
have only one matcher. This isn't true for range filters, which can have 2.

J=SLAP-1088
TEST=manual,auto

test that range filters will now create combined filters, and send them in the request

### DIFF
--- a/src/core/filters/filterregistry.js
+++ b/src/core/filters/filterregistry.js
@@ -123,6 +123,7 @@ export default class FilterRegistry {
   /**
    * Transforms a {@link SimpleFilterNode} to answers-core's {@link Filter} or {@link CombinedFilter}
    * if there are multiple matchers.
+   * TODO(SLAP-1183): remove the parsing for multiple matchers.
    *
    * @param {SimpleFilterNode} filterNode
    * @returns {Filter}
--- a/src/core/filters/filterregistry.js
+++ b/src/core/filters/filterregistry.js
@@ -121,7 +121,8 @@ export default class FilterRegistry {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/core/filters/filterregistry.js
```diff
@@ -121,21 +121,35 @@ export default class FilterRegistry {
...
 
   /**
-   * Transforms a {@link SimpleFilterNode} to answers-core's {@link Filter}
+   * Transforms a {@link SimpleFilterNode} to answers-core's {@link Filter} or {@link CombinedFilter}
+   * if there are multiple matchers.
    *
    * @param {SimpleFilterNode} filterNode
    * @returns {Filter}
    */
   _transformSimpleFilterNode (filterNode) {
```
creotutar, line 131: Is it still considered a SimpleFilterNode if it has multiple matches? I'm reading the docs and it says (below). Multiple matchers seems to not follow the single value part. The reason I bring this up is because I was going to suggest maybe another function based on number of matchers? It almost seems like we should try to recurse a second time in `_transformFilterNodes` to get the CombinedFilter.

```
 * A SimpleFilterNode represents a single, atomic filter.
 * An atomic filter is a filter that filters by a single value on a single field id,
 * and does not contain any children filters.
```

oshi97: Yup still a SimpleFilterNode! I can update the doc

creotutar (approved):
☁️ ☁️ ☁️ LGTM

tmeyer2115 (approved):

PR #1340, opened 2021-03-18
SortOptions on load

oshi97:
This commit adds the ability for SortOptions to be applied
on page load and back navigation, using the new sortBys url param.
Now, the sortBys state is stored in the URL. The SortOptions component
will update it's component state if it sees a persisted sortBys
option that exists in its config.

J=SLAP-1090
TEST=manual, auto

test that I can apply sort options on load, and back nav
see that it works when back/forward naving to either a blank url,
or a url with sortBys in it

### DIFF
--- a/src/core/storage/storagekeys.js
+++ b/src/core/storage/storagekeys.js
@@ -33,7 +33,7 @@ const StorageKeys = {
   SESSIONS_OPT_IN: 'sessions-opt-in',
   VERTICAL_PAGES_CONFIG: 'vertical-pages-config',
   LOCALE: 'locale',
-  SORT_BYS: 'sort-bys',
+  SORT_BYS: 'sortBys',
   NO_RESULTS_CONFIG: 'no-results-config',
   RESULTS_HEADER: 'results-header', // DEPRECATED
   API_CONTEXT: 'context',
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -216,6 +216,7 @@ export default class SortOptionsComponent extends Component {
    * Trigger a search with all filters in storage
    */
   _search () {
+    console.log('trigger sort search')
     this.core.triggerSearch(QueryTriggers.FILTER_COMPONENT);
   }
 
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -19,7 +19,7 @@ export default class SortOptionsComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(assignDefaults(config), systemConfig);
     this.options = this._config.options;
-    this.selectedOptionIndex = this.getPersistedOptionIndex();
+    this.selectedOptionIndex = this.getPersistedSelectedOptionIndex();
     this.options[this.selectedOptionIndex].isSelected = true;
     this.hideExcessOptions = this._config.showMore && this.selectedOptionIndex < this._config.showMoreLimit;
     this.searchOnChangeIsEnabled = this._config.searchOnChange;
@@ -48,7 +48,7 @@ export default class SortOptionsComponent extends Component {
       eventType: 'update',
       storageKey: StorageKeys.HISTORY_POP_STATE,
       callback: () => {
-        const persistedOptionIndex = this.getPersistedOptionIndex();
+        const persistedOptionIndex = this.getPersistedSelectedOptionIndex();
         this._updateSelectedOption(persistedOptionIndex);
         this.setState();
       }
@@ -60,7 +60,7 @@ export default class SortOptionsComponent extends Component {
    *
    * @returns {number|undefined}
    */
-  getPersistedOptionIndex () {
+  getPersistedSelectedOptionIndex () {
     const persistedSortBys = this.core.storage.get(StorageKeys.SORT_BYS) || [];
     const persistedIndex = this._config.options.findIndex(option => {
       return persistedSortBys.find(persistedOption =>
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -216,7 +216,7 @@ export default class SortOptionsComponent extends Component {
    * Trigger a search with all filters in storage
    */
   _search () {
-    console.log('trigger sort search')
+    console.log('trigger sort search');
     this.core.triggerSearch(QueryTriggers.FILTER_COMPONENT);
   }
 
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -216,7 +216,6 @@ export default class SortOptionsComponent extends Component {
    * Trigger a search with all filters in storage
    */
   _search () {
-    console.log('trigger sort search');
     this.core.triggerSearch(QueryTriggers.FILTER_COMPONENT);
   }
 
--- a/src/ui/components/filters/sortoptionscomponent.js
+++ b/src/ui/components/filters/sortoptionscomponent.js
@@ -19,7 +19,7 @@ export default class SortOptionsComponent extends Component {
   constructor (config = {}, systemConfig = {}) {
     super(assignDefaults(config), systemConfig);
     this.options = this._config.options;
-    this.selectedOptionIndex = parseInt(this.core.storage.get(this.name)) || 0;
+    this.selectedOptionIndex = this.getPersistedOptionIndex();
     this.options[this.selectedOptionIndex].isSelected = true;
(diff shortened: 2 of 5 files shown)


src/core/storage/storagekeys.js
```diff
@@ -32,7 +32,7 @@ const StorageKeys = {
   SESSIONS_OPT_IN: 'sessions-opt-in',
   VERTICAL_PAGES_CONFIG: 'vertical-pages-config',
   LOCALE: 'locale',
-  SORT_BYS: 'sort-bys',
+  SORT_BYS: 'sortBys',
```
cea2aj, line 36: for my own edification, what is the reason for this change?

oshi97: In the spec we wanted to have url params that mirror liveApis for setting stuff like sorts/filters on load. The api param is `sortBys` not `sort-bys`, so it seemed like the easiest way to do this was to 
1. rename it to sortBys, so that when it shows up in the urls it's `sortBys` not `sort-bys`, and 
2. do setWithPersists on `SORT_BYS`, instead of having a separate `PERSISTED_SORT_BYS` or something. 

The other PRs had to have new storage keys, because some transformations take place between the data in storage and the data sent to the api, but for `SORT_BYS` what we store in storage is exactly what is sent to the api, so it seemed like we could get away with this instead?

tests/ui/components/filters/sortoptionscomponent.js
```diff
@@ -182,4 +183,60 @@ describe('sort options component', () => {
...
+  it('sets persisted sort option on back navigation', () => {
+    const opts = {
+      ...defaultConfig,
+      options: threeOptions
+    };
+    const component = COMPONENT_MANAGER.create('SortOptions', opts);
+    const wrapper = mount(component);
+    expect(wrapper.find('.yxt-SortOptions-optionSelector').at(0).getDOMNode().checked).toBeTruthy();
+    COMPONENT_MANAGER.core.storage.setWithPersist(StorageKeys.SORT_BYS, [threeOptions[2]]);
+    COMPONENT_MANAGER.core.storage.set(StorageKeys.HISTORY_POP_STATE, {});
```
cea2aj, line 224: Can you expand on how this properly simulates a back nav? Is updating `HISTORY_POP_STATE` sufficient to fully simulate it? Would an acceptance test be a better fit for this? For a back nav test, I would expect the test to check the option on one page, change the page, and then hit the back button and check again

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -43,6 +43,33 @@ export default class SortOptionsComponent extends Component {
...
+      }
+    });
+  }
+
+  /**
+   * Returns the option index matching the persisted sortBys, if one exists.
+   *
+   * @returns {number|undefined}
+   */
+  getPersistedOptionIndex () {
```
cea2aj: To clarify, this function is returning the selected persisted option, correct? If so, what do you think of renaming this to `getSelectedPersistedOptionIndex`?

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -43,6 +43,33 @@ export default class SortOptionsComponent extends Component {
...
+    const persistedSortBys = this.core.storage.get(StorageKeys.SORT_BYS) || [];
```
cea2aj, line 64: Why is persistedSortBys an array? Aren't we just persisting one  (the selected one)?

src/core/storage/storagekeys.js
cea2aj: ah, okay. that makes sense

src/ui/components/filters/sortoptionscomponent.js
oshi97: Technically the option isn't selected yet though. It seems a little weird to call it that when the method has nothing to do with whether the option is selected or not

src/ui/components/filters/sortoptionscomponent.js
oshi97: The api parameter takes in an array, it's just that the SDK is only set up to let you sort by one "sort by" at a time

tests/ui/components/filters/sortoptionscomponent.js
oshi97: Sure, basically the SortOptions component updates itself on backnav by listening for a HISTORY_POP_STATE event. It doesn't do anything else on back nav. SortOptions is a lot simpler than Facets/FilterBox/FilterOptions, and doesn't interact with other components in weird ways, so I thought a regular jest test would be enough here.

src/ui/components/filters/sortoptionscomponent.js
oshi97: The SDK also internally stores an array for SORT_BYS, though it'll only ever have one element, since SortOptions only ever calls core.setSortBys() with one element. (I guess you could directly call core.setSortBys with multiple elements in your config/onReady)

src/ui/components/filters/sortoptionscomponent.js
cea2aj: ah, I see. But we do only persist the selected option, right?

tests/ui/components/filters/sortoptionscomponent.js
```diff
@@ -182,4 +183,60 @@ describe('sort options component', () => {
...
+    wrapper.update();
+    expect(wrapper.find('.yxt-SortOptions-optionSelector').at(3).getDOMNode().checked).toBeTruthy();
```
cea2aj, line 226: Why are we checking at index 3 if there are only three options in the array?

tests/ui/components/filters/sortoptionscomponent.js
```diff
@@ -182,4 +183,60 @@ describe('sort options component', () => {
...
+
+  it('sets persisted sort option on back navigation', () => {
+    const opts = {
+      ...defaultConfig,
+      options: threeOptions
+    };
+    const component = COMPONENT_MANAGER.create('SortOptions', opts);
+    const wrapper = mount(component);
+    expect(wrapper.find('.yxt-SortOptions-optionSelector').at(0).getDOMNode().checked).toBeTruthy();
+    COMPONENT_MANAGER.core.storage.setWithPersist(StorageKeys.SORT_BYS, [threeOptions[2]]);
```
cea2aj, line 223: Should we also be pushing the state to history?

tests/ui/components/filters/sortoptionscomponent.js
cea2aj: okay, that makes sense. I guess I'm still trying to figure out why this test doesn't have three expects: One for the initial page, a second after navigating away to another page, and a third after hitting the back button. And shouldn't the first and third expect have the same option selected?

src/ui/components/filters/sortoptionscomponent.js
oshi97: Ah yeah that's true. What do you think about `getPersistedSelectedOptionIndex` then?

tests/ui/components/filters/sortoptionscomponent.js
oshi97: sort options adds a default option even if you have no options (very weird I know, I think this was part of the original SortOptions spec)

tests/ui/components/filters/sortoptionscomponent.js
oshi97: no, SortOptions doesn't care about the actual history state, just the HISTORY_POP_STATE event. (Also afaik jsdom still doesn't support navigation)

src/ui/components/filters/sortoptionscomponent.js
cea2aj: that works!

tests/ui/components/filters/sortoptionscomponent.js
cea2aj: Actually, I see now. All we really have to do in this test is to ensure that it responds to the history pop state change. What do you think about renaming the test to something like "uses the persisted sorts when the history pop state updates"?

oshi97: This test is just for back navigating from a page. I guess to be a little more accurate, this isn't really for "back navigation" so much as "when HISTORY_POP_STATE is updated, reset the component state using persisted storage". I can add another case that tests going back and forth

src/ui/components/filters/sortoptionscomponent.js
```diff
@@ -190,6 +216,7 @@ export default class SortOptionsComponent extends Component {
    * Trigger a search with all filters in storage
    */
   _search () {
+    console.log('trigger sort search');
```
tmeyer2115: Should this be removed?

oshi97: 🤦  yes, updated!

tmeyer2115 (approved):

PR #1341, opened 2021-03-19
RangeFilters on page load

oshi97:
This commit updates the RangeFilter component to apply filters on page load,
using the new `filters` url param. It will try to look for any relevant
range filters that have been persisted, and if any are found, pick the first
one, apply it, and set the component state using that information.

J=SLAP-1088
TEST=manual,auto

test that range filters will work with backnavigation and page refresh
the filter will be applied to the search, and a removable filter tag will
appear if there is an AppliedFilterss component on the page

### DIFF
--- a/src/core/filters/matcher.js
+++ b/src/core/filters/matcher.js
@@ -1,0 +1,10 @@
+const Matcher = {
+  Equals: '$eq',
+  NotEquals: '!$eq',
+  LessThan: '$lt',
+  LessThanOrEqualTo: '$le',
+  GreaterThan: '$gt',
+  GreaterThanOrEqualTo: '$ge',
+  Near: '$near'
+};
+export default Matcher;
--- a/src/core/filters/matcher.js
+++ b/src/core/filters/matcher.js
@@ -1,3 +1,6 @@
+/**
+ * A Matcher is a filtering operation for {@link Filter}s.
+ */
 const Matcher = {
   Equals: '$eq',
   NotEquals: '!$eq',
--- a/src/core/models/filter.js
+++ b/src/core/models/filter.js
@@ -25,7 +25,7 @@ export default class Filter {
   }
 
   /**
-   * Whether the this filter is a range filter.
+   * Whether this filter is a range filter.
    *
    * @param {Filter} filter
    * @returns {boolean}
--- a/src/core/models/filter.js
+++ b/src/core/models/filter.js
@@ -3,6 +3,13 @@
 import FilterCombinators from '../filters/filtercombinators';
 import Matcher from '../filters/matcher';
 
+const RANGE_MATCHERS = new Set([
+  Matcher.GreaterThan,
+  Matcher.GreaterThanOrEqualTo,
+  Matcher.LessThanOrEqualTo,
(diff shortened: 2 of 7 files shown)


src/core/models/filter.js
```diff
@@ -23,6 +24,27 @@ export default class Filter {
     }
   }
 
+  /**
+   * Whether the this filter is a range filter.
```
creotutar: nit: `the this`

src/core/filters/matcher.js
```diff
@@ -0,0 +1,10 @@
+const Matcher = {
```
creotutar, line 4: Can we add a jsdoc to this enum?

oshi97: yes definitely!

src/core/models/filter.js
```diff
@@ -23,6 +24,27 @@ export default class Filter {
...
+   * @param {Filter} filter
+   * @returns {boolean}
+   */
+  isRangeFilter () {
+    const filterKey = this.getFilterKey();
+    if (!filterKey) {
+      return false;
+    }
+    const matchers = Object.keys(this[filterKey]);
+    const rangeMatchers = new Set([
```
creotutar: do we want to make this every time we call `isRangeFilter`? could we store this as a constant somewhere?

src/core/models/filter.js
```diff
@@ -23,6 +24,27 @@ export default class Filter {
     }
   }
 
+  /**
+   * Whether this filter is a range filter.
+   *
+   * @param {Filter} filter
```
creotutar: nit: param jsdoc without a param

src/ui/tools/filterutils.js
```diff
@@ -19,3 +19,42 @@ export function filterIsPersisted (filter, persistedFilter) {
...
+  }
+  if (Filter.from(persistedFilter).getFilterKey() === fieldId) {
+    return [ persistedFilter ];
+  }
+  return [];
+}
+
+/**
+ * Finds a persisted range filter for the given fieldId, and returns its contents.
+ *
```
creotutar, line 47: nit: can we add jsdocs for the params? possible to have them in the same order as fn above?

tmeyer2115 (approved):

src/core/models/filter.js
oshi97: ah yeah definitely

src/ui/components/filters/rangefiltercomponent.js
```diff
@@ -7,6 +7,9 @@ import FilterNodeFactory from '../../../core/filters/filternodefactory';
 import FilterMetadata from '../../../core/filters/filtermetadata';
 import ComponentTypes from '../../components/componenttypes';
 import TranslationFlagger from '../../i18n/translationflagger';
+import StorageKeys from '../../../core/storage/storagekeys';
```
creotutar, line 11: can we group the imports together have similar paths? e.g. move the core imports up next to FilterMetadata?

creotutar (approved):
🏹 🏹 🏹 LGTM

PR #1342, opened 2021-03-19
DateRangeFilters on load

oshi97:
This commit updates the RangeFilter component to apply filters on page load,
using the new `filters` url param. It will try to look for any relevant
range filters that have been persisted, and if any are found, pick the first
one, apply it, and set the component state using that information.
The main difference between this and RangeFilters on load is that
DateRangeFilters allow you to specify whether to use an inclusive or exclusive range.

J=SLAP-1088
TEST=manual,auto

test that date range filters will work with backnavigation and page refresh
the filter will be applied to the search, and a removable filter tag will
appear if there is an AppliedFilterss component on the page

tested with isExclusive true and false

### DIFF
--- a/src/ui/components/filters/daterangefiltercomponent.js
+++ b/src/ui/components/filters/daterangefiltercomponent.js
@@ -79,6 +79,10 @@ export default class DateRangeFilterComponent extends Component {
     });
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


creotutar (approved):
📅 🏹 📅 🏹 📅 🏹  LGTM

PR #1344, opened 2021-03-22
Allow FilterOptions to be specified with transformFacets

cea2aj:
Allow FilterOptions to be set with transformFacets

Take the transformedFacets and pass that object in as the fieldOverrides
when creating FilterOptions. One caveat is that the DisplayableFacet
model and the FilterOptionsConfig both contain an options field. If an
options field is supplied, we assume the data is in the shape of an array
of DisplayableFacetOption. Another thing to note is that the
`displayName` and the `label` options both do the same thing. If
they both are supplied with transformFacets, the `label` options is
applied because FilterOptions are applied after changes to the core
data models.

J=none
TEST=manual, auto

On a test site, try supplying various filter options using
transformFacets (e.g. enabling searchable), and see them take
effect on the filters which load on the page. Also add a test which
confirms that the filter option 'label' can be specified.

### DIFF
--- a/src/ui/components/filters/facetscomponent.js
+++ b/src/ui/components/filters/facetscomponent.js
@@ -252,10 +252,33 @@ export default class FacetsComponent extends Component {
       updatedFacets = previousFacets;
     }
 
+    updatedFacets = this._transformBooleanFacets(updatedFacets);
+
     return super.setState({
       ...data,
       filters: Facet.fromCore(updatedFacets),
       isNoResults: data.resultsContext === ResultsContext.NO_RESULTS
+    });
+  }
+
+  /**
+   * Applies default formatting to boolean facets
+   *
+   * @param {DisplayableFacet[]} facets from answers-core
+   */
+  _transformBooleanFacets (facets) {
+    return facets.map(facet => {
+      const options = facet.options.map(option => {
+        let displayName = option.displayName;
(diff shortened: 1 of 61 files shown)


src/ui/components/filters/facetscomponent.js
```diff
@@ -251,10 +245,30 @@ export default class FacetsComponent extends Component {
...
+   *
+   * @param {DisplayableFacet | FilterOptionsConfig} transformedFacets a union of the
+   * DisplayableFacet model from ansers-core, and the config options of the FilterOptionsConfig
+   * @returns {Object} config options of the FilterOptionsConfig keyed by filterId
+   */
+  _getFilterOptionsConfigs (transformedFacets) {
+    return transformedFacets.reduce((acc, currentFacet) => {
+      const filterOptions = Object.assign({}, currentFacet);
+      // Don't support options when extracting filterOptions from transformedFacets
+      // because the core library DisplayableFacet isn't compatible with FilterOptionsConfig.options
```
tmeyer2115: What does this mean exactly? Are people unable to configure the options when using `transformFacets`?

cea2aj: I've updated the comment to give some extra context

tmeyer2115 (approved):

PR #1345, opened 2021-03-22
Update dependencies to address vulnerabilities.

tmeyer2115:
Dependabot and BlackDuck suggested that we update the following dependencies:

- @babel/core to v7.13.10.
- @babel/preset-env to v7.13.10.
- @babel/runtime-corejs3 to v7.13.10.
- i18next to v19.9.2.
- css-vars-ponyfill to v2.4.3.
- lodash to v4.17.21.
- gulp-uglify-es to v2.0.0.
- ini to v1.3.8.
- @yext/rtf-converter to v1.3.0.
- rollup-plugin-babl to v4.4.0.
- y18n to v3.2.2.

For each library, I tried as best I could to read through the changelogs and release notes to ensure there were no breaking changes.

TEST=manual

Compared the sizes of the generated bundles before and after the changes. The bundle sizes were comparable (although slightly different). Created a simple experience and verified that it worked in both Chrome and IE11.


package.json
```diff
@@ -23,13 +23,14 @@
...
-    "@yext/rtf-converter": "^1.2.0",
+    "@yext/rtf-converter": "^1.3.0",
     "cross-fetch": "^3.0.6",
-    "css-vars-ponyfill": "^2.3.1",
+    "css-vars-ponyfill": "^2.4.3",
     "gulp-sourcemaps": "^2.6.5",
     "handlebars": "^4.7.2",
     "js-levenshtein": "^1.1.6",
     "kind-of": "^6.0.3",
+    "lodash": "^4.17.21",
```
creotutar: it looks like we are adding `lodash` for the first time here, is that intentional?

package.json
```diff
@@ -66,18 +66,18 @@
     "gulp-replace": "^1.0.0",
     "gulp-rollup-lightweight": "^1.0.10",
     "gulp-sass": "^4.0.2",
-    "gulp-uglify-es": "^1.0.4",
+    "gulp-uglify-es": "^2.0.0",
```
creotutar, line 68: did we want to do major upgrades? i know we said no earlier but if we already tested it, then it LGTM, just making sure

package.json
```diff
@@ -39,14 +40,13 @@
     "template-helpers": "^1.0.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.3.4",
```
creotutar, line 42: it looks like we removed this dependency, is this intentional?

tmeyer2115: This was intentional. I could not find a place it was used. We use `@babel/runtime-corejs3` in our asset creation process, but I couldn't see where `@babel/core` was used. It may be a hold-over from how we used to do poly-fills, but that was removed a while ago.

package.json
tmeyer2115: This is what was recommended by Dependabot. There were no more minors (or patches) on the v1 branch after v1.0.4. I tried my best to make sure the major upgrade didn't bork us. I re-generated assets and ensured that the size was still comparable.

oshi97 (commented):
lgtm!

oshi97 (approved):

PR #1346, opened 2021-03-23
Process facets after transformFacets

cea2aj:
Process facets after transformFacets

Checked with product on this. The purpose of this change is so that
the default formatting for boolean facets (making true and false
uppercase) is applied after transformFacets. This means that the
displayName of boolean facets when passed to transformFacets will
match the API which has true and false lowercase rather than
uppercase

J=none
TEST=manual

Test that default formatting is still applied. Test that in a
transformFacets function, the option displayName for boolean
facets will be lowercase and not uppercase.


src/ui/components/filters/facetscomponent.js
```diff
@@ -272,23 +273,21 @@ export default class FacetsComponent extends Component {
...
-   * Applies default formatting to different types of facets
+   * Applies default formatting to a facet
    *
-   * @param {DisplayableFacet[]} facets from answers-core
-   * @returns {DisplayableFacet[]} from answers-core
+   * @param {DisplayableFacet} facet from answers-core
+   * @returns {DisplayableFacet} from answers-core
    */
-  _processFacets (facets) {
+  _processFacet (facet) {
```
tmeyer2115: I think we may want to give this method a less generic name. It's a bit hard to differentiate between `transformFacets` and `processFacet`. We want a name that also indicates why this should be run after `transformFacets`. What about `applyDefaultFormatting` or something like that?

cea2aj: ya, I like that

tmeyer2115 (approved):

oshi97 (approved):

PR #1347, opened 2021-03-24
Add THIRD-PARTY-NOTICES

cea2aj:
Add a THIRD-PARTY-NOTICES file

- Add an npm command which generates an updated version of the file
- Add a test in circle which verifies that the file is up to date

J=SLAP-1180
TEST=manual

Observe the generated file. Also modify the docs such that they are
out of date, and observe the test fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,6 +116,18 @@ jobs:
           command: |
             sudo apt-get install -qq gettext
             ./.circleci/run_translation_verification.sh
+  # verify that the third party notices file is up to date
+  third_party_notices_test:
+    docker:
+      - image: circleci/node:14.5
+    working_directory: ~/answers
+    steps:
+      - setup-workspace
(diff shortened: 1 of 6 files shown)


tmeyer2115 (commented):
Eventually, do you think we should ensure the THIRD-PARTY-NOTICES is generated automatically (via some Git hook/action or a prePublish hook)? That would allow us to remove this validation layer.

tmeyer2115 (approved):

oshi97 (approved):

cea2aj:
> Eventually, do you think we should ensure the THIRD-PARTY-NOTICES is generated automatically (via some Git hook/action or a prePublish hook)? That would allow us to remove this validation layer.

For the scope of this item, we can stick to this approach. In the future, we could definitely add some automation. Here's a GitHub action which might work: https://github.com/marketplace/actions/add-commit

PR #1354, opened 2021-03-31
Update filterParamsForExperienceLink() with new params

oshi97:
This commit updates filterParamsForExperienceLink() with the
new applied filters from URLs params. A few were missed in the later
stages of the feature. Also removes the older params, which aren't
used anymore.
Also adds a missing case for FilterSearch and the GeolocationFilter, since those components
store both a `query.<componentName>` and a `filter.<componentName>`,
but we were only accounting for `query.<componentName>`

J=SLAP-1209
TEST=manual, auto

checked that navigation tabs don't have the new persisted params
(location radius, filters, facets, sort bys)

### DIFF
--- a/src/core/utils/urlutils.js
+++ b/src/core/utils/urlutils.js
@@ -122,13 +122,14 @@ export function filterParamsForExperienceLink (
     ComponentTypes.GEOLOCATION_FILTER,
     ComponentTypes.FILTER_SEARCH
   ];
-  let paramsToFilter = componentTypesToExclude.flatMap(type => {
-    let params = getComponentNamesForComponentTypes([type]);
-    if (type === ComponentTypes.GEOLOCATION_FILTER || type === ComponentTypes.FILTER_SEARCH) {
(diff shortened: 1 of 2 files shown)


src/core/utils/urlutils.js
```diff
@@ -119,9 +119,6 @@ export function filterParamsForExperienceLink (
   getComponentNamesForComponentTypes
 ) {
   const componentTypesToExclude = [
-    ComponentTypes.RANGE_FILTER,
-    ComponentTypes.DATE_RANGE_FILTER,
-    ComponentTypes.SORT_OPTIONS,
```
creotutar, line 124: Since we're removing everything except for GEOLOCATION_FILTER and FILTER_SEARCH, can we remove the if on L130?

oshi97: ah yeah definitely

creotutar (approved):
☕ ☕ ☕ LGTM

PR #1356, opened 2021-04-01
Use highlightedFields from answers-core

oshi97:
This commit updates the HighlightedFields and AppliedHighlightedFields models
to use the new highlightedFields property on the answers-core Result, which
is just the raw api data.

J=SLAP-1194
TEST=manual, auto

hook up sdk to mt sinai testing repo
https://github.com/yext-pages/mountSinaiJamboTemplateRepo/tree/full-text-highlighting
see in the theme cards that d_highlightedFields is the same as the api response, and
d__highlighted correctly applies strong tags to highlightedFields

updated the highlighted fields jest tests to use .toEqual() instead of .toMatchObject()
because the order and specific structure of highlighted fields matters


src/core/models/appliedhighlightedfields.js
```diff
@@ -9,48 +8,62 @@ export default class AppliedHighlightedFields {
...
   constructor (data = {}) {
     Object.assign(this, data);
   }
 
   /**
-   * Constructs an AppliedHighlightedFields object from an answers-core HighlightedValue array
+   * Constructs an AppliedHighlightedFields object from an answers-core HighlightedField
    *
-   * @param {HighlightedValue[]} highlightedValueArray from answers-core
+   * @param {import('@yext/answers-core').HighlightedFields} highlightedFields from answers-core
```
cea2aj: nit: if we're using the import syntax then we can probably delete the 'from answers-core' part

cea2aj (approved):
lgtm

PR #1357, opened 2021-04-02
v1.7.3

tmeyer2115:
### Bug Fixes
- Interpolation parameters used in run-time translations are now properly sanitized.

### DIFF
--- a/src/ui/rendering/handlebarsrenderer.js
+++ b/src/ui/rendering/handlebarsrenderer.js
@@ -218,8 +218,8 @@ export default class HandlebarsRenderer extends Renderer {
       const language = locale.substring(0, 2);
 
       return isUsingPluralization
-        ? TranslationProcessor.process(pluralizationInfo, interpolationParams, count, language)
-        : TranslationProcessor.process(phrase, interpolationParams);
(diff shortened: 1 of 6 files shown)


src/ui/rendering/handlebarsrenderer.js
```diff
@@ -218,8 +218,8 @@ export default class HandlebarsRenderer extends Renderer {
       const language = locale.substring(0, 2);
 
       return isUsingPluralization
-        ? TranslationProcessor.process(pluralizationInfo, interpolationParams, count, language)
-        : TranslationProcessor.process(phrase, interpolationParams);
+        ? TranslationProcessor.process(pluralizationInfo, interpolationParams, count, language, self.escapeExpression.bind(self))
+        : TranslationProcessor.process(phrase, interpolationParams, null, null, self.escapeExpression.bind(self));
```
oshi97, line 222: would we be able to move escapeExpression to the third parameter?

oshi97: so we don't have to have all of these null placeholders (if I'm too late and this is urgent it's fine)

creotutar (approved):
✌️ ✌️ ✌️ LGTM

PR #1358, opened 2021-04-02
add QueryTriggers.SEARCH_BAR

oshi97:
This commit adds QueryTriggers.SEARCH_BAR,
which the current QUERY_TRIGGER will be set to
when the search bar conducts a search.

J=SLAP-1204
TEST=manual,auto

saw that collapsible-filters would not collapse
on filter changes


cea2aj (approved):
lgtm

PR #1359, opened 2021-04-05
Correct logic for determining if a facet is a Boolean facet or not.

tmeyer2115:
The logic for determining whether or not a facet was a Boolean facet was a bit
off. It looked for the second facet option, instead of the first, and checked
if it's value was a boolean. If a facet had only one boolean option, it would
be incorrectly labeled as a non-Boolean facet. This led to Boolean translations
not appearing intermittently.

J=SLAP-1220
TEST=manual

Used the local test site Bowen reported the issues on. Verified that the
Boolean translations appeared for all languages consistently.


cea2aj (approved):
My bad! LGTM

PR #1363, opened 2021-04-05
fix vertical results rendering during SEARCH_LOADING state

oshi97:
This commit fixes a regression where we were re-mounting the
VerticalResults component during the SEARCH_LOADING state of
vertical results data.

It also removes unneeded functionality from the DOM.removeClass
method which was recently added (as in, does not exist in an officially released
SDK version).

J=SLAP-1217
TEST=manual,auto

check that on universal and vertical, UniversalResults and VerticalResults
still display the search state css classes on their component containers
correctly


cea2aj (approved):
lgtm

PR #1365, opened 2021-04-07
remove flashing from location bias component

oshi97:
This commit adds a searchState property to the
LOCATION_BIAS storage data, and updates the LocationBias
component to only setState
1. the search is not in a loading state and
2. when the data has changed from the previous setState

This improves the performance of the LocationBias component,
and removes unwanted flashing. This flashing was especially
prominent on the vertical-full-page-map in the
answers-hitchhiker-theme, and was making it difficult to
dynamically resize the map.

J=SLAP-1237
TEST=manual

test that, on both universal and vertical, the location bias
component no longer flashes
test that, when starting from a blank page with no visible
LocationBias, running a search, and then hitting the browser
back button, that LocationBias disappears

### DIFF
--- a/src/ui/components/search/locationbiascomponent.js
+++ b/src/ui/components/search/locationbiascomponent.js
@@ -2,6 +2,8 @@ import Component from '../component';
 import StorageKeys from '../../../core/storage/storagekeys';
 import DOM from '../../dom/dom';
 import TranslationFlagger from '../../i18n/translationflagger';
+import isEqual from 'lodash.isequal';
+import SearchStates from '../../../core/storage/searchstates';
 
 const DEFAULT_CONFIG = {
   ipAccuracyHelpText: TranslationFlagger.flag({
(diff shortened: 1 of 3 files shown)


src/ui/components/search/locationbiascomponent.js
```diff
@@ -32,7 +34,16 @@ export default class LocationBiasComponent extends Component {
      * Recieve updates from storage based on this index
      * @type {StorageKey}
      */
-    this.moduleId = StorageKeys.LOCATION_BIAS;
+    this.core.storage.registerListener({
+      storageKey: StorageKeys.LOCATION_BIAS,
+      eventType: 'update',
+      callback: data => {
+        const searchIsLoading = data.searchState === SearchStates.SEARCH_LOADING;
+        if (!searchIsLoading && !isEqual(data, this.getState('locationBias'))) {
```
cea2aj, line 42: Checking that the state is not equal before setting it makes a lot of sense, and I wonder if that could be of use in other components

src/ui/components/search/locationbiascomponent.js
```diff
@@ -32,7 +34,16 @@ export default class LocationBiasComponent extends Component {
      * Recieve updates from storage based on this index
      * @type {StorageKey}
      */
-    this.moduleId = StorageKeys.LOCATION_BIAS;
+    this.core.storage.registerListener({
```
cea2aj, line 37: Are there any possible negative side effects of not setting `moduleId` anymore? I doubt anyone is dependent on it, but i'm just curious.

oshi97: I think the main difference is that when you remove a component, we automatically deregister the moduleId listener for you. We don't do any of that for listeners you register manually. This only really matters for components you would be creating and removing a ton (i.e. child components)

cea2aj (approved):

PR #1367, opened 2021-04-08
use component `uniqueId` for removal

oshi97:
ComponentManager's remove() works by looking for the index of the
component to be removed in _activeComponents using the component's
name. Then, it will remove that index from the _activeComponents
array. If the component's name is not found, then the index will
be set to -1, and the SDK will remove the component found at -1
(the last component in _activeComponents). This should not be happening.
If the component is not found, we should not remove an unrelated component.

Component names are also not unique, and we should use some kind of
unique id system to remove components. Otherwise, when remove() is called
on something like an IconComponent, it's possible that some OTHER IconComponent,
and not the intended one, is removed.

J=SLAP-1231
TEST=manual

test that running repeated searches on no results, which currently
results in the FacetsComponent trying to remove a component that
was already removed, does not remove other unrelated components

### DIFF
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -23,6 +23,7 @@ export default class Component {
 
     /**
      * UUID for the component.
+     * @type {number}
      */
     this.uuid = systemConfig.uuid;
 
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -22,10 +22,10 @@ export default class Component {
     this.moduleId = null;
 
     /**
-     * UUID for the component.
+     * A unique id number for the component.
      * @type {number}
      */
-    this.uuid = systemConfig.uuid;
+    this.uniqueId = systemConfig.uniqueId;
 
     /**
      * Name of this component instance.
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -22,7 +22,12 @@ export default class Component {
     this.moduleId = null;
 
     /**
-     * Unique name of this component instance
+     * UUID for the component.
+     */
+    this.uuid = systemConfig.uuid;
+
+    /**
+     * Name of this component instance
      * Used to distinguish between other components of the same type
      * @type {String}
      */
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -28,8 +28,7 @@ export default class Component {
     this.uuid = systemConfig.uuid;
 
     /**
-     * Name of this component instance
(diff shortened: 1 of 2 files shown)


src/ui/components/component.js
```diff
@@ -22,7 +22,13 @@ export default class Component {
     this.moduleId = null;
 
     /**
-     * Unique name of this component instance
+     * UUID for the component.
+     * @type {number}
```
oshi97, line 26: I think uuid's are normally randomly generated strings but I'm not sure if that's necessary here? javascript's MAX_SAFE_INTEGER  is 9007199254740991 and I don't think we'll ever hit that cap

src/ui/components/componentmanager.js
```diff
@@ -178,16 +185,18 @@ export default class ComponentManager {
...
   }
 
   /**
    * Remove the component with the given name
    * @param {string} name The name of the compnent to remove
    */
   removeByName (name) {
-    const component = this._activeComponents.find(c => c.name === name);
+    const component = Object.values(this._activeComponents).find(c => c.name === name);
     component.remove();
```
oshi97: it's ridiculous but component.remove() ends up calling ComponentManager.remove(this, so we don't need to update this line

src/ui/components/component.js
```diff
@@ -22,7 +22,13 @@ export default class Component {
...
+     */
+    this.uuid = systemConfig.uuid;
+
+    /**
+     * Name of this component instance
      * Used to distinguish between other components of the same type
```
cea2aj: I believe these changes would make this line outdated since name is no longer used to differentiate components of the same type

src/ui/components/componentmanager.js
```diff
@@ -178,16 +185,18 @@ export default class ComponentManager {
...
+    }
   }
 
   /**
    * Remove the component with the given name
    * @param {string} name The name of the compnent to remove
    */
   removeByName (name) {
-    const component = this._activeComponents.find(c => c.name === name);
+    const component = Object.values(this._activeComponents).find(c => c.name === name);
```
cea2aj: Is there a reason for using `Object.values` here?

oshi97: shoot sorry that was from when I was messing with turning this._activeComponents into an object instead of an array

cea2aj (approved):

cea2aj (approved):

PR #1374, opened 2021-04-12
v1.7.4

tmeyer2115:
### Bug Fixes
- Added a unique identifier to each component. This prevents a bug that led to components being unnecessarily removed from the page during a re-render. (#1371)
- Updated the `rtf-converter` dependency to support the new Yext Markdown strike-through syntax. (#1370)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-search-ui",
-  "version": "1.7.3",
+  "version": "1.7.4",
   "description": "Javascript Answers Programming Interface",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-search-ui",
-  "version": "1.7.3",
+  "version": "1.7.4",
```
oshi97, line 3: I think we need to update the version in static/package.json as well! (I forgot this a number of times in the past)

oshi97 (approved):

PR #1387, opened 2021-04-20
Add the entry point for the stand-alone `SearchBar` integration.

tmeyer2115:
This PR adds a new entry point to the SDK's source. This entry point defines
a class, `AnswersSearchBar`, that powers a `SearchBar` only integration. The
class is defined in `answers-search-bar.js`, which bears many similarities to
`answers-umd.js`. There are a few key differences:
- The `FilterNodeFactory`, `formatRichText`, and `processTranslations` methods
are not exported on the top-level object. These are not needed for this type of
integration.
- We will not handle CSS variables in this integration. Styling is very limited
and prescriptive. So, all related ponyfills were removed.
- It is assumed that the client has pulled down the `TemplateBundle` on their
own and we will not need to load it for them. So, the `DefaultTemplateLoader`
is not used.

There may be room for further improvement. I left many of the storage
interactions that are present in `answers-umd.js`, but some of these are likely
irrelevant for a stand-alone `SearchBar`. I will do a closer inspection in a
later PR.

A new registry of `Components`, specific to this type of integration, was
added. This registry was put in a new file, `search-bar-only-registry.js`. At
first, I attempted to export a second registry from the existing `registry.js`.
But, Gulp's tree-shaking was still including all `Component` classes in the
`SearchBar`-only bundle. Creating a new file solved that issue.

Finally, the `ComponentManager` class is no longer a singleton. This is because
it can now accept different `componentRegistry`s. I don't think this should
cause an issue. The `ANSWERS` and `ANSWERS_SEARCH_BAR` instances are already
singletons. So, if you attempt to `init` either twice on your page, you would
already get the same `componentManager`. Likewise if you tried to make multiple
instances of either on your page.

J=SLAP-1271
TEST=manual

Built a new JS and TemplateBundle asset from this entry point. Verified the
following:

- The `SearchBar` appears on the page and works as expected.
- The `Autocomplete` works as expected.
- The `setContext` method could add a context to the search.
- The `redirectUrl` worked as expected and the correct query params were added
  to it.
- The `defaultInitialSearch` worked as expected.
- I could configure the `SearchBar` component.
The `setSessionsOptIn` method worked as expected.

### DIFF
--- a/src/answers-search-bar.js
+++ b/src/answers-search-bar.js
@@ -1,19 +1,14 @@
 /** @module */
 
 import Core from './core/core';
-import cssVars from 'css-vars-ponyfill';
 
 import {
-  DefaultTemplatesLoader,
   Renderers,
-  DOM,
-  SearchParams
+  DOM
 } from './ui/index';
-import Component from './ui/components/component';
 
 import ErrorReporter from './core/errors/errorreporter';
-import ConsoleErrorReporter from './core/errors/consoleerrorreporter';
-import { AnalyticsReporter, NoopAnalyticsReporter } from './core';
+import { AnalyticsReporter } from './core';
 import Storage from './core/storage/storage';
 import { AnswersComponentError } from './core/errors/errors';
 import AnalyticsEvent from './core/analytics/analyticsevent';
@@ -23,15 +18,12 @@ import SearchConfig from './core/models/searchconfig';
 import ComponentManager from './ui/components/componentmanager';
 import VerticalPagesConfig from './core/models/verticalpagesconfig';
 import { SANDBOX, PRODUCTION, LOCALE, QUERY_SOURCE } from './core/constants';
-import RichTextFormatter from './core/utils/richtextformatter';
 import { isValidContext } from './core/utils/apicontext';
-import FilterNodeFactory from './core/filters/filternodefactory';
 import { urlWithoutQueryParamsAndHash } from './core/utils/urlutils';
-import TranslationProcessor from './core/i18n/translationprocessor';
 import Filter from './core/models/filter';
 import SearchComponent from './ui/components/search/searchcomponent';
 import QueryUpdateListener from './core/statelisteners/queryupdatelistener';
-import { COMPONENT_REGISTRY } from './ui/components/registry';
+import { SEARCH_BAR_COMPONENTS_REGISTRY } from './ui/components/search-bar-only-registry';
 
 /** @typedef {import('./core/services/errorreporterservice').default} ErrorReporterService */
 /** @typedef {import('./core/services/analyticsreporterservice').default} AnalyticsReporterService */
@@ -47,30 +39,14 @@ const DEFAULTS = {
 };
 
 /**
- * The main Answers interface
+ * The class that powers the stand-alone SearchBar integration. Note that this class
+ * follows the singleton pattern.
  */
-class Answers {
+class AnswersSearchBar {
   constructor () {
-    if (!Answers.setInstance(this)) {
-      return Answers.getInstance();
+    if (!AnswersSearchBar.setInstance(this)) {
+      return AnswersSearchBar.getInstance();
     }
-
-    /**
-     * A reference to the Component base class for custom
-     * components to extend
-     */
-    this.Component = Component;
-
-    /**
-     * A reference to the AnalyticsEvent base class for reporting
-     * custom analytics
-     */
-    this.AnalyticsEvent = AnalyticsEvent;
-
-    /**
-     * A reference to the FilterNodeFactory class for creating {@link FilterNode}s.
-     */
-    this.FilterNodeFactory = FilterNodeFactory;
 
     /**
      * A reference of the renderer to use for the components
@@ -80,17 +56,10 @@ class Answers {
     this.renderer = new Renderers.Handlebars();
 
     /**
-     * A reference to the formatRichText function.
-     * @type {Function}
-     */
-    this.formatRichText = (markdown, eventOptionsFieldName, targetConfig) =>
-      RichTextFormatter.format(markdown, eventOptionsFieldName, targetConfig);
-
-    /**
      * A local reference to the component manager
      * @type {ComponentManager}
      */
-    this.components = new ComponentManager(COMPONENT_REGISTRY);
+    this.components = new ComponentManager(SEARCH_BAR_COMPONENTS_REGISTRY);
 
     /**
      * A local reference to the core api
@@ -136,13 +105,11 @@ class Answers {
   }
 
   /**
-   * Initializes the SDK with the provided configuration. Note that before onReady
-   * is ever called, a check to the relevant Answers Status page is made.
+   * Initializes the stand-alone SearchBar using the provided configuration.
    *
    * @param {Object} config The Answers configuration.
    */
   init (config) {
-    window.performance.mark('yext.answers.initStart');
     const parsedConfig = this.parseConfig(config);
     this.validateConfig(parsedConfig);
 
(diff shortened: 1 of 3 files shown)


src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+    this._services = null;
+
+    /**
+     * @type {AnalyticsReporterService}
+     * @private
+     */
+    this._analyticsReporterService = null;
+  }
+
+  static setInstance (instance) {
```
cea2aj, line 103: nit: can you add some jsdoc explaining the return value? Specifically that true is returned only if the instance is able to be set

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+   */
+  init (config) {
+    const parsedConfig = this.parseConfig(config);
+    this.validateConfig(parsedConfig);
+
+    parsedConfig.search = new SearchConfig(parsedConfig.search);
+    parsedConfig.verticalPages = new VerticalPagesConfig(parsedConfig.verticalPages);
+
+    const storage = new Storage({
+      updateListener: (data, url) => {
```
oshi97, line 128: would it make sense to git rid of the storage listeners? I don't think we need to do anything when the client's page does back/forward navigation

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+      storage.set(
+        StorageKeys.REFERRER_PAGE_URL,
+        urlWithoutQueryParamsAndHash(document.referrer)
+      );
+    }
+
+    this._services = getServices(parsedConfig, storage);
+
+    this._eligibleForAnalytics = parsedConfig.businessId != null;
+    // TODO(amullings): Initialize with other services
```
cea2aj: nit: this TODO is probably outdated at this point

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+    if (this._eligibleForAnalytics) {
+      this._analyticsReporterService = new AnalyticsReporter(
+        parsedConfig.experienceKey,
+        parsedConfig.experienceVersion,
+        parsedConfig.businessId,
+        parsedConfig.analyticsOptions,
+        parsedConfig.environment);
+
+      // listen to query id updates
+      storage.registerListener({
```
oshi97, line 221: does the redirect searchbar ever get a query id?

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+      experienceVersion: parsedConfig.experienceVersion,
+      locale: parsedConfig.locale,
+      analyticsReporter: this._analyticsReporterService,
+      onVerticalSearch: parsedConfig.onVerticalSearch,
+      onUniversalSearch: parsedConfig.onUniversalSearch,
+      environment: parsedConfig.environment,
+      componentManager: this.components
+    });
+
+    if (parsedConfig.onStateChange && typeof parsedConfig.onStateChange === 'function') {
```
oshi97, line 245: do we need to support onStateChange? I don't think the redirect searchbar ever touches the client's window.history

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+
+    this.renderer.init(parsedConfig.templateBundle, this._getInitLocale());
+    this._onReady();
+    if (!this.components.getActiveComponent(SearchComponent.type)) {
+      this._initQueryUpdateListener(parsedConfig.search);
+    }
+    this._searchOnLoad();
+  }
+
+  _initQueryUpdateListener ({ verticalKey, defaultInitialSearch }) {
```
oshi97, line 269: we technically don't need the QueryUpdateListener because the redirect searchbar never makes search queries

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+    }
+    return this;
+  }
+
+  /**
+   * Conducts a search in the Answers experience
+   *
+   * @param {string} query
+   */
+  search (query) {
```
oshi97, line 378: do we need ANSWERS.search or any of the below methods?

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+      storage)
+  };
+}
+
+/**
+ * Initialize the scroll event listener to send analytics events
+ * when the user scrolls to the bottom. Debounces scroll events so
+ * they are processed after the user stops scrolling
+ */
+function initScrollListener (reporter) {
```
oshi97, line 506: From looking at the code, I thought this would be fired on the client's main page (which seems like incorrect behavior). But then I tried this on a few live client pages and it doesn't look like that's the case. Is there any way we could get rid of this? scroll listeners generally aren't great performance wise

oshi97 (commented):
it looks like there was some intention to keep this new entry relatively similar to answers-umd, but I think it could make the code cleaner long term to remove unneeded portions of answers-umd? Otherwise we could end up with lots of extra noise

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+    return this.instance;
+  }
+
+  /**
+   * Initializes the stand-alone SearchBar using the provided configuration.
+   *
+   * @param {Object} config The Answers configuration.
+   */
+  init (config) {
+    const parsedConfig = this.parseConfig(config);
```
cea2aj, line 121: uh oh. I'm just noticing that Julian's improved error reporting never made it into master: https://github.com/yext/answers-search-ui/pull/1301

tmeyer2115:
> it looks like there was some intention to keep this new entry relatively similar to answers-umd, but I think it could make the code cleaner long term to remove unneeded portions of answers-umd? Otherwise we could end up with lots of extra noise

@oshi97, the intention wasn't to keep this similar to answers-umd.js long term. As I noted in the commit message and as you noted here, I think there are a lot of things (listeners, storage stuff, etc.) that we can probably gut from this new entry point. I wanted to do that as a separate PR. In this first one, I just wanted to get the entry point set up and remove all the low hanging fruit from it (helper methods I knew for sure we wouldn't need).

src/answers-search-bar.js
oshi97: oh... it looks like the feature branch was just never merged in?

oshi97:
> > it looks like there was some intention to keep this new entry relatively similar to answers-umd, but I think it could make the code cleaner long term to remove unneeded portions of answers-umd? Otherwise we could end up with lots of extra noise
> 
> @oshi97, the intention wasn't to keep this similar to answers-umd.js long term. As I noted in the commit message and as you noted here, I think there are a lot of things (listeners, storage stuff, etc.) that we can probably gut from this new entry point. I wanted to do that as a separate PR. In this first one, I just wanted to get the entry point set up and remove all the low hanging fruit from it (helper methods I knew for sure we wouldn't need).

ah sorry that's my bad, pr lgtm then!

oshi97 (approved):

src/answers-search-bar.js
```diff
@@ -0,0 +1,519 @@
...
+  static getInstance () {
+    return this.instance;
+  }
+
+  /**
+   * Initializes the stand-alone SearchBar using the provided configuration.
+   *
+   * @param {Object} config The Answers configuration.
+   */
+  init (config) {
```
cea2aj, line 120: Since so much of this code is the same as answers-umd, I'm a little concerned about maintaining these files in parallel. Ideally we can refactor common pieces of logic into their own classes so that there is less code duplication, however that's out of scope of this PR

src/answers-search-bar.js
tmeyer2115: Great. I will get it merged in separately.

src/answers-search-bar.js
tmeyer2115: In the next PR, I should be stripping out a lot of cruft from the new entry point related to storage and a couple other things. That should really reduce the amount of shared code.

src/answers-search-bar.js
tmeyer2115: Since a search is never actually fired on a redirect searchbar, I don't think there will ever be a query id. So, I will be removing this listener.

src/answers-search-bar.js
tmeyer2115: Removed!

src/answers-search-bar.js
tmeyer2115: I feel like we definitely don't need `search`, I'm meeting with Rose to decide on the other methods.

tmeyer2115: It's unclear to me how conversions/sessions opt-ins would/should work on a redirect search bar.

PR #1389, opened 2021-04-22
Remove unused code from `AnswersSearchBar` entry point.

tmeyer2115:
This PR further optimizes the stand-alone `SearchBar` asset by removing unused
code from its entry point. Specifically, the following was removed:
- Any code related to `VerticalPagesConfig` as there is no navigation or
  no results in this integration.
- The `updateListener` and `resetListener` passed to `Storage`. This
  integration never persists any data in storage. Because of this, if there
  were a pop history event, any URL params would be unrelated to the
  `SearchBar`. This allows us to safely remove the listeners as they will
  never be invoked in this type of integration.
- The `QUERY_ID` storage listener was also removed. This integration doesn't
  actually perform a search, so there will never be a `QUERY_ID`. By the same
  reasoning, references to the `QueryUpdateListener` could also be removed.
- The `scrollListener` was also removed as that is not desired on a
  `SearchBar`-only page.
The other top-level methods, such as `setContext` and `setConversionsOptIn`
were left in, as requested by Product.

J=SLAP-1271
TEST=manual

Ensured the basic functionality of the integration still worked (autocomplete,
clear button, and redirects).

### DIFF
--- a/src/answers-search-bar.js
+++ b/src/answers-search-bar.js
@@ -209,32 +209,6 @@ class AnswersSearchBar {
 
     this.renderer.init(parsedConfig.templateBundle, this._getInitLocale());
     this._onReady();
-    this._searchOnLoad();
-  }
-
-  /**
-   * This guarantees that execution of the SearchBar's search on page load occurs only
-   * AFTER all components have been added to the page. Trying to do this with a regular
-   * onCreate relies on the SearchBar having some sort of async behavior to move the execution
-   * of the search to the end of the call stack. For instance, relying on promptForLocation
-   * being set to true, which adds additional Promises that will delay the exeuction.
-   *
-   * We need to guarantee that the searchOnLoad happens after the onReady, because certain
-   * components will update values in storage in their onMount/onCreate, which are then expected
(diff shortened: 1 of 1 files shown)


src/answers-search-bar.js
```diff
@@ -260,20 +209,9 @@ class AnswersSearchBar {
 
     this.renderer.init(parsedConfig.templateBundle, this._getInitLocale());
     this._onReady();
-    if (!this.components.getActiveComponent(SearchComponent.type)) {
-      this._initQueryUpdateListener(parsedConfig.search);
-    }
     this._searchOnLoad();
```
oshi97: I think we can get rid of searchOnLoad? The SearchComponent's searchAfterAnswersOnReady() guards it from doing anything when there is a redirectUrl.

tmeyer2115: Removed! Thanks for pointing that out.

oshi97 (approved):
lgtm! 🍕🍕🍕

oshi97 (approved):

PR #1395, opened 2021-04-26
Add build chains for `SearchBar`-only assets.

tmeyer2115:
This PR adds the build chains for the modern JS bundle and UMD template bundle
of the `SearchBar` integration. Support for legacy JS bundles and an IIFE
template bundle will be added in the followup epic. Having worked with our
build chain architecture, I think it needs to be cleaned up. The hierarchy of
classes and factories is difficult to follow. I will make a followup item for
this.

J=SLAP-1272
TEST=manual

Ran the acceptance tests to make sure the existing, full SDK assets still were
built properly. Used the `SearchBar`-only assets only a simple site to ensure
they could power that type of integration properly. Generated the search
integration assets in French to make sure they worked correctly on an international
site as well.


conf/gulp-tasks/template/createprecompiletemplates.js
```diff
@@ -28,25 +32,40 @@ function createPrecompileTemplatesTask (locale, translator) {
...
-  const precompiledFileName = getPrecompiledFileName(locale);
+function _precompileTemplates (callback, locale, isSearchBarOnly, translator) {
+  const precompiledFileName = getPrecompiledFileName(locale, isSearchBarOnly);
   const processAST = ast => new TranslateHelperVisitor(translator).accept(ast);
-  return precompileTemplates(callback, precompiledFileName, processAST);
+
+  let templatesWrapper = './conf/templates/handlebarswrapper.txt';
+  let templatesStream = src('./src/ui/templates/**/*.hbs');
+  if (isSearchBarOnly) {
+    templatesWrapper = './conf/templates/handlebarswrapper-search-bar.txt';
```
oshi97: would we be able to modify handlebarswrapper.txt so that it could be shared? the wrapper.txt file is a lodash template passed into gulp-wrap, which I believe we can pass params into

conf/gulp-tasks/template/filenameutils.js
```diff
@@ -36,7 +37,9 @@ exports.getFileName = function (templateType, locale) {
  * @param {string} locale
```
cea2aj, line 37: Can you update the jsdoc to include the new boolean param?

conf/templates/handlebarswrapper-search-bar.txt
```diff
@@ -0,0 +1,31 @@
+import Handlebars from 'handlebars/dist/handlebars.runtime.min.js';
```
oshi97: is this file still used?

conf/gulp-tasks/template/createprecompiletemplates.js
```diff
@@ -28,25 +32,49 @@ function createPrecompileTemplatesTask (locale, translator) {
...
-  return src('./src/ui/templates/**/*.hbs')
+function precompileTemplates (callback, isSearchBarOnly, outputFile, processAST) {
+  let templatesStream = src('./src/ui/templates/**/*.hbs');
+  let handlebarsWrapperData;
+
+  if (isSearchBarOnly) {
+    templatesStream =
+      templatesStream.pipe(filter(['**/search/**/*.hbs', '**/icons/**/*.hbs']));
+    handlebarsWrapperData = {
+      importStatements: "import Handlebars from 'handlebars/dist/handlebars.runtime.min.js';"
```
oshi97, line 66: would it be simpler to pass a boolean into the template and use if statements within the template? I think something like this could work? https://stackoverflow.com/questions/7230470/how-to-use-if-statements-in-underscore-js-templates

conf/templates/handlebarswrapper-search-bar.txt
tmeyer2115: To the best of my knowledge, we still need this, yes. It's the HBS run-time. I believe this is needed to actually render the pre-compiled templates properly at run-time.

conf/gulp-tasks/template/createprecompiletemplates.js
tmeyer2115: I attempted to right a JS conditional in the template that would conditionally import one versus the other. Unfortunately, getting conditional ES6 imports to work is difficult and seems like it would entail a bigger refactor of the whole template. I trying using `require`, which works nicely with conditionals, but we're expecting this file to be ES6 apparently.

conf/templates/handlebarswrapper-search-bar.txt
oshi97: sorry I meant is handlebarswrapper-search-bar.txt still used? it looks like only handlebarswrapper.txt (not handlebarswrapper-search-bar.txt) is being used by gulp-wrap?

conf/gulp-tasks/template/createprecompiletemplates.js
oshi97: I was thinking we could do something like the below in handlebarswrapper.txt? Unless I'm misunderstanding something and this isn't possible (or we just prefer the injecting the strings directly).
```
<% if (data.isSearchBarOnly) { %>
  import Handlebars from 'handlebars/dist/handlebars.runtime.min.js';
<% } else { %>
  import Handlebars from 'handlebars/dist/handlebars.min.js;
  import templateHelpers from 'template-helpers';
<% } %>
```

conf/templates/handlebarswrapper-search-bar.txt
tmeyer2115: Ahhh, gotcha gotcha. Nope, this can be deleted!

oshi97 (approved):

PR #1397, opened 2021-04-30
Allow progressive rendering of child components

cea2aj:
Optimize main thread performance through progressive rendering and by avoiding unnecessary deep clones

By splitting up the rendering of child components with `setTimeout`, it allows the main thread to handle user input, which increases our minimum FPS if the rendering of the children takes a long time. This is is most useful for the mounting of the result cards which take a long time to render.

I added a .yxt-Answers-component--unmounted class to the vertical results which is removed on mount. This allows us to hide the component div until it is rendered which reduces layout shifts.

Finally, I updated the transformData logic so that it only runs when necessary.

J=SLAP-1240
TEST=manual

Enable progressive rendering on the vertical-full-page-map and observe the cards get rendered one-by-one. Also confirm that transform data still works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,8 +138,7 @@
       "test",
       "fixture",
       "CustomEvent",
-      "ANSWERS",
-      "requestAnimationFrame"
+      "ANSWERS"
     ],
     "ignore": [
       "docs/"
--- a/package.json
+++ b/package.json
@@ -138,7 +138,8 @@
       "test",
       "fixture",
       "CustomEvent",
-      "ANSWERS"
+      "ANSWERS",
+      "requestAnimationFrame"
     ],
     "ignore": [
       "docs/"
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -90,6 +90,12 @@ export default class Component {
     this._analyticsOptions = config.analyticsOptions || {};
 
     /**
+     * Allows the main thread to regain control while rendering child components
+     * @type {boolean}
+     */
+    this._progressivelyRenderChildren = config.progressivelyRenderChildren;
+
+    /**
      * A reference to the DOM node that the component will be appended to when mounted/rendered.
      * @type {HTMLElement}
      */
@@ -158,7 +164,7 @@ export default class Component {
      * By default, no transformation happens.
      * @type {function}
      */
-    this.transformData = config.transformData || this.transformData || function () {};
+    this.transformData = config.transformData;
 
     /**
      * The a local reference to the callback that will be invoked when a component is created.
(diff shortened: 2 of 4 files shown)


src/ui/sass/_base.scss
```diff
@@ -11,4 +11,8 @@
 
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+}
+
+.yxt-Answers-component--unmounted {
+  display: none !important;
```
oshi97, line 17: is there some way we could remove this `display: none` by not rendering the parent component until all of its children are rendered? I believe right now we render the parent container before all of the children are done rendering, though this might slow down the overall speed of non locator pages?

src/ui/components/component.js
```diff
@@ -367,12 +369,22 @@ export default class Component {
...
     domComponents.forEach(c => this._createSubcomponent(c, data));
 
-    this._children.forEach(child => {
-      child.mount();
-    });
+    if (this._progressivelyRenderChildren) {
+      this._children.forEach(child => {
+        setTimeout(() => {
+          child.mount();
+        }, 0);
```
oshi97: I think the default is 0, but don't mind specifying it! I feel a little split whether setTimeout defaulting to 0 is something people should just know by heart, which probably means that it's not

package.json
```diff
@@ -138,7 +138,8 @@
       "test",
       "fixture",
       "CustomEvent",
-      "ANSWERS"
+      "ANSWERS",
+      "requestAnimationFrame"
```
oshi97: do we still use requestAnimationFrame in this PR?

src/ui/sass/_base.scss
oshi97: We might need to clone the container node, attach children to the clone, and then replace the original when rendering is done?

cea2aj: I'm not sure that is possible without a larger refactor. Currently handlebars puts the divs for the vertical result cards on the DOM, so we would need to avoid that. That would cause an issue because subcomponents are built by querying the DOM and reading the data on the child divs.

src/ui/components/component.js
cea2aj: ya, this could go either way

package.json
cea2aj: ah, good catch. I ended up removing it from this PR

src/ui/sass/_base.scss
oshi97: that's fair, touching addChild() would be pretty annoying

oshi97: I think this is fine since it's an experimental feature

oshi97 (approved):

PR #1403, opened 2021-05-05
Add IE-11 compliant versions of the `SearchBar` assets.

tmeyer2115:
This PR adds IIFE and UMD bundles of the `SearchBar` that support IE11. As part
of this, I had to add CSS Variable Ponyfills back to the integration's entry
point. Technically, we could create styling for this integration that does not
use CSS variables. But, that would be rather onerous to create and maintain.

J=SLAP-1253
TEST=manual

Created a `SearchBar` site using the IIFE and UMD assets. Verified that the site
looked and worked as expected in IE11.

### DIFF
--- a/conf/gulp-tasks/bundle/bundletaskfactory.js
+++ b/conf/gulp-tasks/bundle/bundletaskfactory.js
@@ -98,7 +98,7 @@ class BundleTaskFactory {
   }
 
   /**
-   * The Gulp task for producing the legacy, IIFE-style SDK bundle.
+   * The Gulp task for producing a legacy, IIFE-style bundle of an SDK asset.
    *
    * @param {function} callback function that will run after the Gulp task
    * @param {BundleType} bundleType The type of SDK asset to generate the bundle for.
@@ -131,7 +131,7 @@ class BundleTaskFactory {
   }
 
   /**
-   * The Gulp task for producing the legacy, UMD-style SDK bundle.
+   * The Gulp task for producing the legacy, UMD-style bundle of an SDK asset.
    *
    * @param {function} callback function that will run after the Gulp task
    * @param {BundleType} bundleType The type of SDK asset to generate the bundle for.
--- a/conf/gulp-tasks/bundle/bundletaskfactory.js
+++ b/conf/gulp-tasks/bundle/bundletaskfactory.js
@@ -37,25 +37,21 @@ class BundleTaskFactory {
    */
   create (bundleType) {
     let bundleFunction;
-    switch (bundleType) {
-      case BundleType.SEARCH_BAR_MODERN:
-        // fall through
-      case BundleType.MODERN:
-        bundleFunction = (callback) => this._modernBundle(callback, bundleType);
-        break;
-      case BundleType.SEARCH_BAR_LEGACY_IIFE:
-        // fall through
-      case BundleType.LEGACY_IIFE:
-        bundleFunction = (callback) => this._legacyBundleIIFE(callback, bundleType);
-        break;
-      case BundleType.SEARCH_BAR_LEGACY_UMD:
-        // fall through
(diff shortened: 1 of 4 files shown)


conf/gulp-tasks/bundle/bundletaskfactory.js
```diff
@@ -41,11 +43,15 @@ class BundleTaskFactory {
       case BundleType.MODERN:
         bundleFunction = (callback) => this._modernBundle(callback, bundleType);
         break;
+      case BundleType.SEARCH_BAR_LEGACY_IIFE:
```
cea2aj: nit: What are your thoughts on changing from a switch statement to a multiple if statements? I feel like that would be simpler

tmeyer2115: Were you thinking something like this: 
```
    if (bundleType === bundleType.MODERN || bundleType === bundleType.SEARCH_BAR_MODERN) {
      bundleFunction = (callback) => this._modernBundle(callback, bundleType);
    } else if (bundleType === BundleType.LEGACY_IIFE || bundleType === bundleType.SEARCH_BAR_LEGACY_IIFE) {
      bundleFunction = (callback) => this._legacyBundleIIFE(callback, bundleType);
    } else if (bundleType === BundleType.LEGACY_UMD || bundleType === BundleType.SEARCH_BAR_LEGACY_UMD) {
      bundleFunction = (callback) => this._legacyBundleUMD(callback, bundleType);
    } else {
      throw new Error('Unrecognized BundleType');
    }
```

I have a slight preference for the `switch` statement over the above. While the `if-elseif-else` is more succinct in terms of lines used, I think it's easier to read the `swtich`.

src/answers-search-bar.js
```diff
@@ -1,10 +1,12 @@
 /** @module */
 
 import Core from './core/core';
+import cssVars from 'css-vars-ponyfill';
```
oshi97, line 4: would it be worth exploring a way to only load cssVars in the legacy bundles?

tmeyer2115: We could make another entry point for the legacy search bar that has this import and the ponyfill support. It's difficult to guard `import` statements with conditionals, so that may be the best option. FWIW, the inclusion of this didn't seem to impact the performance of a page using the modern bundle. I was still seeing a score of 100. The size of the minified modern bundle wasn't impacted much by this addition.

oshi97: sure I'm fine leaving it out then, this is already a significant improvement to what we had before

oshi97 (approved):

conf/gulp-tasks/bundle/bundletaskfactory.js
cea2aj: I may just not be as used to reading switch statements, but up to you. 

I was thinking something like this:

```
const modernBundles = [BundleType.MODERN, BundleType.SEARCH_BAR_MODERN];
const iifeBundles = [BundleType.LEGACY_IIFE, BundleType.SEARCH_BAR_LEGACY_IIFE];
const legacyBundles = [BundleType.LEGACY_UMD, BundleType.SEARCH_BAR_LEGACY_UMD];

if (modernBundles.includes(bundleType)) {
  bundleFunction = (callback) => this._modernBundle(callback, bundleType);
} else if (iifeBundles.includes(bundleType)) {
  bundleFunction = (callback) => this._legacyBundleIIFE(callback, bundleType);
} else if (legacyBundles.includes(bundleType)) {
  bundleFunction = (callback) => this._legacyBundleUMD(callback, bundleType);
} else {
  throw new Error('Unrecognized BundleType');
}
```

tmeyer2115: Done!

PR #1404, opened 2021-05-06
Optimize job order in CircleCI workflows.

tmeyer2115:
This PR ensures that in our workflows, the 3rd party notices and translation test jobs are run before the build job. This ensures that if these tests fail, we don't go through the remainder of the workflow's jobs unnecessarily.

J=SLAP-1256
TEST=manual

For this PR, confirmed only the `build_and_deploy` workflow was run. The jobs in the workflow were executed in the expected order.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,12 +221,7 @@ workflows:
             - headless_acceptance_test
   build_and_deploy_i18n:
     jobs:
-      - translation_test
-      - third_party_notices_test
-      - build_i18n:
-          requires:
-            - translation_test
-            - third_party_notices_test
+      - translation_test:
           filters:
             branches:
               only:
@@ -234,6 +229,18 @@ workflows:
                 - /^hotfix\/.*/
                 - /^feature\/.*-i18n/
                 - /^release\/.*/
+      - third_party_notices_test:
+          filters:
+            branches:
+              only:
+                - /^support\/.*/
+                - /^hotfix\/.*/
+                - /^feature\/.*-i18n/
+                - /^release\/.*/
+      - build_i18n:
+          requires:
+            - translation_test
+            - third_party_notices_test
       - unit_test:
           requires:
             - build_i18n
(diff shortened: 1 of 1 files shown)


cea2aj:
Can we have the build run in parallel to the translation and the licenses test? I think that would speed thing up since the build is not dependent on the other two

tmeyer2115:
> Can we have the build run in parallel to the translation and the licenses test? I think that would speed thing up since the build is not dependent on the other two

My thinking was that we wouldn't run the build unless these two tests passed. We don't have artifact caching right now. Suppose one of these two tests fails. The author puts up another PR to fix the translations or notice. A new run of the workflow is kicked off. This run will re-run the build since the previous artifacts weren't stored. So, the artifacts generated during that first, failed run went to waste and the workflow went on longer than necessary. 

Ultimately, I think we need to figure out a way to cache the generated artifacts, not just the node modules. Once we do that, we should definitely get all three jobs to run in parallel.

cea2aj:
> My thinking was that we wouldn't run the build unless these two tests passed. We don't have artifact caching right now. Suppose one of these two tests fails. The author puts up another PR to fix the translations or notice. A new run of the workflow is kicked off. This run will re-run the build since the previous artifacts weren't stored. So, the artifacts generated during that first, failed run went to waste and the workflow went on longer than necessary.
> 
> Ultimately, I think we need to figure out a way to cache the generated artifacts, not just the node modules. Once we do that, we should definitely get all three jobs to run in parallel.

Is there any issue with a failed run going to waste? Does it slow down subsequent runs?

tmeyer2115:
> > My thinking was that we wouldn't run the build unless these two tests passed. We don't have artifact caching right now. Suppose one of these two tests fails. The author puts up another PR to fix the translations or notice. A new run of the workflow is kicked off. This run will re-run the build since the previous artifacts weren't stored. So, the artifacts generated during that first, failed run went to waste and the workflow went on longer than necessary.
> > Ultimately, I think we need to figure out a way to cache the generated artifacts, not just the node modules. Once we do that, we should definitely get all three jobs to run in parallel.
> 
> Is there any issue with a failed run going to waste? Does it slow down subsequent runs?

Yeah, I think the only downside would be that we have a parallelism limit. So, by having the failed run go on longer than necessary, it might create a backlog for other PR's builds. Of course, my argument is assuming that the build job takes longer to run than the other two jobs. If the three jobs all run in about the same time, it's not really an issue. My sense though was that the build job takes considerably longer.

cea2aj:
> Yeah, I think the only downside would be that we have a parallelism limit. So, by having the failed run go on longer than necessary, it might create a backlog for other PR's builds. 

Makes sense!

cea2aj (approved):

PR #1409, opened 2021-05-07
add Coveralls github workflow

oshi97:
Adds a code coverage workflow using Coveralls
Also adds a code coverage badge to the readme.

J=SLAP-873
TEST=auto

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Answers Search UI
 
-<div>
+<h3>
   <a href='https://coveralls.io/github/yext/answers-search-ui?branch=master'>
     <img src='https://coveralls.io/repos/github/yext/answers-search-ui/badge.svg?branch=master' alt='Coverage Status' />
   </a>
-</div>
+</h3>
 
 1. [Install / Setup](#install-and-setup)
 2. [ANSWERS.init Configuration Options](#answersinit-configuration-options)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,5 +1,13 @@
 # Answers Search UI
+
+<div>
+  <a href='https://coveralls.io/github/yext/answers-search-ui?branch=master'>
+    <img src='https://coveralls.io/repos/github/yext/answers-search-ui/badge.svg?branch=master' alt='Coverage Status' />
+  </a>
+</div>
+
 Outline:
```
cea2aj: nit: can we make this a heading element? Maybe a h3? I think it currently looks a little out of place once the badge is added

oshi97: updated! I also removed the "Outline:" line because the TOC seemed pretty self explanatory

cea2aj: I was actually suggesting to make 'Outline' an h3, however I like your idea better of removing it. With 'Outline' removed, I don't think there is any reason to put the badge in an h3, is there?

oshi97: ~~I don't mind the spacing either way, updated~~ I lied, because the SDk has a list under the badge, instead of text like our other repos, the spacing is a little too tight if it's just a div

cea2aj (approved):

PR #1416, opened 2021-05-24
Direct Answers: Fix IE11 console error

yen-tt:
resolved issue of getting property of undefined/null references in
direct answer component by using Element.getAttribute for SVG elements
in place of className

J=SLAP-556
T=manual

Launched test site in IE11 and confirmed that error
is gone when click on feedback button

### DIFF
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -188,8 +188,8 @@ export default class AutoCompleteComponent extends Component {
   }
 
   isQueryInputFocused () {
-    return document.activeElement &&
-      document.activeElement.className.includes(this._inputEl.substring(1));
+    return document.activeElement && document.activeElement.getAttribute('class') &&
+      document.activeElement.getAttribute('class').includes(this._inputEl.substring(1));
   }
 
   /**


src/ui/components/search/autocompletecomponent.js
```diff
@@ -188,8 +188,8 @@ export default class AutoCompleteComponent extends Component {
   }
 
   isQueryInputFocused () {
-    return document.activeElement &&
-      document.activeElement.className.includes(this._inputEl.substring(1));
+    return document.activeElement && document.activeElement.getAttribute('class') &&
```
cea2aj, line 191: Looks good, Yen! I do have one suggestion which is to write this using optional chaining. Since we run this code through babel, the optional chaining should work on IE11. That version would look like this:  `return document.activeElement?.getAttribute('class')?.includes(this._inputEl.substring(1));`

yen-tt: Since current semistandard version doesn't recognize optional chaining, this change would require updating to v15 and fixing syntax for other files. I made a JIRA ticket to suggest this update in the future!

cea2aj (approved):
Looks good to me!

src/ui/components/search/autocompletecomponent.js
oshi97, line 191: just for my edification how does this relate to direct answers? does this end up getting called somehow when you click the thumbs up/down button?

yen-tt: Yes! This function is used in setState() for direct answers, and get invoke when the feedback button is click.

oshi97 (approved):

PR #1417, opened 2021-05-25
Transition node-sass to dart-sass

yen-tt:
Update package.json to use sass package and remove deprecated node-sass
Set gulp-sass compiler to use dart-sass

J=SLAP-1310
TEST=compile


oshi97 (approved):
lgtm!

PR #1418, opened 2021-05-26
Fix appearance of clear button on empty query in IE11

yen-tt:
update logic for adding/removing classname that hide the clear button since toggle() with second parameter is not supported in IE11

J=SLAP-1143
TEST=manual

Launched test site and confirmed visual changes of clear button on empty and nonempty query in search bar component in IE11, firefox, and chrome.


oshi97 (approved):

oshi97:
This looks great, glad we could get rid of a random todo!

PR #1426, opened 2021-06-04
addComponent no longer throws error, prints to console instead

cjiang2000:
If addComponent fails, we no longer throw an error, prints to console instead

TEST=manual
J=SLAP-1106

Threw an error in the component, and everything else loaded fine

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -15,7 +15,6 @@ import ErrorReporter from './core/errors/errorreporter';
 import ConsoleErrorReporter from './core/errors/consoleerrorreporter';
 import { AnalyticsReporter, NoopAnalyticsReporter } from './core';
 import Storage from './core/storage/storage';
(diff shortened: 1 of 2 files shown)


src/answers-umd.js
```diff
@@ -433,7 +432,7 @@ class Answers {
     try {
       this.components.create(type, opts).mount();
     } catch (e) {
-      throw new AnswersComponentError('Failed to add component', type, e);
+      console.error('Failed to add component', type, e);
```
cea2aj: nit: I'm not sure what the output looks like, but maybe we should put the error message on a separate line?

cjiang2000: We added two new lines so its prettier now.

oshi97 (approved):
nice!!!! 🍍 🍍 🍍 🍍 🍍

PR #1427, opened 2021-06-07
automate npm publish

yen-tt:
GH workflow action which, on successful test and build steps on master branch, will publish to npm registry

J=SLAP-1353
TEST=manual

created mock repo and npm package, made a new tagged release and verified it automatically publish to npm test package.

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -1,0 +1,48 @@
+name: NPM publish on new release
+
+on: 
+  release:
+    types: [created]
+    branches: 
+      - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [15.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run test
+    - run: npm run build
+  
+  npm-publish:
+    
+    needs: build
+    
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [15.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+        registry-url: https://registry.npmjs.org/
+    - run: npm ci
+    - run: npm publish
+      env:
+        NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}


.github/workflows/npm_publish.yml
```diff
@@ -0,0 +1,48 @@
...
+        node-version: [15.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run test
```
tmeyer2115: Do we need to include the test here? If the tests fail will the publish fail? There shouldn't be code in master that fails our tests.

yen-tt: ah I added it in as a safety check, but if we already verify that before pushing to master then I guess this step is not necessary.

.github/workflows/npm_publish.yml
```diff
@@ -0,0 +1,47 @@
+name: NPM publish on new release
+
+on: 
+  release:
+    types: [created]
+    branches: 
+      - master
+
+jobs:
+  build:
```
cea2aj: Is there a reason for having a job for the build step rather than putting `npm run build` after `npm ci` in the npm-publish job?

yen-tt: hmm yeah you are right, it can just be under one job.

.github/workflows/npm_publish.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v2
+      with:
+        node-version: ${{ matrix.node-version }}
+        registry-url: https://registry.npmjs.org/
+    - run: npm ci
+    - run: npm run build
```
tmeyer2115, line 26: Just for my edification, we're not including the contents of `src` or `test` in what we publish to NPM, right? We're essentially just publishing the `dist` directory and some manifest files (`package.json`, etc.)

cea2aj: Currently we include everything except the things specified in .npmignore. In answers-core, we instead use the 'files' param in package.json to specify which files and folders to include. I think we should switch answers-search-ui to do that as well. I think the `dist` folder is the only one we need

yen-tt: I made the change in package.json to include dist/ folder only (npm will also automatically include package.json, license, and readme file when publish)

tmeyer2115: With the `files` addition to the `package.json` do we still need `.npmignore`?

yen-tt: Good point. Would need it if we want to exclude something specific from `files`, but since node_modules is automatically ignored and DS_Store is never going to be push to repo, I will remove `.npmignore`

tmeyer2115 (approved):

PR #1428, opened 2021-06-07
Change semistandard to eslint

cjiang2000:
Use ESLint with the same configuration as Semistandard version 16.0.0
Updated codebase to match new version

Fixed the following rules(automatic):
- Using const when possible
- Strings using single quotes
- Spacing tweaks
- Using dot accessors instead of bracket accessors
- Not using quotes for object properties when possible

Fixed the following rules(manual):
- Using Regex literals instead of RegExp
- Added default values for latter parameters
- Not using hasOwnProperty()
- Using forEach instead of map
- Added brackets around case declarations
- Using const when possible

J = SLAP-1343
TEST = Manual

Built a page in Japanese and opened it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,17 +2,17 @@
    "extends": ["semistandard"],
    "ignorePatterns": ["dist", "docs"],
    "globals": {
-      "beforeEach" : true,
-      "beforeAll" : true,
-      "describe" : true,
-      "it" : true,
-      "expect" : true,
-      "jest" : true,
-      "DOMParser" : true,
-      "ytag" : true,
-      "test" : true,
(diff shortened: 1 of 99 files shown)


cea2aj:
@oshi97 I know we have brought up just using eslint for all our repos. Does switching to that make more sense here? Otherwise I'm happy with this change

oshi97:
> @oshi97 I know we have brought up just using eslint for all our repos. Does switching to that make more sense here? Otherwise I'm happy with this change

yeah I think that makes sense, semistandard is just eslint without configurability so it shouldn't be hard to eject out of it

cea2aj (approved):
lgtm

PR #1430, opened 2021-06-09
v1.8.4

tmeyer2115:
## Fixes
- Fixed a bug that prevented Facets in the URL from being applied on page load. (#1429)


oshi97 (approved):
lgtm!

PR #1433, opened 2021-06-14
Codelabs Core & Unit Tests

cjiang2000:
Test: Manual

### DIFF
--- a/src/core/search/searchapi.js
+++ b/src/core/search/searchapi.js
@@ -192,7 +192,7 @@ export default class SearchApi {
   }
 
   /**
-   * Initiate a search for Connor
+   * Initiate a search for Connor.
    */
   connorSearch () {
     const query = 'Connor';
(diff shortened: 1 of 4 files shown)


src/core/search/searchapi.js
```diff
@@ -242,4 +242,36 @@ export default class SearchApi {
...
       .then(response => response.json());
   }
+
+  /**
+ * Initiate a search
+ * @param {boolean} includeLastName include last name in the search
+ * @param {boolean} useUserName use username instead of real name
+ * @param {string} additionalSearchText more text to add to the search
+ */
+Search(includeLastName, useUserName, additionalSearchText) {
```
cea2aj: Based on what I recall from the codelab, I think the function name is supposed to have your name in it. Something like 'chrisSearch'. Also for function names, we use camelCase so we don't capitalize the first letter

src/core/search/searchapi.js
```diff
@@ -242,4 +242,36 @@ export default class SearchApi {
     return request.get()
       .then(response => response.json());
   }
+
+  /**
+ * Initiate a search
```
cea2aj: I think your indentation is a little off compared to the other functions

src/core/search/searchapi.js
cjiang2000: Wow, good call Connor. I'll be sure to make the changes.

src/core/search/searchapi.js
cjiang2000: I will make sure to stick to semistandard regulations!

src/core/search/searchapi.js
cjiang2000: The fixed have been implemented. Thoughts?

oshi97 (commented):
wow, looks great! +1 from me!!

src/core/search/searchapi.js
cea2aj: looks good!

cea2aj (approved):

PR #1435, opened 2021-06-15
analyticsEventsEnabled flag added

cjiang2000:
Added analyticEventsEnabled flag to analytics reporter. Also added setAnalyticsOptIn(analyticEventsEnabled), a function that modifies analyticsEventsEnabled.

J=SLAP-1392
TEST=manual/auto

Checked website to see if events were reported with flag set to true and false. Also checked to make sure the ANSWERS.setAnalyticsOptIn(analyticEventsEnabled) works as expected.
Created unit tests to see if events were reported.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -474,10 +474,10 @@ class Answers {
 
   /**
    * Opt in or out of analytic events
-   * @param {boolean} bool
+   * @param {boolean} analyticsEventsEnabled
    */
-  setAnalyticsOptIn (bool) {
-    this._analyticsReporterService.setAnalyticsOptIn(bool);
+  setAnalyticsOptIn (analyticsEventsEnabled) {
+    this._analyticsReporterService.setAnalyticsOptIn(analyticsEventsEnabled);
   }
 
   /**
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -41,7 +41,8 @@ import QueryUpdateListener from './core/statelisteners/queryupdatelistener';
 
 const DEFAULTS = {
   locale: LOCALE,
-  querySource: QUERY_SOURCE
+  querySource: QUERY_SOURCE,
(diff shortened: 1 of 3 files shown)


src/core/analytics/analyticsreporter.js
```diff
@@ -26,6 +27,12 @@ export default class AnalyticsReporter {
      */
     this._businessId = businessId;
 
+    /**
+     * Boolean indicating if the Answers Search UI submits analytic click events
+     * @type {boolean}
+     */
+
```
oshi97: nit extra line break!

src/answers-umd.js
```diff
@@ -223,10 +224,8 @@ class Answers {
    * is ever called, a check to the relevant Answers Status page is made.
    *
    * @param {Object} config The Answers configuration.
-   * @param {Object} statusPage An override for the baseUrl and endpoint of the
-   *                            experience's Answers Status page.
    */
-  init (config, statusPage) {
```
tmeyer2115, line 229: Good catch! This should have been removed a while ago.

src/core/analytics/analyticsreporter.js
```diff
@@ -67,8 +74,19 @@ export default class AnalyticsReporter {
     this._globalOptions.queryId = queryId;
   }
 
+  /**
+   * Opt in or out of analytics click events
+   * @param {boolean} bool
+   */
+  setAnalyticsOptIn (bool) {
```
tmeyer2115: According to the spec, we'll need an `ANSWERS.setAnalyticsOptIn` method as well. This means we will need to make a `setAnalyticsOptIn` method available in the `answers-umd` class. It can be a simple pass-through to this.

oshi97: instead of `bool` lets call the param `analyticsEventsEnabled`

oshi97 (approved):

PR #1437, opened 2021-06-17
Update core version to support rich text featured snippets

cea2aj:
Point to core v1.2.0-alpha.2 to support rich text featured snippets

When we release v1.2, we will update to use that one instead

J=SLAP-1385
TEST=manual

Use this version of the SDK with the theme and see rich text featured snippets come through

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,12 @@
     "cssnano": "^4.1.10",
     "del": "^5.1.0",
     "enzyme": "^3.11.0",
+    "eslint": "^7.28.0",
+    "eslint-config-semistandard": "^15.0.1",
+    "eslint-config-standard": "^16.0.3",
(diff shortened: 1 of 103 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   ],
```
oshi97, line 24: would it be possible to release the alpha first so we can run acceptance tests and what not

cea2aj: I did release the version already. I'm re-running all jobs to see if that fixes it. If not, I'll dig deeper, although the unit tests are passing locally for me

cea2aj: I ran a fresh build for 1.2.0-alpha.2 and that one is working!

oshi97 (approved):

PR #1438, opened 2021-06-17
Update SDK to allow DA on vertical search

yen-tt:
add direct answers to vertical search in sdk core data transformer layer
Note: this sdk is using core version 1.3.0-alpha to get vertical search responses with direct answer

J=SLAP-1389
TEST=auto and manual

Update and run jest tests to see that results are formatted properly.
Verify direct answer is fetched properly from answers-core with console log.
Launched theme test site using local sdk bundle, performed a vertical search on people's page and can verify that direct answer is present in network response

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -184,6 +184,7 @@ export default class Core {
       if (!verticalResults || verticalResults.searchState !== SearchStates.SEARCH_LOADING) {
         this.storage.set(StorageKeys.VERTICAL_RESULTS, VerticalResults.searchLoading());
       }
+      this.storage.set(StorageKeys.DIRECT_ANSWER, {});
       this.storage.set(StorageKeys.SPELL_CHECK, {});
       this.storage.set(StorageKeys.LOCATION_BIAS, LocationBias.searchLoading());
     }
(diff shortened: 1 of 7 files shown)


src/core/core.js
```diff
@@ -258,8 +259,12 @@ export default class Core {
           const mergedResults = this.storage.get(StorageKeys.VERTICAL_RESULTS)
             .append(data[StorageKeys.VERTICAL_RESULTS]);
           this.storage.set(StorageKeys.VERTICAL_RESULTS, mergedResults);
+          if (data[StorageKeys.DIRECT_ANSWER].answer) {
```
tmeyer2115, line 262: In this instance we're guarding the setting of `StorageKeys.DIRECT_ANSWER`, but below we just set it without checking `data[StorageKeys.DIRECT_ANSWER].answer`. Do we need to check this value in both cases? If so, we could move that logic outside of the `if(query.append)` conditional and only have it once.

yen-tt: I don't think we need to check this value in both cases. The first result set from this vertical search may or may not have direct answers from live api, which is ok so I didn't place a check there. But with param query.append allowing for search result of new query to add on to the existing results, if this new search result contains a new directAnswer, then we would override it (since we expect only 1 directAnswer return). Otherwise, if this new search result doesn't have a directAnswer, then I think we should keep the previous directAnswer from current result set.

oshi97 (approved):

cea2aj (approved):

PR #1444, opened 2021-06-25
Added universalLimit to universal searches.

cjiang2000:
J=SLAP-1267
TEST=manual/auto

Checked test site to make sure change went through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,8 @@ Below is a list of configuration options related to search, used in the [base co
       verticalKey: 'verticalKey',
       // Optional, the number of results to display per page, defaults to 20. Maximum is 50.
       limit: '20',
-      // Optional, an object containing the number of results to display per vertical key. Minimum is 1. Maximum is 50. No default provided.
+      // Optional, an object containing the number of results to display per vertical key. Minimum is 1. Maximum is 50.
+      // The default value is determined by the API at 10.
       universalLimit: {
         verticalKey1: 20,
         verticalKey2: 49
--- a/README.md
+++ b/README.md
@@ -201,6 +201,11 @@ Below is a list of configuration options related to search, used in the [base co
       verticalKey: 'verticalKey',
       // Optional, the number of results to display per page, defaults to 20. Maximum is 50.
       limit: '20',
+      // Optional, an object containing the number of results to display per vertical key. Minimum is 1. Maximum is 50. No default provided.
+      universalLimit: {
+        verticalKey1: 20,
+        verticalKey2: 49
+      },
       // Optional, Vertical Pages only, a default search to use on page load when the user hasn't provided a query
       defaultInitialSearch: 'What is Yext Answers?',
     },
--- a/src/core/models/searchconfig.js
+++ b/src/core/models/searchconfig.js
@@ -12,6 +12,12 @@ export default class SearchConfig {
     this.limit = config.limit || 20;
 
     /**
+     * The max results for universal search.
+     * @type {number}
+     */
+    this.universalLimit = config.universalLimit;
+
+    /**
      * The vertical key to use for all searches
      * @type {string}
      */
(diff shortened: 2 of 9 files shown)


oshi97 (commented):
lgtm but gonna wait for the core pr to be merged in

src/core/models/searchconfig.js
```diff
@@ -11,6 +11,12 @@ export default class SearchConfig {
      */
     this.limit = config.limit || 20;
 
+    /**
+     * The max results for universal search.
+     * @type {number}
```
tmeyer2115: This would be `@type {Object<string, number>}` right?

src/core/models/searchconfig.js
```diff
@@ -11,6 +11,12 @@ export default class SearchConfig {
      */
     this.limit = config.limit || 20;
```
tmeyer2115: You may want to consider renaming `limit` here. I could see someone getting confused about how `limit` vs `universalLimit` works. They might think `universalLimit` allows us to specify limits per vertical, but `limit` is cumulative over all verticals. 

One way to avoid this confusion could be renaming `limit` to something like `limitForVertical`. Adding more info to the JSDocs about when one is used versus the other could also help disambiguate.

src/core/models/searchconfig.js
```diff
@@ -11,6 +11,12 @@ export default class SearchConfig {
...
 
+    /**
+     * The max results for universal search.
```
tmeyer2115: Add a bit more detail here. It's really the max results per vertical

README.md
```diff
@@ -201,6 +201,11 @@ Below is a list of configuration options related to search, used in the [base co
       verticalKey: 'verticalKey',
       // Optional, the number of results to display per page, defaults to 20. Maximum is 50.
       limit: '20',
+      // Optional, an object containing the number of results to display per vertical key. Minimum is 1. Maximum is 50. No default provided.
+      universalLimit: {
```
tmeyer2115, line 206: Is there an implicit limit in the sense that if no object is provided here, they will get max 50 results per vertical?

src/core/models/searchconfig.js
```diff
@@ -28,7 +34,7 @@ export default class SearchConfig {
   }
 
   validate () {
-    if (typeof this.limit !== 'number' || this.limit < 1 || this.limit > 50) {
+    if (typeof this.limitForVertical !== 'number' || this.limitForVertical < 1 || this.limitForVertical > 50) {
```
tmeyer2115, line 37: Can we add validation for the `universalLimit` as well?

README.md
cea2aj: I think the default is 10 because when I perform a universal query, I'm only seeing 10 results per vertical. Oh, I see that Chris indicated that the limit is 10

src/core/models/searchconfig.js
```diff
@@ -28,8 +34,16 @@ export default class SearchConfig {
...
       throw new AnswersConfigError('Search Limit must be between 1 and 50', 'SearchConfig');
     }
+
+    if (typeof this.universalLimit === 'object' && this.universalLimit !== null) {
+      Object.values(this.universalLimit).forEach((value) => {
+        if (typeof value !== 'number' || value < 1 || value > 50) {
+          throw new AnswersConfigError('Universal limits must be between 1 and 50', 'SearchConfig');
```
cea2aj: nit, I might be a little clearer to say 'Universal limits must be between 1 and 50, inclusive'. And I would add 'inclusive' to line 38 as well

src/core/models/searchconfig.js
cea2aj: Tom, you are suggesting to rename the internal SearchConfig property and not the public config option, correct? Otherwise that would be a breaking change

tmeyer2115 (approved):

PR #1451, opened 2021-07-01
add icons/bakedInIcon partial

oshi97:
This PR replaces all icon usages that only are meant to use a 
baked in icon with the icons/bakedInIcon partial.

The issue with just using iconPartial for everything is that, some icons are intended
to always use a baked in icon instead of switch between an iconName
(for built in icons) and an iconUrl (for custom icons).
However, handlebars partials operate on the context on which they are called, not only
on the variables passed into it. So, if an iconUrl exists on the scope somebody calls
`{{> icons/iconPartial iconName='chevron'}}`, an iconUrl will be passed into the partial
even though none was explicitly passed in.

Since iconUrl has higher priority than iconName, this can cause undesired behavior.
I think the best workaround is just creating a separate partial when we want an icon
that is built into the SDK.

J=SLAP-1297
TEST=manual

test that baked in icons display as expected and do not have this
behavior

see the animated search icon working as expected

manually check that all iconPartial usages have a specified iconUrl

### DIFF
--- a/src/ui/templates/cards/accordion.hbs
+++ b/src/ui/templates/cards/accordion.hbs
@@ -11,7 +11,7 @@ id="{{id}}">
         {{{_config.title}}}
       </h3>
       <div class="yxt-AccordionCard-icon js-yxt-AccordionCard-icon{{#if isExpanded}} yxt-AccordionCard-icon--expanded{{/if}}">
-        {{> icons/iconPartial iconName='chevron'}}
+        {{> icons/bakedInIcon iconName='chevron'}}
       </div>
     </button>
   {{else}}
--- a/src/ui/templates/icons/bakedInIcon.hbs
+++ b/src/ui/templates/icons/bakedInIcon.hbs
@@ -1,0 +1,3 @@
+<div class="Icon Icon--{{iconName}} {{classNames}}" aria-hidden="true">
+  {{icon iconName complexContentsParams}}
+</div>
--- a/src/ui/templates/cards/standard.hbs
+++ b/src/ui/templates/cards/standard.hbs
@@ -76,10 +76,10 @@
           <span>
             {{#if hideExcessDetails}}
               {{_config.showMoreText}}
-              {{> icons/iconPartial iconName='chevron' classNames='Icon--sm Icon-collapseDown'}}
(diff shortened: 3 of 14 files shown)


src/ui/templates/icons/bakedInIcon.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="Icon Icon--{{iconName}} {{classNames}}" aria-hidden="true">
+  {{icon iconName complexContentsParams}}
+</div>
```
cea2aj, line 3: Can we simplify the iconPartial now that we have this, or do we still need the if statement inside of iconPartial?

oshi97: We still need to support the switching behavior, for when both an iconUrl and an iconName is passed in

src/ui/templates/cards/accordion.hbs
```diff
@@ -11,7 +11,7 @@ id="{{id}}">
         {{{_config.title}}}
       </h3>
       <div class="yxt-AccordionCard-icon js-yxt-AccordionCard-icon{{#if isExpanded}} yxt-AccordionCard-icon--expanded{{/if}}">
-        {{> icons/iconPartial iconName='chevron'}}
+        {{> icons/bakedInIcon iconName='chevron'}}
```
cea2aj, line 14: nit: What do you think of 'builtInIcon' instead?

src/ui/templates/icons/bakedInIcon.hbs
oshi97: It makes iconPartial a little more complicated, but lets us avoid lots of guarding statements outside of it

src/ui/templates/cards/accordion.hbs
oshi97: I used bakedInIcon because builtInIcon was hard for my eyes to read, specifically the bu`iltInIc`on part

cea2aj: ya, I actually noticed that when I typed it out. As-is looks good to me

oshi97: bakedInIcon is also not perfect because of the d and I looking similar, but on the other hand it seemed like a slightly more flavorful word

cea2aj (approved):

PR #1452, opened 2021-07-01
Properly implemented sessionTrackingEnabled

cjiang2000:
J=SLAP-1408
TEST=manual

Checked on test page to make sure that session_id and cookies are being passed as expected. When sessionTrackingEnabled is true, the sessionID is passed to the test site and no cookie for sessionID is dropped. When sessionTrackingEnabled is false, the sessionID is generated by the API and no cookie is dropped. When no sessionID is passed and sessionTrackingEnabled is true, a cookie is dropped.


package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "dependencies": {
     "@mapbox/mapbox-gl-language": "^0.10.1",
-    "@yext/answers-core": "^1.2.0",
+    "@yext/answers-core": "file:../answers-core",
```
cea2aj: Once you merge in your core PR, we should publish to NPM and have this PR point to that version

src/core/core.js
```diff
@@ -236,6 +236,7 @@ export default class Core {
         offset: this.storage.get(StorageKeys.SEARCH_OFFSET) || 0,
         skipSpellCheck: this.storage.get(StorageKeys.SKIP_SPELL_CHECK),
         queryTrigger: queryTriggerForApi,
+        sessionId: this.getSessionId(),
```
cea2aj: Is sessionId supposed to be sent even if sessionTrackingEnabled is false?

src/core/core.js
```diff
@@ -746,4 +748,21 @@ export default class Core {
...
     return urls;
   }
+
+  getSessionId() {
+    if (this.storage.get(StorageKeys.SESSIONS_OPT_IN).value) {
+      try {
+        let sessionId = window.sessionStorage.getItem('sessionId');
+        if (!sessionId) {
+          sessionId = uuid();
+          window.sessionStorage.setItem('sessionId', sessionId);
```
cea2aj, line 759: nit: I generally try to avoid putting setters inside of a getters because that means the getter has side effects. What do you think of renaming this function to `getOrSetupSessionId` to try and avoid this confusion?

src/core/utils/uuid.js
```diff
@@ -0,0 +1,10 @@
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Generates a uuid.
+ * 
+ * @returns {String}
+ */
+export function uuid() {
```
cea2aj: nit: This function name is a noun, and I think it's easier to read code when function names start with verbs. Would `generateUUID` work here instead?

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -63,7 +63,7 @@ exports.legacyBundle = function (callback, outputConfig, bundleName, locale, lib
     plugins: [
       resolve({ browser: true }),
       commonjs({
-        include: [/node_modules/]
+        include: [/node_modules/, /answers-core/]
```
tmeyer2115: If we point to a published Core beta with your changes, can we remove this change?

src/core/core.js
```diff
@@ -1,6 +1,6 @@
 /** @module Core */
 import { provideCore } from '@yext/answers-core/lib/commonjs';
-
+import {uuid} from './utils/uuid';
```
tmeyer2115: Stylistic comment: there should be a space between what's imported and the curly braces: `{ uuid }`

src/core/core.js
tmeyer2115: If `sessionTrackingEnabled` is false, we should not send the `sessionId` in the request. Chris, if the value is false, will `getSessionId()` return `null`? If it is `null` and we include it here, is it filtered out by Core and not included in the resulting request?

src/core/core.js
```diff
@@ -345,6 +346,7 @@ export default class Core {
         location: this._getLocationPayload(),
         skipSpellCheck: this.storage.get(StorageKeys.SKIP_SPELL_CHECK),
         queryTrigger: queryTriggerForApi,
+        sessionId: this.getSessionId(),
```
tmeyer2115: One thing the spec is a little unclear on is which requests we need to include `sessionId` with. Is it only the search requests or on autocomplete as well? Rose may need to weigh in there.

tmeyer2115 (commented):
In the case where `sessionTrackingEnabled` is true, but we fail to use session storage, the back-end generates a cookie. Did we have to do anything special to make sure that cookie is passed back on all subsequent requests? Or is that just handled by the browser?

src/core/core.js
cjiang2000: getSessionId() returns null if sessionTrackingEnabled is false, which I think is the same as when we don't pass in sessionId.

src/core/core.js
cjiang2000: I agree that the name of the function could be better. getSessionId() is definitely misleading since it's not just a getter.

src/core/utils/uuid.js
cjiang2000: yeah we can do that.

conf/gulp-tasks/bundle/bundle.js
cjiang2000: yes, it will be removed before the merge.

src/core/core.js
cjiang2000: The browser handles cookies in the case of session storage failing. 

When sessionTrackingEnabled is false, getSessionId() returns null. When that happens, in the browser, I don't see a sessionId query parameter which indicates to me that setting it to null results in it being filtered out. 

I assumed we would only use search requests because a search request is called whenever we refresh the page. Since a search request occurs when the page loads for the first time and we generate a sessionId then. It might be redundant to include it in the autocomplete as well since we don't update our sessionId if a sessionId already exists. So for example, the page loads and a search query happens and we generate our first sessionID. another search happens and our sessionID doesn't change because we already have an existing one.

tmeyer2115: For tracking purposes, they may want to associate various auto-complete calls or question submission calls with a particular `sessionId`. Here, I think it's a question of what the back-end expects. If we don't include a `sessionId` on an auto-complete request, I'm not sure the back-end has any way of knowing which session to associate it with. @mdavish, do you know what the back-end expects here? Should all requests made by the SDK (excluding Analytics) have a `sessionId`?

src/core/models/searchconfig.js
```diff
@@ -9,7 +9,13 @@ export default class SearchConfig {
...
      * @type {number}
      */
-    this.limit = config.limit || 20;
+    this.limitForVertical = config.limit || 20;
+
+    /**
+     * The max results per vertical for universal search.
+     * @type {Object<string, number>}
+     */
+    this.universalLimit = config.universalLimit;
```
oshi97: I think these are showing up as changed because we tried the merge instead of force push, if you reset --hard the merge commit (which had no code changes) and just do a force push do these go away?

src/core/core.js
tmeyer2115: Okay, heard back from Product. We only want `sessionId` for the searches!

tmeyer2115 (approved):

PR #1453, opened 2021-07-02
Add loading indicator to SDK

yen-tt:
Add support for loading indicator and related style changes to SDK

- Update searchBarComponent's animation cycle to include loading indicator and support for custom icon url, if provided in the config data.
- Add loading indicator animation css
- Add opacity to results section on page during loading state (regardless of the use of loading indicator)
- Update SearchBar template to include default and custom loading indicator html
- Update universalComponent to not change results during loading state
- Move searchBar icon related handlebars into its own partial

J=SLAP-1335
TEST=manual

Serve test page in SDk dist/ with the new changes in bundle and tested in universal and vertical page:
- default initial search show loading indicator and switch to magifying glass/Yext icon
- During loading, universal result section have opacity set to 0.5
- hitting enter or apply filter change icon to loading indicator, and change back to magnifying glass after loading
- Click on submit button will change icon to loading indicator, and change back to Yext icon after loading
- Icon is not frozen during loading state transition to non loading state
- Default loading indicator works with custom non-loading icons based on tests above
- Custom loading indicator works with default non-loading icons based on tests above
- Custom loading indicator works with custom non-loading icons based on tests above
- smoke test on ios mobile


src/ui/components/results/universalresultscomponent.js
```diff
@@ -68,6 +68,9 @@ export default class UniversalResultsComponent extends Component {
     const query = this.core.storage.get(StorageKeys.QUERY);
     const searchState = data.searchState || SearchStates.PRE_SEARCH;
     this.updateContainerClass(searchState);
+    if (searchState === SearchStates.SEARCH_LOADING) {
```
tmeyer2115, line 73: For my own edification, what was the issue caused by calling `super.setState` in this case?

src/ui/components/search/searchcomponent.js
```diff
@@ -7,10 +7,14 @@ import SearchParams from '../../dom/searchparams';
 import TranslationFlagger from '../../i18n/translationflagger';
 import QueryUpdateListener from '../../../core/statelisteners/queryupdatelistener';
 import QueryTriggers from '../../../core/models/querytriggers';
+import SearchStates from '../../../core/storage/searchstates';
 
 const IconState = {
-  YEXT: 0,
-  MAGNIFYING_GLASS: 1
+  LOADING: 0,
+  DONE_LOADING: 1,
```
tmeyer2115: Is the `DONE_LOADING` state different than the `YEXT` state? We should show the Yext logo when the search is done loading, right?

src/ui/components/search/searchcomponent.js
```diff
@@ -280,6 +284,32 @@ export default class SearchComponent extends Component {
     if (!this._isTwin) {
       this.initQueryUpdateListener();
     }
+
+    /**
+     * If true, a loading indicator will be displayed
+     * @type {boolean}
+     */
+    this._showLoadingIndicator = (config.loadingIndicator && config.loadingIndicator.display) || false;
```
tmeyer2115: @rosiegrant was this the schema you guys created behind my back? 

```
loadingIndicator: {
  display: boolean
  iconUrl: string
}
```

src/ui/components/search/searchcomponent.js
```diff
@@ -280,6 +284,32 @@ export default class SearchComponent extends Component {
...
+
+    /**
+     * Custom icon url for loading indicator
+     * @type {string}
+     */
+    this.loadIconUrl = (config.loadingIndicator && config.loadingIndicator.iconUrl) || null;
```
tmeyer2115: Here and above, I think you can use the `?` syntax to simplify things. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/ui/components/search/searchcomponent.js
```diff
@@ -280,6 +284,32 @@ export default class SearchComponent extends Component {
...
+        storageKey: this._verticalKey ? StorageKeys.VERTICAL_RESULTS : StorageKeys.UNIVERSAL_RESULTS,
+        callback: results => {
+          results.searchState === SearchStates.SEARCH_LOADING
+            ? this.animateIconToLoading()
+            : this.animateIconToDoneLoading();
+        }
+      });
+    }
+
+    this.onMouseUp = this.oneTimeMouseUpListener.bind(this);
```
tmeyer2115, line 311: What's this used for?

src/ui/components/search/searchcomponent.js
```diff
@@ -344,59 +374,96 @@ export default class SearchComponent extends Component {
...
     if (this.iconState === iconState) {
       return;
     }
+    const prevState = this.iconState;
     this.iconState = iconState;
     if (!this.isRequestingAnimationFrame) {
       this.isRequestingAnimationFrame = true;
       window.requestAnimationFrame(() => {
-        this.forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
-        this.reverseIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
```
tmeyer2115, line 356: Do we need `this.forwardIcon` or `this.reverseIcon` any more or can we remove entirely from the class?

src/ui/components/results/universalresultscomponent.js
yen-tt: this would cause the component to rerender on the page instead of maintaining the old search

src/ui/components/search/searchcomponent.js
yen-tt: We set iconIsFrozen to true/false in `mouseup` and `mousedown` through listeners when user click on elements with `.js-yext-submit` or `.js-yxt-SearchBar-clear`. However, if user somehow move their mouse or lift their finger on mobile that's not on those elements, this will not get trigger and the icons will stay frozen. I updated it to listen on any clicks on `document` but this may get trigger a lot and we don't need to receive/listen on all the `mouseup` events. So I made a `mouseup` listener that get setup when user press down and track `mouseup` event only once.

src/ui/components/search/searchcomponent.js
yen-tt: It's different! It's use to keep track when it's in the transition state from loading to done loading and need to decide on either custom icon, magnifying glass (if searchbar is still in focus), or yext icon.

src/ui/components/search/searchcomponent.js
yen-tt: oh yeah, will do!

src/ui/components/search/searchcomponent.js
yen-tt: no we don't need those variables anymore. I think I already remove them from the class?

src/ui/components/results/universalresultscomponent.js
cea2aj: Can you add a comment in the code that indicates that's why we want to avoid calling setState here?

src/ui/components/results/universalresultscomponent.js
```diff
@@ -68,6 +68,9 @@ export default class UniversalResultsComponent extends Component {
...
+      return;
+    }
     return super.setState(Object.assign(data, {
       isPreSearch: searchState === SearchStates.PRE_SEARCH,
       isSearchLoading: searchState === SearchStates.SEARCH_LOADING,
```
cea2aj, line 78: This `isSearchLoading` variable would be false even when search is loading, however I'm not sure if there's a great way to get around that

src/ui/components/search/searchcomponent.js
```diff
@@ -280,6 +284,32 @@ export default class SearchComponent extends Component {
...
+     * If true, a loading indicator will be displayed
+     * @type {boolean}
+     */
+    this._showLoadingIndicator = config.loadingIndicator?.display || false;
+
+    /**
+     * Custom icon url for loading indicator
+     * @type {string}
+     */
+    this.loadIconUrl = config.loadingIndicator?.iconUrl || null;
```
cea2aj: nit: thoughts on 'loadingIconUrl' rather than 'loadIconUrl'?

yen-tt: sounds good!

src/ui/components/search/searchcomponent.js
```diff
@@ -344,59 +374,96 @@ export default class SearchComponent extends Component {
...
       return;
     }
+    const prevState = this.iconState;
     this.iconState = iconState;
     if (!this.isRequestingAnimationFrame) {
       this.isRequestingAnimationFrame = true;
       window.requestAnimationFrame(() => {
-        this.forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
-        this.reverseIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
+        const iconEls = DOM.queryAll('.js-yxt-SearchBar-Icon');
```
cea2aj: Is it possible to set the parent to this container like this: `DOM.queryAll(this._container, '.js-yxt-SearchBar-Icon');`,  or do we have to search the whole document?

yen-tt: good point, just this container should be fine.

src/ui/components/search/searchcomponent.js
```diff
@@ -344,59 +374,96 @@ export default class SearchComponent extends Component {
...
         }
         this.isRequestingAnimationFrame = false;
       });
     }
   }
 
+  animateIconToLoading () {
+    this.loadingIconUrl
+      ? this.requestIconAnimationFrame(IconState.LOADING, '.js-yxt-SearchBar-loadingImage')
+      : this.requestIconAnimationFrame(IconState.LOADING, '.js-yxt-SearchBar-LoadingIndicator');
```
cea2aj: One of these has loading with a lowercase 'l' and the other has loading with an uppercase 'L'. Is it possible to be consistent with that?

yen-tt: ah I saw in the existing code that the custom images have `yxt-SearchBar-buttonImage` lowercase while the animated ones have uppercase `yxt-SearchBar-AnimatedIcon` so I was kinda following it. I could change them all to Pascal casing if that's preferred?

src/ui/components/search/searchcomponent.js
cea2aj: Does that mean it that the icon can still get frozen if someone lifts their finger off of those elements a second or third time?

yen-tt: every time they press down, it setup another oneTimeMouseUpListener! so oneTimeMouseUpListener would get trigger when they lift their finger off, unfrozen the icon, and remove itself

src/ui/components/search/searchcomponent.js
```diff
@@ -280,6 +284,32 @@ export default class SearchComponent extends Component {
...
+     * If true, a loading indicator will be displayed
+     * @type {boolean}
+     */
+    this._showLoadingIndicator = config.loadingIndicator?.display || false;
+
+    /**
+     * Custom icon url for loading indicator
+     * @type {string}
+     */
+    this.loadingIconUrl = config.loadingIndicator?.iconUrl || null;
```
cea2aj: I konw I just suggested this name, but now I'm thinking renaming this to `customLoadingIconUrl` might be clearer

src/ui/components/search/searchcomponent.js
cea2aj: I think i'd prefer to have them consistent within this file. How about CustomLoadingImage?

src/ui/components/search/searchcomponent.js
yen-tt: yup. make sense, like the other variable name`customIconUrl`..I should've thought of that before

src/ui/components/search/searchcomponent.js
```diff
@@ -344,59 +374,96 @@ export default class SearchComponent extends Component {
     }
   }
 
-  requestIconAnimationFrame (iconState) {
+  requestIconAnimationFrame (iconState, iconElClass) {
```
cea2aj: With the added complexity of of the new animated icon, I think it makes sense to consider creating a new class that manages the styling of the icon component. I don't think it needs to be a full-blown component. It could be called the IconAnimator, and it could provide a function to accept a state, and in response it would toggle the corresponding CSS classes. It would allow us to remove icon-specific styling from the search component class

cea2aj: Is it possible to remove `iconElClass` from this param and instead derive it from the iconState?

src/ui/components/search/searchcomponent.js
cea2aj: Is it possible to remove the DONE_LOADING state since it's just an intermediary state? We could still have a function which sets a done loading state, however that function would simply pick on of the possible done states (CUSTOM_ICON, YEXT, or MAGNIFYING_GLASS)

src/ui/components/search/searchcomponent.js
```diff
@@ -7,10 +7,14 @@ import SearchParams from '../../dom/searchparams';
 import TranslationFlagger from '../../i18n/translationflagger';
 import QueryUpdateListener from '../../../core/statelisteners/queryupdatelistener';
 import QueryTriggers from '../../../core/models/querytriggers';
+import SearchStates from '../../../core/storage/searchstates';
 
 const IconState = {
-  YEXT: 0,
-  MAGNIFYING_GLASS: 1
+  LOADING: 0,
```
cea2aj: Should we have a CUSTOM_LOADING state for consistency with the CUSTOM_ICON state?

src/ui/components/search/searchcomponent.js
yen-tt: The classnames of the icon components to get the right styling in the logic would still be tied to the searchcomponent so it wouldn't be general enough for something else to reuse the function. I think it's alright to keep it like this for now?
for your second comment, LOADING state originally could refer to default or custom icon so there was a need for `iconElClass`. But if I add `CUSTOM_LOADING` to the IconState and adjust some of the logic, then yup we can just derive it from iconState! I will update this

src/ui/components/search/searchcomponent.js
yen-tt: I will make this update so `iconElClass` is not needed when requesting icon animation frame

src/ui/components/search/searchcomponent.js
yen-tt: Right, I already included `animateIconToDoneLoading()` that sets what done possible loading state. The concern of not having DONE_LOADING was calling `animateIconToYext` and `animateIconToMagnifyingGlass` which I added a condition to not animate when it's in loading state. but yeah I could just directly call `requestIconAnimationFrame` to bypass that or avoid setting iconState prematurely.

src/ui/components/search/searchcomponent.js
```diff
@@ -378,25 +440,27 @@ export default class SearchComponent extends Component {
...
     const clickableElementSelectors = ['.js-yext-submit', '.js-yxt-SearchBar-clear'];
     for (const selector of clickableElementSelectors) {
       const clickableEl = DOM.query(this._container, selector);
       if (clickableEl) {
         DOM.on(clickableEl, 'mousedown', () => {
           this.iconIsFrozen = true;
-        });
-        DOM.on(clickableEl, 'mouseup', () => {
-          this.iconIsFrozen = false;
+          DOM.on(document, 'mouseup', this.onMouseUp);
```
oshi97, line 451: so basically we want the mouseup listener to be on the document not the icons?

src/ui/components/search/searchcomponent.js
oshi97: ah yeah I didn't instantiate these in the constructor

src/ui/sass/modules/_Icon.scss
```diff
@@ -160,3 +160,15 @@
     }
   }
 }
+
+.Icon--magnifying_glass svg
+{
```
oshi97: Is this going to affect other usages of the magnifying glass icon. Wondering if we need to scope this css to just the loading indicator

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +378,58 @@ export default class SearchComponent extends Component {
...
+    this.prevState = this.iconState;
     this.iconState = iconState;
     if (!this.isRequestingAnimationFrame) {
       this.isRequestingAnimationFrame = true;
       window.requestAnimationFrame(() => {
-        this.forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
-        this.reverseIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
+        const iconEls = DOM.queryAll(this._container, '.js-yxt-SearchBar-Icon');
+        iconEls.forEach((el) => {
+          el.classList.contains('js-yxt-SearchBar-LoadingIndicator')
```
oshi97: I know I didn't do this for the original animated search icon, but would we be able to move the state management into another class? If this was React we would probably have it be another component, but child components do not work nicely in the SDK so we're not able to divide responsibility like that. It's kind of hard to tell what's going on here. I think generally speaking it's cleaner to keep state as some javascript variables and not rely on css classes in the DOM, otherwise you can end up in jQuery-like spaghetti land

src/ui/templates/search/search.hbs
```diff
@@ -36,25 +36,39 @@
         </button>
         <button {{#if useForm}}type="submit"{{else}}type="button"{{/if}}
           class="js-yext-submit yxt-SearchBar-button">
+          {{#if showLoadingIndicator}}
```
oshi97: could we move the icon handlebars into it's own partial? Fun fact, any file you make in the src/ui/templates folder is automatically registered as both a partial and template in the templates bundle! @tmeyer2115 iirc in order to fulfill the "no inline JS" csp, we're only not allowed to use inline partials, but regular partials are fine right? 
If we can't use regular partials in the search bar, we'll also need to use IconComponent instead of iconPartial (this was something I didn't think about for the iconPartial PR).

src/ui/templates/search/search.hbs
```diff
@@ -98,24 +112,44 @@
...
               {{> icons/iconPartial iconUrl=_config.customIconUrl}}
             </div>
           {{else}}
-            <div class="yxt-SearchBar-AnimatedIcon yxt-SearchBar-AnimatedIcon--paused js-yxt-AnimatedForward
-              {{#if autoFocus}} yxt-SearchBar-AnimatedIcon--inactive{{/if}}"
-              data-component="IconComponent"
-              data-opts="{{json forwardIconOpts}}">
+            <div class="yxt-SearchBar-AnimatedIcon yxt-SearchBar-AnimatedIcon--paused js-yxt-AnimatedForward js-yxt-SearchBar-Icon
+              {{#unless autoFocus}} yxt-SearchBar-Icon--inactive{{/unless}}">
+              {{> icons/bakedInIcon
```
oshi97: (I just renamed this to builtInIcon after some comments from Tom sorry about affecting your pr :( )

src/ui/components/search/searchcomponent.js
yen-tt: Yup. If user press down on the icon (which freeze them), and somehow nudge their mouse or lift their finger not on the icon, it would get stuck. It's best to check mouseup on the document instead.

src/ui/sass/modules/_Icon.scss
yen-tt: This is just a static magnifying glass without the animation from yext to it, but will replace the original animated magnifying glass icon if it's a transition from loading state to non-loading. since this not used in a children div of the loading indicator, so by scoping it, you mean renaming it to something related to loading indicator?

src/ui/templates/search/search.hbs
yen-tt: No worries! I will update it, thanks for letting me know!

src/ui/templates/search/search.hbs
```diff
@@ -98,24 +112,44 @@
...
           class="js-yext-submit yxt-SearchBar-button">
+          {{#if showLoadingIndicator}}
+            {{#if customLoadingIconUrl}}
+              <div class="yxt-SearchBar-CustomLoadingImage js-yxt-SearchBar-CustomLoadingImage js-yxt-SearchBar-Icon yxt-SearchBar-Icon--inactive">
+                {{> icons/iconPartial iconUrl=customLoadingIconUrl}}
+              </div>
+            {{else}}
+              <div class="yxt-SearchBar-AnimatedIcon js-yxt-SearchBar-LoadingIndicator yxt-SearchBar-LoadingIndicator js-yxt-SearchBar-Icon 
+                yxt-SearchBar-LoadingIndicator--inactive Icon--lg">
+                <svg class= "yxt-SearchBar-LoadingIndicator-AnimatedIcon" viewBox="0 0 72 72">
```
oshi97, line 123: what do you think about adding a "circle" icon to the list of built in icons?

src/ui/sass/modules/_Icon.scss
yen-tt: updated

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +376,46 @@ export default class SearchComponent extends Component {
...
     }
+    this.prevState = this.iconState;
     this.iconState = iconState;
     if (!this.isRequestingAnimationFrame) {
       this.isRequestingAnimationFrame = true;
       window.requestAnimationFrame(() => {
-        this.forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
-        this.reverseIcon.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
+        DOM.query(this._container, this.prevState).classList.add('yxt-SearchBar-Icon--inactive');
+        const iconEl = DOM.query(this._container, this.iconState);
```
oshi97: could we call this "activeIcon" instead?

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +376,46 @@ export default class SearchComponent extends Component {
...
+        iconEl.classList.remove('yxt-SearchBar-Icon--inactive');
+        iconEl.classList.remove('yxt-SearchBar-AnimatedIcon--paused');
```
oshi97: would it be possible to remove `yxt-SearchBar-AnimatedIcon--paused` only when the iconState is MAGNIFYING_GLASS or YEXT?

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +376,46 @@ export default class SearchComponent extends Component {
...
         if (this.iconState === IconState.MAGNIFYING_GLASS) {
-          this.forwardIcon.classList.remove('yxt-SearchBar-AnimatedIcon--inactive');
-          this.reverseIcon.classList.add('yxt-SearchBar-AnimatedIcon--inactive');
-        } else if (this.iconState === IconState.YEXT) {
-          this.forwardIcon.classList.add('yxt-SearchBar-AnimatedIcon--inactive');
-          this.reverseIcon.classList.remove('yxt-SearchBar-AnimatedIcon--inactive');
+          const forwardEl = DOM.query(this._container, '.Icon--yext_animated_forward');
```
oshi97: is `forwardEl` the same DOM element as `iconEl`?

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +376,46 @@ export default class SearchComponent extends Component {
...
+          this.prevState === IconState.LOADING || this.prevState === IconState.CUSTOM_LOADING
```
oshi97: maybe we could do something like
`const wasLoading = [ IconState.LOADING, IconState.CUSTOM_LOADING ].includes(this.prevState)`?

src/ui/components/search/searchcomponent.js
```diff
@@ -348,26 +376,46 @@ export default class SearchComponent extends Component {
...
+            ? forwardEl.classList.add('yxt-SearchBar-MagnifyingGlass')
```
oshi97: Ok I just thought of this but would adding a `--static` suffix make things a little clearer? Maybe we add a comment here saying why we need this, since it's kind of a specific, nonsensical (from a code perspective) product request?

src/ui/components/search/searchcomponent.js
```diff
@@ -524,10 +574,12 @@ export default class SearchComponent extends Component {
       }
     }
 
-    inputEl.blur();
-    DOM.query(this._container, '.js-yext-submit').blur();
-    if (this.isUsingYextAnimatedIcon) {
-      this.animateIconToYext();
+    if (!this._showLoadingIndicator) {
+      inputEl.blur();
```
oshi97: is it ok to move inputEl.blur() into this if statement?

Actually for fun I went digging and this originally came from https://github.com/yext/answers-search-ui/pull/174/files
I wonder if this was related to the weird mobile keyboard techops we had a while back that Connor looked into

src/ui/components/search/searchcomponent.js
yen-tt: yup

src/ui/components/search/searchcomponent.js
yen-tt: Not quite, it's the div inside iconEl

src/ui/components/search/searchcomponent.js
yen-tt: hmm sounds good, will do!

src/ui/components/search/searchcomponent.js
oshi97: ahh got it. could we query iconEl instead of this._container?

src/ui/components/search/searchcomponent.js
yen-tt: ah ok I see. yeah it would bring back that techop issue. 
I made an update. `inputEl.blur()` gets call without that if statement (got rid of the call `animateIconToYext ()` that is right after this, since the blur will trigger the `focusout` listener to animate it anyway). But, with this change, the call to `window.requestAnimationFrame(..)` will not get register each time since the time frame between each call is too quick for the function to set `this.isRequestingAnimationFrame` back to false. This mean we can't rely on hiding just the previous iconState. I had to change the code back to hiding all icons with `.js-yxt-SearchBar-Icon` instead of just the prevState.

yen-tt: Now that I'm thinking about it, the original idea was that pressing enter while typing on the searchbar should still remain focus on the searchbar. But if this lead to weird mobile keyboard, then with this update to always blur on submit, technically we will never have a scenario where we go from loading state to magnifying glass but instead to yext..darn. Updated to have static yext icon instead of mag glass.

src/ui/components/search/searchcomponent.js
yen-tt: sure thing

oshi97 (approved):
looks good to me!

PR #1454, opened 2021-07-02
Spellcheck integration testing

cea2aj:
Convert a spellcheck acceptance test into integration tests

This test coverage has two parts. In the first part, our test coverage confirms that the spellcheck components renders with the correct query params in the hyperlink. To accomplish this, I converted the previous spellcheck tests to more of an integration test by mounting the component with the enzyme adapter.

In the second part, I confirm that when ANSWERS initializes with the spellcheck params in the URL, a search is fired with those spellcheck params. To make this possible, this PR sets up integration testing for answers-umd which previously did not exist in our jest tests. This PR mocks the core library so that we don't actually fire HTTP requests when running these tests.

J=SLAP-1062
TEST=auto

Add integration test coverage

### DIFF
--- a/tests/answers-umd.js
+++ b/tests/answers-umd.js
@@ -36,7 +36,14 @@ function mockWindowWithData (data) {
     performance: {
       mark: jest.fn()
     },
+    sessionStorage: {
+      getItem: jest.fn(),
+      setItem: jest.fn()
+    },
     addEventListener: jest.fn(),
+    history: {
+      replaceState: jest.fn()
+    },
     ...data
   }));
 }
--- a/tests/answers-umd.js
+++ b/tests/answers-umd.js
@@ -1,0 +1,56 @@
+import ANSWERS from '../src/answers-umd';
+
+let windowSpy;
+
+describe('ANSWERS instance integration testing', () => {
+  beforeEach(() => {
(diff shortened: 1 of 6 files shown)


tmeyer2115 (approved):
LGTM! Really like the mocked Core pattern. Maybe we can leverage that and the new `answers-umd` test suite to convert other acceptance tests over.

tests/answers-umd.js
```diff
@@ -0,0 +1,63 @@
+import ANSWERS from '../src/answers-umd';
+
+let windowSpy;
+
+describe('ANSWERS instance integration testing', () => {
+  beforeEach(() => {
+    windowSpy = jest.spyOn(window, 'window', 'get');
```
oshi97, line 7: is this needed so we can reset the window after each test?

cea2aj: ya, it's setup this way so that on each test we can have a fresh window to work with

oshi97 (approved):

oshi97:
like the mock for the answers-core!

cea2aj:
> like the mock for the answers-core!

I was surprised to find that without the mock, answers-core will actually fire real HTTP requests when this test is ran

PR #1460, opened 2021-07-07
Dev/sessiontest

cjiang2000:
added sessionId testing

J=SLAP-1409
TEST=automatic

ran tests and they work as expected.

### DIFF
--- a/tests/core/core.js
+++ b/tests/core/core.js
@@ -3,6 +3,11 @@ import SearchConfig from '../../src/core/models/searchconfig';
 import Storage from '../../src/core/storage/storage';
 import StorageKeys from '../../src/core/storage/storagekeys';
 
+jest.mock('../../src/core/utils/uuid');
+const uuid = require('../../src/core/utils/uuid');
+// uuid is a mock function
+uuid.generateUUID = jest.fn(() => '42');
+
 describe('Search requests are created properly', () => {
   window.performance.mark = jest.fn();
   const context = {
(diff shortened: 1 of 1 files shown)


oshi97:
we'll wanna rebase this guy on top of develop so that it doesn't include changes from #1452 (properly implement sessionTrackingEnabled)

package-lock.json
```diff
@@ -5158,6 +5158,10 @@
           "integrity": "sha512-tCfb2WLjqFAtXn4KEdxIhalnRtoKFN7nAwj0B3ZXCbQloV2tq5eDbcTmT68JJD3nRJq24/XgxtQKFIpQdtvmVw==",
           "dev": true,
           "requires": {
+            "commander": "^2.7.1",
+            "lodash.get": "^4.0.0",
+            "lodash.isequal": "^4.0.0",
+            "validator": "^8.0.0"
```
tmeyer2115: This seems to be causing a build error. There's no `,` after this line, so `npm` fails ti parse the `package-lock`

tests/core/core.js
```diff
@@ -41,6 +46,44 @@ describe('Search requests are created properly', () => {
   });
 });
 
+describe('Search requests are created properly', () => {
```
tmeyer2115: We shouldn't have two `describe` blocks with the same name. We'd either put these tests in the block above or give this `describe` block another name.

tests/core/core.js
```diff
@@ -41,6 +46,44 @@ describe('Search requests are created properly', () => {
...
+  it('sessionId is passed in vertical search', () => {
+    const mockCore = getMockCore();
+    mockCore.storage.set(StorageKeys.SESSIONS_OPT_IN, { value: true });
+    mockCore.verticalSearch();
+    expect(mockCore._coreLibrary.verticalSearch).toHaveBeenCalledWith(
+      expect.objectContaining({ sessionId: '42' })
+    );
+  });
+
+  it('sessionId is passed in universal search', () => {
```
tmeyer2115: Here and below, we're verifying that `sessionId` is not passed in search

tests/core/core.js
```diff
@@ -3,6 +3,11 @@ import SearchConfig from '../../src/core/models/searchconfig';
 import Storage from '../../src/core/storage/storage';
 import StorageKeys from '../../src/core/storage/storagekeys';
 
+jest.mock('../../src/core/utils/uuid');
+const foo = require('../../src/core/utils/uuid');
```
oshi97: does importing generateUUID like 
```js
const { generateUUID } = require('../../src/core/utils/uuid')
```
work? If that doesn't work we should rename `foo` to `uuid`

tests/core/core.js
```diff
@@ -3,6 +3,11 @@ import SearchConfig from '../../src/core/models/searchconfig';
 import Storage from '../../src/core/storage/storage';
 import StorageKeys from '../../src/core/storage/storagekeys';
 
+jest.mock('../../src/core/utils/uuid');
```
tmeyer2115, line 6: In his Spellcheck integration test PR, Connor created a `__mocks__` folder for the SDK. You can read a bit about how that works here: https://jestjs.io/docs/manual-mocks. The tl;dr is that there are multiple ways to manually mock a module. But, we should pick one way and be consistent in our libraries. Since we've already got the `__mocks__` pattern established, let's go with that. I think you'll need to take what you have here and make a `__mocks__/core/utils/uuid` file.

oshi97 (approved):

PR #1465, opened 2021-07-14
Testcafe acceptance test wait for results

yen-tt:
Update testCafe acceptance test

- Add waitOnSearchComplete functionality in verticalResults and UniversalResults component to wait for query response(s) and searchComplete state appears on page.
- Register requestLogger to each test to track vertical/universal requests and response to Answers API endpoint
- Update acceptance tests to wait for results on search

J=SLAP-1425
TEST= auto

ran testCafe locally, verify logger requests and wait time in tests

### DIFF
--- a/tests/acceptance/acceptancesuites/acceptancesuite.js
+++ b/tests/acceptance/acceptancesuites/acceptancesuite.js
@@ -24,18 +24,23 @@ import StorageKeys from '../../../src/core/storage/storagekeys';
  */
 
 fixture`Universal search page works as expected`
+  .beforeEach(async t => {
+    await UniversalPage.getUniversalResultsComponent().registerLogger(t);
+  })
   .before(setupServer)
   .after(shutdownServer)
   .page`${UNIVERSAL_PAGE}`;
 
 test('Basic universal flow', async t => {
   const searchComponent = UniversalPage.getSearchComponent();
+  const resultComponent = UniversalPage.getUniversalResultsComponent();
 
   await searchComponent.enterQuery('Tom');
   await searchComponent.clearQuery();
 
   await searchComponent.enterQuery('ama');
   await searchComponent.getAutoComplete().selectOption('amani farooque phone number');
+  await resultComponent.waitOnSearchComplete();
 
   const sections =
         await UniversalPage.getUniversalResultsComponent().getSections();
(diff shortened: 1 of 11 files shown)


oshi97:
can we change cafetest to testcafe in the commit msg

tests/acceptance/acceptancesuites/acceptancesuite.js
```diff
@@ -46,84 +51,120 @@ test('Basic universal flow', async t => {
...
   await t.expect(pageNum).eql('Page 2');
+
   await browserRefreshPage();
+  await resultComponent.waitOnSearchComplete();
   pageNum = await paginationComponent.getActivePageLabelAndNumber();
   await t.expect(pageNum).eql('Page 2');
 });
 
 test('spell check flow', async t => {
+  const searchQueryUrl = VerticalPage.getVerticalResultsComponent().getSearchQueryUrl();
```
oshi97: I think the url should go into some kind of constants file instead of on the VerticalResultsComponent. either one of the existing utils files or a new one

tests/acceptance/blocks/universalresultscomponent.js
```diff
@@ -23,4 +28,42 @@ export default class UniversalResultsComponentBlock {
 
     return sections;
   }
+
+  /**
+   * Wait for results to load on page by checking query response status and searchComplete state
+   * (timeout is set to 10 seconds)
+   */
+  async waitOnSearchComplete () {
```
oshi97: What do you think about making a ResultsLogger wrapper that would handle waiting for a request? Feels a little weird to sandwich it with the component classes

cea2aj:
How much of a problem was the waiting for results for the SDK acceptance tests? I am a little concerned that we are adding more complexity by adding this to nearly every test, and we are trying to solve a problem that's not really there. If there are specific tests that are having an issue with not waiting for the results to load, maybe we should just add this wait to those tests?

tests/acceptance/blocks/universalresultscomponent.js
cea2aj: I agree, I think we should isolate this request logger code so that we don't have to duplicate it in both the unviversalresultscomponent and in the verticalresultscomponent

yen-tt:
> How much of a problem was the waiting for results for the SDK acceptance tests? I am a little concerned that we are adding more complexity by adding this to nearly every test, and we are trying to solve a problem that's not really there. If there are specific tests that are having an issue with not waiting for the results to load, maybe we should just add this wait to those tests?

From Oliver's investigation, there isn't a specific acceptance test failure in SDK (mainly from theme). In theory, we can rely on testcafe [built-in wait mechanism](https://testcafe.io/documentation/402827/guides/concepts/built-in-wait-mechanisms) for the SDK acceptance test, which would take care of most(?) of the tests in SDK. There are some scenarios that the built-in wait doesn't apply, like asserting for selector's count and exists property, and the default 3 seconds wait for requests only before a test action and not necessarily before an assertion unless a specific timeout option is provided. I can remove at some spots to reduce the complexity a bit.

tests/acceptance/blocks/universalresultscomponent.js
yen-tt: yup, will do!

tests/acceptance/searchrequestlogger.js
```diff
@@ -0,0 +1,66 @@
...
+   */
+  async waitOnSearchComplete () {
+    const responseWaitTimeout = 10000;
+    const waitTimeInterval = 200;
+    let totalWaitTime = 0;
+    while (totalWaitTime < responseWaitTimeout && !this.isLoggerResultsPresent()) {
+      await t.wait(waitTimeInterval);
+      totalWaitTime += waitTimeInterval;
+    }
+    await t.expect(this._searchComplete.exists).ok();
```
oshi97: sorry if somebody already asked but do we need to wait for the search complete state? I'm sure at worst it doesn't hurt to have it but I feel like we might not need it

yen-tt: `exists` property is evaluated immediately regardless of selector timeout so we wouldn't be waiting in this assert statement. Yeah the results response check is probably sufficient, this is to confirm that the UI setup on page is completely loaded.

yen-tt: hmm it would just be an extra check so it's not needed, will remove it!

oshi97 (approved):

PR #1466, opened 2021-07-14
add opacity to DirectAnswer during search loading state

yen-tt:
Add opacity to DirectAnswer while search is loading

- Update DirectAnswer to have searchState 
- add the corresponding css class to container based on searchState for opacity styling

J=SLAP-1445
TEST=auto&manual

- added jest test to verify the right css class is set for different loading state
- Launched test site and verified that opacity is added to direct answer on universal page

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -329,7 +329,6 @@ export default class Core {
       }
     }
 
-    this.storage.set(StorageKeys.DIRECT_ANSWER, {});
     const universalResults = this.storage.get(StorageKeys.UNIVERSAL_RESULTS);
     if (!universalResults || universalResults.searchState !== SearchStates.SEARCH_LOADING) {
       this.storage.set(StorageKeys.UNIVERSAL_RESULTS, UniversalResults.searchLoading());
@@ -337,6 +336,7 @@ export default class Core {
     this.storage.set(StorageKeys.QUESTION_SUBMISSION, {});
     this.storage.set(StorageKeys.SPELL_CHECK, {});
     this.storage.set(StorageKeys.LOCATION_BIAS, LocationBias.searchLoading());
+    this.storage.set(StorageKeys.DIRECT_ANSWER, DirectAnswer.searchLoading());
 
     const queryTrigger = this.storage.get(StorageKeys.QUERY_TRIGGER);
     const queryTriggerForApi = this.getQueryTriggerForSearchApi(queryTrigger);
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -329,6 +329,7 @@ export default class Core {
       }
     }
 
+    this.storage.set(StorageKeys.DIRECT_ANSWER, DirectAnswer.searchLoading());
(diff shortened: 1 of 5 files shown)


src/ui/components/results/directanswercomponent.js
```diff
@@ -210,7 +229,8 @@ export default class DirectAnswerComponent extends Component {
   }
 
   eventOptions (data) {
-    if (!data || Object.keys(data).length === 0) {
+    if (!data || Object.keys(data).length === 0 ||
+      (Object.keys(data).length === 1 && Object.keys(data)[0] === 'searchState')) {
```
oshi97: do we need this extra guarding?

src/core/core.js
```diff
@@ -329,14 +329,14 @@ export default class Core {
...
 
-    this.storage.set(StorageKeys.DIRECT_ANSWER, {});
     const universalResults = this.storage.get(StorageKeys.UNIVERSAL_RESULTS);
     if (!universalResults || universalResults.searchState !== SearchStates.SEARCH_LOADING) {
       this.storage.set(StorageKeys.UNIVERSAL_RESULTS, UniversalResults.searchLoading());
     }
     this.storage.set(StorageKeys.QUESTION_SUBMISSION, {});
     this.storage.set(StorageKeys.SPELL_CHECK, {});
     this.storage.set(StorageKeys.LOCATION_BIAS, LocationBias.searchLoading());
+    this.storage.set(StorageKeys.DIRECT_ANSWER, DirectAnswer.searchLoading());
```
oshi97: do we need to change the order here? I think it's probably fine in this case, but we try to avoid changing this because messing with the ordering can change behavior in unexpected ways

src/ui/components/results/directanswercomponent.js
yen-tt: yeah, if there's no direct answer found in live api response after a complete search, directAnswer will have a SEARCH_COMPLETE state but no actual directAnswers `data` field, which `eventOptions` would throw an error trying to access `data.relatedItem.verticalConfigId`

src/core/core.js
yen-tt: ah no it's not needed. I don't think it would change any behavior, just want to organize it with other `set` call. I will keep it in mind!

src/ui/components/results/directanswercomponent.js
oshi97: could we either guard this by the existence of relattedItem.verticalConfigId then, or use optional chaining?

src/core/core.js
oshi97: okay let's move it back then if it's not needed then!

oshi97: changing the order will affect the order that components are rendered, which could break some weird thing consulting is doing

src/ui/components/results/directanswercomponent.js
yen-tt: hmm optional chaining would work. Updated!

oshi97 (approved):

PR #1468, opened 2021-07-15
add locale defaulting for SpeechRecognition and support check method

oshi97:
This commit adds a method to handle locale defaulting on microsoft edge.
Chrome and Safari both have nice handling for strange locales like en_FAKE.
Edge is more sensitive, and will throw an error if the locale is not one
of its specific supported ones, or if the locale uses underscores instead of dashes.

The list of locales supported in edge was generated programmatically using a script.
There is documentation for the list of locales supported by Azure's Speech service,
but I could not find anything saying that the SpeechRecognition implementation on Edge
uses this service directly (though it seems to be the case).

J=SLAP-1441
TEST=manual,auto

tested the isMicrosoftEdge() function on different browsers by exposing it on ANSWERS
did the same thing with transformSpeechRecognitionLocaleForEdge()


src/core/speechrecognition/locales.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @param {string} locale
+ * @returns {string}
+ */
+export function transformSpeechRecognitionLocaleForEdge (locale) {
+  const underscoreIndex = locale.indexOf('_');
+  if (underscoreIndex === -1) {
+    return locale;
+  }
+  const isCompatibleWithEdge = SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE.includes(locale.toLowerCase());
```
cea2aj, line 18: Is is possible that the supplied locale has a dash and therefore it wouldn't be found in the list of supported locales?

oshi97: Yeah, but that's only possible if it's a user supplied locale. That's why I added the ` Expects the input locale to use an underscore separator as opposed to a dash separator.` part of the jsdoc. I can add handling if you think it's necessary, I was just a little worried about locales that have multiple separators like uz-Latn-UZ.

oshi97: I don't think we deal with any locales like that to be fair so I'll just add the handling. We already do locale canonicalization in jambo so any issues we run into here would probably be mirrored there anyways.

tests/core/speechrecognition/locales.js
```diff
@@ -0,0 +1,26 @@
+import { transformSpeechRecognitionLocaleForEdge } from '../../../src/core/speechrecognition/locales';
+
+jest.mock('../../../src/core/constants', () => ({
+  SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE:
+    require('../../../conf/i18n/constants').SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE
```
cea2aj, line 5: this is a cool way of doing it

tests/core/speechrecognition/locales.js
```diff
@@ -0,0 +1,26 @@
...
+});
+
+it('replaces underscores with dashes for supported locales', () => {
+  expect(transformSpeechRecognitionLocaleForEdge('en_US')).toEqual('en-US');
+  expect(transformSpeechRecognitionLocaleForEdge('en_GB')).toEqual('en-GB');
+});
+
+it('is case insensitive', () => {
+  expect(transformSpeechRecognitionLocaleForEdge('en_us')).toEqual('en-us');
+  expect(transformSpeechRecognitionLocaleForEdge('EN_AI')).toEqual('EN');
```
cea2aj, line 29: to confirm, Edge doesn't care about the casing of the locale?

oshi97: nope!

tests/core/speechrecognition/locales.js
oshi97: I just wanted all the locale stuff to be colocated - I'm a little paranoid about us forgetting to test new locales for SpeechRecognition

src/core/utils/useragent.js
```diff
@@ -0,0 +1,15 @@
...
+ * using the User-Agent string if clients hints are not supported.
+ *
+ * @returns {boolean}
+ */
+export function isMicrosoftEdge () {
+  const brands = navigator.userAgentData?.brands;
+  if (brands && brands.length > 0) {
+    return !!brands.find(b => b.brand === 'Microsoft Edge');
+  } else if (navigator.userAgent) {
+    return navigator.userAgent.includes('Edg');
```
cea2aj, line 13: Do you think just checking these 3 characters might lead to some false positives? Is it practical to instead list out the possible full strings?

oshi97: I spent quite a while trying to find better ways, but this is the recommended way for checking for Edge. I agree it's strange because any other word that starts with Edg will give us a false positive. I don't think it's practical to list out all possible strings.

oshi97: On the bright side it looks like the future is User-Agent Client Hints!

src/core/speechrecognition/locales.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param {string} locale
+ * @returns {string}
+ */
+export function transformSpeechRecognitionLocaleForEdge (locale) {
+  const underscoreIndex = locale.indexOf('_');
+  if (underscoreIndex === -1) {
+    return locale;
+  }
+  locale = locale.replace('_', '-');
```
cea2aj, line 17: nit: modifying the locale here means this function has the side-effect of modifying the input param. It's just a string so it's probably okay, but still something to watch out for

cea2aj (approved):

src/core/speechrecognition/locales.js
oshi97: this won't modify the original locale though! You can test this in console with a simple function like
```js
function mutate(a) { a = '123' }
```
and passing in a string

oshi97: I think this kind of mutation is only an issue when you touch the data inside an object, reassignment (primitive or object) shouldn't affect the original variable

cea2aj: ah, you're right!

PR #1470, opened 2021-07-15
fix waitOnSearchComplete loop issue

yen-tt:
- update hook registration to use `requestHook` on fixture instead of `beforeEach` for each test
- update waitOnSearchComplete to use current testInstance from test suite instead of the imported testInstance reference from testcafe

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,22 +80,58 @@
       }
     },
     "@babel/helper-annotate-as-pure": {
-      "version": "7.10.4",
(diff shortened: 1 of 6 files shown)


oshi97:
oh fun fact you can quickly iterate on testcafe tests using live mode! https://testcafe.io/documentation/402830/guides/basic-guides/run-tests#live-mode
just add the -L flag when using the testcafe cli. I would recommend combining it with the -t or -T flags to just run a single test

oshi97 (approved):

oshi97 (approved):

PR #1473, opened 2021-07-16
Voice Search Icon

cea2aj:
Add the voice search icon for voice search support

There are two voice search icon SVGs. One is the microphone and the other are the listening dots. The animations are stored in the SVGs and they are activated by the VoiceSearchController. Both the microphone and the listening icon can be swapped out with custom icons. When a custom icon is supplied, it will simply fade in or fade out with CSS animations. This is a little different than the built-in icons which use SVG animations.

In a future PR, the animated dots will be modified to move when listening to sound.

When the listening state is change, the screen reader text is updated to be either "Start Voice Search" or "Stop Voice Search". Translations will be added once we receive them. A focus state is also present for keyboard WCAG support.

J=SLAP-1442
TEST=manual

Load a test site with voice search enabled in the config and see the icon appear. See that the transitions work properly when clicking on the icon. Also try specifying custom mic and custom listening icons and see them properly fade in and out during the click. Observe the screen reader text change when the button is clicked.


src/ui/icons/voice_search_dots.js
```diff
@@ -0,0 +1,95 @@
+import SVGIcon from './icon.js';
+export default new SVGIcon({
+  name: 'voice_search_dots',
+  viewBox: '0 0 33 27',
+  complexContents: `
+    <svg x="16.25" y="13.75" width="0.01" height="0.01" style ="overflow:visible" opacity="1" label="anchorPointContainer">
```
oshi97: I know we don't do this for other icons, but is it possible to import the complexContents from an svg file directly?
like
```js
import VoiceSearchDots from './svg/voice_search_dots.svg'
```
I think this could slightly decrease the bundle size of the svg too, if importing it like this removes whitespace that would otherwise be in the complexContents string?

src/ui/sass/modules/_Icon.scss
```diff
@@ -43,6 +43,56 @@
   }
 }
 
+.yxt-SearchBar-voiceSearch {
```
oshi97: should this go in the searchbar's scss file? or maybe half and half?

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -107,13 +107,13 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
   &-clear
   {
-    padding-left: calc(var(--yxt-base-spacing) / 2);
-    padding-right: calc(var(--yxt-base-spacing) / 2);
+    padding-left: calc(var(--yxt-base-spacing) * 3/4);
```
oshi97, line 110: do we always want this change, or just when somebody has the voice search icon?

src/ui/controllers/voicesearchcontroller.js
```diff
@@ -0,0 +1,135 @@
...
+      });
+    } else if (this._state === State.LISTENING) {
+      this._dotsFadeInAnimations.forEach(animation => {
+        animation.beginElement();
+      });
+    }
+  }
+
+  _applyCustomMicIconStyling () {
+    if (this._state === State.NOT_LISTENING) {
```
oshi97, line 105: would it be weird if we just toggled one css class on the `yxt-SearchBar-voiceIconWrapper` container, instead of toggling these classes on the individual divs?

src/ui/controllers/voicesearchcontroller.js
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+      this._executeDefaultListeningIconAnimations();
+    }
+    this._updateScreenReaderText();
+    this._voiceSearchElement.blur();
+  }
+
+  _executeDefaultMicIconAnimations () {
+    if (this._state === State.NOT_LISTENING) {
+      this._micFadeInAnimation.beginElement();
```
oshi97, line 86: this is cool! did you have to manually write the SVG to use `<animate>` or was UX able to do that all for us?

src/ui/icons/voice_search_dots.js
oshi97: if this works we could also update the other big icons in another item

src/ui/controllers/voicesearchcontroller.js
oshi97: not sure which way would be cleaner overall, on one hand I think just toggling one css class is nice, but the css also has to "know more" in that case

src/ui/icons/voice_search_dots.js
cea2aj: I found a rollup plugin that can handle the import, but it doesn't remove the whitespace. I think it makes sense to go with that since working with an SVG file is nicer than working with the template inline

src/ui/sass/modules/_Icon.scss
cea2aj: I think the serachbar's scss file make the most sense. It's also consistent with the loading indicator work.

src/ui/controllers/voicesearchcontroller.js
cea2aj: The SVGs that UX provided had the the animate tags in them. I just had to add a css class and `begin="indefinate"` so that I could start them with Javascript rather than on page load.

src/ui/controllers/voicesearchcontroller.js
cea2aj: I tried the way you suggested and you're right, the javascript is a little simpler but the styling is a little more complicated. I don't really have a preference as it can really go either way

src/ui/sass/modules/_SearchBar.scss
cea2aj: I think it's simpler to always have this change. Even when voice search is disabled, the visual spacing between the clear button and the search button is the same as before. The only downside is a small decrease in the width of the clickable search button by 4px which I don't think is a problem

package.json
```diff
@@ -142,7 +144,11 @@
       "**/tests/ui/**/*.js",
       "**/tests/conf/**/*.js",
       "**/tests/answers-umd.js"
-    ]
+    ],
+    "transform": {
+      "^.+\\.js$": "babel-jest",
+      ".+\\.(svg)$": "jest-transform-stub"
```
oshi97: just asking for myself but why does the babel-jest regex need a caret? And do we need parenthesis around the "svg" part?

cea2aj: I was following the example from the jest-transform-stub repo, but we don't need the caret so I'll remove it. We don't need the parenthesis either, but we can add those back if we ever want to add more types of assets like this: `".+\\.(css|styl|less|sass|scss|png|jpg|ttf|woff|woff2)$"`

oshi97 (approved):

PR #1474, opened 2021-07-20
add no-unsafe-eval percy snapshot

oshi97:
We cannot use testcafe to testcafe to create a no-unsafe-eval test, because
testcafe internals rely on eval.
Instead, this commit adds a percy snapshot which tests the same thing - that
the page loads even with a CSP that disables unsafe-eval.

Note we only guarantee no unsafe-eval's for the SearchBar component.

J=SLAP-1440
TEST=auto


tests/acceptance/percy/snapshots.mjs
```diff
@@ -0,0 +1,32 @@
+import puppeteer from 'puppeteer';
+import percySnapshot from '@percy/puppeteer';
```
yen-tt, line 2: Is there a difference between using @percy/script (PercyScript setup from theme) vs @percy/puppeteer?

cea2aj (approved):
+1

tests/acceptance/percy/snapshots.mjs
oshi97: they've very recently deprecated percyscript https://docs.percy.io/docs/percyscript
percyscript is basically puppetter with some VERY thing wrapping, so it should be incredibly easy to migrate
I mean just looking at the code there's almost no differences besides the boilerplate

yen-tt (approved):

PR #1475, opened 2021-07-21
Add translations POT file

cea2aj:
Add the POT file to the repo so that we can easily access it

Also update the translation verification script to throw an error if the extracted translations are out of date

NOTE: The translation test is failing because we don't have all the translations in yet.

J=none
TEST=manual

See that when the messages.pot file in the repo is out of date, the test fails

### DIFF
--- a/.circleci/run_translation_verification.sh
+++ b/.circleci/run_translation_verification.sh
@@ -12,8 +12,6 @@ if test $diff_exit_code -eq 1
 then
   echo "Extracted translations are out of date. Run 'npm run extract-translations' and commit the updated pot file."
   exit 1
-else
-  echo "Extracted translations are up to date"
 fi
 
 cd conf/i18n/translations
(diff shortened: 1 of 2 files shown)


.circleci/run_translation_verification.sh
```diff
@@ -1,9 +1,22 @@
 #!/bin/bash
 
+# Verifiy that the messages.pot file in the repo is up to date
+# If there are any git diffs after translations are extracted, the checked-in file is out of date
+
 npm run extract-translations
 
+git diff --exit-code > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
```
oshi97: can we diff just messages.pot?

cea2aj: sure

oshi97 (approved):
~looks great!~ still looks great but the translation test and unsafe-eval snapshot are derping out

cea2aj:
> ~looks great!~ still looks great but the translation test and unsafe-eval snapshot are derping out

the translation test is supposed to be failing because we don't have all the translations in yet

It looks like the unsafe-eval snapshot has been blank since that PR was merged in. (I didn't notice it because it was auto-approved by percy)

oshi97 (approved):

PR #1478, opened 2021-07-23
State Machine for SearchBar Icon

yen-tt:
Move logic around searchBar icons to a state machine system

- Added a general state machine model and state class
- Added a searchBarIconController that holds a state machine and manages all events and behavior surrounding the search bar icons.
- Added 3 states for different behavior in search bar icon (loading, search, and default)
- Removed related search bar icon code from searchBarComponent
- Removed RequestAnimationFrame calls since we are relying on CSS animations

J=SLAP-1437
TEST=manual & auto

- serve page from sdk dist/ and verified icon changes during search, click outside of search bar, type query, and click submit. Verified that custom icons for default and loading states work.
- Added Jest tests for the controller

### DIFF
--- a/src/ui/controllers/searchbariconcontroller.js
+++ b/src/ui/controllers/searchbariconcontroller.js
@@ -35,25 +35,27 @@ export default class SearchBarIconController {
     this.searchBarContainer = config.searchBarContainer;
     this.onMouseUp = this.oneTimeMouseUpListener.bind(this);
 
-    const loadingState = new LoadingIconState(this, IconState.LOADING_ICON);
-    const searchState = new SearchIconState(this, IconState.SEARCH_ICON, config.useCustomIcon);
-    const defaultState = new DefaultIconState(this, IconState.DEFAULT_ICON, config.useCustomIcon);
+    const stateMap = {
+      [IconState.LOADING_ICON]: new LoadingIconState(this, IconState.LOADING_ICON),
+      [IconState.DEFAULT_ICON]: new DefaultIconState(this, IconState.DEFAULT_ICON, config.useCustomIcon),
+      [IconState.SEARCH_ICON]: new SearchIconState(this, IconState.SEARCH_ICON, config.useCustomIcon)
+    };
 
     const possibleTransitions = {
       [IconState.LOADING_ICON]: {
-        SEARCH_COMPLETE: defaultState
+        [IconEvent.SEARCH_COMPLETE]: IconState.DEFAULT_ICON
       },
       [IconState.DEFAULT_ICON]: {
-        SUBMIT: loadingState,
-        FOCUS: searchState
+        [IconEvent.SUBMIT]: IconState.LOADING_ICON,
+        [IconEvent.FOCUS]: IconState.SEARCH_ICON
       },
       [IconState.SEARCH_ICON]: {
-        SUBMIT: loadingState,
-        FOCUS_OUT: defaultState
+        [IconEvent.SUBMIT]: IconState.LOADING_ICON,
+        [IconEvent.FOCUS_OUT]: IconState.DEFAULT_ICON
       }
     };
-    const initialIconState = config.isFocus ? searchState : defaultState;
-    this._fsm = new StateMachine(initialIconState, possibleTransitions);
+    const initialIconState = config.isFocus ? stateMap[IconState.SEARCH_ICON] : stateMap[IconState.DEFAULT_ICON];
+    this._fsm = new StateMachine(initialIconState, stateMap, possibleTransitions);
   }
 
   handleEvent (event, context) {
--- a/src/ui/controllers/searchbariconcontroller.js
+++ b/src/ui/controllers/searchbariconcontroller.js
@@ -7,6 +7,25 @@ import DefaultIconState from '../statemachine/defaulticonstate';
 import { StateMachine } from '../statemachine/statemachine';
 
 /**
+ * Events to trigger state transitions in search bar icon state machine
+ */
+const IconEvent = {
+  SEARCH_COMPLETE: 'SEARCH_COMPLETE',
+  SUBMIT: 'SUBMIT',
+  FOCUS: 'FOCUS',
+  FOCUS_OUT: 'FOCUS_OUT'
+};
+
+/**
+ * Possible states in search bar icon state machine
+ */
+const IconState = {
+  LOADING_ICON: 'LOADING_ICON',
+  SEARCH_ICON: 'SEARCH_ICON',
+  DEFAULT_ICON: 'DEFAULT_ICON'
+};
+
+/**
  * Responsible for controlling the search bar icon behavior
  */
 export default class SearchBarIconController {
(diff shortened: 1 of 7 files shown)


cea2aj:
Looks great! I like this approach a lot

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -0,0 +1,83 @@
...
+
+    const IconState = {
+      LOADING: new LoadingIconState(this),
+      SEARCH: new SearchIconState(this, config.useCustomIcon),
+      DEFAULT: new DefaultIconState(this, config.useCustomIcon)
+    };
+
+    const transitions = {
+      [IconState.LOADING.id]: {
+        SEARCH_COMPLETE: IconState.DEFAULT
```
cea2aj: What do you think of having an enum for the 'SEARCH_COMPELTE', 'SUBMIT', 'FOCUS', and 'FOCUS_OUT' events?

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -0,0 +1,83 @@
...
+ * Responsible for controlling the search bar icon behavior
+ */
+export default class SearchBarIconController {
+  constructor (config) {
+    this.iconIsFrozen = false;
+    this.inputEl = config.inputEl;
+    this.searchBarContainer = config.searchBarContainer;
+    this.onMouseUp = this.oneTimeMouseUpListener.bind(this);
+
+    const IconState = {
```
cea2aj: nit: To me this reads like it's an enum where the values are strings, and that throws me off a little since the values are class instances. Is it possible to make this a regular enum? Each state class could have the enum set as its own id.  Each instance of the state could be passed to the state machine. Then when the state machine transitions it could switch to the correct state instance. That would also let us remove the ids from the transitions object

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -0,0 +1,83 @@
...
+      LOADING: new LoadingIconState(this),
+      SEARCH: new SearchIconState(this, config.useCustomIcon),
+      DEFAULT: new DefaultIconState(this, config.useCustomIcon)
+    };
+
+    const transitions = {
```
cea2aj: Would `allowedTransitions` or `possibleTransitions` be accurate and potentially more descriptive?

src/ui/statemachine/defaulticonstate.js
```diff
@@ -0,0 +1,45 @@
...
+    return !this._controller.iconIsFrozen && !focusStillInSearchbar;
+  }
+
+  onEnter (prevState) {
+    this._useCustomIcon ? this._handleTransitionToCustom() : this._handleTransitionToYext(prevState);
+  }
+
+  _handleTransitionToYext (prevState) {
+    this._defaultIconElement.classList.remove('yxt-SearchBar-Icon--inactive');
+    const svgWrapper = DOM.query(this._defaultIconElement, '.Icon--yext_animated_reverse');
```
cea2aj: Can we move some of these `DOM.query` calls to the constructor? I think it's a little better for performance if we don't make a DOM query on every transition if it's not needed

src/ui/statemachine/statemachine.js
```diff
@@ -0,0 +1,45 @@
+export class State {
```
cea2aj, line 5: Can you add a jsdoc?

src/ui/statemachine/defaulticonstate.js
```diff
@@ -0,0 +1,45 @@
+import { State } from './statemachine';
+import LoadingIconState from './loadingiconstate';
+import DOM from '../dom/dom';
+
+/**
+ * Defines behavior for search bar icon in default state and during state transitions
```
cea2aj: nit: Can you add that by default, this state is the Yext logo?

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -0,0 +1,83 @@
...
+export default class SearchBarIconController {
+  constructor (config) {
+    this.iconIsFrozen = false;
+    this.inputEl = config.inputEl;
+    this.searchBarContainer = config.searchBarContainer;
+    this.onMouseUp = this.oneTimeMouseUpListener.bind(this);
+
+    const IconState = {
+      LOADING: new LoadingIconState(this),
+      SEARCH: new SearchIconState(this, config.useCustomIcon),
```
cea2aj: thoughts on adding `_ICON` to these so that they match the class names?

tests/ui/components/controllers/searchbariconcontroller.js
```diff
@@ -0,0 +1,163 @@
...
+    const defaultIcon = wrapper.find('.js-yxt-AnimatedReverse');
+    const searchIcon = wrapper.find('.js-yxt-AnimatedForward');
+    const defaultSVGWrapper = wrapper.find('.Icon--yext_animated_reverse');
+    const searchSVGWrapper = wrapper.find('.Icon--yext_animated_forward');
+
+    expect(defaultSVGWrapper.getDOMNode().classList.contains('yxt-SearchBar-Yext--static')).toBeTruthy();
+    expect(defaultIcon.getDOMNode().classList.contains('yxt-SearchBar-Icon--inactive')).toBeFalsy();
+
+    controller.handleEvent('FOCUS');
+    expect(defaultIcon.getDOMNode().classList.contains('yxt-SearchBar-Icon--inactive')).toBeTruthy();
```
cea2aj: Can we move some of these selectors which are used a lot into string variables?

src/ui/controllers/searchbariconcontroller.js
yen-tt: I see what you mean. I will have the enum store the id and instantiate the states separately in the update.

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -16,27 +35,25 @@ export default class SearchBarIconController {
...
+    const loadingState = new LoadingIconState(this, IconState.LOADING_ICON);
+    const searchState = new SearchIconState(this, IconState.SEARCH_ICON, config.useCustomIcon);
+    const defaultState = new DefaultIconState(this, IconState.DEFAULT_ICON, config.useCustomIcon);
 
-    const transitions = {
-      [IconState.LOADING.id]: {
-        SEARCH_COMPLETE: IconState.DEFAULT
+    const possibleTransitions = {
+      [IconState.LOADING_ICON]: {
+        SEARCH_COMPLETE: defaultState
```
cea2aj: do you know if it's possible to use the enum value as the key here?

src/ui/controllers/searchbariconcontroller.js
```diff
@@ -16,27 +35,25 @@ export default class SearchBarIconController {
...
       },
-      [IconState.DEFAULT.id]: {
-        SUBMIT: IconState.LOADING,
-        FOCUS: IconState.SEARCH
+      [IconState.DEFAULT_ICON]: {
+        SUBMIT: loadingState,
```
cea2aj: nit: can we also use the enum for these values as well? (e.g. we would have IconState.LOADING_ICON rather than loadingState) I believe we would need to pass a map to the state machine which maps the enums to the state instances

cea2aj (approved):

PR #1480, opened 2021-07-23
Speech Recognition Functionality

cea2aj:
Add speech recognition functionality to the searchbar

This PR includes:
- Connecting the speech recognizer to the voice search button
- Adding analytics for clicking on the voice search button which includes not sending the queryId for the voice search analytics
- Checking that `navigator.mediaDevices` exists so that the search bar will hide on Edge without HTTPs
- Update the autocomplete component to know when voice search is active. This allows the autocomplete dropdown to appear while voice search is active
- Add `submitVoiceSearch()` to the search bar so that a search can be made with the VOICE_SEARCH query trigger

J=SLAP-1444
TEST=manual

Test speech recognition on Chrome, Safari, and Edge. Firefox is not supported. Note: The site must be on HTTPS for Edge to work.


src/core/speechrecognition/support.js
```diff
@@ -9,5 +9,5 @@
  * @returns {boolean}
  */
 export function speechRecognitionIsSupported () {
-  return !!(window.SpeechRecognition || window.webkitSpeechRecognition);
+  return !!(window.webkitSpeechRecognition && navigator.mediaDevices);
```
yen-tt, line 12: we don't need to check for window.SpeechRecognition anymore?

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -233,11 +233,18 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     }
   }
 
+  &-voiceIconWrapper
+  {
+    width: 32px;
```
yen-tt: Icon scss file have some width and height styling (sm and lg). Can we not use that for voice search icons or is it necessary to have a specific width here?

src/ui/speechrecognition/speechrecognizer.js
```diff
@@ -0,0 +1,146 @@
...
+    } catch (err) {
+      console.warn(err);
+    }
+  }
+
+  /**
+   * Stops the speech recognizer if it is active
+   */
+  stop () {
+    this._recognitionActive && this._speechRecognition.stop();
```
yen-tt, line 120: Should this._recognitionActive be set to false?

yen-tt: oh wait nvm, I guess this._speechRecognition.stop() trigger an 'end' event that would set this to false.

src/core/speechrecognition/support.js
cea2aj: On the browsers that we support (Edge, Chrome and Safari) the api is available at window.webkitSpeechRecognition and not at window.SpeechRecognition, so I think we're fine to remove that. @oshi97 Are there any reasons that you can think of for leaving it in?

src/ui/sass/modules/_SearchBar.scss
cea2aj: oh, you're right. i just tried that and that works better. Updating!

src/ui/speechrecognition/speechrecognizer.js
cea2aj: Correct, I found it simplest to just have the 'start' and 'end' events toggle `this._recognitionActive'

src/core/speechrecognition/support.js
oshi97: I think in the future we might need to check for SpeechRecognition since that's the official API spec, so I would probably leave it in. On the other hand, if something changes about the API when it moves from webkitSpeechRecognition to just SpeechRecognition we will need to make changes anyways. I'm fine either way

src/ui/components/search/searchcomponent.js
```diff
@@ -789,20 +796,26 @@ export default class SearchComponent extends Component {
     return DOM.query(this._container, '.js-yxt-SearchBar-clear');
   }
 
+  hideClearButton () {
+    this._showClearButton = false;
```
oshi97, line 800: how were we hiding the clear button before?

src/ui/speechrecognition/speechrecognizer.js
```diff
@@ -0,0 +1,146 @@
...
+      this.stop();
+    }
+  }
+
+  /**
+   * Handles a speech recognition error
+   * @param {SpeechRecognitionError} event
+   */
+  _handleErrorEvent (event) {
+    if (event.error === 'service-not-allowed') {
```
oshi97, line 94: did you find out whether this error code ever happens for other reasons?

src/ui/speechrecognition/speechrecognizer.js
```diff
@@ -0,0 +1,146 @@
+/* eslint-disable new-cap */
+import { isMicrosoftEdge } from '../../core/utils/useragent';
+import { transformSpeechRecognitionLocaleForEdge } from '../../core/speechrecognition/locales';
+import TranslationFlagger from '../i18n/translationflagger';
+
+/**
+ * Responsible for recognizing speech
+ */
+export default class SpeechRecognizer {
```
oshi97, line 9: this class is pretty Pog

src/core/speechrecognition/support.js
cea2aj: I'd prefer that we manually verify that the official `SpeechRecognition` API works the same rather than having us automatically start using that API once it's available so I think we're good to leave as-is

src/ui/speechrecognition/speechrecognizer.js
cea2aj: Ya, it can happen for a variety of reasons including "the user agent doesn't support it or because of reasons of security, privacy or user preference". Do you think we should include more of that info in our error message? My intent here was to have a general message of 'Speech Recognition is not available.' and also include a message in case the problem was Siri not being enabled

src/core/analytics/analyticsreporter.js
```diff
@@ -99,7 +99,14 @@ export default class AnalyticsReporter {
       throw new AnswersAnalyticsError('Tried to send invalid analytics event', event);
     }
 
-    event.addOptions(this._globalOptions);
+    if (includeQueryId) {
+      event.addOptions(this._globalOptions);
```
oshi97, line 109: is this needed for when a user specifies a queryId in ANSWERS.init's config.analyticsOptions?

cea2aj: By default we send the last queryId with every analytics event. For voice search, we shouldn't send one so I added this so that we could send the analytics events without one

src/ui/speechrecognition/speechrecognizer.js
oshi97: ah ok, maybe we could update it slightly with something like `Note: Siri must be ---`? Otherwise to me it implies that the reason voice search isn't working is because of siri, as opposed to a "make sure to check" kind of thing.

src/core/analytics/analyticsreporter.js
oshi97: For fun I was wondering when this was added https://github.com/yext/answers-search-ui/pull/196/files
I was a little worried this would encourage people to add more flags in the future, but eh I can't think of a better way than what you have

src/ui/speechrecognition/speechrecognizer.js
oshi97: I was thinking we *could* try sniffing the browser but probably not worth it just for this?

cea2aj: I can update the message! I try to avoid browser sniffing when possible

src/core/analytics/analyticsreporter.js
cea2aj: Would a `reportWithoutQueryId` function work better than a flag?

oshi97: yeah I like that more!

cea2aj: I actually just updated the function to have an optional `options` object as the second parameter which is clearer than the lone boolean flag and it more easily lets us add additional functionality. What do you think of that approach?

oshi97 (approved):
lgtm!

PR #1485, opened 2021-07-27
Don't require translation verification on develop and dev branches

cea2aj:
Update the config so that the translation verification is no longer required in order to deploy assets for develop and dev branches

We still run the test so that we have visibility into whether or not translations are present, but they shouldn't prevent a deploy

J=SLAP-1476
TEST=manual

See that this PR deploys despite the translation verification failing. Check the Circle CI workflow graph and see that the translation verification runs but is not required for a deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,7 +211,6 @@ workflows:
             - unit_test
             - browserstack_acceptance_test
             - headless_acceptance_test
-            - translation_test
             - third_party_notices_test
   build_and_deploy_i18n:
     jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -211,7 +211,6 @@ workflows:
             - unit_test
             - browserstack_acceptance_test
             - headless_acceptance_test
-            - translation_test
```
oshi97, line 214: I think in the future we might run the portion of translation_test that verifies the .pot file to run on all commits. We could probably use the [add-and-commit github action](https://github.com/EndBug/add-and-commit) that Vaccine is using for their third party licenses generation

oshi97 (approved):

.circleci/config.yml
cea2aj: This PR does still run the translation_test. It's just doesn't block the deployment if it fails

PR #1486, opened 2021-07-27
Speech recognition integration test

cea2aj:
Add some integration testing that checks that clicking on the icon starts and stops the speech recognition

Due to the async nature of the code, I had to tweak the speech recognition code so that we can wait for the code to execute before continuing the test.

J=none
TEST=auto

Confirm tests pass


src/ui/speechrecognition/voicesearchcontroller.js
```diff
@@ -66,6 +69,7 @@ export default class VoiceSearchController {
       });
     } else if (this._state === State.LISTENING) {
       this._enterNotListeningState();
+      return Promise.resolve();
```
oshi97: This technically "could" return undefined instead of a promise if this._state somehow isn't either of the 2 states.
Could we move return Promise.resolve() outside the statements? We could also use async await though I kind of like traditional promises

tests/ui/components/search/searchcomponent.js
```diff
@@ -132,15 +135,86 @@ describe('SearchBar component', () => {
...
+        mediaDevices: {
+          getUserMedia: () => Promise.resolve({
+            getTracks: () => ([{ stop: jest.fn() }])
+          })
+        }
+      }));
+    });
+
+    afterEach(() => {
+      windowSpy.mockRestore();
```
oshi97: does `jest.restoreAllMocks` work here?

tests/ui/components/search/searchcomponent.js
```diff
@@ -132,15 +135,86 @@ describe('SearchBar component', () => {
...
+      const component = COMPONENT_MANAGER.create('SearchBar', {
+        ...defaultConfig,
+        voiceSearch: {
+          enabled: true
+        }
+      });
+      mount(component);
+      await component._voiceSearchController.handleIconClick();
+      await component._voiceSearchController.handleIconClick();
+      expect(webkitSpeechRecognitionMock.start).toHaveBeenCalledTimes(1);
```
oshi97, line 201: ~Could we switch lines 215 and 216? Maybe even throw in a `expect(webkitSpeechRecognitionMock.stop).toHaveBeenCalledTimes(0);`~ nevermind you cover that in the above test case. Would it make sense to combine these two cases?

src/ui/speechrecognition/voicesearchcontroller.js
cea2aj: true. updated!

tests/ui/components/search/searchcomponent.js
cea2aj: A combo of restoreAllMocks and clearAllMocks did the trick

tests/ui/components/search/searchcomponent.js
cea2aj: Sure, I just combined these into one test

tests/ui/components/search/searchcomponent.js
```diff
@@ -132,15 +135,71 @@ describe('SearchBar component', () => {
...
+      start: jest.fn(() => {
+        'start' in events && events.start();
+      }),
+      stop: jest.fn(() => {
+        'stop' in events && events.stop();
+      })
+    };
+
+    beforeEach(() => {
+      jest.spyOn(global, 'window', 'get').mockImplementation(() => ({
```
cjiang2000, line 153: is the purpose of jest.spyOn purely for setting a mockImplementation? and if so, why not just jest.fn()?

package.json
```diff
@@ -112,6 +112,7 @@
     "docs": "jsdoc -R README.md -d docs/ -r src/",
     "lint": "eslint .",
     "test": "eslint . && stylelint src/**/*.scss && jest",
+    "jest": "jest",
```
oshi97: this is just a passthrough to npx right?

tests/ui/components/search/searchcomponent.js
cea2aj: I tried putting a mock function on the window and I wasn't able to get it working, however I was able to get the spy working so I went with that

package.json
cea2aj: I believe it uses the jest that's installed locally. It's using the same one as the test command. I just found it quicker for testing to not have to run the linter every time

oshi97: I think running `npx jest` will do the exact same thing though (I also like to skip the linter!)

cea2aj: oh, I see what you mean. we don't really need this because I can just use npx instead. I'll remove!

cea2aj: removed!

oshi97 (approved):

PR #1488, opened 2021-07-27
Update translation check

cea2aj:
Update the translation check to correspond with the changes to the hitchhikers theme's check

J=none
TEST=manual

Confirm that the test fails when the messages.pot file is out of date or if translations are missing


cjiang2000 (approved):
lgtm

PR #1490, opened 2021-07-29
Voice search async fix

cea2aj:
Fix a bug that caused voice search to break due to the async keyword

The keyword is not necessary, and it causes a 'regeneratorRuntime' error in browsers. Removing it fixes the issue

J=none
TEST=manual

Smoke test voice search


yen-tt (commented):
Just curious, would there be any issues if someone click on this multiple times? would that mess with the transition process going between entering listening and not listening state?

cea2aj:
> Just curious, would there be any issues if someone click on this multiple times? would that mess with the transition process going between entering listening and not listening state?

I've testing trying to click the voice search button many times very quickly and I can't get it to break. It was a problem before, but I was able to fix that with some updates to the SearchRecognizer class.

yen-tt (approved):

PR #1492, opened 2021-07-29
RTL language support in map

yen-tt:
- add mapbox-gl-rtl-text plugin to support right to left languages

J=SLAP-1459
TEST=manual

from feature/arabic-i18n branch, built ar-* bundle and use it in theme to spin up ar map page (locations.html), and see that arabic is displayed properly. see that en and es is displayed as normal.

### DIFF
--- a/src/ui/components/map/providers/mapboxmapprovider.js
+++ b/src/ui/components/map/providers/mapboxmapprovider.js
@@ -33,7 +33,11 @@ export default class MapBoxMapProvider extends MapProvider {
       onload: () => {
         this._isLoaded = true;
         mapboxgl.accessToken = this._apiKey;
-
+        mapboxgl.setRTLTextPlugin(
+          'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js',
+          null,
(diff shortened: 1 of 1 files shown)


src/ui/components/map/providers/mapboxmapprovider.js
```diff
@@ -33,7 +33,11 @@ export default class MapBoxMapProvider extends MapProvider {
       onload: () => {
         this._isLoaded = true;
         mapboxgl.accessToken = this._apiKey;
-
+        mapboxgl.setRTLTextPlugin(
+          'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js',
```
oshi97, line 37: Could we use https://www.npmjs.com/package/@mapbox/mapbox-gl-rtl-text?activeTab=dependents instead? If we're trying to lazy load it I think we could try something like https://rollupjs.org/guide/en/#dynamic-import

oshi97: actually eh probably not worth the hassle for the SDK

oshi97 (approved):

cea2aj (approved):

PR #1493, opened 2021-07-30
Add RTL CSS bundle

cea2aj:
Add an RTL CSS bundle

The RTL bundle does a few things including putting the filter checkboxes on the right side. We probably don't need to build the RTL bundle unless we are building for all languages or locales, however it is so quick that I think it makes sense to just build it during dev mode since it makes developing RTL styling much faster.

J=SLAP-1465
TEST=manual

Build a site with the RTL styling and see the filter checkboxes on the right side

### DIFF
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -2,6 +2,8 @@ const { series, parallel, src, dest, watch } = require('gulp');
 const path = require('path');
 const postcss = require('gulp-postcss');
 const sass = require('gulp-sass');
+const rtlcss = require('gulp-rtlcss');
+const rename = require('gulp-rename');
 sass.compiler = require('sass');
 
 const getLibraryVersion = require('./utils/libversion');
@@ -111,11 +113,34 @@ async function createBundleTaskFactory (locale) {
 }
 
 function compileCSS () {
+  return new Promise(resolve => {
+    return parallel(compileLTRCSS, compileRTLCSS)(resolve);
+  });
+}
+
+/**
+ * Compiles the standard left to right CSS
+ */
+function compileLTRCSS () {
   return src('./src/ui/sass/**/*.scss')
     .pipe(sass({
       outputStyle: 'compressed'
     }).on('error', sass.logError))
     .pipe(postcss())
+    .pipe(dest('./dist/'));
+}
+
+/**
+ * Compiles the right to left CSS which is used for languages such as Arabic
+ */
+function compileRTLCSS () {
+  return src('./src/ui/sass/**/*.scss')
+    .pipe(sass({
+      outputStyle: 'compressed'
+    }).on('error', sass.logError))
+    .pipe(rtlcss())
+    .pipe(postcss())
+    .pipe(rename('answers.rtl.css'))
     .pipe(dest('./dist/'));
 }
 
--- a/conf/gulp-tasks/library.gulpfile.js
+++ b/conf/gulp-tasks/library.gulpfile.js
@@ -2,7 +2,7 @@ const { series, parallel, src, dest, watch } = require('gulp');
 const path = require('path');
 const postcss = require('gulp-postcss');
 const sass = require('gulp-sass');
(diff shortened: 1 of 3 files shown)


conf/gulp-tasks/library.gulpfile.js
```diff
@@ -111,11 +113,34 @@ async function createBundleTaskFactory (locale) {
...
+      outputStyle: 'compressed'
+    }).on('error', sass.logError))
+    .pipe(postcss())
+    .pipe(dest('./dist/'));
+}
+
+/**
+ * Compiles the right to left CSS which is used for languages such as Arabic
+ */
+function compileRTLCSS () {
```
yen-tt: not sure if it would make much difference in efficiency if we are doing this in parallel but what do you think about making the rtl bundle based off the already compiled ltr css so we don't have to go through the sass compilation and compression again?

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -111,11 +113,34 @@ async function createBundleTaskFactory (locale) {
...
   return src('./src/ui/sass/**/*.scss')
     .pipe(sass({
       outputStyle: 'compressed'
     }).on('error', sass.logError))
+    .pipe(rtlcss())
     .pipe(postcss())
+    .pipe(rename('answers.rtl.css'))
```
yen-tt: I guess it will always be answers.css as the name by default for ltr version?

conf/gulp-tasks/library.gulpfile.js
cea2aj: that might be possible. I'll try it out

conf/gulp-tasks/library.gulpfile.js
cea2aj: yes, the name is always answers.css for the ltr version

conf/gulp-tasks/library.gulpfile.js
cea2aj: So this does work for the most part, however it doesn't preserve the CSS comments properly that we need to be able to specify custom rtl behavior such as flipping the chevrons. Also we should run the postcss plugin after rtlcss, so at this point I say we keep them as parallel builds

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -111,11 +113,34 @@ async function createBundleTaskFactory (locale) {
...
 function compileCSS () {
+  return new Promise(resolve => {
+    return parallel(compileLTRCSS, compileRTLCSS)(resolve);
+  });
+}
+
+/**
+ * Compiles the standard left to right CSS
+ */
+function compileLTRCSS () {
```
oshi97: Maybe LTR and RTL shouldn't be all caps? since right now `LTRCSS` is all mushed together as one all-caps thing

conf/gulp-tasks/library.gulpfile.js
```diff
@@ -111,11 +113,34 @@ async function createBundleTaskFactory (locale) {
...
+
+/**
+ * Compiles the right to left CSS which is used for languages such as Arabic
+ */
+function compileRTLCSS () {
   return src('./src/ui/sass/**/*.scss')
     .pipe(sass({
       outputStyle: 'compressed'
     }).on('error', sass.logError))
+    .pipe(rtlcss())
```
oshi97, line 141: (github won't let me start a new conversation on the same line as an already resolved one)
is there an easy way to pass in the file name? in webpack it would be something like `[filename].rtx.[ext]`

conf/gulp-tasks/library.gulpfile.js
cea2aj: sure, i'll update to `compileLtrCss` and `compileRtlCss`

conf/gulp-tasks/library.gulpfile.js
cea2aj: yep, updated to append '.rtl' to the basename

oshi97 (approved):

PR #1494, opened 2021-07-30
Flip chevrons on RTL

cea2aj:
Flip chevrons for RTL

J=SLAP-1460
TEST=SLAP-1460

See that chevrons appear are reversed when using the rtl bundle

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -304,7 +304,7 @@ MIT License
 
 The following NPM packages may be included in this product:
 
- - @yext/answers-core@1.3.0-beta.2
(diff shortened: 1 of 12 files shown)


yen-tt (approved):

oshi97 (approved):

oshi97:
oh could you double check that the percy snapshot is ok and not just something spurious?

cea2aj:
> oh could you double check that the percy snapshot is ok and not just something spurious?

hmm, that's odd. I'll take a look. Update: It looks fine locally. I'l re-run the snapshots. 2nd update: It looks like it was something spurious

PR #1501, opened 2021-08-03
Create multi entry points for SDK bundles

yen-tt:
- Add a script to generate the exports field in package.json that specified import paths to css, js, and template bundles for different language. This will execute in build step.

J=SLAP-1422
TEST=manual

Create jest test to see that exports field is created as expected
import answers-search-ui locally in another repo. Tried import statements for different bundles and verified no NoModuleFound error. Bundled with webpack, serve page and console.log Answers


package.json
```diff
@@ -164,5 +165,33 @@
...
+    "./esm.min": "./dist/answers-modern.min.js",
+    "./template": "./dist/answerstemplates.compiled.min.js",
+    "./es": "./dist/es-answers-umd.js",
+    "./es/esm": "./dist/es-answers-modern.js",
+    "./es/esm.min": "./dist/es-answers-modern.min.js",
+    "./es/template": "./dist/es-answerstemplates.compiled.min.js",
+    "./fr": "./dist/fr-answers-umd.js",
+    "./fr/esm": "./dist/fr-answers-modern.js",
+    "./fr/esm.min": "./dist/fr-answers-modern.min.js",
+    "./fr/template": "./dist/fr-answerstemplates.compiled.min.js",
```
oshi97, line 183: What do the import paths look like with these exports? Would it be possible to just rely on the natural filepath from the dist folder instead of having a custom path for every asset?

yen-tt: It's possible to use the natural filepaths. After the SDK multilang bundle investigation meeting, it was decided we should provide custom entry points to make it easier to access the multilang bundles

oshi97: ah got it, would it be easier to just reorganize the dist folder before an npm publish? Also what is the "exports" field? I couldn't find it in https://docs.npmjs.com/cli/v7/configuring-npm/package-json#exports 
I do see it [here](https://nodejs.org/api/packages.html#packages_exports) though, would people need to be on a newer version of node/npm to use exports?

yen-tt: hmm you mean like reorganize dist folder for different languages without the custom entry points? 
Yea, from Node.js 12+, otherwise it just ignores it and use `main` field as normal

oshi97: Yeah I was thinking it might let us avoid adding a huge chunk to the package.json, but I think what you have might be better!
Would it be hard to add test coverage for this? Also is this something that we would only run as part of the prepublishOnly script, which currently just swaps readmes, or would it be always be checked into the package.json? If the latter I think we want to add some kind of verification test like with third party notices, maybe we can try https://github.com/EndBug/add-and-commit here

yen-tt: @tmeyer2115 what's your thought on just reorganizing the dist folder?
I can add test to check if it generates the expected export field. Export field would need update if we ever add new languages to SDK (like arabic soon) or update script if there's new bundle (like rtl-css). Probably just in prepublishOnly should be ok?

oshi97: oh just to clarify I'm cool with having this exports field in the package.json! Was just throwing out an alternative and wondering if it would stick or not.

cea2aj: That's an interesting idea. On NPM I don't think we need the source code, so maybe it makes sense to not even have a dist folder on NPM and we instead move everything from dist into the top level

conf/npm/generateentrypoints.js
```diff
@@ -0,0 +1,72 @@
...
+  Object.assign(exportJSON.exports, cssEntries);
+}
+
+/**
+ * Add entry points to js and template bundles to the export field for default language en
+ *
+ * @param {Object} exportJSON
+ * @param {string} lang
+ */
+function addDefaultLanguageBundleEntries (exportJSON) {
```
cea2aj: I'm not sure that we need a function for this and for the CSS. This could be an object and in appendExportsField we use the spread operator to add to the exports

conf/npm/generateentrypoints.js
```diff
@@ -0,0 +1,72 @@
...
+  Object.assign(exportJSON.exports, bundleEntries);
+}
+
+/**
+ * Add entry points to js and template bundles to the export field for the provided language
+ *
+ * @param {Object} exportJSON
+ * @param {string} lang
+ */
+function addLanguageBundleEntries (exportJSON, lang) {
```
cea2aj: nit: Could this be a getter instead which just takes in a language and returns an object? We could use the spread operator to add the object to the exports. That would let us avoid having this function need to know about the shape of exportJSON

conf/npm/generateentrypoints.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Append the exports field in the package.json. This defines a public interface for the package,
+ * mapping entry points (path from import requests) to bundles. Any other paths not specified
+ * in exports field will lead to a ModuleNotFound Error or ERR_PACKAGE_PATH_NOT_EXPORTED Error.
+ *
+ * @param {string} inputFile name of file to append exports field to
+ */
+function appendExportsField (inputFile) {
+  const packageData = JSON.parse(fs.readFileSync(inputFile, 'utf8'));
+  const exportJSON = { exports: {} };
```
cea2aj: would it be simpler to just have an exports object like `const exports = {}`, and then add it to `packageData` like: `Object.assign(packageData, { exports: exports })`?

yen-tt: ah yeah, make sense

conf/npm/prepublishOnly.sh
```diff
@@ -4,3 +4,4 @@
 
 mv ./README.md ./README.tmp
 cp ./conf/npm/README.md ./README.md
+npm run generate-entry-points
```
oshi97: if we're running this as part of the prepublish, should we still check the exports into the package.json? like I think we should decide whether exports are **always** part of the package.json, or only added in as part of the publish. I'm fine with having them always checked in if that's simpler!

cea2aj (approved):
LGTM so long as product wants to go this route (as opposed to reorganizing the dist folder as oshi suggested)

conf/npm/prepublishOnly.sh
cea2aj: I think keeping it always checked in is simpler. Maybe it would make sense to run the generation as part of the build script though?

oshi97: I was a little worried about running generate-entry-points blindly as part of the prepublish, since npm doesn't let you undo a publish. On the other hand we can just publish a patch version if something goes wrong, though it's a little annoying.

oshi97 (approved):

cea2aj (changes requested):
After thinking a little more about Oshi's comment, I think it would be a little safer for us generate the file at a different spot (maybe the build step) so that we don't risk breaking it right before the npm publish

conf/npm/prepublishOnly.sh
yen-tt: oh yeah, generating it in build step would be better. Will update!

cea2aj (approved):

PR #1502, opened 2021-08-03
Disable loading icon when inactive

cea2aj:
Disable the loading icon when it is inactive in order to prevent the icon from flashing during transitions

Note: The search bar will still sometimes flash the very first time the icon is clicked because of a long tasks associated with creating the components and waiting for the autocomplete response, but that's a separate issue

J=SLAP-1473
TEST=manual

See that when the loading indicator is enabled, the icon will no longer flash when clicking on and off the search bar.

### DIFF
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -216,6 +216,11 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     transform-origin: 50% 50%;
     animation: dash 1.4s ease-in-out infinite;
   }
+
+  &-Icon--inactive &-LoadingIndicator-AnimatedIcon 
+  {
+    display: none;
+  }
   
   @keyframes dash 
   {


src/ui/sass/modules/_SearchBar.scss
```diff
@@ -216,6 +216,11 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     transform-origin: 50% 50%;
     animation: dash 1.4s ease-in-out infinite;
   }
+
+  &-Icon--inactive &-LoadingIndicator-AnimatedIcon
```
yen-tt, line 220: I think if you just remove line [187-192](https://github.com/yext/answers-search-ui/pull/1502/files#diff-dd8e4dd237e4b3042803a44c5d46b4ee3cebe20b8199ef3b3c0e1e9a31ae81ecL187-L192) (which was used to keep the animation running while not on display), the display: none should work on the loading indicator without additional scss

cea2aj: If we keep the PR as-is, the starting position of the animation will actually change as if the animation was running the whole time. If we delete those lines then the loading icon will always start from the same spot

yen-tt (approved):

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -216,6 +216,11 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     transform-origin: 50% 50%;
     animation: dash 1.4s ease-in-out infinite;
   }
+
```
oshi97, line 219: So unless I'm misunderstanding something, this was only a problem when somebody was using the searchbar without the loading indicator turned on?

cea2aj: The problem occurs when the icon transitions while the loading indicator animation is running. Since the loading indicator was always running in the background, this was always happening. By setting the animated icon to `display: none` while it is inactive, it is no longer possible for the loading indicator to be running while the icon is transitioning from the magnifying glass to the yext logo

PR #1503, opened 2021-08-03
Don't run dots animation when hidden

cea2aj:
Don't run the dots animation when voice search is not active. This improves performance and prevents icon flickering when clicking on and off of the search bar

I also removed some SVG animations which weren't doing anything because all of the values were set to -0

J=SLAP-1473
TEST=manual

Enable voice search and see that the yext logo icon doesn't flicker when clicking on and off of the icon. Test clicking the voice search icon really fast and see that I'm unable to get the icon animations into a broken state.


src/ui/speechrecognition/listeningiconstylist.js
```diff
@@ -28,13 +32,11 @@ export default class ListeningIconStylist {
       this._dotsFadeInAnimations.forEach(animation => {
         animation.beginElement();
       });
-      if (this._hasDotsAnimationStarted) {
-        return;
-      }
+      // Clear the timeout to prevent it from stopping the animation
+      clearTimeout(this._stopListeningAnimationTimeout);
```
oshi97: would we want to give this a quick null guard? from googling it sounds like clearTimeout(null) is fine but haven't looked too deep into it

cea2aj: it should be okay, but i can add them anyway

oshi97 (approved):

PR #1504, opened 2021-08-03
Set fixed custom icon size

cea2aj:
Set an explicit custom icon width so that the voice and clear button don't shift around when the loading indicator appears

J=SLAP-1473
TEST=manual

Enable the loading indicator and set a custom icon. See that the voice icon and the clear button stay in the same spot

### DIFF
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -264,6 +264,12 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
       position: relative;
     }
   }
+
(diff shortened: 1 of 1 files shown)


oshi97 (approved):
think we generally try to use px instead of em or rem but I wouldn't be able to tell you why, and the SDK uses em and rem directly in a number of places anyways

cea2aj:
> think we generally try to use px instead of em or rem but I wouldn't be able to tell you why, and the SDK uses em and rem directly in a number of places anyways

makes sense. I was going with em here because that's the same unit used by other icon styling, but i'll keep that in mind. We may want to have a discussion about pixels vs em vs rem for theme 2.0. I know product likes rem because it scales well when zooming in

PR #1505, opened 2021-08-04
Display email validation error in QuestionSubmission

yen-tt:
- surface email error from backend with code 22 and display corresponding error message in hbs template
- update questionSubmission URLs to use liveapi version
- minor fix: added optional chaining in case analyticsReporter is not set

J=SLAP-1020
TEST=manual

- serve page from dist folder with QASubmission component, submit with invalid email domain and see that the error message is displayed.

### DIFF
--- a/src/ui/components/questions/questionsubmissioncomponent.js
+++ b/src/ui/components/questions/questionsubmissioncomponent.js
@@ -169,6 +169,15 @@ const DEFAULT_CONFIG = {
   }),
 
   /**
+   * The default email error text, shown when there is an email is invalid with the QA Submission
+   * request.
+   * @type {string}
+   */
+  emailInvalidErrorText: TranslationFlagger.flag({
+    phrase: 'Your email address was not valid. Please refresh the page and try submitting your question again.'
+  }),
+
+  /**
    * Whether or not this component is expanded by default.
    * @type {boolean}
    */
@@ -293,7 +302,7 @@ export default class QuestionSubmissionComponent extends Component {
 
     const questionText = DOM.query(formEl, '.js-question-text');
     DOM.on(questionText, 'focus', () => {
-      this.analyticsReporter.report(this.getAnalyticsEvent('QUESTION_FOCUS'));
+      this.analyticsReporter?.report(this.getAnalyticsEvent('QUESTION_FOCUS'));
     });
   }
 
@@ -305,7 +314,7 @@ export default class QuestionSubmissionComponent extends Component {
   bindFormSubmit (formEl) {
     DOM.on(formEl, 'submit', (e) => {
       e.preventDefault();
-      this.analyticsReporter.report(this.getAnalyticsEvent('QUESTION_SUBMIT'));
+      this.analyticsReporter?.report(this.getAnalyticsEvent('QUESTION_SUBMIT'));
 
       // TODO(billy) we probably want to disable the form from being submitted twice
       const errors = this.validate(formEl);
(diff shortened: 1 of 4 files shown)


src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -168,6 +168,15 @@ const DEFAULT_CONFIG = {
     phrase: 'We\'re sorry, an error occurred.'
   }),
 
+  /**
+   * The default email error text, shown when there is an email is invalid with the QA Submission
+   * request.
+   * @type {string}
+   */
+  emailInvalidErrorText: TranslationFlagger.flag({
+    phrase: 'Your email address was not valid. Please refresh the page and try submitting your question again.'
```
cea2aj: nit: what do you think about changing that first sentence to: "Invalid email address"?

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -323,10 +332,12 @@ export default class QuestionSubmissionComponent extends Component {
         questionDescription: formData.questionDescription
       })
         .catch(error => {
+          let errorMessage = this._config.networkErrorText;
```
cea2aj: do you think it makes sense to rename networkErrorText to defaultErrorText?

src/ui/components/questions/questionsubmissioncomponent.js
```diff
@@ -323,10 +332,12 @@ export default class QuestionSubmissionComponent extends Component {
...
+          if (error.code === 22) {
+            errorMessage = this._config.emailInvalidErrorText;
+          }
           this.setState(
-            new QuestionSubmission(formData, {
-              network: 'We\'re sorry, an error occurred.'
-            })
+            new QuestionSubmission(formData, { network: errorMessage })
```
cea2aj, line 340: Setting network to errorMessage seems a little odd to me. How about `{ errorMessage: errorMessage }` instead?

src/ui/templates/questions/questionsubmission.hbs
```diff
@@ -38,7 +38,7 @@
         {{#if errors.network}}
           <div class="yxt-QuestionSubmission-formRow">
             <div class="yxt-QuestionSubmission-formError" role="alert">
-              {{_config.networkErrorText}}
+              {{errors.network}}
```
cea2aj, line 41: Because errors is plural here, it suggests to me that this is an array. I'm not sure if changing the QuestionSubmission model would have other effects if we changed it to just `error`

src/ui/components/questions/questionsubmissioncomponent.js
yen-tt: I think keeping the error name as network will be a bit more specific and can differentiate the backend network errors from the error check we have in the frontend for QuestionSubmission (nameError, questionPolicyError, questionTextError, etc.)

src/ui/templates/questions/questionsubmission.hbs
yen-tt: I think it's ok to keep it as `errors`, our validate function in questionsubmissioncomponent uses this field as well and it's possible we can more than one error store in there

src/ui/components/questions/questionsubmissioncomponent.js
yen-tt: that would work too

src/ui/components/questions/questionsubmissioncomponent.js
yen-tt: that sounds fine too

cea2aj (approved):

cea2aj:
Solves https://github.com/yext/answers-search-ui/issues/570

PR #1508, opened 2021-08-13
Automate update to third party notice file

yen-tt:
- create github action to add commit with updated 3rd party notices if it's not up to date
- remove the check from circleci

J=SLAP-1505
TEST=auto

push a package.json with version different from what's listed in the 3rd party notice file and see that a automated commit is made in this pr


.github/workflows/third_party_notice_check.yml
```diff
@@ -0,0 +1,30 @@
...
+  license-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - name: Install generate-license-file
+        run: npm install -g generate-license-file
+      - uses: actions/checkout@v2
+      - name: Npm install
```
oshi97: we probably don't need a name for the steps where the name is just the literal command being run

.github/workflows/third_party_notice_check.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: npm install
```
oshi97: can we use npm ci?

.github/workflows/third_party_notice_check.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Check if there are changes
+        id: changes
+        uses: UnicornGlobal/has-changes-action@v1.0.11
+      - name: Update THIRD-PARTY-NOTICES
+        uses: EndBug/add-and-commit@v7
+        with:
+          message: 'Automated update to THIRD-PARTY-NOTICES'
+          add: 'THIRD-PARTY-NOTICES'
+          push: true
+          default_author: github_actions
```
oshi97: maybe we can say something about this being the third party notices check?

.github/workflows/third_party_notices_check.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - name: Install generate-license-file
+        run: npm install -g generate-license-file
+      - uses: actions/checkout@v2
+      - run: npm ci
+      - name: Generate license file
+        run: generate-license-file --input package.json --output ./THIRD-PARTY-NOTICES
+      - name: Git diff
```
oshi97: Would it be ok to remove the Git diff and Install generate-license-file step names too?

.github/workflows/third_party_notices_check.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: git diff
+      - name: Check if there are changes
+        id: changes
```
oshi97: for my edification what is id here?

.github/workflows/third_party_notices_check.yml
```diff
@@ -0,0 +1,29 @@
+name: Check Third Party Notices File
+
+on: pull_request
+
+jobs:
+  license-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
```
oshi97: can we use v2? our other actions are on v2

package.json
```diff
@@ -23,7 +23,7 @@
     "dist"
   ],
   "dependencies": {
-    "@mapbox/mapbox-gl-language": "^0.10.1",
+    "@mapbox/mapbox-gl-language": "0.10.0",
```
oshi97, line 26: did we mean to update this?

.github/workflows/third_party_notices_check.yml
yen-tt: it's a unique identifier for the step that can be to reference the step in contexts. This portion is copied from vaccine's github action, where I assumed they would have want to use that id to check if there's changes before calling the next step to commit changes (`if: steps.changes.outputs.changed == 1`). However, I looked through `EndBug/add-and-commit@v7`. Seems like if there's no changes to the targeted file, a commit can't be made anyway so this step is not necessary.

package.json
```diff
@@ -194,4 +194,4 @@
     "./ja/modern.min": "./dist/ja-answers-modern.min.js",
     "./ja/template": "./dist/ja-answerstemplates.compiled.min.js"
   }
-}
\ No newline at end of file
+}
```
oshi97, line 197: could you try running npm install again? Just to make sure the package-lock needs the changes it has

yen-tt: yup! no changes detected after npm i

oshi97 (approved):

PR #1510, opened 2021-08-13
Fix inverted check mark in filter

yen-tt:
J=SLAP-1519
TEST=manual

see that check mark is not inverted in filter box from en and ar locale

### DIFF
--- a/src/ui/sass/modules/_FilterOptions.scss
+++ b/src/ui/sass/modules/_FilterOptions.scss
@@ -170,7 +170,8 @@
       left: -20px;
       width: 3px;
       height: 9px;
-      transform: rotate(45deg) translateY(-50%);
+      transform: rotate(45deg) translateY(-50%) #{"/*rtl:rotate(45deg) translateX(-100%)*/"};
+      /*!rtl:ignore*/
       border-right: 1px solid $color-brand-hover;
       border-bottom: 1px solid $color-brand-hover;
       border-color: var(--yxt-filter-options-checkmark-color);


src/ui/sass/modules/_FilterOptions.scss
```diff
@@ -170,7 +170,8 @@
       left: -20px;
       width: 3px;
       height: 9px;
-      transform: rotate(45deg) translateY(-50%);
+      transform: rotate(45deg) translateY(-50%) #{"/*rtl:rotate(45deg) translateX(-100%)*/"};
+      /*!rtl:ignore*/
```
oshi97, line 174: is this ignoring the border-right?

yen-tt: yup

cea2aj: For my own edification, why do we need to ignore the border for RTL?

yen-tt: rtl plugin would change this right border styling to left border styling. But we want the right border to show instead of the left one so it's not inverted when we rotate/translate the symbol

oshi97: do we need to make it border-left for rtl then?

oshi97: ohh I see ignoring is different than removing, I thought this would just remove the line for rtl

oshi97 (approved):

PR #1513, opened 2021-08-17
Safari specific voice search alert message

oshi97:
This commit makes the "Note: for Safari users, Siri must be
enabled" message only appear on Safari, using user-agent sniffing.
Generally speaking you want to avoid UA sniffing where you can, but
here it doesn't seem like there are any alternatives.

I tried using https://www.npmjs.com/package/bowser for sniffing,
but that adds an extra 8KB to our modern bundle (128KB -> 134KB).
Our use case is pretty small so it seems a little overkill to use a fully featured
library. Bowser also does not try to use userAgentData internally (AKA user agent clients hints),
which are more robust than the regular user-agent string.

J=SLAP-1522
TEST=manual,auto

test that the Safari specific not only shows up in Safari
I couldn't figure out how to naturally get the 'service-not-allowed'
error to show up on safari/chrome, so I changed the error switching
code to always go into this case.
For Chrome I disconneted from wifi to trigger the 'network' error.
For Edge I ended the voice recognition before it was done to trigger the 'no-speech' error.

Instead of writing unit tests for specific UA strings, which aren't
particularly useful for making sure your sniffing works in a real
scenario (since issues normally arise from the strings becoming out
of date, NOT somebody changing the source code), I added a useragent
acceptance test suite. This suite spins up instances of specific browsers,
and checks that the sniffer methods work as expected on each of them.
If need be, we can run these tests on specific browser versions.

I moved the quarantine-mode flag to only the tests run by CI.
Locally, we typically don't want to run the tests in quarnatine mode.
Also these tests don't need to run in quarantine mode.

The useragentsuite is in a different folder than the other acceptance suites,
because it's meant to be run separately with a specific suite of browsers.


src/core/utils/useragent.js
```diff
@@ -13,3 +13,38 @@ export function isMicrosoftEdge () {
...
+
+/**
+ * Returns whether the current browser is Safari.
+ * Currently, Safari does not support User-Agent clients hints (userAgentData).
+ * Chrome likes to pretend that it's Safari.
+ *
+ * @returns {boolean}
+ */
+export function isSafari () {
+  if (isMicrosoftEdge()) {
```
yen-tt: How come we need to check for microsoft edge here?

oshi97: Edge also likes to pretend it's safari

oshi97: Adding to the jsdoc!

tests/acceptance/useragent/useragentsuite.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!(t.browser.name in browserNameToSniffingMethod)) {
+    throw new Error(
+      `Unexpected testcafe browser "${t.browser.name}".\n` +
+      `Expected one of ${Object.keys(browserNameToSniffingMethod).join(', ')}`);
+  }
+  for (const [browserToTest, sniffer] of Object.entries(browserNameToSniffingMethod)) {
+    const [userAgent, userAgentData] = await ClientFunction(() => {
+      return [window.navigator.userAgent, window.navigator.userAgentData];
+    })();
+    global.navigator = { userAgent, userAgentData };
```
yen-tt, line 25: can we not directly access window.navigator.[userAgent|userAgentData] in isSafari or isChrome? is navigator different from window.navigator

oshi97: We can't run isSafari directly in the browser through testcafe (I tried that initially), ClientFunction doesn't have any reference to the code outside of the browser

oshi97: like this is what happens
```js
function someFunctionIWantInTheBrowser() {
  console.log('blah');
}
await ClientFunction(() => {
  someFunctionIWantInTheBrowser()
  // Error is thrown "someFunctionIWantInTheBrowser is not defined"
})();
```

cea2aj:
I believe product was okay with simply not including the Safari-specific messaging, so we could go that route if we don't want to add this complexity to the repo. Although now that we have the code for it I think we should move forward with it because I think it will lead to a better UX

cea2aj:
Product was able to get the safari message to pop up on iphone on google chrome which was why the QA item was created

yen-tt (approved):

cea2aj (approved):

PR #1519, opened 2021-08-18
update edge supported locales for voice search

oshi97:
Ran my codepen that brute force checks locales on edge, which works by
by turning on speech recognition for 10s, and adding the locale
to a list if a 'network' occurs. After 10s, the speech recognition
is aborted and a 'no-speech' error is expected.

Also added a zh-cn and zh-tw to the supported locales list, even
though they're not currently locales the SDK has built in, in case
a user manually supplies these locales.

J=SLAP-1532
TEST=manual,auto

manually test a number of the locales (~a third of them)

### DIFF
--- a/conf/i18n/constants.js
+++ b/conf/i18n/constants.js
@@ -186,5 +186,28 @@ exports.SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE = [
   'de-at',
   'de-ch',
   'de-de',
-  'ja-jp'
+  'ja-jp',
+  'ar-ae',
+  'ar-bh',
+  'ar-dz',
+  'ar-iq',
+  'ar-kw',
+  'ar-ly',
+  'ar-om',
+  'ar-ps',
+  'ar-qa',
+  'ar-sa',
+  'nl-be',
+  'nl-nl',
+  'pt-br',
+  'pt-pt',
+  'sv-se',
+  'zh-hans-cn',
+  'zh-hans-hk',
+  'zh-hant-hk',
+  'zh-hant-tw',
+
+  // Below are locales that are not the SDK's built-in locales, but are supported by edge
+  'zh-cn',
+  'zh-tw'
 ];
--- a/src/core/speechrecognition/locales.js
+++ b/src/core/speechrecognition/locales.js
@@ -1,4 +1,5 @@
 import { SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE } from '../constants';
+import { parseLocale } from '../utils/i18nutils';
 
 /**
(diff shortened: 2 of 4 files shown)


conf/i18n/constants.js
```diff
@@ -186,5 +186,28 @@ exports.SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE = [
...
+  'ar-om',
+  'ar-ps',
+  'ar-qa',
+  'ar-sa',
+  'nl-be',
+  'nl-nl',
+  'pt-br',
+  'pt-pt',
+  'sv-se',
+  'zh-hans-cn',
```
cea2aj, line 205: just confirming 'zh-hans-sg` and 'zh-hant-mo' aren't supported?

src/core/speechrecognition/locales.js
```diff
@@ -9,15 +10,39 @@ import { SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE } from '../constants';
...
-  const isCompatibleWithEdge = SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE.includes(locale.toLowerCase());
+  const locale = formatLocaleForEdge(language, modifier, region);
+  const isCompatibleWithEdge = SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE.includes(locale);
   if (isCompatibleWithEdge) {
     return locale;
   }
-  return locale.substring(0, underscoreIndex);
+  if (modifier) {
+    return formatLocaleForEdge(language, modifier);
+  }
```
cea2aj, line 25: If a locale isn't in the compatibility list but it has a modifier, we return it? I would expect that if the locale isn't in the compatibility list, we return just the language, not a language and modifier.

oshi97: If zh-hans-bob isn't supported, we can return zh-hans instead of just zh
by default edge will interpret zh on its own as zh-hans, so this only matters for zh-hant-[unsupported-region]

conf/i18n/constants.js
oshi97: they are not, you can very quickly test them here https://codepen.io/oshi97/pen/RwVKKzB

oshi97: or by hacking the searchbar's SpeechRecognition instance through dev tools with `ANSWERS.components._activeComponents.find( // find the searchbar and go through its instance variables`

src/core/speechrecognition/locales.js
cea2aj: alright, got it

cea2aj (approved):

PR #1525, opened 2021-08-19
Additional Language Map Support

cea2aj:
Add support for additional languages and language scripts

Most notably, this adds support for Simplified and Traditional Chinese for the maps

J=1534
TEST=manual

Load a test site with traditional and simplifed Chinese for mapbox and google maps

### DIFF
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -41,7 +41,7 @@ export default class GoogleMapProvider extends MapProvider {
    */
   getLanguage (localeStr) {
     const googleMapsCustomLanguages =
-      ['zh-CN', 'zn-HK', 'zh-TW', 'en-AU', 'en-GB', 'fr-CA', 'pt-BR', 'pt-PT', 'es-419'];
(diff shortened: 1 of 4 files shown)


tmeyer2115 (approved):

yen-tt (commented):
zn-HK should be zh-HK in google's list of supported locales. 
Update: There could be cases where region is provided, like zh_hant_TW and zh_hant_HK so we should check that as well?

cea2aj:
> zn-HK should be zh-HK in google's list of supported locales. And there could be cases where region is provided, like zh_hans_TW and zh_hans_HK so we should check that as well?

For locales such as 'zh-Hans-TW' and 'zh-Hans-HK', they will be converted to 'zh-CN'
 because Google uses 'zh-CN' and 'zh-TW' to represent simplified and traditional. Google maps mostly expects just the language code, except for the few locales which it supports which also include a region, and those extra cases are listed in the array. Edit: Updated so that any Chinese HK locales will use 'zh-HK'

yen-tt (approved):

oshi97 (approved):

PR #1529, opened 2021-08-23
Search bar only asset deployment

cea2aj:
Setup Circle CI to build and deploy the search bar only assets

The deploy process for the search bar is based on the same deployment process of the regular assets. I added a deploy-bucket param to the deploy-to-aws so that we can specify which bucket to deploy the assets to.

I updated the github action for the npm publish so that we only publish the full answers assets to NPM and not the search bar assets.

J=SLAP-1509
TEST=manual

Create a test release (v0.0.1) and see the tests pass in circle CI and see the assets deploy under the appropriate version (v0, v0.0, and v0.0.1)


.circleci/testcafe_search_bar.json
```diff
@@ -0,0 +1,5 @@
+{
+  "src": ["!tests/acceptance/acceptancesuites/*.js", "tests/acceptance/acceptancesuites/searchbaronlysuite.js"],
```
oshi97: do we need the first negation? I think we might not even need a config file for the search bar, if the only difference is the src

.circleci/run_headless_acceptance_search_bar.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+npx testcafe -c 3 "chrome:headless,firefox:headless" --config-file ./.circleci/testcafe_search_bar.json -q
```
oshi97: Do we need a bash script for a one liner?

.circleci/config.yml
```diff
@@ -174,6 +198,29 @@ jobs:
       - deploy-to-aws:
           subdirectory: '/${MAJOR_VERSION}'
           cache-control: 'max-age=43200'
+  # deploys assets to a long-lived-cache folder in the S3 bucket named by release tag
```
oshi97, line 203: I was thinking it could be nice if we extracted the shared logic for deploy_version and deploy_search_bar_version, but think it's unneeded unless we plan to add additional "sub-libraries" (I'm making up words)

.circleci/config.yml
```diff
@@ -319,3 +366,37 @@ workflows:
...
+      - build_search_bar:
+          filters:
+            tags:
+              only: /^search-bar-v.*/
+            branches:
+              ignore: /.*/
+      - headless_acceptance_test_search_bar:
+          filters:
+            tags:
+              only: /^search-bar-v.*/
```
oshi97, line 382: I forget why, but we're not able to just have the first step have the `filters: tags: only: /^search-bar` option right, with the other steps automatically getting the filter?

.github/workflows/npm_publish.yml
```diff
@@ -1,10 +1,9 @@
 name: NPM publish on new release
 
 on: 
-  release:
-    types: [created]
-    branches: 
-      - master
+  push:
```
oshi97, line 4: would it be possible to still have this on the `release` event, but with the tag filtering?

cea2aj: No, that's not possible. I tried that initially, but the filter would not take effect. I found in the docs that you can only filter the tags on push or the pull_request events

.circleci/testcafe_search_bar.json
yen-tt: can we remove `!tests/acceptance/acceptancesuites/*.js"`? I think just specifying tests/acceptance/acceptancesuites/searchbaronlysuite.js should ignore other test files

.github/workflows/npm_publish.yml
oshi97: ah ok, got it!

.circleci/config.yml
cea2aj: We are planning on reworking the logic for deploying the major version since it doesn't take into account whether or not it's the latest version, so we could consolidate then

.circleci/config.yml
cea2aj: Filters don't automatically pass through the same way branches do. You need to include filters at every step, but branches will automatically pass through

.circleci/testcafe_search_bar.json
cea2aj: I can remove that first negation. But how would we remove the config file? Would we point it to the same one as the regular acceptance test and also pass a flag in the testcafe command to override the src? At that point, I think the config file would be cleaner

.circleci/run_headless_acceptance_search_bar.sh
cea2aj: That's true, I can update

.circleci/testcafe_search_bar.json
cea2aj: updated!

oshi97: ok! I think I like reducing redundancy a little more than being explicit, but either way works for me!

oshi97 (approved):

PR #1530, opened 2021-08-23
Deprecate the serving of the latest assets

cea2aj:
This PR deprecates the serving of the latest assets off of the master branch

Rather than pinning to latest, developers should pin to a version of Answers.

This PR also makes the `subdirectory` param required for the `deploy-to-aws` command. In the `build_i18n` step of `build_and_deploy_hold`, now ignores all branches as only version tags should trigger it. Because we are ignoring all branches at the start of the workflow (in `build_i18n`), we no longer need to ignore all branches in the `deploy_version` step

J=SLAP-1536
TEST=none

I don't think we can fully test this until it gets merged into master. However we know the config is valid because otherwise Circle CI would throw an error

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,6 @@ commands:
     parameters:
       subdirectory:
         type: string
-        default: ''
       cache-control:
         type: string
         default: "no-cache"
@@ -125,13 +124,6 @@ jobs:
           command: |
             sudo apt-get install -qq gettext
             ./.circleci/run_translation_verification.sh
-  # deploy to S3 in the root folder, overwriting the existing latest version
-  deploy_latest:
-    docker:
-      - image: circleci/python:2.7
-    working_directory: ~/answers
-    steps:
-      - deploy-to-aws
   # deploys assets to an uncached folder in the S3 bucket named by branch
   deploy_branch:
     docker:
@@ -262,8 +254,7 @@ workflows:
             tags:
               only: /^v.*/
             branches:
-              only:
-                - master
+              ignore: /.*/
       - unit_test:
           filters:
             tags:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -23,7 +23,6 @@ commands:
     parameters:
       subdirectory:
         type: string
-        default: ''
```
yen-tt, line 26: for my own understanding, where do we get the subdirectory parameter? is that always given so there's no need for a default?

.circleci/config.yml
```diff
@@ -262,8 +254,7 @@ workflows:
             tags:
               only: /^v.*/
             branches:
-              only:
-                - master
+              ignore: /.*/
```
yen-tt, line 257: this essentially ignore every branch right? can we remove this?

oshi97:
In light of sites actually using answers/latest, I'm wondering whether it would be possible have the answers/latest folder point to the answers/v1 folder in S3. That way we could safely delete this workflow and still keep the exact same behavior (until we release an SDK v2 aka never)

.circleci/config.yml
cea2aj: All of the calls to `deploy-to-aws` specify a subdirectory. Be removing the default, I'm simply making it a required parameter since we no longer want to update the files at the top level of the folder

cea2aj:
> In light of sites actually using answers/latest, I'm wondering whether it would be possible have the answers/latest folder point to the answers/v1 folder in S3. That way we could safely delete this workflow and still keep the exact same behavior (until we release an SDK v2 aka never)

I fear that would make the S3 bucket more complicated because we would have to remember that we have a pointer in S3 (also, i'm not sure if that's supported)

oshi97:
> > In light of sites actually using answers/latest, I'm wondering whether it would be possible have the answers/latest folder point to the answers/v1 folder in S3. That way we could safely delete this workflow and still keep the exact same behavior (until we release an SDK v2 aka never)
> 
> I fear that would make the S3 bucket more complicated because we would have to remember that we have a pointer in S3 (also, i'm not sure if that's supported)

Hmm, we could also just add it adjacent to the major version deploy. I think that would work?

oshi97 (approved):

PR #1544, opened 2021-08-30
Add register custom partial functionality

nmanu1:
Allow custom Handlebars partials to be registered, similar to templates and helpers.

J=SLAP-1544
TEST=manual

Local testing by adding and using a partial in theme

### DIFF
--- a/README.md
+++ b/README.md
@@ -2319,6 +2319,41 @@ ANSWERS.addComponent('MyCustomComponent', {
 });
 ```
 
+# Custom Partials
+
+[Handlebars partials](https://handlebarsjs.com/guide/partials.html) are
+templates that can be reused and called directly from other templates. The 
+SDK offers an `ANSWERS.registerPartial` function. This will map a partial string 
+to an entry in the Answers handlebars renderer.
+
+```js
+  /**
+   * Add a partial
+   * @param {string} partialName The unique name for the partial
+   * @param {string} partial The handlebars partial string
+   */
+  registerPartial (partialName, partial)
+```
+
+The default handlebars renderer uses a mapping from partial name strings to
+handlebars partial strings. Once registered, this custom partial can be called
+from other templates.
+
+For example,
+```js
+  // create custom partial
+  ANSWERS.registerPartial(
+    'thank-you-message',
+    '<div class="thanks">Thank you!</div>'
+  );
+
+  // use new partial in a template
+  ANSWERS.registerTemplate(
+    'cards/custom',
+    '<p>Main information</p> {{> thank-you-message }}'
+  );
+```
+
 # Template Helpers
 
 When using handlebars templates, Answers ships with a bunch of pre-built template helpers that you can use. You can learn more about them [here](https://github.com/jonschlinkert/template-helpers).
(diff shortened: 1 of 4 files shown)


src/ui/rendering/handlebarsrenderer.js
```diff
@@ -51,6 +51,15 @@ export default class HandlebarsRenderer extends Renderer {
     this._handlebars.registerHelper(name, cb);
   }
 
+  /**
+   * register a partial to the internal Handlebars compiler
+   * @param {*} partialName The unique name for the partial
```
tmeyer2115: Update the param types here

src/ui/rendering/handlebarsrenderer.js
```diff
@@ -51,6 +51,15 @@ export default class HandlebarsRenderer extends Renderer {
     this._handlebars.registerHelper(name, cb);
   }
 
+  /**
+   * register a partial to the internal Handlebars compiler
```
tmeyer2115: Capitalize register

tmeyer2115 (commented):
We should add a blurb about this in the README.

README.md
```diff
@@ -2319,6 +2319,41 @@ ANSWERS.addComponent('MyCustomComponent', {
 });
 ```
 
+# Custom Partials
+
+[Handlebars partials](https://handlebarsjs.com/guide/partials.html) are
+templates that can be reused and called directly from other templates. The 
+SDK offers an `ANSWERS.registerPartial` function. This will map a partial string
```
cea2aj: nit: Since this repo is no longer officially called the SDK, I would reword this to "The ANSWERS object offers an `ANSWERS.registerPartial` function for custom partials."

README.md
```diff
@@ -2319,6 +2319,41 @@ ANSWERS.addComponent('MyCustomComponent', {
 });
 ```
 
+# Custom Partials
+
+[Handlebars partials](https://handlebarsjs.com/guide/partials.html) are
+templates that can be reused and called directly from other templates. The
```
tmeyer2115: We need to be careful with the verbiage here. Templates have a very specific meaning in Handlebars and are distinct from Partials. A partial can be called from a template or another partial.

README.md
```diff
@@ -2319,6 +2319,41 @@ ANSWERS.addComponent('MyCustomComponent', {
 });
 ```
 
+# Custom Partials
```
cea2aj: Can we add this to the README table of contents? It's located at the top of the README.

Since this section is similar to the "Template Helpers" section, what do you think of making an H1 section called "Custom Rendering", and inside that we have H2 sections for "Custom Partials" and "Template Helpers"?

nmanu1: Yeah, I was also thinking it might be nice to group them. Would the H1 title be confusing since there is also a "Using a Custom Renderer" section that might look similar?

cea2aj: Maybe we have the H1 be "Extending the built-in renderer"? That might be distinct enough from "Using a custom renderer"

README.md
nmanu1: Good to know! I've reworded this a little in the latest commit, but let me know if I should change it more

tmeyer2115 (approved):

PR #1548, opened 2021-09-15
Fix google map inValidValueError

yen-tt:
- allow config to pass a pin label that is an empty string instead of defaulting to SDK built-in label got GoogleMapProvider (only use when label is not specified at all in config)

J=SLAP-655
TEST=manual

launched theme's test site, and see that theme's pin label config (empty string) is used instead of default marker label


oshi97 (approved):
nice!!

cea2aj (approved):

PR #1549, opened 2021-09-17
custom alert replace window.alert

yen-tt:
- Use sweetAlert library to provide framework for browser dialogs and notifications.
- Add custom css (to match with current window popup UI) to answers.scss
- create alert function to wrap around sweetAlert functions with default configuration setup
- update all window.alert calls to use alert from alert.js

J=SLAP-1524
TEST=manual

- serve page using voicesearch functionality with mic access block, see that an alert is display on page with proper UI
- check with screen reader and see that the popup message and button can be read


cea2aj:
Unfortunately we can't use alertifyJS because it uses the GPL-3.0 License which is a copyleft license which forces us to follow certain conditions. The licenses pre-approved by legal are MIT, BSD 2, BSD 3, and Apache.

src/answers-umd.js
```diff
@@ -79,6 +80,11 @@ class Answers {
      */
     this.renderer = new Renderers.Handlebars();
 
+    /**
+     * A reference of AlertifyJS library for browser dialogs and notifications.
+     */
+    this.alertify = alertify;
```
cea2aj: Based on the other PR, it doesn't look like the theme needs access to an alert function, so I don't think we need to add it to ANSWERS yet. Tom likes to keep the functionality on ANSWERS to a minimum and we can always add it in the future if needed

src/ui/components/filters/geolocationcomponent.js
```diff
@@ -264,7 +265,7 @@ export default class GeoLocationComponent extends Component {
     this.setState({ geoError: true });
     const { enabled, message } = this._geolocationTimeoutAlert;
     if (enabled) {
-      window.alert(message);
+      alertify.alert(message);
```
cea2aj: Can we wrap the library's alert function inside our own alert() helper function? That way our code doesn't depend directly on alertify, and if we need to swap out the alert library, we can make the change in a single file

yen-tt: will do!

yen-tt:
Window popup:
Screen Shot 2021-09-20 at 2 37 00 PM
SweetAlert + custom css:
Screen Shot 2021-09-20 at 2 36 34 PM

src/ui/alert.js
```diff
@@ -0,0 +1,16 @@
+import swal from 'sweetalert';
+import DOM from './dom/dom';
+
+const defaultAlertOptions = {
+  closeOnClickOutside: false,
+  closeOnEsc: false
+};
+
+export default class Notify {
```
cea2aj: nit: Can we export a function called alert rather than wrap it in a class? I also like to match the file name with the export for default exports, so exporting a single function named 'alert' would make the default export match the filename

yen-tt: updated!

cea2aj (approved):

oshi97 (approved):
peeked at the npm package and this looks sweet! is there a difference between sweetalert and sweetalert2?

yen-tt:
> peeked at the npm package and this looks sweet! is there a difference between sweetalert and sweetalert2?

sweetalert2 is forked off sweetalert when there was a period it wasn't being maintained. From the documentation, sweetalert2 seems to have more configuration and themes options, but it is a bigger package. For our purpose, I think sweetalert should be sufficient!

PR #1550, opened 2021-09-20
Add distance to card data

yen-tt:
- add distance to result data in card component for dataMapping
- provide a function in ANSWERS that format + localized distance string with km or mi as unit of length based on locale, if distance is given from liveAPI data.
- functions to get distance from locale is mainly extracted from Theme with some changes

J=1573
TEST=manual
- test with standard card component on index.html with universal search, after modifying the template to display distance, and see that distance is porperly formatted on each result.
- jest test


oshi97 (approved):

package.json
```diff
@@ -194,6 +196,46 @@
     "./ja": "./dist/ja-answers-umd.js",
     "./ja/modern": "./dist/ja-answers-modern.js",
     "./ja/modern.min": "./dist/ja-answers-modern.min.js",
-    "./ja/template": "./dist/ja-answerstemplates.compiled.min.js"
+    "./ja/template": "./dist/ja-answerstemplates.compiled.min.js",
```
tmeyer2115, line 199: I think we may need to issue a hotfix with these new entry points. The current SDK version published to NPM is supposed to have support for all these languages. But, without these entry points, the new language bundles are inaccessible. 

Can we have this change in a separate PR then, pointing to a hotfix branch?

src/core/utils/i18nutils.js
```diff
@@ -50,3 +50,116 @@ export function parseLocale (localeCode) {
...
+ * Convert distance from meters to kilometers
+ *
+ * @param {number} distance
+ * @param {string} locale
+ * @returns {string} a string consist of distance and km as unit of length
+ */
+function _toKilometers (distance, locale) {
+  const distanceInKilometers = distance / 1000; // Convert meters to kilometers
+  return new Intl.NumberFormat(locale,
+    { style: 'decimal', maximumFractionDigits: 1, minimumFractionDigits: 1 })
```
tmeyer2115: Here and below move `locale` to this line. We either have the args all on one line, or if that's too long, we have one arg per line. I think we should also make sure that `locale` is not flush with `.format`. It should have an additional indent. That makes the code a bit easier to read.

tmeyer2115 (commented):
I think we need to check with Rose about the scope of this item. I'm assuming she wants `distance` as an attribute that's available when specifying a `dataMapping` for a card. Does she want the raw data or the formatted data? Or does she want the raw data, but a method available in the SDK to format it if the user so chooses?

yen-tt:
> I think we need to check with Rose about the scope of this item. I'm assuming she wants `distance` as an attribute that's available when specifying a `dataMapping` for a card. Does she want the raw data or the formatted data? Or does she want the raw data, but a method available in the SDK to format it if the user so chooses?

Confirmed with her this item should provide raw data for dataMapping and then a method for formatting

src/ui/components/cards/cardcomponent.js
```diff
@@ -111,6 +111,10 @@ export default class CardComponent extends Component {
 
     // Use the cardType as component name if it is not a built-in type
     const cardComponentName = cardTypes[cardType] || cardType;
+
+    // additional field(s) for dataMappings config option
+    this.result._raw.distance = this.result.distance;
```
oshi97: would it make sense to add this to answers-core's ResultFactory.ts? Instead of having the base card component mutate the raw data.

yen-tt: this seems more of a SDK specific implementation for user customization purposes using dataMappings, so I think it might not be necessary to modify answers-core's result format

oshi97: ok! should we do this in the constructor instead then?

yen-tt: oh yeah, make sense. Updated!

src/ui/components/cards/cardcomponent.js
```diff
@@ -64,6 +64,9 @@ export default class CardComponent extends Component {
      * @type {string}
      */
     this.verticalKey = data.verticalKey;
+
+    // additional field(s) for dataMappings config option
+    this.result._raw.distance = this.result.distance;
```
oshi97: I think this will error out if data.result is undefined, since in that case this.result will be set to `{}` (also can we move this to right below this.result is first set)

yen-tt: ah yup, you are right. Updated!

oshi97 (approved):

PR #1551, opened 2021-09-21
update SDK Mapbox to same version as Theme

yen-tt:
- updated version to v1.13.1 for Mapbox js and css api
- checked that there's no breaking change with the version update
- confirmed with product of the change in pricing model

TEST=manual
- add map to universal search component on page, serve page using locale answer bundle and see that mapbox work properly
- test that dark-v10 style works for mapbox


tmeyer2115 (approved):
LGTM, we should just make sure we get the okay from Rose on the pricing.

yen-tt:
> LGTM, we should just make sure we get the okay from Rose on the pricing.

She said it's fine to proceed with the change

PR #1558, opened 2021-09-28
Search Rate Tracking

cea2aj:
Allow search rate tracking by sending impression analytics events when the experience loads

J=SLAP-1600
TEST=manual, integration

View the network tab and see that the answers impression event is fired on load. For the search bar only assets, see that 'standAlone' is true. Also see that 'standaAlone' is true if the regular SDK assets are used but only the search bar component is added to the page. Confirm the other fields are included such as experienceKey, experienceVersion, searcher, and verticalConfigId. Add integration test which confirms that the report function is called properly for both the standard integration and the search bar integration

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -305,15 +305,19 @@ class Answers {
     return asyncDeps.finally(() => {
       this._onReady();
 
-      if (!this.components.getActiveComponent(SearchComponent.type)) {
+      const isSearchBarActive = this.components.getActiveComponent(SearchComponent.type);
+      const numActiveComponents = this.components.getNumActiveComponents();
+
+      if (isSearchBarActive) {
+        const impressionEvent = createImpressionEvent({
+          verticalKey: parsedConfig.search?.verticalKey,
+          // check that 1 or 2 components are active because the search bar also creates the autocomplete component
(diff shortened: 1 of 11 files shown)


src/answers-umd.js
```diff
@@ -303,9 +304,17 @@ class Answers {
     const asyncDeps = this._loadAsyncDependencies(parsedConfig);
     return asyncDeps.finally(() => {
       this._onReady();
+
       if (!this.components.getActiveComponent(SearchComponent.type)) {
         this._initQueryUpdateListener(parsedConfig.search);
       }
+
+      const impressionEvent = createImpressionEvent({
```
tmeyer2115: Here, do we need to look at what components are on the actual page first? If there is no `SearchBar` component on the page, do we fire an impression event? It's also possible that someone could pull down the full SDK and only instantiate a `SearchBar` component. In that case, should we send `standAlone` is true?

cea2aj: Those are good points. I'll make some updates

cea2aj: Updated. I'm assuming that the standalone assets will always put the search bar on the page, so those assets will always set `standalone: true`

tmeyer2115 (approved):

PR #1561, opened 2021-10-01
fix defaultInitialSearch

oshi97:
Prior to this change, if you didn't have a businessId, searches on page load
would error out because a search impression analytics event would try to be fired
using a null analytics reporter service.

J=SLAP-1616
TEST=manual,auto

I can run default initial searches, and searches using the query url param again, in my local
SDK test site and also the theme's test-site

tweaked the universalinitialsearch acceptance suite to not have a business id

### DIFF
--- a/src/answers-search-bar.js
+++ b/src/answers-search-bar.js
@@ -215,11 +215,13 @@ class AnswersSearchBar {
     this._handlePonyfillCssVariables(parsedConfig.disableCssVariablesPonyfill)
       .finally(() => {
         this._onReady();
-        const impressionEvent = createImpressionEvent({
-          verticalKey: parsedConfig.search?.verticalKey,
-          standAlone: true
-        });
-        this._analyticsReporterService.report(impressionEvent, { includeQueryId: false });
+        if (this._analyticsReporterService) {
+          const impressionEvent = createImpressionEvent({
+            verticalKey: parsedConfig.search?.verticalKey,
+            standAlone: true
+          });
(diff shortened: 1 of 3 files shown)


yen-tt (commented):
nit: maybe just `activeSearchBar` since this return a component instead of boolean?

oshi97:
> nit: maybe just `activeSearchBar` since this return a component instead of boolean?

ah that's a good point, I wrapped it as a boolean!

yen-tt (approved):

src/answers-search-bar.js
```diff
@@ -215,11 +215,13 @@ class AnswersSearchBar {
...
-          verticalKey: parsedConfig.search?.verticalKey,
-          standAlone: true
-        });
-        this._analyticsReporterService.report(impressionEvent, { includeQueryId: false });
+        if (this._analyticsReporterService) {
+          const impressionEvent = createImpressionEvent({
+            verticalKey: parsedConfig.search?.verticalKey,
+            standAlone: true
+          });
+          this._analyticsReporterService.report(impressionEvent, { includeQueryId: false });
```
cea2aj, line 223: nit: I believe we can do something like `this._analyticsReporterService?.report()`, if we want to avoid writing the entire if statement for this, but it's really not necessary

cea2aj (approved):
lg, my bad

src/answers-search-bar.js
oshi97: oh yeah I just didn't want to create the event if we didn't need to

PR #1562, opened 2021-10-01
remove node-sass dependency

oshi97:
This commit removes the unused node-sass dep by upgrading gulp-sass to the latest
version. Now you can use any node version you want, yay!

Originally I wanted to upgrade the lockfile to v2, but that breaks some of our github actions
that still use npm 6 and the npm ci command. Not sure whether it's ok to just upgrade our actions so
I reverted that change.

J=NONE
TEST=manual

ran a build before this change, saw node-sass complain about my node version being incompabitible
ran a build just find after the change

ran `npm ls node-sass` before this change and saw a node-sass dep pop up under gulp-sass
ran it after the change and saw it go away


yen-tt (approved):
nice!

PR #1564, opened 2021-10-05
Add visitor support

nmanu1:
Accept visitor object in `init` and allow visitor to be set or changed using `setVisitor`. If a visitor is set, it is passed to all analytics events and to a new Answers Core instance.

J=SLAP-1620
TEST=auto, manual

Add jest tests for setting and changing the visitor. Check if visitor is sent during analytics events and searches in test-site.


src/answers-umd.js
```diff
@@ -693,6 +697,11 @@ class Answers {
         return value;
     }
   }
+
+  setVisitor (visitor) {
+    this._analyticsReporterService?.setVisitor(visitor);
+    this.core.storage.set(StorageKeys.VISITOR, visitor);
```
yen-tt: does visitor needs to be in storage? Once this is pass as part of the config param to core during init, that can be use for every core request. Looking at connor's pr, he's pulling it from the stored config in core, not the request argument

src/answers-umd.js
```diff
@@ -285,6 +285,10 @@ class Answers {
       componentManager: this.components
     });
 
+    if (parsedConfig.visitor) {
+      this.setVisitor(parsedConfig.visitor);
```
yen-tt: maybe we should check if visitor.id is present?

yen-tt:
should update answers-search-bar as well to support visitor

README.md
```diff
@@ -129,6 +129,8 @@ function initAnswers() {
     apiKey: '<API_KEY_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
+    // Optional, visitor interacting with the experience, see Visitor Configuration below
+    visitor: '<VISTOR_HERE>',
```
oshi97: since the visitor is supposed to be an object, let's just put some example visitor, or even just the example you have below, instead of `<VISITOR_HERE>`. `<API_KEY_HERE>` and `<EXPERIENCE_KEY_HERE>` are like that because they're strings, but the visitor is supposed to be an object.

nmanu1: should I just move the example and embed it here instead of having it below?

src/answers-umd.js
nmanu1: I can add that

README.md
oshi97: yeah that also works!

nmanu1: I decided to keep it below as well because there are additional details

src/answers-umd.js
nmanu1: changed to pass in the config instead of as a request argument

src/answers-search-bar.js
```diff
@@ -450,6 +454,16 @@ class AnswersSearchBar {
         return value;
     }
   }
+
+  setVisitor (visitor) {
+    if (visitor.id) {
+      this._analyticsReporterService?.setVisitor(visitor);
+      this.core.storage.set(StorageKeys.VISITOR, visitor);
+      this.core.init();
```
oshi97: why does setVisitor reinit core?

src/core/core.js
```diff
@@ -133,6 +133,7 @@ export default class Core {
       experienceKey: this._experienceKey,
       locale: this._locale,
       experienceVersion: this._experienceVersion,
+      visitor: this.storage.get(StorageKeys.VISITOR),
```
oshi97: like Yen said it's probably unnecessary to store the visitor in storage, and we already have a few too many storage keys

src/answers-search-bar.js
nmanu1: the visitor is passed in the config when making an instance of the core library. so if the visitor is changed, we have to reinit

src/answers-umd.js
```diff
@@ -297,7 +297,11 @@ class Answers {
 
     this._setDefaultInitialSearch(parsedConfig.search);
 
-    this.core.init();
+    if (parsedConfig.visitor) {
+      this.setVisitor(parsedConfig.visitor);
```
oshi97, line 301: maybe it would make sense for core.init to take in a visitor param?
I was thinking something like
```js
this.core.init({
  visitor: parsedConfig.visitor
});
```

nmanu1: updated!

src/core/core.js
nmanu1: changed so it's passed into init instead of storing in storage

oshi97 (approved):

src/core/core.js
```diff
@@ -127,12 +127,13 @@ export default class Core {
   /**
    * Initializes the {@link Core} by providing it with an instance of the Core library.
    */
-  init () {
+  init (visitor) {
```
cea2aj: Can we have a config param instead that we spread into the params object? That way it's not specific to the visitor and the init could be written as oshi suggested: `this.core.init({ visitor: parsedConfig.visitor });`?

nmanu1: updated!

cea2aj (approved):

PR #1566, opened 2021-10-11
Add support for Auth Token to SDK

yen-tt:
- update Answers (answers-umd and answers-search-bar)and Core class to support token field pass in from config.
- update readme

J=SLAP-1627
TEST=manual & auto

see that token is in header when making a request

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ Add the Javascript library and placeholder elements for [Answers components](#co
 <div id="UniversalResultsContainer"></div>
 ```
 
-Add an initialization script with an apiKey, experienceKey and onReady function. In the example below, we've initialized two
-basic components: [SearchBar](#searchbar-component) and [UniversalResults](#universal-results-component).
+Add an initialization script with an apiKey or token, experienceKey and onReady function. In the example below, we've initialized two
+basic components with apiKey: [SearchBar](#searchbar-component) and [UniversalResults](#universal-results-component).
 ```js
 function initAnswers() {
   ANSWERS.init({
@@ -125,8 +125,10 @@ The configuration provided here is configuration that is shared across component
 ```js
 function initAnswers() {
   ANSWERS.init({
-    // Required, your Yext Answers API key
+    // Required*, your Yext Answers API key. *Do NOT provide apiKey if token is used.
     apiKey: '<API_KEY_HERE>',
+    //Required*, custom auth token. *Do NOT provide token if apiKey is used.
+    token: '<TOKEN_HERE>',
     // Required, the key used for your Answers experience
     experienceKey: '<EXPERIENCE_KEY_HERE>',
     // Optional, initialize components here, invoked when the Answers component library is loaded/ready.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -125,8 +125,10 @@ The configuration provided here is configuration that is shared across component
 ```js
 function initAnswers() {
   ANSWERS.init({
-    // Required, your Yext Answers API key
+    // Required*, your Yext Answers API key. *Do NOT provide apiKey if token is used.
     apiKey: '<API_KEY_HERE>',
+    //Required*, custom auth token. *Do NOT provide token if apiKey is used.
```
tmeyer2115: Space after //. For my edification what do the '*'s do?

src/answers-search-bar.js
```diff
@@ -247,11 +248,12 @@ class AnswersSearchBar {
     }
     parsedConfig.sessionTrackingEnabled = sessionTrackingEnabled;
 
+    const authIdKey = parsedConfig.apiKey ? 'apiKey' : 'token';
```
tmeyer2115, line 255: I think we should make this behavior clear in the README. It's not an invalid state if you provide both an `apiKey` and a 'token`, we'll just ignore the `token`.

src/answers-umd.js
```diff
@@ -419,8 +421,12 @@ class Answers {
...
     // TODO (tmeyer): Extract this method into it's own class. Investigate the use of JSON schema
     // to validate these configs.
-    if (typeof config.apiKey !== 'string') {
-      throw new Error('Missing required `apiKey`. Type must be {string}');
+    if (typeof config.apiKey !== 'string' && typeof config.token !== 'string') {
+      throw new Error('Missing required `apiKey` or `token`. Type must be {string}');
+    }
+
+    if (typeof config.apiKey === 'string' && typeof config.token === 'string') {
+      throw new Error('Both apiKey and token are present. Only one authentication method should be provided');
```
tmeyer2115, line 433: Is this really an exceptional state? I think we should ask Rose. I feel like we should just pick one and document that.

yen-tt: The spec stated that if both a token and an apiKey are provided, an error should be thrown, so I think this is ok to keep

README.md
yen-tt: I following the approach similar to how `businessId` is documented with 'Optional*' below. It indicates that it's required, with an exception if another authentication method is used. So in this case, token is required unless apiKey is used.

src/answers-search-bar.js
yen-tt: Spec noted it would be an invalid state if both is provided. If that changed then I will update this

src/answers-umd.js
tmeyer2115: Rose and I discussed. Initially, she wanted to change the behavior so that we gracefully default to `token` if both are specified. Then she remembered that she wrote "blow up" in the spec. So, we're going with that!

README.md
tmeyer2115: Okay, we still need a space before "Required" though.

tmeyer2115 (approved):

PR #1574, opened 2021-10-14
Build all assets before publish

cea2aj:
Update the prepublishOnly script to build alll languages and generate entrypoints before publishing

When releasing, I noticed that a standard build wasn't sufficient before publishing because it doesn't generate all the language assets. This updates the package script so that we can run an `npm publish` (or a `slap publish`) and have all of the correct assets built. This PR also checks in the updated entrypoints into the repo

J=none
TEST=manual

Run the prepublishOnly script and see the assets being generated and see the entrypoints being generated


cea2aj:
I'm not sure why the third party notices check is failing, however that should resolve once we merge it into develop. If not, I can take a closer look.

package.json
```diff
@@ -120,7 +120,7 @@
     "fix": "eslint . --fix",
     "extract-translations": "gulp extractTranslations",
     "preinstall": "npx npm-force-resolutions",
-    "prepublishOnly": "./conf/npm/prepublishOnly.sh",
+    "prepublishOnly": "npm run build-languages && npm run generate-entry-points && ./conf/npm/prepublishOnly.sh",
```
yen-tt: nit: should we just put those commands into the `prepublishOnly.sh` script?

cea2aj: the prepublishOnly script actually sets up the README for npm, and I would prefer to keep it doing just that one thing, so i'll rename that script to indicate what it's does

yen-tt: sounds good!

yen-tt (approved):

PR #1577, opened 2021-10-19
Automate WCAG testing

yen-tt:
- add new Github Action to run WCAG tests on push or pull request to develop, hotfix, and support branches
  - will add this to generator-slapshot repo after pr approval
- add pageNavigator, similar to theme's pageNavigator, which uses puppeteer API to automate browser navigation. 
- use axe-core/puppeteer library to generate WCAG report on each page, exit with error if there are violations
  - this library have license of type MPL-2.0 (reviewed and approved, for test environment only)

Note: 
- This pr does not include all possible ui components in the test yet, should make a new item/pr!
- This pr does not address all violations listed in the report, should make a new item/pr!

J=SLAP-1634
TEST=manual/auto

- run `npm run wcag` locally, and see that a wcag report is properly generated for each url, and listed out the violations to be addressed.
- See that github action ran on this pr


.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,24 @@
+name: WCAG tests
+
+on:
+  push:
```
tmeyer2115: I think just having this run on PRs is sufficient for now! We can adjust if need be later on.

.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches: [develop, hotfix/*, support/*]
+  pull_request:
+    branches: [develop, hotfix/*, support/*]
```
tmeyer2115: May as well throw `release/*` in here too. That way we aren't shocked to find a bunch of WCAG issues when we try and merge the `release` branch into `develop`.

tests/acceptance/wcag/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+const puppeteer = require('puppeteer');
+const { AxePuppeteer } = require('@axe-core/puppeteer');
+const PageNavigator = require('./pagenavigator');
+const WcagReporter = require('./wcagreporter');
+const handler = require('serve-handler');
+const http = require('http');
+const PORT = 9999;
+
+const axeCoreConfig = {
```
tmeyer2115, line 19: We should add high-level JSDoc explaining this config. Which WCAG standards does it test against? What does 'no-passes' mean?

tests/acceptance/wcag/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+  await browser.close();
+  await server.close();
+
+  if (failedResults.length > 0) {
+    console.error(`${failedResults.length} pages failed to be WCAG complaint:`);
+    console.log(JSON.stringify(failedResults, null, 2));
+    process.exit(1);
+  }
+})();
```
tmeyer2115: Small nit: Could we define the `async` function and invoke it separately? Something like:
```
const wcagTester = async () => ... ;
wcagTester();
```
I think that improves readability compared to having everything be in-line.

tests/acceptance/wcag/pagenavigator.js
```diff
@@ -0,0 +1,31 @@
+const { waitTillHTMLRendered, getQueryParamsString } = require('./utils');
+
+/**
+ * Responsible for navigating pages
+ */
+class PageNavigator {
```
tmeyer2115: Can/should we use this `PageNavigator` for our Percy snapshots as well? Right now, we manually navigate between pages when getting the snapshots.

tests/acceptance/wcag/utils.js
```diff
@@ -0,0 +1,44 @@
+
+/**
+ * Waits until the content of the page stops changing.
```
tmeyer2115, line 3: For my own edification, the `goto` method does not wait until the DOM of the page is fully loaded? Instead of rolling our own wait here, could we use `page.waitForNavigation()`? It's referenced a couple places:
- https://stackoverflow.com/questions/52497252/puppeteer-wait-until-page-is-completely-loaded
- https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitfornavigationoptions

yen-tt: No, goto doesn't wait for content to be fully loaded. I think we tried that and with waitForNavigation/other waitFor approach before in theme and it didn't work well so we resolve it with our custom wait. I tested waitForNavigation out locally just now with different network and DOM content loaded options, it's always stuck on the first page.

tests/acceptance/wcag/pagenavigator.js
yen-tt: sure, I updated to use PageNavigator

tests/acceptance/pagenavigator.js
```diff
@@ -22,8 +22,7 @@ class PageNavigator {
    * @param {string} selector The CSS selector to click on
    */
   async click (selector) {
-    // await this._page.click(selector);
-    await this._page.evaluate((selector) => document.querySelector(selector).click(), selector);
+    await this._page.click(selector);
     await waitTillHTMLRendered(this._page);
```
tmeyer2115, line 26: There are other built-in waiting mechanisms for Puppeteer, such as `waitFor`, `waitForSelector`, and `waitForFunction`. Have we tried all these options? I would prefer we use something built-in, rather than roll our own.

yen-tt: `waitFor` would require each navigation/click to wait for a fixed time, which may not accurately wait till things are loaded and cause tests to have longer wait time in general. I tried waitForSelector, but it doesn't work super well. Other puppeteer-extra plugins doesn't fit our use case here. I think for now, we should keep this wait till render function

cea2aj: The tricky part about `waitForSelector` or `waitForFunction` is having a reliable way of indicating that the page is done loading. We could have the SDK put something in the SDK or storage that we could use to indicate that all of the rendering is done, but that's potentially out of scope here. The waitTillHTMLRendered function is a pretty robust way of waiting for rendering to stop without requiring internal SDK changes

tests/acceptance/wcag/index.js
```diff
@@ -0,0 +1,70 @@
...
+ * - set reporter to 'no-passes' to only return violation results
+ * - set runOnly with tag values below to run WCAG standards:
+ *    - WCAG 2.0 Level A, AA, AAA
+ *    - WCAG 2.1 Level A, AA
+ */
+const axeCoreConfig = {
+  reporter: 'no-passes',
+  runOnly: {
+    type: 'tag',
+    values: ['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa', 'wcag2aaa']
```
cea2aj, line 23: I would check with Rose, but I think currently we are only trying to meet WCAG AA compliance. (We can leave AAA in here for now, but when we go to address these concerns, we may want to bump down to just AA)

tmeyer2115 (approved):

cea2aj:
Maybe we should merge this PR into a feature branch rather than develop until we address the failures? If we merge into develop, every PR will fail to pass the tests

tests/acceptance/wcag/index.js
yen-tt: will do

tmeyer2115:
> Maybe we should merge this PR into a feature branch rather than develop until we address the failures? If we merge into develop, every PR will fail to pass the tests

IMO, I would rather we merge this in now. Long-lived feature branches can become cumbersome. Merge conflicts can arise, etc. You have to constantly rebase. Especially here. If we don't rebase constantly, then the WCAG checks wouldn't catch any new failures added in `develop`. If possible, I think we should ensure that WCAG failures do not block the merging of a PR. They only register as warnings (if that's a thing). Once we are fully WCAG compliant, we can change it so that failure would block a PR.

cea2aj:
> > Maybe we should merge this PR into a feature branch rather than develop until we address the failures? If we merge into develop, every PR will fail to pass the tests
> 
> IMO, I would rather we merge this in now. Long-lived feature branches can become cumbersome. Merge conflicts can arise, etc. You have to constantly rebase. Especially here. If we don't rebase constantly, then the WCAG checks wouldn't catch any new failures added in `develop`. If possible, I think we should ensure that WCAG failures do not block the merging of a PR. They only register as warnings (if that's a thing). Once we are fully WCAG compliant, we can change it so that failure would block a PR.

Makes sense. We'll just want to watch out for the boy who cried wolf so that we don't start ignoring tests because they are always failing. It doesn't look like warnings are a thing except for code annotations which probably isn't what we are looking for.

tests/acceptance/wcag/index.js
yen-tt: Rose said Level AA for WCAG 2.0 and 2.1 should be good!

PR #1578, opened 2021-10-19
Improve autocomplete screen reader support 

nmanu1:
Move `span` with `aria-live` label to be top-level for the autocomplete component so that it is present on page-load, even with empty content. Add instructions for navigating autocomplete to the search bar. Update the autocomplete component to correctly announce number of autocomplete options found for no results and for sectioned filter search.

Note: there should be no screen reader text for number of options found or autocomplete instructions when the search bar is not active. Also, there should be no announcement for number of options found when user focuses on the search bar and there are no default autocomplete options shown.

J=SLAP-1653, 1402
TEST=manual

Smoke testing with theme test-site to check that screen reader correctly announces instructions and number of autocomplete options found.

### DIFF
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -143,11 +143,6 @@ export default class AutoCompleteComponent extends Component {
      * @type {boolean}
      */
     this._isVoiceSearchActive = false;
-
-    /**
-     * Indicates that the user hasn't typed anything in the autocomplete input yet
-     */
-    this._beforeSearch = true;
   }
 
   /**
@@ -184,20 +179,18 @@ export default class AutoCompleteComponent extends Component {
       this._isOpen = true;
     }
 
+    let _onOpen = false;
+
     if (wasOpen && !this._isOpen) {
       this._onClose();
-      this._beforeSearch = true;
     } else if (!wasOpen && this._isOpen) {
       this._onOpen();
-    }
-
-    if (queryInputEl.value && this.isQueryInputFocused()) {
-      this._beforeSearch = false;
+      _onOpen = true;
     }
 
     super.setState(Object.assign({}, data, {
       hasResults: this.hasResults(data),
-      beforeSearch: this._beforeSearch,
+      isAlreadyOpen: this._isOpen && !_onOpen,
       sectionIndex: this._sectionIndex,
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
@@ -248,7 +241,8 @@ export default class AutoCompleteComponent extends Component {
     // When a user focuses the input, we should populate the autocomplete based
     // on the current value
     DOM.on(queryInput, 'focus', () => {
-      this.reset();
+      this._sectionIndex = 0;
+      this._resultIndex = -1;
       this.autoComplete(queryInput.value);
     });
 
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -179,18 +179,18 @@ export default class AutoCompleteComponent extends Component {
       this._isOpen = true;
     }
 
-    let _onOpen = false;
+    // let _onOpen = false;
 
     if (wasOpen && !this._isOpen) {
       this._onClose();
     } else if (!wasOpen && this._isOpen) {
       this._onOpen();
-      _onOpen = true;
+      // _onOpen = true;
     }
 
     super.setState(Object.assign({}, data, {
       hasResults: this.hasResults(data),
-      isAlreadyOpen: this._isOpen && !_onOpen,
+      isAlreadyOpen: this._isOpen && wasOpen,
       sectionIndex: this._sectionIndex,
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -143,6 +143,11 @@ export default class AutoCompleteComponent extends Component {
      * @type {boolean}
      */
     this._isVoiceSearchActive = false;
+
+    /**
+     * Indicates that the user hasn't typed anything in the autocomplete input yet
+     */
+    this._beforeSearch = true;
   }
 
   /**
@@ -181,12 +186,18 @@ export default class AutoCompleteComponent extends Component {
 
     if (wasOpen && !this._isOpen) {
       this._onClose();
+      this._beforeSearch = true;
     } else if (!wasOpen && this._isOpen) {
       this._onOpen();
     }
 
+    if (queryInputEl.value && this.isQueryInputFocused()) {
+      this._beforeSearch = false;
+    }
+
     super.setState(Object.assign({}, data, {
       hasResults: this.hasResults(data),
+      beforeSearch: this._beforeSearch,
       sectionIndex: this._sectionIndex,
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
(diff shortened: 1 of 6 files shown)


src/ui/components/search/autocompletecomponent.js
```diff
@@ -181,12 +186,18 @@ export default class AutoCompleteComponent extends Component {
...
       this._onOpen();
     }
 
+    if (queryInputEl.value && this.isQueryInputFocused()) {
+      this._beforeSearch = false;
+    }
+
     super.setState(Object.assign({}, data, {
       hasResults: this.hasResults(data),
+      beforeSearch: this._beforeSearch,
```
yen-tt: instead of using this new variable `beforeSearch`, can we use `_isOpen` instead?

nmanu1: I tried using that initially, but the state is updated more than once when focus is put on the search bar. so, the first time _isOpen is true, but after that, _isOpen is false even if the user hasn't typed anything yet

src/ui/templates/search/filtersearch.hbs
```diff
@@ -5,7 +5,13 @@
          type="text"
          name="query"
          value="{{query}}"
-         {{#if _config.placeholderText}}placeholder="{{_config.placeholderText}}"{{/if}}>
+         {{#if _config.placeholderText}}placeholder="{{_config.placeholderText}}"{{/if}}
+         aria-describedby="instructions">
   </input>
+  <div id="instructions" class="yxt-SearchBar-instructions sr-only">
```
yen-tt, line 11: is `instructions` classname and id + css styling needed? I think the mixin styling for `sr-only` would help hide it already

src/ui/templates/search/autocomplete.hbs
```diff
@@ -1,4 +1,21 @@
 <div class="yxt-AutoComplete-wrapper js-yxt-AutoComplete-wrapper" aria-expanded="{{hasResults}}">
+  <span class="yxt-AutoComplete-resultsCount sr-only" aria-live="assertive">
+    {{#unless beforeSearch}}
+      {{#if hasResults}}
+        {{#each sections}}
+          {{translate
+            phrase='[[resultsCount]] [[label]] autocomplete option found.'
```
tmeyer2115, line 7: Outside the context of filter search, there is no `label` attribute. I think out of an abundance of caution, we should have a phrase for when `label` is present and another phrase for when there is no `label`. Having a single phrase like this, we may run into issues for certain languages when `label` is the empty string or undefined.

src/ui/templates/search/filtersearch.hbs
nmanu1: based on the accessible360 link for autocomplete, we would ideally want autocomplete instructions to be unreachable by the screen reader except when the user focuses on the input field in order to reduce screen-reader clutter. using the `sr-only` styling would still have the instructions be reachable if the search bar is not in focus. adding this new css styling makes the instructions unreachable from the screen reader except for when it is found from `aria-describedby`

src/ui/templates/search/autocomplete.hbs
```diff
@@ -1,4 +1,21 @@
...
+            pluralForm='[[resultsCount]] [[label]] autocomplete options found.'
+            resultsCount=resultsCount
+            label=label
+            count=resultsCount
+          }}
+        {{/each}}
+      {{else}}
+        {{translate phrase='0 autocomplete options found.'}}
```
tmeyer2115: Do we need to special case this? Or can it be handled by the more generic '[resultsCount]] autocomplete options found' we will have above?

src/ui/components/search/autocompletecomponent.js
yen-tt: it makes sense to me that _isOpen is false if user hasn't typed anything yet. and it's true when there's an input in the search bar, which is what you want, right?

nmanu1: I check the behavior again and my previous comment wasn't entirely correct. _isOpen is true whenever focus is on the search bar, regardless of whether there is any input text or not

src/ui/templates/search/autocomplete.hbs
nmanu1: In the case where there are no results, there are no sections for filterSearch, so the generic '[resultsCount]] autocomplete options found' would not be reached

src/ui/templates/search/autocomplete.hbs
nmanu1: Updated!

src/ui/templates/search/autocomplete.hbs
tmeyer2115: Understood. But do we need a specific, new phrase in this case? Could we not use 
```
            {{translate
              phrase='[[resultsCount]] autocomplete option found.'
              pluralForm='[[resultsCount]] autocomplete options found.'
              resultsCount=resultsCount
              count=resultsCount
            }}
```
here too? That would save us having to have an additional translation in Smartling just for this.

tmeyer2115: The `resultCount` would be 0 here.

nmanu1: yes, I can change that

src/ui/components/search/autocompletecomponent.js
yen-tt: if we use a `onOpen` variable under the condition `(!wasOpen && this._isOpen)`, that should work. In regard to the toggle value from true to false (update more than once). I see that this is triggered by `DOM.on(queryInput, 'focus',...)` where reset() update state once, and the following autocomplete call to core update state again. Let's change it to use the reset code but without the updateState call

nmanu1: updated to use `onOpen` instead

src/ui/templates/search/autocomplete.hbs
```diff
@@ -1,29 +1,34 @@
...
+        {{else}}
+          {{translate
+            phrase='[[resultsCount]] autocomplete option found.'
+            pluralForm='[[resultsCount]] autocomplete options found.'
+            resultsCount=resultsCount
+            count=resultsCount
+          }}
+        {{/if}}
+      {{/each}}
+    {{else}}
```
yen-tt: update to else if

src/ui/components/search/autocompletecomponent.js
```diff
@@ -184,20 +179,18 @@ export default class AutoCompleteComponent extends Component {
...
-
-    if (queryInputEl.value && this.isQueryInputFocused()) {
-      this._beforeSearch = false;
+      _onOpen = true;
     }
 
     super.setState(Object.assign({}, data, {
       hasResults: this.hasResults(data),
-      beforeSearch: this._beforeSearch,
+      isAlreadyOpen: this._isOpen && !_onOpen,
```
yen-tt: try `this._isOpen && wasOpen`

src/ui/components/search/autocompletecomponent.js
```diff
@@ -237,7 +238,8 @@ export default class AutoCompleteComponent extends Component {
     // When a user focuses the input, we should populate the autocomplete based
     // on the current value
     DOM.on(queryInput, 'focus', () => {
-      this.reset();
+      this._sectionIndex = 0;
+      this._resultIndex = -1;
```
tmeyer2115, line 242: I see we replaced the call to `this.reset` should these two things be done as part of `this.reset()`?

yen-tt: yes, but since this.reset() is also triggering stateUpdate, and then call `core.autoComplete` in the following line, this would cause state update twice and the values of wasOpen and isOpen to toggle quickly when click on searchBar. This wouldn't reflect the expected behavior, and it causes problem with the screen reader when there's no result on first click to search bar with empty query

tmeyer2115: I see.

tmeyer2115 (approved):

yen-tt (approved):

PR #1579, opened 2021-10-26
Add aria-activedescendant support in autocomplete

yen-tt:
- update searchbar input element's attributes to include aria-activedescedant based on selected autocomplete option using UP/DOWN key.

J=SLAP-1650
TEST=manual

see that search bar input contains aria-activedescendant with id of the selected autocomplete
see that search bar input remains in focus when moving throught the autocomplete option
tab over components and see that autocomplete options is not registered individually
see that aria-activedescedant is removed when move back to input element + when hit enter/click submit


src/ui/components/search/autocompletecomponent.js
```diff
@@ -185,6 +185,12 @@ export default class AutoCompleteComponent extends Component {
       this._onOpen();
     }
 
+    queryInputEl.removeAttribute('aria-activedescendant');
```
cea2aj: Is there a reason for doing this in setState rather than updating the autocomplete template to include the 'aria-activedescendant' attribute during rendering?

yen-tt: yeah, so the sectionIndex and resultIndex is located in autocompletecomponent but the actual input element is in the searchcomponent

cea2aj: got it. +1 from me. I'm wondering what @oshi97 thoughts are as well

oshi97: It's unfortunate we can't put it in the template, but that's a failure of the SDK we've seen in a few other places as well. We could have the searchbar pass a callback to autocomplete but I don't think that's easier to reason about. I think we can argue "this is something every usage of autocomplete will want" so I'm also good with it being here!

oshi97: It might make sense to move this into the keydown listener or maybe handleNavigationResults (or maybe do some kind of check to make sure the section actually changed) and also move this into its own method, but I'm pretty happy as is.

oshi97 (approved):

src/ui/components/search/autocompletecomponent.js
yen-tt: updated to move the code into its own method! I keep it in setState since `_sectionIndex` and `_resultIndex` may be change in there depending on if autocomplete should be hidden, also putting it in keydown listener or handleNavigationResults would spit the remove/add attribute into different logic path so overall I think leaving it in setState should be good.

PR #1580, opened 2021-10-26
Add tests for analytics opt in getter

nmanu1:
Add jest tests for the previously added getter function for whether analytics is enabled or not (#1568).

J=none
TEST=auto

Check that the new tests pass


tests/answers-umd.js
```diff
@@ -85,4 +85,26 @@ describe('ANSWERS instance integration testing', () => {
...
+
+  it('Set and get analytics opt in', async () => {
+    mockWindow(windowSpy, {
+      location: {
+        search: '?query=test'
+      }
+    });
+    await initAnswers(ANSWERS);
+    ANSWERS.setAnalyticsOptIn(true);
+    ANSWERS.getAnalyticsOptIn();
```
cea2aj, line 97: What do you think about updating this test so that it first sets it to true and then uses the getter to confirm the value is true? And then it could set it to false and use the getter to confirm the value is false. That approach might be better at capturing the behavior of the functions compared to checking if internal functions are called

nmanu1: The analytics reporter is mocked in this file, so the getter wouldn't return the correct value.

cea2aj: got it. In that case, lgtm

cea2aj (approved):

PR #1597, opened 2021-11-04
Support consumer auth in sandbox

yen-tt:
- Add a new field `environment` to set if the current Answer experience should be running with 'production' or 'sandbox' endpoints.
- make sure it's still backward compatible with the prefix in apikey when there's no `environment` field
- always remove `sandbox-` prefix if apikey is provided

J=SLAP-1663
TEST=manual

- apikey as 'sandbox-someKey'  ⇒ use sandbox endpoints (backward compatible)
- apikey as 'someKey'  ⇒ use production endpoints (default)
- token as 'some.random.token'  ⇒ use production endpoints (default)
- apikey as 'someKey' and environment as 'sandbox'/production ⇒ use sandbox/production endpoints
- token as 'some.random.token' and environment as 'sandbox'/production ⇒ use sandbox/production endpoints
- apikey as 'sandbox-someKey' and environment as 'production' ⇒ use production endpoints (prefix is removed)


src/answers-search-bar.js
```diff
@@ -252,12 +253,15 @@ class AnswersSearchBar {
     }
     parsedConfig.sessionTrackingEnabled = sessionTrackingEnabled;
 
-    const authIdKey = parsedConfig.apiKey ? 'apiKey' : 'token';
-    const sandboxPrefix = `${SANDBOX}-`;
-    parsedConfig[authIdKey].includes(sandboxPrefix)
-      ? parsedConfig.environment = SANDBOX
-      : parsedConfig.environment = PRODUCTION;
-    parsedConfig[authIdKey] = parsedConfig[authIdKey].replace(sandboxPrefix, '');
+    if (parsedConfig.apiKey) {
```
oshi97, line 256: so basically we don't need to check for the sandbox- prefix if a token is used instead of an apiKey?

oshi97 (approved):

src/answers-search-bar.js
yen-tt: yeah, I believe token will never have a prefix. We only need to take care of apiKey prefix for the purpose of backward compatibility

PR #1600, opened 2021-11-11
Update icon size

yen-tt:
- searchbar icon reduce from 2em to 1.5em
- update other searchbar icons to use searchbar styling instead of default _Icon.scss. moved the scss module for Icon above searchbar scss module since it make sense that specific component's styling should have precedent over the generic icon styling, like in this case with searchbar

J=SLAP-1697
TEST=manual

serve page from SDK with magnifying_glass submitIcon. launch test site from theme with a custom submitIcon. See that both pages have smaller icon size from before

### DIFF
--- a/src/ui/sass/answers.scss
+++ b/src/ui/sass/answers.scss
@@ -9,8 +9,8 @@
 @import "base";
 
 // Component Styling
-@import "modules/SearchBar";
 @import "modules/Icon";
+@import "modules/SearchBar";
 @import "modules/Nav";
 @import "modules/DirectAnswer";
 @import "modules/Results";
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -270,7 +270,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     }
   }
 
-  &-buttonImage .Icon svg, .Icon-image
+  &-AnimatedIcon svg, &-buttonImage .Icon svg, .Icon-image
   {
     width: 1.5em;
     height: 1.5em;
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -272,8 +272,8 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
   &-buttonImage .Icon svg, .Icon-image
   {
-    width: 2em;
-    height: 2em;
+    width: 1.5em;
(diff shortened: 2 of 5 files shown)


cea2aj:
Have you also checked how this looks on a theme site?

yen-tt:
> Have you also checked how this looks on a theme site?

yeah, icon looks smaller than the site from QA

cea2aj:
The reason behind the 2em was that it prevented the voice and clear button from shifting around when the loading indicator appears. Can you make sure that's not an issue with 1.5em? For reference: https://github.com/yext/answers-search-ui/pull/1504

yen-tt:
> The reason behind the 2em was that it prevented the voice and clear button from shifting around when the loading indicator appears. Can you make sure that's not an issue with 1.5em? For reference: #1504

Yeah, I see that it's because the loading indicator and voice search icons follow 'Icon--lg' instead of search-bar specified size. Updated! I tested it out in SDK and theme. the icons doesn't shift around with loading indicator appears

src/ui/sass/answers.scss
```diff
@@ -9,8 +9,8 @@
 @import "base";
 
 // Component Styling
-@import "modules/SearchBar";
 @import "modules/Icon";
+@import "modules/SearchBar";
```
cea2aj, line 13: It looks like import order matters here? If so, is there a way we can handle this without relying on the import order? It may be easy for us to accidentally switch the order

yen-tt: it does for css styling, where anything that comes later in the overall css file would override the previous styling if both are pointing to the same element. There's another way of doing this, which is to make `Icon--md` styling in _Icon.scss with 1.5em size. But I think this may also be confusing since there are two spots that determines icon size for search bar, but for every other component, their own styling takes precedent. Would adding a comment help?

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -270,10 +270,10 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
     }
   }
 
-  &-buttonImage .Icon svg, .Icon-image
+  &-AnimatedIcon svg, &-buttonImage .Icon svg, .Icon-image
```
cea2aj, line 273: This looks like it's shrinking the regular logo on percy since this styling affects it too

src/ui/sass/answers.scss
cea2aj: Sure, a comment would help

yen-tt:
from product discussion, icon size will stay as 2em.

PR #1604, opened 2021-11-16
Version 1.12

nmanu1:
### Features
- Allow search rate tracking (#1558)
- Add support for setting and changing the visitor and passing it to answers-core (#1564)
- Add support for the auth token that is passed in from the config (#1566)
- Add an `environment` field to support consumer auth in Sandbox (#1597)
- Allow components to override the beforeMount function (#1547)
- Add distance to the card data and a function to format it (#1550)
- WCAG updates (allow pagination with Enter (#1575), identify current page in navigation tab (#1576), update autocomplete screen reader support (#1578, #1579))

### Changes
- Update directAnswers component data to include the searcher (#1596)
- Use custom alerts instead of window.alert (#1549)
- Update Mapbox version to match the Theme (#1551)
- Internal repo changes (#1562, #1577)

### Bugfixes
- Fix console error which would appear on google maps (#1548)
- Fix FAQ expansion when default is expanded (#1553)
- Fix error for searches on page load with no businessId (#1561)


yen-tt (approved):
Product said we are good to proceed with the cut

PR #1605, opened 2021-11-16
Update core config

yen-tt:
- only pass a token or apiKey field if it is defined

TEST=manual

see error without it, see local page work after the change

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -136,8 +136,8 @@ export default class Core {
    */
   init (config) {
     const params = {
-      token: this._token,
-      apiKey: this._apiKey,
+      ...(this._token && { token: this._token }),
+      ...(this._apiKey && { apiKey: this._apiKey }),
       experienceKey: this._experienceKey,
       locale: this._locale,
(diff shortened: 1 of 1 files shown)


src/core/core.js
```diff
@@ -136,8 +136,8 @@ export default class Core {
    */
   init (config) {
     const params = {
-      token: this._token,
-      apiKey: this._apiKey,
+      ...(this._token && { token: this._token }),
```
cea2aj, line 139: I don't remember, do we need the parenthesis here? Otherwise, LGTM

cea2aj (approved):

src/core/core.js
oshi97: it looks like you technically don't but I slightly prefer having it

oshi97 (approved):

src/core/core.js
yen-tt: it's not needed, but I think it's easier to see that it's spreading out the object in the second condition instead of the first field

PR #1648, opened 2021-12-23
Update map controls

nmanu1:
Add zoom control buttons to MapBox maps. Remove controls except for zoom control buttons from Google map to match the Theme map. These can be added back if specified in the config. Switch the zoom buttons position for RTL, as in the theme.

J=SLAP-1765
TEST=manual

Check that only zoom buttons now appear on all maps and switch position for RTL pages.

### DIFF
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -128,8 +128,8 @@ export default class GoogleMapProvider extends MapProvider {
       const container = DOM.query(el);
 
       const zoomControlPosition = isRTL(this._language)
-        ? google.maps.ControlPosition.LEFT_BOTTOM
-        : google.maps.ControlPosition.RIGHT_BOTTOM;
+        ? google.maps.ControlPosition.LEFT_TOP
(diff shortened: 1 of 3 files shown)


src/ui/components/map/providers/googlemapprovider.js
```diff
@@ -126,9 +126,22 @@ export default class GoogleMapProvider extends MapProvider {
     // Only here for demo purposes, so we'll fix later.
     setTimeout(() => {
       const container = DOM.query(el);
+
+      const zoomControlPosition = isRTL(this._language)
+        ? google.maps.ControlPosition.LEFT_BOTTOM
+        : google.maps.ControlPosition.RIGHT_BOTTOM;
+
```
yen-tt: In theme we have `LEFT_TOP` and `RIGHT_TOP` so let's be consistent and make either top or bottom for both. Similar thing with mapbox one.

nmanu1: Switched to match the theme

yen-tt (approved):

PR #1650, opened 2022-01-13
Fix loading indicator in safari and ie11

yen-tt:
- animation in the loading indicator's svg tag fail to restart in safari after display is set to none. Since the display was used to stop the animation, and doesn't affect what is display, changing it to `animation: none;` fixed it.
- EI11 does not support `stroke-dashoffset` property in css animation.  Added a workaround for ie11 by setting a smaller length for the loading circle's stroke to simulate a fixed gap on rotation instead of a dynamic shrink/expand animation with `stroke-dashoffset`

J=SLAP-1755
TEST=manual

load page and perform search in EI11, safari, chrome, and firefox. See that loading indicator behave as expected.

### DIFF
--- a/src/ui/sass/modules/_SearchBar.scss
+++ b/src/ui/sass/modules/_SearchBar.scss
@@ -213,13 +213,21 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-LoadingIndicator-AnimatedIcon 
   {
     stroke-dasharray: 208;
+    /* IE11 specific styling */  
+    _:-ms-fullscreen, :root & { 
+      /**
+       * IE11 does not support stroke-* properties in animation, so use smaller stroke
+       * length to simulate a fixed gap instead of using stroke-dashoffset in animation.
+       */
+      stroke-dasharray: 150;
+   }
     transform-origin: 50% 50%;
     animation: dash 1.4s ease-in-out infinite;
   }
 
   &-Icon--inactive &-LoadingIndicator-AnimatedIcon 
   {
-    display: none;
+    animation: none;
   }
   
   @keyframes dash 


src/ui/sass/modules/_SearchBar.scss
```diff
@@ -213,13 +213,21 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
...
+      stroke-dasharray: 150;
+   }
     transform-origin: 50% 50%;
     animation: dash 1.4s ease-in-out infinite;
   }
 
   &-Icon--inactive &-LoadingIndicator-AnimatedIcon 
   {
-    display: none;
+    animation: none;
```
oshi97, line 230: out of curiosity what happens if you have both display: none and animation: none?

yen-tt: it would behave the same as before with the bug, so it wouldn't display, and the animation would stop playing in the background even after Icon--inactive classname is removed.

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -213,13 +213,21 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
   &-LoadingIndicator-AnimatedIcon 
   {
     stroke-dasharray: 208;
+    /* IE11 specific styling */  
+    _:-ms-fullscreen, :root & { 
+      /**
+       * IE11 does not support stroke-* properties in animation, so use smaller stroke
+       * length to simulate a fixed gap instead of using stroke-dashoffset in animation.
+       */
+      stroke-dasharray: 150;
```
tmeyer2115, line 222: Out of curiosity, could we remove the `stroke-dashoffset` from the animation SVG and simply use this styling regardless of browser? Would that be contrary to best practices? I ask because having browser-specific styling can get a bit tricky.

yen-tt: it would simplify the loading indicator (i.e. the line that spin in a circle wouldn't shrink and expand) in comparison to what product wanted from the spec. I could hold off on this and ask rose when she's back about this change

tmeyer2115: Ignore me.

tmeyer2115 (approved):

PR #1652, opened 2022-01-20
Github action build and unit test workflows

cea2aj:
Create callable build and unit test workflows for GH actions

These are split into callable workflows so that they can be reused for other workflows (e.g. i18n builds, and search bar assets)

Note: A current limitation of callable workflows is that they currently need to specify the full repo and the branch. The PM for Github Actions said that they would be releasing an update in a couple of weeks which would allow us to call workflows from within the same repo. We can keep this migration in a feature branch until they release that feature and then we can update the code and complete the migration

Some additional optimizations include adding caching to NPM node modules and setting up concurrency so that new commits will cancel the GH action jobs for previous commits

J=SLAP-1811, SLAP-1812
TEST=manual

Observe that in the 'Actions' tab the build and unit tests run successfully


.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+      - develop
+      - master
+      - support/**
+      - hotfix/**
+      - feature/**-i18n
+      - release/**
+
+jobs:
+  call_build:
+    uses: yext/answers-search-ui/.github/workflows/build.yml@dev/build-workflow
```
oshi97, line 15: is it okay that this is always pointing to `dev/build-workflow`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
+name: Build Answers assets
+
+on:
+  workflow_call:
```
oshi97, line 4: this overall process seems cool!

.github/workflows/build_and_deploy.yml
cea2aj: No, but the PM for Github Actions reusable workflows said that it will soon be possible to have it point to the same branch: https://github.community/t/ref-head-in-reusable-workflows/203690 Since we aren't in a rush to complete this migration, I was thinking we could hold off on completing the migration until the feature is released. We can point the workflows at the feature branch in the mean time (I can update to point to the feature branch once these workflows are merged into the feature branch)

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,23 @@
+name: Build and deploy # Note: Deployment will be added in a future PR 
+
+on:
+  push:
+    branches-ignore:
+      - develop
+      - master
```
yen-tt, line 7: how come we are ignoring master for this build and deploy workflow?

cea2aj: This is based on our circle CI config where master has a separate deploy process (build_and_hold) which requires manual approval. That config is scheduled for future items

yen-tt: ah gotcha

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  call_build:
+    uses: yext/answers-search-ui/.github/workflows/build.yml@dev/build-workflow
+
+  call_unit_test:
+    uses: yext/answers-search-ui/.github/workflows/unit_test.yml@dev/build-workflow
+    needs: call_build
+
+concurrency:
+  group: ci-build-and-deploy-${{ github.ref }}-1
```
yen-tt, line 22: what's the `-1` for?

oshi97: I think that's something like a version number, similar to our `v3` in our circleci cache key

cea2aj: It's just a version number that we can increment if we ever need to make a new group key for the same github ref. It's not really necessary but it also doesn't hurt to have it. It's similar to a cache key, and it's common to have a number on cache keys so that you can manually bust the cache.

yen-tt (approved):
apart from waiting for the feature release to update workflow path to use local reference from current branch instead of dev/build-workflow. lgtm!

cea2aj:
> apart from waiting for the feature release to update workflow path to use local reference from current branch instead of dev/build-workflow. lgtm!

yep, i'll keep the work in the feature branch for now and update in the next PR to point to the feature branch

PR #1653, opened 2022-01-24
Add GitHub Action deployment

cea2aj:
Add GitHub Action deployment

Add a deploy.yml file and connect it to the build_and_deploy workflow. Comment out the corresponding build_and_deploy from CircleCI. When the migration is complete, we can delete the commented out portions from the CircleCI config. I added the AWS credentials to the repo's secrets. I also set up branch name formatting where any '/' is replaced with '-' in order to match the CircleCI behavior.

J=SLAP-1816
TEST=manual

Observe the build and deploy in the Github Actions panel. Point the theme test site to the assets and see the page work correctly.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -18,31 +18,24 @@ jobs:
     uses: yext/answers-search-ui/.github/workflows/unit_test.yml@feature/gh-action-migration
     needs: call_build
 
-  format_branch_name:
+  call_deploy:
     runs-on: ubuntu-latest
+    #needs: call_unit_test
     steps:
       - name: Format branch name # replace '/' with '-'
         id: vars
         run: |
           FORMATTED_BRANCH="$(echo ${GITHUB_REF_NAME} | sed "s/\//-/g")" 
-          echo $FORMATTED_BRANCH
           echo ::set-output name=formatted_branch::${FORMATTED_BRANCH}
-  print_formatted_branch_name:
-    runs-on: ubuntu-latest
-    needs: format_branch_name
-    steps:
       - name: Print formatted branch name
         run: |
           echo "${{ steps.vars.outputs.formatted_branch }}"
-
-  # call_deploy:
-  #   uses: yext/answers-search-ui/.github/workflows/deploy.yml@dev/gh-s3-deploy
-  #   needs: call_unit_test
-  #   with:
-  #     directory: ${{ github.ref_name }}
-  #   secrets:
-  #     AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  #     AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      # - uses: yext/answers-search-ui/.github/workflows/deploy.yml@dev/gh-s3-deploy
+      #   with:
+      #     directory: ${{ github.ref_name }}
+      #   secrets:
+      #     AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #     AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
 concurrency:
   group: ci-build-and-deploy-${{ github.ref }}-1
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -18,9 +18,10 @@ jobs:
     uses: yext/answers-search-ui/.github/workflows/unit_test.yml@feature/gh-action-migration
     needs: call_build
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -12,12 +12,35 @@ on:
...
+        run: |
+          FORMATTED_BRANCH="$(echo ${GITHUB_REF_NAME} | sed "s/\//-/g")" 
+          echo $FORMATTED_BRANCH
+          echo ::set-output name=formatted_branch::${FORMATTED_BRANCH}
+
+  call_deploy:
+    needs: 
+      - call_unit_test
+      - format_branch_name
+    uses: yext/answers-search-ui/.github/workflows/deploy.yml@dev/gh-s3-deploy
```
cea2aj: When GithubActins supports pointing to the HEAD ref, we'll update this: https://github.community/t/ref-head-in-reusable-workflows/203690

.github/workflows/build_and_deploy.yml
```diff
@@ -12,12 +12,35 @@ on:
 
 jobs:
   call_build:
-    uses: yext/answers-search-ui/.github/workflows/build.yml@dev/build-workflow
+    uses: yext/answers-search-ui/.github/workflows/build.yml@feature/gh-action-migration
```
oshi97: I only thought of this now, but what are the benefits of downloading the build assets instead of having each workflow individually build them? It definitely feels smarter, and it probably uses overall less computation, but since github actions are completely free for public repos, the only benefit I could see is the added flexibility if we wanted to make the SDK private? Is the point that having a separate build workflow let us just have one file to put optimizations and what not in, instead of copy pasting them all over the place? We don't have any optimizations yet but I was thinking we try building multi-lang in parallel across a few github runners

cea2aj: Ya, I think you listed most of the important reasons. It's DRY, and offers strong separation of concerns. Changes to the build happen in one spot. With just this one workflow (build_and_deploy), it's not necessary. However separating these out will make it super easy to create the i18n, search bar, and the versioned release workflows.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: build-output
+          path: dist/
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
```
oshi97, line 39: did we always specify the aws region?

.github/workflows/build_and_deploy.yml
cea2aj: Update: They just released the feature however users are reporting that it's not working yet

.github/workflows/deploy.yml
cea2aj: Yes, the CircleCI environment variables include AWS_REGION: xxxxst-1 which I assume is us-east-1. I checked AWS and assets.sitescdn.net is in us-east-1

.github/workflows/build_and_deploy.yml
cea2aj: While it is free, the max number of concurrent jobs for enterprise is 180, so minimizing computation helps keep us under that. Extra jobs get put in queue. [Source](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration)

oshi97 (approved):

.github/workflows/deploy.yml
tmeyer2115: I'm 99% certain that it is `us-east-1`, but I'm trying to verify the contents of that env var. I'm a bit paranoid when it comes to region stuff. In Okta, when I log into S3, it automatically puts me in the East region. So anything I look at in S3 will already be in that region. That doesn't mean those buckets don't exist in other regions though. 

I don't think AWS_REGION could be some sort of wild card that allows for East and West, but I want to make sure.

tmeyer2115 (approved):

.github/workflows/build_and_deploy.yml
cea2aj: 2nd Update: The feature appears to be working now, so I updated the code!

tmeyer2115 (approved):

.github/workflows/build_and_deploy.yml
```diff
@@ -12,12 +12,35 @@ on:
 
 jobs:
   call_build:
-    uses: yext/answers-search-ui/.github/workflows/build.yml@dev/build-workflow
+    uses: ./.github/workflows/build.yml
```
oshi97, line 15: pog

PR #1655, opened 2022-01-26
support custom prompts in autocomplete

yen-tt:
this pr provide support for user to add custom autocomplete prompts in addition to the autocomplete results from core. 

User can either specify an array of string for custom prompts through searchbar.autocomplete config (`ANSWERS.addComponent('SearchBar', {..., autocomplete: {customPrompts: ['prompt']} }`) or use `setCustomPrompts` (use in theme). In autocomplete's setState, custom prompts will transform into right data format and merge with core results. This is then set in storage to trigger a new state update with the combined result. Custom prompts appear on empty search and persist when user type similar substring as the prompt.

J=SLAP-1803
TEST=manual

- serve page and add SearchBar with customPrompts configuration, with and without hardcoded prompts from backend. See that custom prompts appear as expected. 
  - Type same starting substrings and see that custom prompts still display (levenshtein match) 
  - Type and remove input, see that custom prompts appear again.
- Hook SDK changes to theme. spin up test site and see that custom prompts appear as expected


src/ui/components/search/autocompletecomponent.js
```diff
@@ -179,6 +186,15 @@ export default class AutoCompleteComponent extends Component {
       this._isOpen = true;
     }
 
+    if (this.customPrompts && !queryInputEl.value && data.sections) {
```
tmeyer2115: Does this conditional mean that we don't show the custom prompts if there's some text in the `input`? What if the text in the input matches what's in some of the prompts?

src/ui/components/search/autocompletecomponent.js
```diff
@@ -358,6 +374,23 @@ export default class AutoCompleteComponent extends Component {
     this._isVoiceSearchActive = value;
   }
 
+  setCustomPrompts (prompts) {
+    this.customPrompts = prompts;
+  }
+
+  combineResultsWithCustomPrompts (data) {
+    const newSections = data.sections.filter(result => result.resultsCount !== 0);
```
tmeyer2115: Small nit: Could we break this method up? Maybe have a `createSectionFromCustomPrompts` and then the combine method?

src/ui/components/search/autocompletecomponent.js
yen-tt: Yes. Product was ok with custom prompts disappearing when user start typing. Although, they also ask to match the behavior of the hardcoded prompts that user can configure in the backend, which does include the prompts if input have some sort of text matching. I can add that check using the levenshtein functions we used in headless

src/ui/components/search/autocompletecomponent.js
yen-tt: sure thing

src/ui/components/search/autocompletecomponent.js
```diff
@@ -179,6 +187,18 @@ export default class AutoCompleteComponent extends Component {
       this._isOpen = true;
     }
 
+    // should show IF it's empty search OR levenstien match on a non empty search
```
oshi97: levenshtein

src/ui/components/search/autocompletecomponent.js
```diff
@@ -179,6 +187,19 @@ export default class AutoCompleteComponent extends Component {
...
+        isCloseMatch(prompt, queryInputEl.value)
+      );
+      if (!hasCustomPrompts && matchedCustomPrompts.length > 0) {
+        const dataWithCustomPrompts = this.combineResultsWithCustomPrompts(data, matchedCustomPrompts);
+        this.core.storage.set(`${StorageKeys.AUTOCOMPLETE}.${this.name}`, dataWithCustomPrompts);
+        return;
+      }
+    }
+
     if (wasOpen && !this._isOpen) {
```
oshi97, line 203: Do we need to worry about this wasOpen and isOpen stuff for custom prompts?

src/ui/tools/stringmatching.js
```diff
@@ -0,0 +1,66 @@
+import levenshtein from 'js-levenshtein';
```
oshi97, line 1: Should we update the FilterOptions component to use these functions? For a bit of history, the whole levenshtein stuff we do actually originated there from when Alexis did searchable facets. It's kind of funny how it's come full circle. I don't think it should be too hard to update and test, but if it is I think it's totally fine if we say it's not worth the developer time.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -358,6 +379,28 @@ export default class AutoCompleteComponent extends Component {
     this._isVoiceSearchActive = value;
   }
 
+  setCustomPrompts (prompts) {
+    this.customPrompts = prompts;
+  }
+
+  createSectionForCustomPrompts (customPrompts) {
+    const customAutocompleteResults = customPrompts
+      .map(prompt => AutoCompleteResponseTransformer._transformAutoCompleteResult({ value: prompt }));
```
oshi97: Could we remove the underscore from `_transformAutoCompleteResult` if it's being used "publicly" now?

src/ui/components/componentmanager.js
```diff
@@ -200,6 +200,10 @@ export default class ComponentManager {
     return this._activeComponents.find(c => c.constructor.type === type);
   }
 
+  getActiveComponents (type) {
```
oshi97, line 203: For my edification how is this being used? I didn't see any usages in the fresh code

src/ui/components/search/autocompletecomponent.js
```diff
@@ -179,6 +187,19 @@ export default class AutoCompleteComponent extends Component {
...
 
+    // show custom prompts if it's an empty search or levenshtein match on a non empty search
+    if (this.customPrompts && data.sections) {
+      const hasCustomPrompts = data.sections.find(result => result.isCustom);
+      const matchedCustomPrompts = this.customPrompts.filter(prompt =>
+        isCloseMatch(prompt, queryInputEl.value)
+      );
+      if (!hasCustomPrompts && matchedCustomPrompts.length > 0) {
+        const dataWithCustomPrompts = this.combineResultsWithCustomPrompts(data, matchedCustomPrompts);
+        this.core.storage.set(`${StorageKeys.AUTOCOMPLETE}.${this.name}`, dataWithCustomPrompts);
```
oshi97, line 198: Is this kind of like doing a useLayoutEffect in react, where we update the state without a paint and then trigger a re-render?

src/ui/components/componentmanager.js
oshi97: Ah ok I see it in the hitchhiker-theme PR now

src/ui/tools/stringmatching.js
yen-tt: There are some small differences with FilterOptions (such as needing the offset of substrings in the search term) that would require a bit of changes. I think I would prefer if we make another item/do it in another pr for this, if that's alright?

src/ui/components/search/autocompletecomponent.js
yen-tt: sure!

src/ui/components/search/autocompletecomponent.js
yen-tt: yea, we update the state to include the custom prompts section (this is needed so keyboard navigation and query submission registers the right sections and indexes), and the return statement following this line ensures the old data doesn't get render.

src/ui/tools/stringmatching.js
oshi97: 👍  yeah!

oshi97 (approved):

PR #1656, opened 2022-01-27
use GH actions for acceptance tests

oshi97:
Added ANSWERS.setGeolocation to all acceptance testing fixtures.

Without these, a number of acceptance tests would pretty consistently fail.
This seems to have been due to GH runners having a different location than
CircleCI runners, resulting in different API responses.
We already set the geolocation in the theme's acceptance tests.

I included the useragent tests in the acceptance test workflow, despite them also being considered
part of "miscellaneous tests". I felt like it's nice to have all the browserstack tests in one place.

J=SLAP-1815
TEST=manual,auto

see the acceptance tests run in parallel
saw that the `if` statement in the acceptance script works https://github.com/yext/answers-search-ui/pull/1657


cea2aj (approved):
🚀

.github/run_browserstack_acceptance.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export BROWSERSTACK_USE_AUTOMATE="1"
+export BROWSERSTACK_PROJECT_NAME="Answers SDK"
+export BROWSERSTACK_BUILD_ID="${GITHUB_REF_NAME} - ${GITHUB_RUN_ID} + ${CIRCLE_BUILD_NUM}"
```
nmanu1: Why do we use `CIRCLE_BUILD_NUM` here?

oshi97: oh shoot this was from debugging when I wanted to differentiate between the two in the browserstack automate UI (and I couldn't get GITHUB_BRANCH to show up), I'll remove!

nmanu1 (approved):

PR #1660, opened 2022-01-28
add call_misc_tests GH workflow

oshi97:
Also fixes a bug in the translation test bash script, where a non-zero exit code would only be returned
if the LAST .po file in the for loop had a msgcmp error.

J=SLAP-1815
TEST=manual,auto

tested using https://github.com/yext/answers-search-ui/pull/1659
tests run in a support branch
tests fail if any of the translations in a pot file are missing, in either the last po file or a different one
short circuits if pot file is not up to date

### DIFF
--- a/.github/run_translation_verification.sh
+++ b/.github/run_translation_verification.sh
@@ -24,7 +24,8 @@ if [[ $GITHUB_REF_NAME == release/*
   || $GITHUB_REF_NAME == support/* ]]
 then
   for po_file in *.po
-    do msgcmp $po_file messages.pot
+  do
+    msgcmp $po_file messages.pot
     exit_code=$(( exit_code + $? ))
   done
 else
--- a/.github/run_translation_verification.sh
+++ b/.github/run_translation_verification.sh
@@ -25,8 +25,7 @@ if [[ $GITHUB_REF_NAME == release/*
 then
   for po_file in *.po
   do
-    msgcmp $po_file messages.pot
-    exit_code=$(( exit_code + $? ))
+    msgcmp $po_file messages.pot || exit_code=1
   done
 else
   echo "Skipping the verification that all translations are present"
--- a/.github/run_translation_verification.sh
+++ b/.github/run_translation_verification.sh
@@ -16,6 +16,8 @@ fi
 
 # Verify that translations are present for all languages
 cd conf/i18n/translations
+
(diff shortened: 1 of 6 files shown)


tmeyer2115 (approved):

.github/run_translation_verification.sh
```diff
@@ -16,14 +16,20 @@ fi
...
+if [[ $GITHUB_REF_NAME == release/*
+  || $GITHUB_REF_NAME == hotfix/*
+  || $GITHUB_REF_NAME == master
+  || $GITHUB_REF_NAME == support/* ]]
 then
   for po_file in *.po
-    do msgcmp $po_file messages.pot
+  do
+    msgcmp $po_file messages.pot
+    exit_code=$(( exit_code + $? ))
```
yen-tt: Can this be `msgcmp $po_file messages.pot || exit_code=1` instead incrementing the exit code this way?

tests/acceptance/blocks/filtersearchcomponent.js
```diff
@@ -23,6 +23,6 @@ export default class FilterSearchComponentBlock {
       .typeText(this._input, displayValue)
       .pressKey('space')
       .wait(500)
-      .click(this._selector.find('.js-yext-autocomplete-option').withExactText(displayValue));
```
yen-tt: How come the text from clicking the option does not match the option's exact display value?

.github/run_translation_verification.sh
oshi97: yes, updated! I thought it might be nice to have the exit code reflect how many po files had errors, but that's probably not standard exit code behavior (though from reading online it sounds like exit codes in general are a little fuzzy)

tests/acceptance/blocks/filtersearchcomponent.js
oshi97: think it was just a spurious failure, our current usages should always be able to match exactly. I changed it from `withText` to `withExactText` in my PR right before this

oshi97: if this pops up frequently we can try bumping the selector timeout, since this was a headless acceptance failure we don't have the browserstack video to see what happened during this run

yen-tt (approved):

PR #1661, opened 2022-01-28
test build_and_deploy_hold workflow

yen-tt:
this pr add a build and deploy workflow with hold state. 

- Requires an environment `production` set in repo's settings with a 'require reviewers' protection rule on for team slapshot as qualify members to approve before workflow start deployment.
- `deploy_hold.yml` is made to use the environment `production` setting. The github action `build_and_deploy_hold`, after receiving approval, will deploy three set of assets in folders with name based on tag version, major version and minor version. (/v1.2.3, /v1.2, /v1)
  - For tag release for support branches, such as cases where v1.2.1 is release but there's a newer version v1.3.0 previously release, we don't want to deploy to v1 (major version) and only to v1.2 (minor version) and v1.2.1 (tag version). Added `should_deploy_major_version` job to ensure this behavior

Note: There are some limitations in how secrets and environments can work with reusable workflows (discussions in this [github post](https://github.community/t/reusable-workflows-secrets-and-environments/203695)). There's also a [repo](https://github.com/AllanOricil/workflow-template-bug/blob/master/.github/workflows/workflow-inplementation.yml) with some examples to show issues with environment and secrets when use certain ways.

To summarize:
I can't set `environment` in the top level job and use `uses` to call `deploy` workflow. That's not supported and seems to [error out during testing](https://github.com/yext/answers-search-ui/actions/runs/1763550565/workflow): `workflow is not valid, unexpected value 'uses'`. This mean `environment` have to either:
1) get pass into the reusable workflow `deploy.yml` as an input (similar to my-workflow-job-4 in the example repo):
   - this requires a default `staging` environment without approval restriction and this would spam prs with messages of deploying to staging (as shown in this pr), which could get annoying and cluttering up prs. But if the team is alright with this, I can update to use this approach.
2) hardcode into the workflow (similar to my-workflow-job-3 in the example repo):
   - this is what I have in `deploy_hold.yml` workflow where environment is hardcoded to `production`. Supposedly, when trying to access secrets in an action with an environment specified, github would check for the environment secrets first before checking repo secrets automatically so we don't have to pass in the secrets. But that [didn't work during testing](https://github.com/yext/answers-search-ui/pull/1661/commits/850e86ec1dfe76832586cce08d1a6b9713f790ed) so I kept the secrets as inputs.

I also can't combine all the deploy jobs into one with 'steps' field because `secrets` input field for workflows in `steps` field is not supported. But I think it make sense for them to be separate jobs anyway since they are not dependent on each other, even though it feels a little like duplicating code.

J=SLAP-1818
TEST=manual

push a tag to repo, and see build_and_deploy_hold workflow ran: https://github.com/yext/answers-search-ui/actions/runs/1763784557. Check in s3 buckets from this branch's dir and see the expected assets in folders with major/minor/full versions in there.
push two tags: test-v3.4.0 and test-v3.3.0 in order. See that the first tag deploy all versions and the second tag cancel the major version deployment ([test](https://github.com/yext/answers-search-ui/actions/runs/1774931889))


.github/workflows/build_and_deploy_hold.yml
```diff
@@ -0,0 +1,94 @@
...
+    uses: ./.github/workflows/acceptance.yml
+    needs: call_build
+    secrets:
+      BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
+      BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+
+  call_format_branch_name:
+    uses: ./.github/workflows/format_branch_name.yml
+
+  call_extract_versions:
```
cea2aj: I was using the `call` convention to indicate that we were using the callable workflows feature. Since this is a standard job, could we instead call this `extract_versions`?

.github/workflows/build_and_deploy_hold.yml
```diff
@@ -44,16 +44,36 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: check for possible next minor version from current tag version
+        id: vars
+        run: |
+          MINOR_VERSION=$(echo "${{ github.ref_name }}" | cut -d '.' -f 2) 
+          MAJOR_VERSION=$(echo "${{ github.ref_name }}" | cut -d '.' -f 1)
+          NEXT_MINOR_VERSION=$(( $MINOR_VERSION + 1 ))
+          OUTPUT=$(git tag -l "$MAJOR_VERSION.$NEXT_MINOR_VERSION.*")
```
cea2aj: Does this need 'v' in front of the major version?

cea2aj: nit: Can we use '--list' instead of '-l' to make it slightly clearer?

yen-tt: sure thing

yen-tt: git tag would need 'v' and MAJOR_VERSION should already include 'v' since it just the substring cut off from the first dot in the tag.

.github/workflows/build_and_deploy_hold.yml
```diff
@@ -0,0 +1,116 @@
...
+
+  should_deploy_major_version:
+    runs-on: ubuntu-latest
+    outputs:
+      deploy_major_version: ${{ steps.vars.outputs.deploy_major_version }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: check if the next minor version from current tag version exists
```
cea2aj: nit: Can we update this to indicate when a deploy should happen? In this case, we should deploy the next major version when the next minor doesn't exist

yen-tt: something like this? `allow for major version deployment if next minor version from current tag doesn't exist`

cea2aj: sure, that works

.github/workflows/deploy_hold.yml
```diff
@@ -0,0 +1,45 @@
+name: Deploy assets to AWS S3 with hold state setup in github's production environment
```
oshi97, line 1: for my edification how does the hold state work here?

.github/workflows/build_and_deploy_hold.yml
```diff
@@ -0,0 +1,116 @@
...
+  call_deploy_minor_version:
+    needs: 
+      - call_unit_test
+      - call_format_branch_name
+      - call_acceptance
+      - extract_versions
+      - call_misc_tests
+    uses: ./.github/workflows/deploy_hold.yml
+    with:
+      directory: ${{ needs.extract_versions.outputs.minor_version }}
```
cea2aj, line 108: I was just about to approve but just wanted to double check, if the version is say 1.12.1, will the `minor_version` here be 'v1.12', or '12'? Because we should be deploying into the 'v1.12' directory and not the '12' directory: edit: I missed the `-f 1.2`

yen-tt: major version and minor version output coming from extract_versions job would be proper strings for the deployment directory. So v1, v1.12 and v1.12.1! The strings in should_deploy_major_version job is slightly different and would not be use as part of the dir path

.github/workflows/deploy_hold.yml
yen-tt: There's some explanation in the pr description (i know it's kinda chunky), but basically you can specify an 'environment' field in a github workflow. This environment is created in repo's settings manually. I setup one call 'production' with a 'require reviewers' protection rule on for team slapshot as qualify members to approve before workflow start deployment!

cea2aj (approved):
🔥🔥

.github/workflows/deploy_hold.yml
oshi97: oh ok! I was a little confused but I think I kind of get it now

oshi97 (approved):

PR #1662, opened 2022-01-31
hide voice search icon on mobile except for select browsers

oshi97:
This PR updates the SearchBar to only display the voice search
icon on mobile for select browsers that are known to have support
for it. Previously, the voice search icon could appear even when
SpeechRecognition was not supported, for example on Chrome for iOS.

Since we are now using the Bowser package for user agent sniffing,
we no longer need our userAgent acceptance test suite.
After this PR I can remove them from the gh-action migration feature branch.
edit: it'll probably be better to do this as part of merging the feature branch into develop

J=SLAP-1704
TEST=manual

go on browserstack/personal phone
use mkcert and the http-server with the `-S` flag for https, so that `navigator.mediaDevices` will exist
see that voice search appears for safari ios, chrome android
see that voice search does NOT appear for chrome ios, firefox android
for samsung internet for android add a window.alert to show that the bowser's useragent sniffing works as expected, since this browser does not have a "visit anyway" for https sites with self signed certificates
voice search icon still shows up on edge, chrome, safari desktop browsers, and does not show up on firefox desktop


cea2aj (approved):
LGTM!

PR #1663, opened 2022-02-01
Search bar asset deployment on Github Actions

cea2aj:
Deploy the search bar assets on Github Actions

The logic is based on the logic from the Circle CI config, however this also checks whether or not the major version should be deployed.

The search bar relies on much of the same deploy logic as the regular build_and_deploy_hold workflow, so I extracted out the extract_version and the should_deploy_major_version workflows. In order to support the search bar, I added an `ignore_prefix` to those workflows.

Remove the circle CI config because this is the last piece of the migration.

J=SLAP-1819
TEST=manual

For testing, update the search bar deploy workflow to run based off of the branch name, and create a search-bar-v0.1.0 branch and see it deploy v0, v0.1, and v0.1.0. Create a test tag to confirm the build and deploy hold script didn't break


.github/workflows/build_and_deploy_hold.yml
```diff
@@ -24,52 +24,19 @@ jobs:
...
-          then
-            echo ::set-output name=deploy_major_version::true
-          fi
+  call_should_deploy_major_version:
+    uses: ./.github/workflows/should_deploy_major_version.yml
 
   call_deploy_tag_version:
     needs: 
       - call_unit_test
       - call_format_branch_name
```
yen-tt: should remove this since we don't need it, or this would error out without the job defined

.github/workflows/build_and_deploy_search_bar.yml
```diff
@@ -0,0 +1,80 @@
...
+    uses: ./.github/workflows/extract_versions.yml
+    with:
+      ignore_prefix: search-bar-
+
+  call_should_deploy_major_version:
+    uses: ./.github/workflows/should_deploy_major_version.yml
+    with:
+      ignore_prefix: search-bar-
+
+  call_deploy_tag_version:
```
yen-tt: nit: can we rename to call_deploy_full_version the two deploy and hold workflow? mainly since we ignore prefix 'search-bar-' in tag version for this case

.github/workflows/extract_versions.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      minor_version: ${{ steps.vars.outputs.minor_version }}
+      major_version: ${{ steps.vars.outputs.major_version }}
+      full_version: ${{ steps.vars.outputs.full_version }}
+    steps:
+      - name: extract major and minor version substrings
+        id: vars
+        run: |
+          MAJOR_VERSION="$(echo "${GITHUB_REF_NAME##${{ inputs.ignore_prefix }}}" | cut -d '.' -f 1)"
```
yen-tt, line 29: does `##` mean removing whatever specified on the right from the left side?

.github/workflows/build_and_deploy_search_bar.yml
cea2aj: Sure, I can rename. Are you suggesting to rename it in this search bar workflow and in the regular build_and_deploy_hold workflow?

.github/workflows/extract_versions.yml
cea2aj: In this case the `##` removes the inputs.ignore_prefix from the `GITHUB_REF_NAME`. So if the tag is search-bar-v1.12.0, the prefix would be ignored and you would get v1 as the major version. If the ignore_prefix is empty, the GITHUB_REF_NAME will be used (no effect)

.github/workflows/build_and_deploy_hold.yml
cea2aj: ah, ya I can remove this

.github/workflows/build_and_deploy_search_bar.yml
yen-tt: yea

.github/workflows/extract_versions.yml
yen-tt: gotcha, I'm just curious what the syntax ## in general would do. sounds good!

tmeyer2115 (approved):

PR #1668, opened 2022-02-11
allow search to occur if "near me" request fails

yen-tt:
this pr update the middleware function `promptForLocation` to gracefully handle rejected promises related to fetching near me intents or getting user location. 

If fails, a search request should still get executed. To do so, catch in the then() call chain to handle rejected promise cases and provide a console warn message. Since the handler function doesn't return anything, the promise returned by then gets resolved with an undefined value.

J=SLAP-1885
TEST=manual

return a rejected promise for `fetchQueryIntents` function to mock a failed autocomplete request to compute near me intents. See that a console warn message appear indicating the error, but the search is still executed and see that results for the query is shown on page.

### DIFF
--- a/src/ui/components/search/searchcomponent.js
+++ b/src/ui/components/search/searchcomponent.js
@@ -579,30 +579,35 @@ export default class SearchComponent extends Component {
   promptForLocation (query) {
     if (this._promptForLocation) {
       return this.fetchQueryIntents(query)
-        .then(queryIntents => queryIntents.includes('NEAR_ME'))
-        .then(queryHasNearMeIntent => {
-          if (queryHasNearMeIntent && !this.core.storage.get(StorageKeys.GEOLOCATION)) {
-            return new Promise((resolve, reject) =>
-              navigator.geolocation.getCurrentPosition(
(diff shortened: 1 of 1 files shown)


src/ui/components/search/searchcomponent.js
```diff
@@ -579,30 +579,35 @@ export default class SearchComponent extends Component {
...
+                    const { enabled, message } = this._geolocationTimeoutAlert;
+                    if (enabled) {
+                      alert(message);
+                    }
+                  },
+                  this._geolocationOptions)
+              );
+            }
+          },
+          error => console.warn('Unable to determine user\'s location. Error occured using geolocation API.', error)
```
tmeyer2115: I don't think this will actually catch an error thrown by the `Geolocation` block above. That's because it's part of the same `then` call. This `onRejected` would only fire if the proceeding `then` call (checking for near me intent) is rejected.

yen-tt: right, updated!

tmeyer2115 (approved):

PR #1669, opened 2022-02-14
Gracefully handle LiveAPI failures

cea2aj:
Update the code to prevent live api failures from causing unhandled promise rejections resulting in a broken search experience

Promise chains needs a `.catch`, or a rejected handler or else an unhandled promise rejection event will be sent to the window. Adding a catch will prevent that issue. I added the caches to our requests in core.js. Note: If someone adds a `.then` to one of our network requests, the `.then` will execute even if the rejection handler in the catch block was called. In this case, the data passed to the `.then` handler will be undefined. That is why I added null checking to the search component after checking for the near me intent.

The question submission component was built to read the live API error code so that it could display the proper error message. Because of that, I made the questionSubmission function throw the error in the catch block so that the Question submission component could read the error.

J=SLAP-1884
TEST=manual

Mock the core library's universalAutocomplete function to return a rejected promise. Use a local theme test site and point it to this SDK version. See that after this code change, the experience doesn't break and regular search queries can be fired and the results will update. Test the question submission and see that the error message shows up properly and doesn't break the experience.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -303,6 +303,8 @@ export default class Core {
         }
         this.updateHistoryAfterSearch(queryTrigger);
         window.performance.mark('yext.answers.verticalQueryResponseRendered');
+      }).catch(error => {
+        console.error('Vertical search failed with the following error: ' + error);
       });
   }
 
@@ -389,6 +391,8 @@ export default class Core {
         }
         this.updateHistoryAfterSearch(queryTrigger);
         window.performance.mark('yext.answers.universalQueryResponseRendered');
+      }).catch(error => {
+        console.error('Universal search failed with the following error: ' + error);
       });
   }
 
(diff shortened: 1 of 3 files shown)


src/core/core.js
```diff
@@ -510,6 +520,9 @@ export default class Core {
         this.storage.set(
           StorageKeys.QUESTION_SUBMISSION,
           QuestionSubmission.submitted());
+      }).catch(error => {
+        console.error('Question submission failed with the following error: ' + error);
+        throw error;
```
cea2aj, line 525: I'm opting for the slight inconsistency of only throwing the error for question submission because only the question submission component needs to be able to read the error code. If we were to throw the error for all network requests, all consumers of those functions would need to use a .catch block or else the unhandled promise rejection issue could pop up again. If we would prefer to be 100% consistent, then I can have all of these functions throw and i'll update the consumers to have catch blocks

tmeyer2115: I think it's okay to have this inconsistency. It's no big deal. I'd rather the file be slightly inconsistent than have to add `.catch` blocks everywhere else with the potential for code duplication around error handling.

tmeyer2115 (approved):

PR #1678, opened 2022-03-09
add FOLLOW_UP_QUERY analytics event

yen-tt:
Update core.js' search and verticalsearch function to send a FOLLOW_UP_QUERY analytics event for subsequent searches from the initial search. The previous queryId is used in the analytics event instead.

J=SLAP-1916
TEST=manual

See that initial query did not fire a FOLLOW_UP_QUERY analytics event.
See that the second query did fire a FOLLOW_UP_QUERY analytics event containing the previous queryId and correct searcher type.
See that analytic event did fire from a new search due to a filter change (new queryId generated).
See that analytic event did not fire from a new search due to pagination (no new queryId generated)

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -175,6 +175,21 @@ export default class Core {
   }
 
   /**
+   * Send a FOLLOW_UP_QUERY analytics event for subsequent searches from the initial search.
+   * The following search must contains a different query id from the previous search.
+   *
+   * @param {string} newQueryId - id of the new query
+   * @param {string} searcher - searcher type of the new query ("UNIVERSAL" or "VERTICAL")
+   */
+  _reportFollowUpQueryEvent (newQueryId, searcher) {
+    const previousQueryId = this.storage.get(StorageKeys.QUERY_ID);
+    if (previousQueryId && previousQueryId !== newQueryId) {
+      const event = new AnalyticsEvent('FOLLOW_UP_QUERY').addOptions({ searcher });
+      this._analyticsReporter.report(event);
+    }
+  }
+
+  /**
    * Search in the context of a vertical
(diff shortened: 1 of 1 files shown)


src/core/core.js
```diff
@@ -234,6 +235,12 @@ export default class Core {
     const queryTrigger = this.storage.get(StorageKeys.QUERY_TRIGGER);
     const queryTriggerForApi = this.getQueryTriggerForSearchApi(queryTrigger);
 
+    const queryId = this.storage.get(StorageKeys.QUERY_ID);
+    if (queryId) {
+      const event = new AnalyticsEvent('FOLLOW_UP_QUERY').addOptions({ searcher: Searcher.VERTICAL });
```
oshi97: I think this will also trigger for pagination changes, just checking that that's what we want?

oshi97: also filter changes

yen-tt: for talking to product, they want to log this when there's a new queryId. So filter changes would constitute a follow_up_query but not pagination. Updated!

cea2aj (dismissed):
LGTM

src/core/core.js
```diff
@@ -370,6 +377,13 @@ export default class Core {
       })
       .then(response => SearchDataTransformer.transformUniversal(response, urls, this._fieldFormatters))
       .then(data => {
+        const previousQueryId = this.storage.get(StorageKeys.QUERY_ID);
```
oshi97: can we consolidate this logic somewhere? maybe a shared function somewhere on the class?

yen-tt: sure!

cea2aj (approved):

PR #1684, opened 2022-03-15
implement mobile/desktop visibilitychange workflow

yen-tt:
As part of the search duration metrics epic, this pr add implementations for HIDDEN_RESULTS/UNHIDDEN_RESULTS for desktop and mobile (with some exceptions in ios). The implementation relies on VisibilityChange API and unload event.

Browser specific implementation details on **desktop**:
- Safari: When user place a search, switch to new tab, and eventually close browser (UC 1.5), safari desktop would trigger visibility change event twice (for switching tab, and closing browser). Product said they uses the final event to calculate duration. So, `documentVisibilityState` variable is used to ensures HIDDEN_RESULTS is sent only if page is not hidden (not to fire another event if user already navigate away from the search page).
- IE11: IE11 does not trigger VisibilityChange listener when user close tab/browser. As suggested in MDN webdocs, pagehide event should be the next best alternatives. However, when testing it, pagehide does not fire as expected (strangely it only works when I have unload listener set as well..). So, unload event is used along with VisiblityChange API to track page closing scenarios, which seems consistent so far during testing. With two listeners on ie11, similar issue occur with Safari on UC 1.5. This is resolve with `documentVisibilityState`.

Browser specific details on **mobile**:
- Firefox: close tab/browser doesn’t work.
- Safari: close tab doesn’t work.
- Safari & Chrome: can’t produce consistent results when close from app switcher mode. Certain cases (e.g. switch tab, return to search page, and then close from app switcher) never fire visibilityChange API event, and some cases (e.g. search and then close app switcher without switch tab) was seen firing a few times only.

After some discussions with product, it's decided we would have partial ios support for now with what is currently working and will make a follow up epic to address the rest of the use cases.

J=SLAP-1917
TEST=manual

Setup for the following tests: 
Serve html page that uses local SDK bundle. Use postman mock server to track mock analytics event firing. Use ngrok to test local changes in ios and android. Afterward, confirm that real analytics endpoint requests send successfully by switching tab in chrome desktop.

Screen Shot 2022-03-15 at 1 19 17 PM
Screen Shot 2022-03-15 at 1 19 52 PM
Screen Shot 2022-03-15 at 1 32 28 PM

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -772,8 +772,8 @@ class Answers {
       return;
     }
     const searcher = verticalKey ? Searcher.VERTICAL : Searcher.UNIVERSAL;
-    const event2 = new AnalyticsEvent(eventName).addOptions({ searcher });
-    this._analyticsReporterService.report(event2);
+    const event = new AnalyticsEvent(eventName).addOptions({ searcher });
+    this._analyticsReporterService.report(event);
   }
 }
 
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -33,6 +33,7 @@ import { COMPONENT_REGISTRY } from './ui/components/registry';
 import { localizedDistance, parseLocale } from './core/utils/i18nutils';
 import createImpressionEvent from './core/analytics/createimpressionevent';
 import Searcher from './core/models/searcher';
+import { isIE } from './core/utils/useragent';
 
 /** @typedef {import('./core/services/errorreporterservice').default} ErrorReporterService */
 /** @typedef {import('./core/services/analyticsreporterservice').default} AnalyticsReporterService */
@@ -271,7 +272,7 @@ class Answers {
 
       this.components.setAnalyticsReporter(this._analyticsReporterService);
       initScrollListener(this._analyticsReporterService);
-      initVisibilityChangeListener(this._analyticsReporterService, parsedConfig.search?.verticalKey);
+      this._initVisibilityChangeListeners(parsedConfig.search?.verticalKey);
     }
 
     this.core = new Core({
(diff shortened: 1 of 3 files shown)


src/answers-umd.js
```diff
@@ -726,6 +727,54 @@ class Answers {
...
+  /**
+   * Send visibility change related analytics event.
+   */
+  _reportVisibilityChangeEvent (verticalKey, eventName) {
+    const queryId = this._analyticsReporterService.getQueryId();
+    if (!queryId) {
+      return;
+    }
+    const searcher = verticalKey ? Searcher.VERTICAL : Searcher.UNIVERSAL;
+    const event2 = new AnalyticsEvent(eventName).addOptions({ searcher });
```
oshi97: why event2 instead of just event?

src/answers-umd.js
```diff
@@ -726,6 +727,54 @@ class Answers {
       console.error(`Invalid visitor. Visitor was not set because "${JSON.stringify(visitor)}" does not have an id.`);
     }
   }
+
+  /**
+   * Initialize visibility change event listener(s) to send analytics
+   * events when user close or navigate away from a result page.
+   */
+  _initVisibilityChangeListeners (verticalKey) {
```
oshi97: Maybe this init method could go in its own file? Part of me wants to say that using a `let`-declared variable to hold state is hacky but I think I actually like it lol. Alternatively we could have a class or something

src/answers-umd.js
```diff
@@ -726,6 +727,54 @@ class Answers {
...
+    });
+
+    /**
+     * VisibilityChange API does not register when page is terminated (close tab/browser) in IE11.
+     * So, unload event is used to capture those RESULTS_HIDDEN scenarios.
+     */
+    if (isIE()) {
+      window.addEventListener('unload', () => {
+        if (documentVisibilityState !== 'hidden') {
+          documentVisibilityState = 'hidden';
```
oshi97: do we need to update the documentVisibilityState if the page is about the be closed?

oshi97: I also feel like it could be nice if somebody closing the page is a different analytics dimension, I'm kinda curious if the spec says anything about that but eh

src/answers-umd.js
yen-tt: ah I was testing something, and forgot to rename it. Updated!

src/answers-umd.js
yen-tt: are you suggesting another analytics event when page terminate instead of in background? That could be neat but the specs don't seems to mention about differentiating that. I think the current plan with search duration calculations is just based on RESULTS_HIDDEN and RESULTS_UNHIDDEN only (and follow up query events) so they are treating hide and close page to be the same thing.

yen-tt: Good point, if it's closing anyway, I guess we don't need to update documentVisibilityState. I will remove it

yen-tt: Good point. If it's closing anyway, we don't need to update documentVisibilityState. I will remove it

src/answers-umd.js
yen-tt: Moved into a class!

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -0,0 +1,60 @@
...
+      }
+    });
+
+    /**
+     * VisibilityChange API does not register when page is terminated (close tab/browser) in IE11.
+     * Unload event is used to capture those RESULTS_HIDDEN scenarios.
+     */
+    if (isIE()) {
+      window.addEventListener('unload', () => {
+        if (this._documentVisibilityState !== 'hidden') {
```
oshi97, line 41: do we need this internal variable or can we just rely on `document.visibilityState`?

yen-tt: document.visibilityState would reflect the new visibility state (in this case 'hidden'). We would need the internal variable to know what was the visibility state before visibilitychange/unload listener was triggered. If it's a transition from visible to hidden, then we fire an analytics event here and in line 28.

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -0,0 +1,60 @@
...
+          this._reportVisibilityChangeEvent('RESULTS_HIDDEN_UNLOAD');
```
oshi97: is RESULTS_HIDDEN_UNLOAD an ie11 specific event? it looks like it's only ever fired here?

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Safari desktop listener and IE11 listeners fire visibility change event twice when switch
+     * to new tab and then close browser. This variable is used to ensure RESULTS_HIDDEN analytics event
+     * does not get send again if the page is already hidden.
+     */
+    document.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'hidden' && this._documentVisibilityState !== 'hidden') {
+        this._documentVisibilityState = 'hidden';
+        this._reportVisibilityChangeEvent('RESULTS_HIDDEN');
+      } else if (document.visibilityState === 'visible') {
```
oshi97, line 29: do we need to check that `this._documentVisibilityState` is not `'visible'` here as well?

src/core/analytics/visibilityanalyticshandler.js
oshi97: ohh ok got it, the point is so we don't get extra events

src/core/analytics/visibilityanalyticshandler.js
yen-tt: omg, removed it but didn't save file so it didn't get to the commit. no it's just for testing purposes after refactoring into a class

src/core/analytics/visibilityanalyticshandler.js
yen-tt: I don't see any scenarios where it go from visible to visible state that it would double fire the analytics event so I believe we don't need to check it. The reason it is needed for the hidden state is that safari desktop and unload listener from ie11 trigger visibility change event when browser is close even through user have switch to a different tab so there would be a chance of it going from hidden to hidden state again.

oshi97 (approved):

PR #1686, opened 2022-03-16
pass SDK-Client HTTP header and allow custom key-value pairs

oshi97:
Our tests are emitting a number of console.errors, however it looks like
this has been the case for some time now. It's probably not worth making an item
to fix these since we're deprecating the SDK, and the tests still test what they need to.

I tried using the compact and preferConst options for rollup/json-plugin, but they didn't
decrease the bundle size at all.

J=SLAP-1948
TEST=manual,auto

test that universal/vertical autocomplete and searches, filter search,
and question submission all pass the Client-SDK header and any custom headers
(tested both with and without customClientSdk set in ANSWERS.init)
tested this works with the hitchhiker-theme's test site

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ function initAnswers() {
     // Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD', 'OVERLAY', or arbitrary strings. Defaults to 'STANDARD'
     querySource: 'STANDARD',
     // Optional, additional key-value pairs for the Client-SDK HTTP header.
-    customSdkClient: {}
+    customClientSdk: {}
   })
 }
 ```
--- a/README.md
+++ b/README.md
@@ -174,8 +174,12 @@ function initAnswers() {
     disableCssVariablesPonyfill: false,
     // Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD', 'OVERLAY', or arbitrary strings. Defaults to 'STANDARD'
     querySource: 'STANDARD',
-    // Optional, additional key-value pairs for the Client-SDK HTTP header.
-    customClientSdk: {}
+    // Optional, additional values for the HTTP headers listed below. Headers other than those listed below will be ignored.
+    additionalHttpHeaders: {
+      // Additional key-value pairs to send in the Client-SDK header.
+      // ANSWERS_CORE and ANSWERS_SEARCH_UI_SDK are automatically set and cannot be overwritten.
+      'Client-SDK': {}
+    }
   })
 }
 ```
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ function initAnswers() {
     disableCssVariablesPonyfill: false,
     // Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD', 'OVERLAY', or arbitrary strings. Defaults to 'STANDARD'
     querySource: 'STANDARD',
+    // Optional, additional key-value pairs for the Client-SDK HTTP header.
+    customSdkClient: {}
   })
 }
 ```
--- a/conf/gulp-tasks/bundle/bundle.js
+++ b/conf/gulp-tasks/bundle/bundle.js
@@ -11,6 +11,7 @@ const {
   TRANSLATION_FLAGGER_REGEX,
   SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE
 } = require('../../i18n/constants');
+const rollupJson = require('@rollup/plugin-json');
 
 const TranslateCallParser = require('../../i18n/translatecallparser');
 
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -174,6 +174,8 @@ function initAnswers() {
     disableCssVariablesPonyfill: false,
     // Optional, the analytics key describing the Answers integration type. Accepts 'STANDARD', 'OVERLAY', or arbitrary strings. Defaults to 'STANDARD'
     querySource: 'STANDARD',
+    // Optional, additional key-value pairs for the Client-SDK HTTP header.
```
tmeyer2115: Maybe note that the SDK's agent will be include by default. I'm still not totally sold on the naming referring to the Snowflake keys. `additionalAnswersAgents` feels a bit more intuitive and user-friendly to me.

conf/gulp-tasks/bundle/bundle.js
```diff
@@ -11,6 +11,7 @@ const {
   TRANSLATION_FLAGGER_REGEX,
   SPEECH_RECOGNITION_LOCALES_SUPPORTED_BY_EDGE
 } = require('../../i18n/constants');
+const rollupJson = require('@rollup/plugin-json');
```
tmeyer2115, line 14: For my own edification, was this necessary for the PR or was this a separate attempt at optimizing our bundle size?

README.md
tmeyer2115: I think maybe we should get Product to be the tie-breaker here. Ask Rose's opinion.

conf/gulp-tasks/bundle/bundle.js
oshi97: it was necessary because answers-core now imports its own package.json in its src, which typescript just puts directly in the dist folder
I'm not entirely happy that if somebody wants to use rollup with core they need to include an extra plugin
if you want we can revisit the answers-core implementation

oshi97: this is what the answers-core looks like in the node_modules now
image

src/core/core.js
```diff
@@ -121,6 +121,9 @@ export default class Core {
      * @type {ComponentManager}
      */
     this._componentManager = config.componentManager;
+
+    /** @type {{ [agent: string]: string, ANSWERS_CORE?: never, ANSWERS_SEARCH_UI_SDK?: never }} */
```
yen-tt: I see that in answers-core, CustomClientSDK have the `agent` field as `[agent: string]: string | undefined`, can it be undefined?

src/core/core.js
```diff
@@ -815,4 +824,11 @@ export default class Core {
     }
     return null;
   }
+
+  _getMergedClientSdk () {
+    return {
+      ...this._customClientSdk,
+      ANSWERS_SEARCH_UI_SDK: LIB_VERSION
```
yen-tt: to confirm my own understanding, we are getting this LIB_VERSION during the build process by checking git tag?

oshi97: yup!

src/core/core.js
oshi97: yeah, I'll update after Rose gets back with the updated config name

src/core/utils/mergeAdditionalHttpHeaders.js
```diff
@@ -0,0 +1,25 @@
...
+ * Merges a given additionalHttpHeaders with the SDK's built in value,
+ * and performs run-time validation.
+ *
+ * @param {Record<string, string>} additionalHttpHeaders
+ * @returns {import('@yext/answers-core').AdditionalHttpHeaders}
+ */
+export function mergeAdditionalHttpHeaders (additionalHttpHeaders = {}) {
+  const ignoredHeaders = Object.keys(additionalHttpHeaders).filter(v => v !== CLIENT_SDK_HEADER);
+  ignoredHeaders.forEach(h => console.warn(
+    `Found unknown HTTP header "${h}", ignoring it. ` +
```
yen-tt: nit: can we rephrase this sentence? maybe like `Ignored unknown HTTP header "${h}"`

src/core/utils/mergeAdditionalHttpHeaders.js
```diff
@@ -0,0 +1,25 @@
...
+    `Only the "${CLIENT_SDK_HEADER}" header is recognized in additionalHttpHeaders.`
+  ));
+
+  return {
+    [CLIENT_SDK_HEADER]: {
```
yen-tt, line 36: I know we have specific typings in core to prevent ANSWERS_CORE in the object, but for here with js, do we need to manually check if ANSWERS_CORE is provided and should be ignored if it is?

src/core/utils/mergeAdditionalHttpHeaders.js
oshi97: I like that better! more concise

src/core/utils/mergeAdditionalHttpHeaders.js
oshi97: added a bunch more runtime checking! think Rose wants that as well

yen-tt (approved):

PR #1687, opened 2022-03-16
fix for having both google and mapbox maps on the same page

oshi97:
Before this fix, you could not have both a google and a mapbox map
on the page at the same time, because they were using the same DOM id.

Giving them unique ids lets them exist simultaneously. This is a very
rare usecase that probably isn't worth a hotfix.

J=SLAP-1969
TEST=manual

hooked up local SDK to theme test-site, saw both google and mapbox map on the
same universal search page

### DIFF
--- a/src/ui/components/map/providers/googlemapprovider.js
+++ b/src/ui/components/map/providers/googlemapprovider.js
@@ -81,7 +81,7 @@ export default class GoogleMapProvider extends MapProvider {
       return;
     }
 
-    let script = DOM.query('#yext-map-js');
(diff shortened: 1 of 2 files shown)


oshi97:
You can see this issue if you go to https://master-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=
the google map does not load and throws an error, while the mapbox one below it loads fine.
it's possible the api results shuffled around to end up with two maps on the universal page, or that we just never noticed one of the maps was not functioning

yen-tt (approved):

PR #1689, opened 2022-03-18
use npm overrides to update dependencies

oshi97:
This PR updates a number of dependencies to reduce snyk detected vulnerabilities.
Some packages could not be directly updated to remove vulnerabilities in transient deps, due to no newer versions existing.
For example, the latest version gulp-rtlcss uses an older version of rtlcss which is still on an older postcss version.

Instead, we can use the new npm "overrides" option, which lets you override the version of a transient dependency.
This feature was added in npm v8, so in order to take advantage of it I updated the package-lock to version 2, and also updated the circleci containers to use node 16.

Using npm overrides lets us drop usage of the npm-force-resolutions package. Dropping npm-force-resolutions should also fix issues issues with installing answers-search-ui via npm. Currently, answers-search-ui needs to be installed using the `npm install --ignore-scripts` flag. This is due to our npm-force-resolutions preinstall script, which looks for the SDK's package-lock.json, which isn't currently being tracked by our npm package.

I tried to update jest from 24 to 27, however even upgrading from 24 to 25 results in issues with the sortoptionscomponent tests. I tried updating our enzyme adapter's simulateEvent handler, but was not able to resolve the issue. 
note: when upgrading to jest 27, it's necessary to manually specify the `testEnvironment` as "jsdom"

J=SLAP-1937
TEST=manual

smoke tested a test site
hooked up to a local theme test-site, smoke tested general functionality


src/ui/dom/dom.js
```diff
@@ -95,19 +95,19 @@ export default class DOM {
   /**
    * createEle will create a {HTMLElement} and apply the properties attributes through an object provided.
    * @param {string} el The element `tag` name to construct
-   * @param {Object} opts_data Optional attributes to apply to the new HTMLElement
+   * @param {Object} optsData Optional attributes to apply to the new HTMLElement
    */
-  static createEl (el, opts_data = {}) {
+  static createEl (el, optsData = {}) {
```
oshi97, line 100: this change was originally from when I tried to update some of the eslint config dependencies we inherit from, from when I tried to update our eslint- related dependencies without using npm overrides

I think we don't need it anymore but I think I prefer having opts_data in camel case anyways

.github/workflows/wcag_test.yml
```diff
@@ -6,11 +6,11 @@ on:
 
 jobs:
   WCAG-test:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
nmanu1: for my own understanding, why is this change needed?

oshi97: I was running into this error when running the wcag tests via CI, but couldn't reproduce locally
https://github.com/yext/answers-search-ui/runs/5606193735?check_suite_focus=true
from looking things up online it seems to be a lower level error that people don't really know the solution for, so I started off by seeing if making the environment closer to my local one, and for whatever reason it seems to be working now

oshi97: eh seems to be working now, maybe was a github error?

.postcssrc.json
```diff
@@ -6,7 +6,7 @@
       "mediaQuery": true
     },
     "autoprefixer": {
-      "browsers": ["last 2 versions"],
+      "overrideBrowserslist": ["last 2 versions"],
```
yen-tt, line 9: was looking at what this do in the [docs](https://github.com/postcss/autoprefixer#options), looks like its use is discouraged? The best practice they recommend is to use .browserslistrc config or browserslist key in package.json

tests/setup/setup.js
```diff
@@ -1,5 +1,5 @@
 import { configure } from 'enzyme';
 import AnswersAdapter from './enzymeadapter';
-import 'regenerator-runtime/runtime';
-
+import regeneratorRuntime from 'regenerator-runtime/runtime';
+global.regeneratorRuntime = regeneratorRuntime;
```
yen-tt, line 4: I thought it's already defined globally with the way we import before. why do we need to update it to set to global object this way?

conf/npm/generateentrypoints.js
```diff
@@ -46,7 +46,7 @@ function appendExportsField (inputFile) {
     }
   });
   Object.assign(packageData, { exports: exports });
-  fs.writeFileSync(inputFile, JSON.stringify(packageData, null, 2));
+  fs.writeFileSync(inputFile, JSON.stringify(packageData, null, 2) + '\n');
```
yen-tt, line 49: noice

tests/setup/setup.js
oshi97: it was very strange, but we actually had different versions of regenerator-runtime in our repo EVEN with npm-force-resolutions
this was probably due to having 0.13.7 in our package.json but 0.13.1 in our "resolutions"
image

we were using 0.13.1 for any transient deps, but our direct dep was using 0.13.9, which is the one used here in test setup
When using npm overrides, npm will complain if you try to override a version that doesn't match up with your package.json. For example, prior to this PR we had 0.13.7 in our package.json but 0.13.1 in our "resolutions". npm didn't like that, so we're now using 0.13.1 for both our source code AND our test setup.

With regards to test setup, 0.13.1 and 0.13.9 need to be specified differently. Since the package is still v0 the maintainer was free to make breaking changes like this between 0.13.1 and 0.13.9

.postcssrc.json
oshi97: I tried it out and it looks like that might affect the js file contents? The docs say that it's recommended since it also can apply to babel and stylelint. But since we weren't doing it before, it's possible this could change some behavior? I took a sha1 hash of answers.js before and after adding browserslist to the package.json and got different values, while the hash for answers.css was identical before and after. I want to push off adding a shared browserslist setting to later if that's okay?

yen-tt: sure, maybe make an item or add to the idea sheet?

yen-tt (approved):

.postcssrc.json
oshi97: added to the sheet!

PR #1692, opened 2022-03-25
fix json parsing from eventoptions data in autocomplete component

yen-tt:
This pr address a JSON parse error when autocomplete option's text contains special characters, such as double quotes. When the autocomplete value is read by handlebars and placed in `data-eventoptions` to directly construct a JSON object in string form, the special characters are in escaped-HTML character format. When page load, those characters are no longer escaped. This causes problem when component try to perform JSON.parse on `data-eventoptions` string to send an analytic event.

I briefly checked other areas where we might do insert user custom text to data-eventoptions. Beside verticalkey and pagenumbers, it seems like this is the main place where we would run to this problem. We should check the repo, in another item, for instances this approach of passing JSON data through handlebars occur, outside of data-eventoptions attribute.

Screen Shot 2022-03-25 at 10 44 43 AM

J=SLAP-2013
TEST=manual

see that new jest test failed with previous code, and pass with the new changes
use SDK searchbar in an html page, added a universal prompt with special characters (double quotes) in the text, see that error appear with previous code, and function normally with the new changes. Click on that universal prompt and see that an analytics event with that text value fired.

### DIFF
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -192,8 +192,22 @@ export default class AutoCompleteComponent extends Component {
       sectionIndex: this._sectionIndex,
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
-      listLabelIdName: this.listLabelIdName
+      listLabelIdName: this.listLabelIdName,
+      eventOptions: this.eventOptions(data)
     }));
+  }
+
+  eventOptions (data) {
+    const eventOptions = [];
+    data.sections?.forEach(section => {
+      const sectionEventOptions = section.results.map(result => {
+        return {
+          suggestedSearchText: result.value
+        };
+      });
+      eventOptions.push(sectionEventOptions);
+    });
+    return eventOptions;
   }
 
   updateAriaAttribute () {
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -199,7 +199,7 @@ export default class AutoCompleteComponent extends Component {
 
   eventOptions (data) {
     const eventOptions = data.sections?.map(section =>
-      section.results.map(result => {
+      section.results?.map(result => {
         return { suggestedSearchText: result.value };
       })
     );
--- a/src/ui/components/search/autocompletecomponent.js
+++ b/src/ui/components/search/autocompletecomponent.js
@@ -198,15 +198,11 @@ export default class AutoCompleteComponent extends Component {
   }
 
   eventOptions (data) {
-    const eventOptions = [];
-    data.sections?.forEach(section => {
-      const sectionEventOptions = section.results.map(result => {
-        return {
-          suggestedSearchText: result.value
-        };
-      });
-      eventOptions.push(sectionEventOptions);
-    });
+    const eventOptions = data.sections?.map(section =>
+      section.results.map(result => {
+        return { suggestedSearchText: result.value };
+      })
+    );
     return eventOptions;
   }
 
--- a/src/ui/templates/search/autocomplete.hbs
+++ b/src/ui/templates/search/autocomplete.hbs
@@ -53,7 +53,7 @@
               data-short="{{shortValue}}"
               data-filter='{{json filter}}'
               data-eventtype="AUTO_COMPLETE_SELECTION"
-              data-eventoptions='{"suggestedSearchText": "{{value}}"}'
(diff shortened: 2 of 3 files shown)


src/ui/components/search/autocompletecomponent.js
```diff
@@ -192,10 +192,24 @@ export default class AutoCompleteComponent extends Component {
...
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
-      listLabelIdName: this.listLabelIdName
+      listLabelIdName: this.listLabelIdName,
+      eventOptions: this.eventOptions(data)
     }));
   }
 
+  eventOptions (data) {
+    const eventOptions = [];
+    data.sections?.forEach(section => {
```
oshi97: can we use a map for the outer loop too?

src/ui/templates/search/autocomplete.hbs
```diff
@@ -53,7 +53,7 @@
               data-short="{{shortValue}}"
               data-filter='{{json filter}}'
               data-eventtype="AUTO_COMPLETE_SELECTION"
-              data-eventoptions='{"suggestedSearchText": "{{value}}"}'
```
oshi97, line 56: For my edification, does handlebars just automatically remove escape characters? I tried on https://handlebarsjs.com/playground.html and it looks like it replaces them with `&quot'`s?

I was wondering if we could just use triple braces instead to fix this issue? it seeeems to work like we'd want but I haven't tested anything beyond this screenshot
image

oshi97: ^I'm also not sure if this approach would have security issues so maybe just having the explicit `eventOptions` data is better?

yen-tt: I attempted the triple curly braces (didn't work) before going with this approach. I do think having an explicit eventOptions data is better though

yen-tt: Yea, when I logged out value from handlebars log helper in template, the double braces would perform HTML escaped, but when it's rendered on page, it will remove the escaped characters. I think someone tried to do something similar with JSON directly as string and it's [not recommended](https://github.com/handlebars-lang/handlebars.js/issues/1301)

oshi97: got it!

oshi97: thanks for looking into it

oshi97 (approved):

src/ui/components/search/autocompletecomponent.js
```diff
@@ -192,10 +192,20 @@ export default class AutoCompleteComponent extends Component {
...
-      listLabelIdName: this.listLabelIdName
+      listLabelIdName: this.listLabelIdName,
+      eventOptions: this.eventOptions(data)
     }));
   }
 
+  eventOptions (data) {
+    const eventOptions = data.sections?.map(section =>
+      section.results.map(result => {
+        return { suggestedSearchText: result.value };
```
cea2aj: Are we sure that `results` is always an array (as opposed to undefined)? And that `section` and `result` are always objects? It looks like that's the case, but if it's not then this could crash at runtime

yen-tt: from the methods in `autocompleteresponsetransformer.js`, it looks like `section` and `result` should always be objects. The way we generate `results` field there does make sure it's always be an array. But I do see `hasResults()` in autocomplete check if it's defined anyway, so I will add an optional chain guard for results here.

src/ui/components/search/autocompletecomponent.js
```diff
@@ -192,10 +192,20 @@ export default class AutoCompleteComponent extends Component {
...
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
-      listLabelIdName: this.listLabelIdName
+      listLabelIdName: this.listLabelIdName,
+      eventOptions: this.eventOptions(data)
     }));
   }
 
+  eventOptions (data) {
+    const eventOptions = data.sections?.map(section =>
```
cea2aj, line 201: If `data.sections` is falsy and returns undefined, would the lookup in the HBS fail?

yen-tt: the HBS template uses `{{#each sections}}` and `{{#each results}}` to drill into data.sections[index1].results[index2]. If data.sections is undefined, I believe it wouldn't call the lookup functions since it will not enter the two loops at all

cea2aj: Got it. Even if it tried to lookup undefined data, it would simply return nothing

cea2aj (approved):

PR #1698, opened 2022-03-29
Merge master (v1.12.3) into develop

yen-tt:
merge hotfix and search-bar-only acceptance test update from master to develop

J=none
T=none


oshi97 (dismissed):

package-lock.json
```diff
@@ -24785,7 +24785,7 @@
       "integrity": "sha512-L6rvG9GDxaLgFjg41K+5Yv9OMrU98sWe+Ykmc6FDJW/+vYZMhdOMKkISgzptMaERHvS2Y2lw9MDRm2gHhlQQoA==",
       "dev": true,
       "requires": {
-        "regenerator-runtime": "0.13.1"
+        "regenerator-runtime": "^0.13.4"
```
nmanu1: is there a reason the package version changed? I thought it was pinned to 0.13.1

and similarly for the other packages we use overrides for

oshi97: Oo I think master doesn't have the npm overrides changes yet, and is still built using npm 6. For develop I think we want to keep these versions the same as before? I iiimagine running npm install would undo these?

oshi97: I gave it a go and don't think just running npm install sets it back to 0.13.1, interestingly I did an `npm ls regenerator-runtime` and it says all usages are 0.13.1?

yen-tt: Running it in develop has the same result, i'm getting invalid for the 0.13.4 version and deduped for 0.13.1 so I think it's ok?
Screen Shot 2022-03-31 at 5 22 24 PM
Would you all prefer if I revert package-lock (except the SDK version change) to merge this in? We can make a separate item to dig into this maybe

oshi97: sure that works for me

oshi97: The "no unsafe eval CSP" percy snapshot still looks correct so we have some protection
https://percy.io/cd93112c/yext-answers/builds/16917527/search/preview/950769252?browser=firefox&searchParam=unsafe&viewLayout=side-by-side&viewMode=new&width=1280

nmanu1 (approved):

PR #1702, opened 2022-04-04
set SEARCH_COMPLETE state for necessary models when an API error occurs

oshi97:
Now when an API error occurs, the SEARCH_COMPLETE state is set for models
that are set to SEARCH_LOADING.

J=SLAP-1467
TEST=manual

set an invalid apiKey, see the search loading state start and stop as expected
see expected console log for both universal + vertical searches

also do the same thing with an invalid verticalKey for vertical searches

see that no location bias or direct answer renders after an API error

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,14 +522,14 @@ ANSWERS.addComponent('SearchBar', {
     // Optional, callback invoked when the autocomplete component changes from closed to open.
     onOpen: function() {},
   },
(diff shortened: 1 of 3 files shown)


cea2aj (approved):
LGTM. I made the same acceptance test fix here: https://github.com/yext/answers-search-ui/pull/1703

PR #1704, opened 2022-04-05
move aria-expanded attribute from autocomplete div to text input

oshi97:
This commit updates the SearchBar, FilterSearch, and GeolocationFilter
components to have the aria-expanded attribute on the text input instead
of on the container div for the autocomplete.

We cannot use the SDK's setState to do this, because setState does not
work properly for text input, and will reset the user's text cursor location.
This is a known issue.
Instead, an onMount was added to AutoComplete component usages that manually makes
a DOM manipulation.

J=SLAP-2015
TEST=manual

see that for FilterSearch, SearchBar, and GeolocationFilter the input will have
aria-expanded true when there are autocomplete results visible, and false when not visible
tested autocomplete queries with no results
tested autocomplete queries with results that were visible
tested autocomplete queries with results, but that weren't currently visible due to clicking
on the parent document

### DIFF
--- a/src/ui/components/filters/geolocationcomponent.js
+++ b/src/ui/components/filters/geolocationcomponent.js
@@ -224,7 +224,17 @@ export default class GeoLocationComponent extends Component {
       inputEl: inputSelector,
       verticalKey: this._config.verticalKey,
       searchParameters: this.searchParameters,
-      onSubmit: (query, filter) => this._handleSubmit(query, filter)
+      onSubmit: (query, filter) => this._handleSubmit(query, filter),
+      onMount: () => {
+        const inputEl = DOM.query(this._container, inputSelector);
+        if (!inputEl) {
+          return;
+        }
+        const hasAutocompleteResults = this._autocomplete.getState('hasResults');
+        if (inputEl.getAttribute('aria-expanded') !== hasAutocompleteResults) {
+          inputEl.setAttribute('aria-expanded', hasAutocompleteResults);
+        }
+      }
     });
   }
 
--- a/src/ui/components/filters/geolocationcomponent.js
+++ b/src/ui/components/filters/geolocationcomponent.js
@@ -226,14 +226,7 @@ export default class GeoLocationComponent extends Component {
       searchParameters: this.searchParameters,
       onSubmit: (query, filter) => this._handleSubmit(query, filter),
       onMount: () => {
-        const inputEl = DOM.query(this._container, inputSelector);
-        if (!inputEl) {
-          return;
-        }
-        const hasAutocompleteResults = this._autocomplete.getState('hasResults');
-        if (inputEl.getAttribute('aria-expanded') !== hasAutocompleteResults) {
-          inputEl.setAttribute('aria-expanded', hasAutocompleteResults);
-        }
+        this._autocomplete.updateAriaExpanded(DOM.query(this._container, inputSelector));
       }
     });
   }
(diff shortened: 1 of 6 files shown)


src/ui/components/filters/geolocationcomponent.js
```diff
@@ -224,7 +224,17 @@ export default class GeoLocationComponent extends Component {
...
       searchParameters: this.searchParameters,
-      onSubmit: (query, filter) => this._handleSubmit(query, filter)
+      onSubmit: (query, filter) => this._handleSubmit(query, filter),
+      onMount: () => {
+        const inputEl = DOM.query(this._container, inputSelector);
+        if (!inputEl) {
+          return;
+        }
+        const hasAutocompleteResults = this._autocomplete.getState('hasResults');
+        if (inputEl.getAttribute('aria-expanded') !== hasAutocompleteResults) {
```
tmeyer2115: Can the AutoComplete Component mount initially with no results? I'm asking because if it does, then `aria-expanded` would not initially be present on the `inputEl`, that's different than having `aria-expanded` be present and set to `false`. I'm wondering if that has WCAG implications.

oshi97: from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#menu
> When the menu is hidden, aria-expanded can be omitted. If specified when the menu is hidden, it should be set as aria-expanded="false".

So I think having `aria-expanded="false"` even when no autocomplete results exist is correct behavior! It's also the current behavior of the SDK, prior to this change.

src/ui/components/search/searchcomponent.js
```diff
@@ -566,6 +566,16 @@ export default class SearchComponent extends Component {
...
+      },
+      onMount: () => {
+        const inputEl = DOM.query(this._container, inputSelector);
+        if (!inputEl) {
+          return;
+        }
+        const hasAutocompleteResults = this._autocomplete.getState('hasResults');
+        if (inputEl.getAttribute('aria-expanded') !== hasAutocompleteResults) {
+          inputEl.setAttribute('aria-expanded', hasAutocompleteResults);
+        }
```
yen-tt: do we need to guard this? or can it just be
`inputEl.setAttribute('aria-expanded', !!hasAutocompleteResults);`

src/ui/components/search/filtersearchcomponent.js
```diff
@@ -220,6 +220,16 @@ export default class FilterSearchComponent extends Component {
...
+      },
+      onMount: () => {
+        const inputEl = DOM.query(this._container, inputSelector);
+        if (!inputEl) {
+          return;
+        }
+        const hasAutocompleteResults = this.autoCompleteComponent.getState('hasResults');
+        if (inputEl.getAttribute('aria-expanded') !== hasAutocompleteResults) {
+          inputEl.setAttribute('aria-expanded', hasAutocompleteResults);
+        }
```
yen-tt: maybe it would be good to have some utils file for /search to avoid duplication if it's used in different search component?

src/ui/components/search/searchcomponent.js
oshi97: I didn't want to do extra dom mutations, though I can remove the guard if you want?

oshi97: From some light testing it doesn't look like there's any harm removing the guard (I don't see the elements "flash" in dev tools or anything) but I thought it was a little safer?

src/ui/components/search/filtersearchcomponent.js
oshi97: updated and retested!

src/ui/components/search/searchcomponent.js
yen-tt: I see, if the guard here serves a purpose of avoiding extra dom mutations then that's fine! I just wasn't sure why we didn't directly set it before

yen-tt (approved):

PR #1707, opened 2022-04-05
ensure support for numerical facets

oshi97:
Adds tests for numerical facets.
Also manually tested using hardcoded data.

J=SLAP-2026
TEST=manual,auto

tested setting and applying numerical facets, using fixtures from answers-core

### DIFF
--- a/tests/ui/components/filters/facetscomponent.js
+++ b/tests/ui/components/filters/facetscomponent.js
@@ -7,6 +7,28 @@ import Storage from '../../../../src/core/storage/storage';
 import FilterRegistry from '../../../../src/core/filters/filterregistry';
 import ResultsContext from '../../../../src/core/storage/resultscontext';
 
+function setupDOM() {
+  const bodyEl = DOM.query('body');
+  DOM.empty(bodyEl);
+  DOM.append(bodyEl, DOM.createEl('div', { id: 'test-component' }));
+}
+
+function setupComponentManager() {
+  const storage = new Storage().init();
+  const filterRegistry = new FilterRegistry(storage);
+
+  const mockCore = {
+    verticalSearch: jest.fn(),
+    enableDynamicFilters: jest.fn(),
+    storage: storage,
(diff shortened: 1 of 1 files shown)


tests/ui/components/filters/facetscomponent.js
```diff
@@ -154,3 +160,78 @@ describe('Facets Component', () => {
...
+  const wrapper = mount(component);
+
+  it('renders numerical facets correctly', () => {
+    const optionLabelNodes = wrapper.find('.js-yxt-FilterOptions-optionLabel--name');
+    expect(optionLabelNodes).toHaveLength(2);
+    expect(optionLabelNodes.at(0).text().trim()).toEqual('0 - 23');
+    expect(optionLabelNodes.at(1).text().trim()).toEqual('23 - 60');
+  });
+
+  it('applies numerical facets correctly', () => {
```
cea2aj, line 211: Did you consider checking the params passed to core's `verticalSearch`? It might let us test this without needing to be aware of FACET_FILTER_NODES

oshi97: I tried that originally, but Facets doesn't call `core.verticalSearch` but `core.triggerSearch`, which itself triggers a search not by calling `verticalSearch` but by updating the query param in storage. I can update the mocked triggerSearch to just call verticalSearch though

oshi97: Actually, I think it would be a little messy to try to check the answers-core verticalSearch call, since we would have to use a real instance of the SDK's Core class. Right now we are just using an object with mocked out methods and properties, including `core.verticalSearch` (not `answers-core.verticalSearch`). I updated it to check for `filterRegistry.getFacetsPayload` instead, which ultimately ends up checking the same data, if that's ok?

cea2aj: Sounds good

cea2aj (approved):

PR #1675, opened 2022-02-23
resolve snyk issues

yen-tt:
[Prototype Pollution issue with serve@11.3.2](https://app.snyk.io/org/answers/project/49f79ce2-9224-433d-b61a-5832b913b5d5#issue-SNYK-JS-AJV-584908). Removed since there’s no direct usage of it (testcage uses npx serve)

[Arbitrary Code Injection](https://app.snyk.io/org/answers/project/49f79ce2-9224-433d-b61a-5832b913b5d5#issue-SNYK-JS-UNDERSCORE-1080984) and [ReDos issues](https://app.snyk.io/org/answers/project/49f79ce2-9224-433d-b61a-5832b913b5d5#issue-SNYK-JS-MARKED-584281) with underscore@1.13 from jsdoc@3.6.6. Confirmed with product that docs folder is no longer use. Removed jsdoc.

[Regular Expression Denial of Service issues](https://app.snyk.io/org/answers/project/49f79ce2-9224-433d-b61a-5832b913b5d5#issue-SNYK-JS-STYLELINT-1585622) with stylelint@13.8.0. Updated stylelint package version. instead of stylelint-scss, v14 uses share configs like `stylelint-config-standard-scss`. switch to extending that config (require postcss as peerdepedency)

J=none
TEST=manual

smoke test SDK build in theme test site after lint fix.

PR #1709, opened 2022-04-05
Merge v1.13.0 into develop

cea2aj:
Resolve merge conflicts to merge v1.13.0 into develop


nmanu1 (commented):
Looks like a Snyk check is failing. Did the package-lock changes revert some of the package upgrades for fixing vulnerabilities?

oshi97:
I think it's possible the package-lock changes are upsetting snyk?

cea2aj:
> Looks like a Snyk check is failing. Did the package-lock changes revert some of the package upgrades for fixing vulnerabilities?

I checked the package.json, and the synk fixes from #1675 are still present. I looked through the reported issues in synk, and they are coming from jest@24.9.0, enzyme@3.11.0, @percy/cli@1.0.0-beta.65, and gulp@4.0.2. These versions are present in master and in develop (with the exception of the percy/cli which uses a newer version of the percy/cli in develop). As a result, I believe the synk issues are new issues reported on Snyk for existing dependencies, and not a result of this PR which would be the role of the slapshot responder to address.

oshi97:
should the package-lock be getting updated here? I would have thought develop would have priority

cea2aj:
> should the package-lock be getting updated here? I would have thought develop would have priority

Master uses package-lock version 1 and develop uses version 2, so there were significant conflicts. I deleted the package-lock and made a new one on node 14.17.6 and npm 8.6.0.

oshi97:
> > should the package-lock be getting updated here? I would have thought develop would have priority
> 
> Master uses package-lock version 1 and develop uses version 2, so there were significant conflicts. I deleted the package-lock and made a new one on node 14.17.6 and npm 8.6.0.

Deleting and regenerating the package-lock will update a number of dependencies, but I'm fine with that

oshi97:
> > should the package-lock be getting updated here? I would have thought develop would have priority
> 
> Master uses package-lock version 1 and develop uses version 2, so there were significant conflicts. I deleted the package-lock and made a new one on node 14.17.6 and npm 8.6.0.

Alternatively, since no dependencies in the package.json were changed at all, I don't think we would need to address merge conflicts and could just use the package-lock on develop verbatim. But, again, I think it's fine if we want to regenerate it. At this point the SDK has good test coverage, if a little messy, but we shouldn't run into issues for say updating a babel plugin to the latest hotfix.

oshi97 (approved):

PR #1714, opened 2022-04-08
add truncatedSuffix param to formatRichText

oshi97:
This is necessary so that the truncated suffix appears WITHIN the actual rich text content.

Also change the RichTextFormatter class from a singleton instance class to just a class with a static method on it.
It did not hold any state, so there really was not any point of `format` being an instance method. This simplifies
things upstream a little.

J=SLAP-2039
TEST=manual,auto

test that I can use ANSWERS.formatRichText and the suffix appears when truncating

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -86,8 +86,9 @@ class Answers {
      * A reference to the formatRichText function.
      * @type {Function}
      */
-    this.formatRichText = (markdown, eventOptionsFieldName, targetConfig, truncatedLength) =>
-      RichTextFormatter.format(markdown, eventOptionsFieldName, targetConfig, truncatedLength);
+    this.formatRichText = (markdown, eventOptionsFieldName, targetConfig, truncatedLength, truncatedSuffix) =>
+      RichTextFormatter.format(
+        markdown, eventOptionsFieldName, targetConfig, truncatedLength, truncatedSuffix);
 
     /**
      * A reference to the localizeDistance function.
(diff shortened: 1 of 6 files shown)


src/answers-umd.js
```diff
@@ -86,8 +86,9 @@ class Answers {
      * A reference to the formatRichText function.
      * @type {Function}
      */
-    this.formatRichText = (markdown, eventOptionsFieldName, targetConfig, truncatedLength) =>
-      RichTextFormatter.format(markdown, eventOptionsFieldName, targetConfig, truncatedLength);
+    this.formatRichText = (markdown, eventOptionsFieldName, targetConfig, truncatedLength, truncatedSuffix) =>
```
yen-tt: if this is just passing through the arguments, I wonder if it would make sense to do this? 
`this.formatRichText = RichTextFormatter.format`?

oshi97: Ah ok I just tried it and it changes the `this` context of RichTextFormatter.format from the RichTextFormatter instance to I think the answers-umd instance
image

This could be pretty easily fixed by pulling out the `this._urlTransformer` and `this._generatePluginName` methods out of the RichTextFormatter class, or even just changing it from a class (RichTextFormatter is stateless) to a function. I opted for just making the format method a static method, to try to prevent people from falling into surprising `this` context behavior.

yen-tt: I see, yea having it as a static method makes sense!

yen-tt (approved):

PR #1722, opened 2022-04-15
Add remaining mocked responses for `Basic universal flow` test.

tmeyer2115:
This PR adds the ability to mock Universal AutoComplete responses. These
mocked reponses are used by the `Basic universal flow` test, in place of
hitting LiveAPI. The entire test is now using fully mocked LiveAPI responses.

TEST=auto

Ensured the acceptance test ran fine.


tests/acceptance/fixtures/responses/universal.js
```diff
@@ -1,5 +1,44 @@
...
+        matchedSubstrings: [],
+        queryIntents: [],
+        verticalKeys: []
+      }
+    ];
+  } else if (prompt.includes('a')) {
+    mockedResponse.response.results = [
+      { value: 'amani farooque phone number', matchedSubstrings: [], queryIntents: [], verticalKeys: [] }
+    ];
+  }
```
yen-tt: do we need to separate the autocomplete response based on prompt like this? I think the one test in acceptancesuite.js with `searchComponent.enterQuery('ama');` should be able to select the second option of the mocked results set in the if statement above with `getAutoComplete().selectOption(...)`. We may only need to return the same mocked autocomplete response for all the tests so that would simplify things here.

edit: then again, if we want to simulate the exact behavior that this should return only 1 autocomplete result for 'ama' input then this is fine! Just a small nit, maybe use startsWith('a') instead of includes() so query like Virginia wouldn't get the same thing?

tmeyer2115: Technically, the tests would pass if we just used one autocomplete response. But, that wouldn't be a realistic mocking of LiveAPI. In the test we also enter the query 'Tom'. Returning query suggestions not containing that string would be strange. I worry that if we don't realistically mock LiveAPI, that could lead us to write bad tests. As an example, someone might add some logic to our acceptance tests that looks for this query suggestion when someone enters "Tom". But, in the real world, "Tom" produces no query suggestions.

tmeyer2115: Good call! Added the `startsWith`.

tests/acceptance/fixtures/responses/universal.js
```diff
@@ -484,3 +523,15 @@ export const MockedUniversalSearchRequest = RequestMock()
     return urlRegex.test(request.url) && request.method === 'get';
   })
   .respond(UniversalSearchResponse, 200, CORSHeaders);
+
+export const MockedUniversalAutoCompleteRequest = RequestMock()
+  .onRequestTo(async request => {
+    const urlRegex = /^https:\/\/liveapi-cached.yext.com\/v2\/accounts\/me\/answers\/autocomplete/;
```
oshi97: I was hoping we could do
```ts
.onRequestTo({ url: /regex/, method: 'GET', isAjax: true })
```
though I'm not sure if this would work?

tests/acceptance/fixtures/responses/universal.js
```diff
@@ -484,3 +523,15 @@ export const MockedUniversalSearchRequest = RequestMock()
     return urlRegex.test(request.url) && request.method === 'get';
   })
   .respond(UniversalSearchResponse, 200, CORSHeaders);
+
+export const MockedUniversalAutoCompleteRequest = RequestMock()
```
oshi97: would it make sense to put the universal search and universal autocomplete in separate files?

tests/acceptance/fixtures/responses/universal.js
tmeyer2115: I don't think that will work unfortunately. The first arg is a string, not a regex. I believe the string is a literal match as well.

tests/acceptance/fixtures/responses/universal.js
tmeyer2115: Sure! I could make a `responses/universal` folder with two separate files.

yen-tt (approved):

oshi97 (approved):

PR #1723, opened 2022-04-18
update package-lock to respect overrides

oshi97:
There were a number of issues with npm overrides not being
respected by the npm install command. Supposedly, these issues
have been fixed in npm 8.7.0.

See https://github.com/npm/cli/pull/4709

This PR deletes and regenerates the package-lock so that overrides are respected.

J=SLAP-2022
TEST=manual

see that the unsafe-eval test page loads properly
rerunning npm install does not undo overrides


cea2aj:
Since release/v1.14 has been cut, should we merge this there? Otherwise, LGTM

oshi97:
> Since release/v1.14 has been cut, should we merge this there? Otherwise, LGTM

Ah I was thinking I didn't want this to go into the release branch, but I think you're right!

cea2aj (approved):

PR #1725, opened 2022-04-25
fire follow_up_query analytics on tab nav

yen-tt:
This pr adds support for firing FOLLOW_UP_QUERY event when user perform a search and proceed to click or hit enter on a tab in the navigation bar. Here, we assume the expected behavior of tab navigation of SDK to a relative url path is that it always fire a new search, whether it's an empty search or using the same query registered when navigating away. If an absolute url path is provided for the tab, the origin of the tab url must be the same as the current page url to fire a FOLLOW_UP_QUERY.

Alternatively, we could have queryId persist in the browser url when navigate to new page/tab. The page after navigation would extract the previous queryId from the url if there's a new search fired. This would be more accurate since we would know for sure if a new search is actually fired from a new Answers instance on the new page/tab. However, this required adding new param to the browser url to keep track of previous queryId, I opted not to do this approach.

J=SLAP-2054
TEST=manual

navigate between tabs in theme, see that FOLLOW_UP_QUERY event is fired for all/vertical tab navigation.
tested similar behavior in ios and android chrome browser

### DIFF
--- a/src/ui/components/navigation/navigationcomponent.js
+++ b/src/ui/components/navigation/navigationcomponent.js
@@ -9,6 +9,8 @@ import DOM from '../../dom/dom';
 import { mergeTabOrder, getDefaultTabOrder, getUrlParams } from '../../tools/taborder';
 import { filterParamsForExperienceLink, replaceUrlParams } from '../../../core/utils/urlutils.js';
 import TranslationFlagger from '../../i18n/translationflagger';
+import AnalyticsEvent from '../../../core/analytics/analyticsevent';
+import Searcher from '../../../core/models/searcher';
 
 /**
  * The debounce duration for resize events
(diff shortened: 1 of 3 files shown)


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -236,12 +238,31 @@ export default class NavigationComponent extends Component {
...
     this.unbindOverflowHandlers();
   }
 
+  reportFollowUpQueryEvent (tabIndex) {
+    const searcher = tabIndex === 0 ? Searcher.UNIVERSAL : Searcher.VERTICAL;
+    const tabs = this.getState('tabs');
+    const tabUrlString = tabs?.[tabIndex]?.url || '';
+    let sameOriginForTabUrl = true;
+    if (tabUrlString.startsWith('http')) {
+      const tabUrl = new URL(tabUrlString);
```
oshi97, line 256: do we need a polyfill for URL?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -236,12 +238,31 @@ export default class NavigationComponent extends Component {
...
+      sameOriginForTabUrl = tabUrl.origin === window.location.origin;
```
oshi97, line 257: I'm preeetty sure this comparison will work for iframes?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -236,12 +238,31 @@ export default class NavigationComponent extends Component {
...
+      DOM.onClickOrEnter(selector, this.reportFollowUpQueryEvent.bind(this, index));
+    });
   }
 
   onUnMount () {
     this.unbindOverflowHandlers();
   }
 
+  reportFollowUpQueryEvent (tabIndex) {
+    const searcher = tabIndex === 0 ? Searcher.UNIVERSAL : Searcher.VERTICAL;
```
nmanu1, line 251: is it always guaranteed that the universal tab will be the first one? or is that customizable (maybe with the `isFirst` option for vertical pages)?

src/ui/components/navigation/navigationcomponent.js
yen-tt: ah ie11, I will just parse the origin directly then

src/ui/components/navigation/navigationcomponent.js
yen-tt: It's not guaranteed the first one will be universal. But in navigation handlebars, we already assume the first one is a universal page since we are sending `ALL_TAB_NAVIGATION` instead of `VERTICAL_TAB_NAVIGATION`. I will look for another way to check for this

src/ui/components/navigation/navigationcomponent.js
yen-tt: yup, it works, the window here will be from the iframe where the nav bar is

src/ui/components/navigation/navigationcomponent.js
yen-tt: I don't see a good way of knowing if a tab will lead to a universal or a vertical page without making some assumption. Each tab object does contains a verticalKey value, but it's optional (serverside vertical config), and will be assign to the tab url if a `verticalKey` not provided so it's hard to distinguish. I will check with product on the assumptions we can make or if it may be better to go with the new param on url solution

yen-tt:
Decided to go with the approach of add query id as a param in url to persist through page refresh and tab navigation.

PR #1727, opened 2022-04-27
FOLLOW_UP_QUERY event for tab navigation and refresh

yen-tt:
This pr add support to fire `FOLLOW_UP_QUERY` analytics event when user perform a search and then click on tab navigation OR refresh the page.
- `FOLLOW_UP_QUERY` analytics event requires the previous query id in is payload. To have the previous query id persist during page navigation/refresh, the id is store as a new param in the URL using the logic of persistentStorage already use in the SDK every time a search is made and a new id is generated. 
- With this approach, there could be a case of user copying a link with query and query-id in the url to a new tab, in which an initial search is performed and a follow_up_query event is fired even though there's only one actual search. To prevent this, query-id param is deleted on load IF the page navigation is not because of a href link click (check through document.referrer) OR if it's not a page refresh (check using PerformanceNavigationTiming API)
 
J=SLAP-2054 & SLAP-2055
TEST=manual

Attempted to test through localhost with theme site, there were issues with .html extensions that cut off the search param during href navigation (localhost:5042/jobs?query=test would preserve the search params BUT localhost:5042/jobs.**html**?query=test navigates to localhost:5042/jobs.html)
Instead, tested with other approaches:
- [serve theme site in storm](https://devtestfollowupquery-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=&query-id=01806ba7-670a-a7f9-9e7c-375566d79940) through a dev branch that points to this SDK dev branch.
  - site performed a search onload by default. See that the query-id appears in the URL. I clicked on a vertical tab, which triggers another search onload. See that the query-id updates according to the new search and a `FOLLOW_UP_QUERY` event is fired with the old query-id in its payload.
  - refresh the page after the site performed a search. See that query-id updates accordingly and a `FOLLOW_UP_QUERY` event is fired with the old query-id in its payload.
  - Copy link with query-id to another tab. see that no `FOLLOW_UP_QUERY` was fired from initial search. smoke test in different browsers (desktop chrome, firefox, safari, ie; mobile ios safari, android chrome)
 
 This also work when tested with `python -m SimpleHttpServer 5042` instead of `npm run serve-test-site` in theme

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -195,9 +195,13 @@ class Answers {
      * search on load.
      */
     const searchParams = new SearchParams(window.location.search);
+    console.log('searchParams', searchParams);
     if (searchParams.has(StorageKeys.QUERY_ID)) {
       const isPageReload = window.performance.getEntriesByType('navigation')?.[0]?.type === 'reload';
+      console.log('isPageReload', isPageReload);
+      console.log('document.referrer', document.referrer);
       if (!document.referrer && !isPageReload) {
+        console.log('delete QUERY_ID');
         searchParams.delete(StorageKeys.QUERY_ID);
       }
     }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -195,13 +195,9 @@ class Answers {
      * search on load.
      */
     const searchParams = new SearchParams(window.location.search);
-    console.log('searchParams', searchParams);
     if (searchParams.has(StorageKeys.QUERY_ID)) {
       const isPageReload = window.performance.getEntriesByType('navigation')?.[0]?.type === 'reload';
-      console.log('isPageReload', isPageReload);
-      console.log('document.referrer', document.referrer);
       if (!document.referrer && !isPageReload) {
-        console.log('delete QUERY_ID');
         searchParams.delete(StorageKeys.QUERY_ID);
       }
     }
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -186,7 +186,23 @@ class Answers {
       },
       persistedValueParser: this._parsePersistentStorageValue
     });
-    storage.init(window.location.search);
+
+    /**
+     * Guard the case of user copying a link with query-id param to a new tab
+     * and firing a FOLLOW_UP_QUERY analytics event from their first search.
+     * This is different from redirecting a href click into a new tab or a page
+     * refresh, which should fire a FOLLOW_UP_QUERY analytics event upon next
+     * search on load.
+     */
+    const searchParams = new SearchParams(window.location.search);
+    if (searchParams.has(StorageKeys.QUERY_ID)) {
+      const isPageReload = window.performance.getEntriesByType('navigation')?.[0]?.type === 'reload';
+      if (!document.referrer && !isPageReload) {
+        searchParams.delete(StorageKeys.QUERY_ID);
+      }
+    }
+
+    storage.init(searchParams.toString());
     storage.set(StorageKeys.SEARCH_CONFIG, parsedConfig.search);
     storage.set(StorageKeys.VERTICAL_PAGES_CONFIG, parsedConfig.verticalPages);
     storage.set(StorageKeys.LOCALE, parsedConfig.locale);
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -281,7 +281,6 @@ export default class Core {
         this._persistLocationRadius();
         this._reportFollowUpQueryEvent(data[StorageKeys.QUERY_ID], Searcher.VERTICAL);
 
-        console.log('queryId from search', data[StorageKeys.QUERY_ID]);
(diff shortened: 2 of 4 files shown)


src/core/core.js
```diff
@@ -280,7 +281,7 @@ export default class Core {
         this._persistLocationRadius();
         this._reportFollowUpQueryEvent(data[StorageKeys.QUERY_ID], Searcher.VERTICAL);
 
-        this.storage.set(StorageKeys.QUERY_ID, data[StorageKeys.QUERY_ID]);
+        this.storage.setWithPersist(StorageKeys.QUERY_ID, data[StorageKeys.QUERY_ID]);
```
oshi97, line 284: Did product consider any potential issues with persisting the query id?

tmeyer2115: Nope, Davish seemed cool with it: https://yext.slack.com/archives/C016ZKY42CF/p1651009932053909?thread_ts=1651008679.567229&cid=C016ZKY42CF

oshi97: I was thinking along the lines of somebody bookmarking the page, copy pasting the URL to send to somebody else, or copy pasting the url to be linked to from a different page.

oshi97: Oh I wrote my above comment before seeing Tom's (github race condition), sgtm!

oshi97 (dismissed):

tmeyer2115 (commented):
Maybe I'm overthinking things, but there's a particular use-case here that bugs me. What if someone copied the link with the previous `queryId` and sent it to someone else. When that person pastes the URL in, it would cause a `FOLLOW_UP_QUERY` event to fire, but that seems disingenuous. They've just fired the one query.

src/core/core.js
yen-tt: I was thinking in the case of someone perform a search and open the vertical page in a new tab, which would still consider valid to fire a follow_up_query. This distinction between opening a vertical page in a new tab vs copying the link to a new tab can be resolve with checking document.referrer. My concern is on page refresh which is hard to distinguish from a page load form a copied link, I will look into this further.

yen-tt: I added logic to guard the case you and tom mentioned, by ensuring that the navigation is due to a refresh or from a tab href link. For refresh, I did use an experimental [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type) API which was recommended as oppose to the deprecated [PerformanceNavigation API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation/type).

yen-tt:
> Maybe I'm overthinking things, but there's a particular use-case here that bugs me. What if someone copied the link with the previous `queryId` and sent it to someone else. When that person pastes the URL in, it would cause a `FOLLOW_UP_QUERY` event to fire, but that seems disingenuous. They've just fired the one query.

Addressed in oshi's comment aboved. Added logic to delete query-id from url to prevent `FOLLOW_UP_QUERY` from firing in this case, where it's neither a tab navigation or a refresh.

tmeyer2115 (approved):

src/core/core.js
oshi97: experimental but also supported in ie11 :O very cool!

src/answers-umd.js
```diff
@@ -186,7 +186,23 @@ class Answers {
...
+     * and firing a FOLLOW_UP_QUERY analytics event from their first search.
+     * This is different from redirecting a href click into a new tab or a page
+     * refresh, which should fire a FOLLOW_UP_QUERY analytics event upon next
+     * search on load.
+     */
+    const searchParams = new SearchParams(window.location.search);
+    if (searchParams.has(StorageKeys.QUERY_ID)) {
+      const isPageReload = window.performance.getEntriesByType('navigation')?.[0]?.type === 'reload';
+      if (!document.referrer && !isPageReload) {
+        searchParams.delete(StorageKeys.QUERY_ID);
```
oshi97, line 201: Is there a chance for the url to get out of sync with persistent storage? Since we're manually updating the URL instead of using storage

yen-tt: Are you saying if the persistent storage would store a set of a different params from the modified url here? The storage is initialized based on the searchParams [here after it is modified ](https://github.com/yext/answers-search-ui/pull/1727/files/3a73dd4b18891862fa697ece1c5383c507d920ab#diff-151f0f77fa3617c80e488938e18494f85ca7eb1cea04bc981728ce3408ff10f2R205) so I don't think that would be an issue

yen-tt:
After a discussion with product and tom, we decided to use VERTICAL_TAB_NAVIGATION and ALL_TAB_NAVIGATION to demarcate the end of a query for tab navigation use cases and the existing RESULTS_HIDDEN event fire for page refresh use case. So, this pr is no longer needed.

PR #1729, opened 2022-04-29
handle "navigate away" use cases for RESULTS_HIDDEN

yen-tt:
Add support for firing results visibility events on "navigate away" use cases. The expected scenarios are:
1. Enter and navigate to a new URL.
2. Click a link that opens in the same tab.
3. Back/Forward away from the page.

Case 1 and 2 is already covered with the existing code. During testing, the analytics event does not get capture properly in the inspector network tab for Chrome and Safari (it does appear in Firefox). However, the analytics event does show up in snowflake database confirming SDK does fire visibility change analytics events on case 1 and 2.

For case 3, b/f navigation between different url origin works as expected with visibilitychange API. However, b/f navigation between different url params from the same answers page will NOT trigger visibilitychange event since [browser will not do a complete reload the page on pushState/replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#:~:text=the%20browser%20won%27t%20attempt%20to%20load%20this%20URL%20after%20a%20call%20to%20pushState()), which we use for updating url on search. So, popState is use to capture when a page exit (RESULTS_HIDDEN) from b/f navigation. Since SDK may call 'clearResults' in `storage`'s popState listener function, queryId is stored in POP_STATE_HISTORY's data before everything is cleared so `visibilitychangehandler`'s popState can access the queryId of the popped state. A storage listener is also set on query_id to fire RESULTS_UNHIDDEN for the new page/state if the page previously fired RESULTS_HIDDEN from a popstate event.

J=SLAP-2053
TEST=manual

In network tab, open t[heme site](https://devtestresultshidden-theme-slapshot-pagescdn-com.preview.pagescdn.com/) points to this SDK branch, with no initial search on load, then ran search 'test', and 'test2'. navigate back to no search page and navigate forward to 'test2' page. See that the series of analytics event and their queryId are as expected.

navigate back 'test2' => 'test': 
 -fired `RESULTS_HIDDEN` with 'test2' queryId and `RESULTS_UNHIDDEN` with new 'test' queryid
navigate back 'test' => first page with no search: 
 -fired `RESULTS_HIDDEN` with 'test' queryId
navigate forward first page with no search => 'test': 
 -fired `RESULTS_UNHIDDEN` with new 'test' queryId
navigate forward 'test' => 'test2': 
 -fired `RESULTS_HIDDEN` with 'test' queryId and `RESULTS_UNHIDDEN` with new 'test2' queryId

Inspect in snowflake using answers key 'slanswers-test-analytics', query for user event on today's date. Smoke test navigate away cases on Desktop - Chrome, Safari, Firefox, IE, IOS - Chrome, Safari, Firefox, and Android - Chrome, Firefox.

### DIFF
--- a/src/answers-umd.js
+++ b/src/answers-umd.js
@@ -160,6 +160,9 @@ class Answers {
         this.core.storage.delete(StorageKeys.SORT_BYS);
         this.core.filterRegistry.clearAllFilterNodes();
 
+        const queryId = this.core.storage.get(StorageKeys.QUERY_ID);
+        data.set('pop-state-queryId', queryId);
+
         if (!hasQuery) {
           this.core.clearResults();
         } else {
@@ -274,7 +277,7 @@ class Answers {
         this._analyticsReporterService,
         parsedConfig.search?.verticalKey
       );
-      visibilityAnalyticsHandler.initVisibilityChangeListeners();
+      visibilityAnalyticsHandler.initVisibilityChangeListeners(storage);
     }
 
     this.core = new Core({
--- a/src/core/analytics/visibilityanalyticshandler.js
+++ b/src/core/analytics/visibilityanalyticshandler.js
@@ -60,9 +60,11 @@ export default class VisibilityAnalyticsHandler {
      */
     window.addEventListener('popstate', () => {
       const poppedStateQueryId = storage.get(StorageKeys.HISTORY_POP_STATE)?.get('pop-state-queryId');
+      const storageQueryId = this._analyticsReporterService.getQueryId();
       this._analyticsReporterService.setQueryId(poppedStateQueryId);
       this._previousResultsVisibilityEvent = RESULTS_VISIBILITY_EVENT.HIDDEN;
       this._reportVisibilityChangeEvent(RESULTS_VISIBILITY_EVENT.HIDDEN);
+      this._analyticsReporterService.setQueryId(storageQueryId);
     });
 
     storage.registerListener({
--- a/src/core/analytics/visibilityanalyticshandler.js
+++ b/src/core/analytics/visibilityanalyticshandler.js
@@ -21,6 +21,8 @@ export default class VisibilityAnalyticsHandler {
   /**
    * Initialize visibility change event listener(s) to send analytics events
    * when a result page have become visible or have been hidden.
+   *
+   * @param {Storage} storage - a container around application state
    */
   initVisibilityChangeListeners (storage) {
     /**
@@ -77,6 +79,8 @@ export default class VisibilityAnalyticsHandler {
 
   /**
    * Send visibility change related analytics event.
+   *
+   * @param {string} eventName - the name of the analytics event to report
    */
   _reportVisibilityChangeEvent (eventName) {
     const queryId = this._analyticsReporterService.getQueryId();
--- a/src/core/analytics/visibilityanalyticshandler.js
+++ b/src/core/analytics/visibilityanalyticshandler.js
@@ -1,13 +1,19 @@
 import { isIE } from '../utils/useragent';
 import AnalyticsEvent from '../analytics/analyticsevent';
 import Searcher from '../models/searcher';
+import StorageKeys from '../storage/storagekeys';
+
+const RESULTS_VISIBILITY_EVENT = {
+  HIDDEN: 'RESULTS_HIDDEN',
+  UNHIDDEN: 'RESULTS_UNHIDDEN'
+};
 
 /**
  * Manages the document's visibility status and handles any visibility related analytics events.
  */
 export default class VisibilityAnalyticsHandler {
   constructor (analyticsReporterService, verticalKey) {
-    this._documentVisibilityState = undefined;
+    this._previousResultsVisibilityEvent = undefined;
     this._analyticsReporterService = analyticsReporterService;
     this._verticalKey = verticalKey;
   }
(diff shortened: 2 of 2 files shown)


src/answers-umd.js
```diff
@@ -160,6 +160,9 @@ class Answers {
         this.core.storage.delete(StorageKeys.SORT_BYS);
         this.core.filterRegistry.clearAllFilterNodes();
 
+        const queryId = this.core.storage.get(StorageKeys.QUERY_ID);
+        data.set('pop-state-queryId', queryId);
```
cea2aj, line 164: Would it also be accurate to call this the 'prevous-queryId'?

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -16,19 +22,19 @@ export default class VisibilityAnalyticsHandler {
    * Initialize visibility change event listener(s) to send analytics events
    * when a result page have become visible or have been hidden.
    */
-  initVisibilityChangeListeners () {
+  initVisibilityChangeListeners (storage) {
```
cea2aj, line 27: nit: can you add the type of 'storage' to the jsdoc?

src/answers-umd.js
yen-tt: If we are navigating back a page, I thought it may be a bit confusing if 'previous' would mean the query id from the page we are navigating from, or the query id from page we are navigating back to. This is technically storing the latest queryId as the next one have not been executed yet, if there is a search on load from the page we are navigating back/forward to.

src/core/analytics/visibilityanalyticshandler.js
yen-tt: sure thing!

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -38,15 +46,41 @@ export default class VisibilityAnalyticsHandler {
...
+
+    storage.registerListener({
+      eventType: 'update',
+      storageKey: StorageKeys.QUERY_ID,
+      callback: id => {
+        if (this._previousResultsVisibilityEvent !== RESULTS_VISIBILITY_EVENT.UNHIDDEN) {
+          this._previousResultsVisibilityEvent = RESULTS_VISIBILITY_EVENT.UNHIDDEN;
+          this._reportVisibilityChangeEvent(RESULTS_VISIBILITY_EVENT.UNHIDDEN);
+        }
+      }
```
cea2aj: Should this check that the `visibilityState` is 'visible' before sending the UNHIDDEN event? Also, does this listener fire after the popstate listener?

yen-tt: query_id update when there's a search execute and I believe it can be done only if the page is visible so I didn't put in that check. So far it has been visible status when triggered. However, it could be the case that the response takes a while to load and it finish loading in the background when user switch to a different tab. I can include this check! 

I believe this listener will fire after popstate listener if there's a new query executed. However, in storage's popstate function, the call `clearResults`, when navigate to a page with no search, would trigger this before the visibilitychangehandler's popstate. But since there's no new query_id, _reportVisibilityChangeEvent wouldn't fire an analytics event for this page.

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -1,34 +1,42 @@
...
+     * RESULTS_HIDDEN analytics event does not get send again if the page is already hidden.
      */
     document.addEventListener('visibilitychange', () => {
-      if (document.visibilityState === 'hidden' && this._documentVisibilityState !== 'hidden') {
-        this._documentVisibilityState = 'hidden';
-        this._reportVisibilityChangeEvent('RESULTS_HIDDEN');
-      } else if (document.visibilityState === 'visible') {
-        this._documentVisibilityState = 'visible';
-        this._reportVisibilityChangeEvent('RESULTS_UNHIDDEN');
+      if (document.visibilityState === 'hidden' && this._previousResultsVisibilityEvent !== RESULTS_VISIBILITY_EVENT.HIDDEN) {
```
oshi97: Is there some way we could make these conditions a little more readable? I was thinking maybe some kind of helper method so we can give it a proper name, otherwise as somebody not familiar with the code it's hard to grasp exactly what is going on and why

cea2aj:
Can you list out the HIDDEN and UNHIDDEN events which were fired during the test example that you described in the description where you fired off two searches and navigated back and forward?

yen-tt:
> Can you list out the HIDDEN and UNHIDDEN events which were fired during the test example that you described in the description where you fired off two searches and navigated back and forward?

sure thing, updated the pr description

src/core/analytics/visibilityanalyticshandler.js
yen-tt: sure, I updated to separate some of these conditions in if statements to its own helper method. Hopefully that's clearer

src/core/analytics/visibilityanalyticshandler.js
```diff
@@ -38,15 +46,61 @@ export default class VisibilityAnalyticsHandler {
...
+      eventType: 'update',
+      storageKey: StorageKeys.QUERY_ID,
+      callback: id => {
+        if (this._resultsVisibilityChangeToVisible()) {
+          this._previousResultsVisibilityEvent = RESULTS_VISIBILITY_EVENT.UNHIDDEN;
+          this._reportVisibilityChangeEvent(RESULTS_VISIBILITY_EVENT.UNHIDDEN);
+        }
+      }
+    });
+  }
```
tmeyer2115, line 80: I think we should try using a state machine approach here. We could maintain a Mapping of Query Ids to a state: visible or hidden. I think that would simplify a lot of the logic here. We wouldn't be dependent on the toggling back and forth of the `_previousResultsVisibilityEvent`. I think we could also get rid of the `storage` listener with an appropriate state machine.

src/answers-umd.js
oshi97, line 164: should this be a storage key?

yen-tt: 'pop-state-queryId' is part of the `data` is going to to be set in StorageKeys.HISTORY_POP_STATE (a bit lower in the code here). Are you suggesting to separate it out into its own StorageKey?

tmeyer2115 (approved):

src/core/analytics/visibilityanalyticshandler.js
yen-tt: talked over slack. decided the current approach of the pr is sufficient - didn't want to sink more time into refactoring at this moment.

oshi97 (approved):

cea2aj (approved):

PR #1731, opened 2022-05-05
modify tab url based on given parent url

yen-tt:
This pr is made in tandem with the [theme changes](https://github.com/yext/answers-hitchhiker-theme/pull/1067) to support iframe experience's tab navigation where opening one in a new browser tab should still be under the domain of the iFrame's parent for the new page's url.

if a `_parentUrl` is provided, the tabs are updated with their href having the domain of the parent url and a verticalUrl param attached to indicate which page to navigate to. The target is also updated to '_parent' to handle the case of tab navigation in the same browser tab.
a function `setParentUrl` is added to the navigation component for user to set `_parentUrl`, which will cause a rerender of the component to update the tabs's link

J=SLAP-2089
TEST=manual

Tested along with the changes in theme. 
Deployed two test sites in storm, with [one being from the theme dev branch](https://devtabnaviniframe-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=) that points to the SDK pr dev branch, under the staging commit: dev/tab-nav-in-iframe[@65f09b10](https://github.com/yext/answers-hitchhiker-theme/commit/65f09b101f882c597389717d4a8b8d814edb1bda). That is used in the src script of iframe_test.html of the [other site](https://devyttesting-theme-slapshot-pagescdn-com.preview.pagescdn.com/?query=virginia&referrerPageUrl=) under the staging commit: dev/yt-testing[@517e852f](https://github.com/yext/answers-hitchhiker-theme/commit/517e852fe3ff2f7961704090e11118273ba162b3). Navigate to different tabs in the same tab and opening them in new tab. See that the page is still under the domain of the iFrame's parent.

### DIFF
--- a/src/ui/components/navigation/navigationcomponent.js
+++ b/src/ui/components/navigation/navigationcomponent.js
@@ -390,7 +390,7 @@ export default class NavigationComponent extends Component {
       const urlParser = document.createElement('a');
       tabs.forEach(tab => {
         urlParser.href = tab.url;
-        const tabParams = new SearchParams(tab.url);
+        const tabParams = new SearchParams(urlParser.search);
         const verticalUrl = urlParser.pathname.replace(/^\//, '');
         tabParams.set('verticalUrl', verticalUrl);
         tab.url = parentUrlWithoutParams + '?' + tabParams.toString();
--- a/src/ui/components/navigation/navigationcomponent.js
+++ b/src/ui/components/navigation/navigationcomponent.js
@@ -9,6 +9,7 @@ import DOM from '../../dom/dom';
 import { mergeTabOrder, getDefaultTabOrder, getUrlParams } from '../../tools/taborder';
 import { filterParamsForExperienceLink, replaceUrlParams } from '../../../core/utils/urlutils.js';
 import TranslationFlagger from '../../i18n/translationflagger';
+import SearchParams from '../../dom/searchparams';
 
 /**
  * The debounce duration for resize events
@@ -238,6 +239,11 @@ export default class NavigationComponent extends Component {
     }
   }
 
+  setParentUrl (parentUrl) {
+    this._parentUrl = parentUrl;
+    this.setState(this._state.get());
+  }
+
   onUnMount () {
     this.unbindOverflowHandlers();
   }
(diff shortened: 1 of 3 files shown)


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -379,6 +385,19 @@ export default class NavigationComponent extends Component {
       }
     }
 
+    if (this._parentUrl) {
+      const parentUrlWithoutParams = this._parentUrl.split('?')[0];
+      const urlParser = document.createElement('a');
```
tmeyer2115, line 390: Is it necessary that we create an anchor tag here just to update the `tabs`? That seems like a bit of a code smell. Every time this component is re-rendered, will more and more of the anchor tags get appended to the DOM?

src/ui/components/navigation/navigationcomponent.js
```diff
@@ -238,6 +239,11 @@ export default class NavigationComponent extends Component {
     }
   }
 
+  setParentUrl (parentUrl) {
```
tmeyer2115, line 242: I'm sort of torn here. The reason we have this as a setter is for the Theme. Conceptually, though, I think this probably makes more sense as a config option for the Component. I can't really see a use-case where it's set more than once. If we made it part of the Component config, though, we couldn't get the Theme portion working, correct?

oshi97: Could this be a config option if it was a getter? so I was thinking `{ getParentUrl: () => /* fancy logic */ }`

oshi97: In general calling the component instance directly isn't common. I don't think there are specific reasons for it besides the SDK already giving too many degrees of freedom

src/ui/components/navigation/navigationcomponent.js
yen-tt: the use of the anchor tag here is not to append to DOM but to help parse pathname. This will not be append to the dom. From looking online, I found this to be a lightweight alternative to manually parse the url ourselves, which would be tricky since we have to deal with relative and absolute urls and their different formats as well. Alternatively we could use URL function but we would need a url polyfill lib for IE11.

src/ui/components/navigation/navigationcomponent.js
yen-tt: the iframe resizer is loaded async from the ANSWERS initialization and component creation portion. I don't think we can guarantee to get the parentUrl from the iframe's parent to put it as part of the component config at the time of adding/creating navigation component.

oshi97: I was thinking we could put parent URL in RuntimeConfig, and then have the the getParentUrl config option call RuntimeConfig, but I'm now realizing the nav component would still have to setState again to update the URLs. I'm fine with the current approach!

oshi97 (dismissed):

tmeyer2115 (approved):

PR #1733, opened 2022-05-05
ie11 support and relative links in constructRedirectUrl

yen-tt:
ie11 on browserstack works with url api but it's not actually supported as specified in mdn - testing it through local window desktop confirmed that this would not work. This pr replace the usage of URL api. This update also ensures cases of user providing relative link format for their redirectUrl.

J=none
TEST=manual

see new jest tests passed.
add redirectUrl (absolute/relative path) to searchBar and see that it navigates as expected in Chrome, safari, firefox.
use ngrok to serve a page with the sdk change, and tested redirectUrl in a local window laptop with ie browser.

### DIFF
--- a/src/ui/tools/urlutils.js
+++ b/src/ui/tools/urlutils.js
@@ -6,16 +6,17 @@ import SearchParams from '../dom/searchparams';
  *
  * @param {string} redirectUrl user-specified redirect url
  * @param {SearchParams} params url params saved in answers storage
- * @returns {URL} new redirect url including params from answers storage
+ * @returns {string} new redirect url including params from answers storage
  */
 export function constructRedirectUrl (redirectUrl, params) {
-  const newRedirectUrl = new URL(redirectUrl);
-  const redirectUrlParams = new SearchParams(newRedirectUrl.search);
+  const urlParser = document.createElement('a');
+  urlParser.href = redirectUrl;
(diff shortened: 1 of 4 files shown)


src/ui/tools/urlutils.js
```diff
@@ -6,16 +6,17 @@ import SearchParams from '../dom/searchparams';
  *
  * @param {string} redirectUrl user-specified redirect url
  * @param {SearchParams} params url params saved in answers storage
- * @returns {URL} new redirect url including params from answers storage
+ * @returns {string} new redirect url including params from answers storage
  */
 export function constructRedirectUrl (redirectUrl, params) {
-  const newRedirectUrl = new URL(redirectUrl);
-  const redirectUrlParams = new SearchParams(newRedirectUrl.search);
+  const urlParser = document.createElement('a');
```
oshi97, line 12: what is the createElement for?

yen-tt: the use of the anchor tag here is to help parse and construct url. This is one of the online [suggested approach](https://www.geeksforgeeks.org/how-to-parse-url-using-javascript/) as oppose to manually parse the url ourselves or have a url polyfill lib for IE11. I had this in the SDK iframe experience pr as well.

oshi97 (approved):
small typo in the pr title but lgtm besides that!!

PR #1743, opened 2022-05-12
Update build for i18n deploy workflow

nmanu1:
This PR updates the `build_and_deploy_i18n` workflow to run a simplified build before the acceptance tests instead of the full i18n build. This shortens the amount of time it takes before we see the results of the acceptance tests. The simplified build is the same as the `dev` build, but without the watch task. The i18n build runs in parallel and still must complete successfully before the deploy jobs are run.

The `build_and_deploy_hold` workflow also runs the i18n build, but another item (SLAP-2103) plans to remove the acceptance tests from this workflow, so it doesn't seem necessary to update it to run a simplified build.

J=SLAP-2099
TEST=auto

Tested on another branch (`test-basic-build`) and saw that the acceptance and unit tests run once the basic build is complete, the i18n build runs in parallel, and the deploy waits until both the tests and the i18n build finish.

### DIFF
--- a/conf/gulp-tasks/templates.gulpfile.js
+++ b/conf/gulp-tasks/templates.gulpfile.js
@@ -33,21 +33,18 @@ async function devTemplates (shouldWatch = true) {
   const translator = await createTranslator(DEFAULT_LOCALE);
   const precompileTemplates = createPrecompileTemplatesTask(DEFAULT_LOCALE, translator);
 
-  const nonWatchTasks = [precompileTemplates, bundleTemplatesUMD, cleanFiles];
+  const templateTasks = [precompileTemplates, bundleTemplatesUMD, cleanFiles];
 
   function watchTemplates () {
     return watch(['./src/ui/templates/**/*.hbs'], {
       ignored: './dist/'
-    }, series(precompileTemplates, bundleTemplatesUMD, cleanFiles));
+    }, series(...templateTasks));
   }
 
+  const tasks = shouldWatch ? [...templateTasks, watchTemplates] : templateTasks;
+
   return new Promise(resolve => {
-    return shouldWatch
-      ? series(
-        ...nonWatchTasks,
-        watchTemplates
-      )(resolve)
-      : series(...nonWatchTasks)(resolve);
+    return series(...tasks)(resolve);
   });
 }
 
--- a/conf/gulp-tasks/templates.gulpfile.js
+++ b/conf/gulp-tasks/templates.gulpfile.js
@@ -23,14 +23,17 @@ async function createTranslator (locale) {
  * Precompiles templates together, bundles them together,
  * cleans up any intermediary files, and kicks off the watch task.
  *
+ * @param {boolean} shouldWatch Whether or not to watch for changes. Defaults to true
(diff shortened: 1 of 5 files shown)


conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -23,32 +23,38 @@ async function createTranslator (locale) {
...
  */
-async function devTemplates () {
+async function devTemplates (shouldWatch = true) {
   const bundleTemplatesUMD =
     new BundleTemplatesTaskFactory(DEFAULT_LOCALE).create(TemplateType.UMD);
   const cleanFiles = createCleanFilesTask(DEFAULT_LOCALE);
   const translator = await createTranslator(DEFAULT_LOCALE);
   const precompileTemplates = createPrecompileTemplatesTask(DEFAULT_LOCALE, translator);
 
+  const nonWatchTasks = [precompileTemplates, bundleTemplatesUMD, cleanFiles];
```
yen-tt: nit: we could consider making a createTemplateTasks for 30-34 and split this into devTemplates and basicTemplates, similar to how you have a separate function `basicJSBundle` (`devJSBundle` but without the watch tasks) to be consistent. Otherwise, what do you think about calling this templatesTasks of sort instead? and we can spread this array inside series() in line 41 too.

package.json
```diff
@@ -116,6 +116,7 @@
     "build-languages": "gulp buildLanguages && size-limit",
     "build-locales": "gulp buildLocales && size-limit",
     "build-search-bar-only": "gulp buildSearchBarOnlyAssets && size-limit",
+    "build-basic": "gulp basic",
```
yen-tt: nit: basic is ok too but I'm thinking we can have a bit more descriptive name for this build, maybe related to the bundle type?

nmanu1: something like `build-unminified-default-locale`?

yen-tt: maybe `build-unminified-legacy`? I thinkk we can assume it's default locale unless the build have -languages/locales attached

conf/gulp-tasks/templates.gulpfile.js
nmanu1: I updated `nonWatchTasks` to `templateTasks` and spread it in line 41 and simplified some of the promise logic. I don't think we need to split it into `devTemplates` and `basicTemplates` because this is similar to the pattern lower in `templates.gulpfile.js` where the locales build and search-bar-only build both use the same function, just with a differing boolean param passed in. but, if you feel strongly about it, I can change it

package.json
```diff
@@ -116,7 +116,7 @@
     "build-languages": "gulp buildLanguages && size-limit",
     "build-locales": "gulp buildLocales && size-limit",
     "build-search-bar-only": "gulp buildSearchBarOnlyAssets && size-limit",
-    "build-basic": "gulp basic",
+    "build-unminified-legacy": "gulp basic",
```
yen-tt: nit: can we update the gulp function `basic` to match the new name as well?

nmanu1: sure, updated

yen-tt (approved):

PR #1744, opened 2022-05-13
speed up i18n build in github actions

yen-tt:
This pr speed up the process of building i18n assets by splitting the work between multiple jobs in github action. Updated `build_and_deploy_i18n.yml`, `build_and_deploy_hold.yml`, and `build_and_deploy_search_bar.yml`. Based on times from testing, the total time to build i18n assets dropped from 17-20 mins to 2-3 mins across all jobs.

Attempted to use gulp.parrallel and ran into an error addressed in this [pr](https://github.com/yext/answers-search-ui/pull/1101). Gulp command line does accept node v8 options as arguments and adding --max-old-space-size=7168 (or higher) did fix the error, but the time did not improve (time from github actions in this commit [ae3bb9a](https://github.com/yext/answers-search-ui/pull/1744/commits/ae3bb9a0c0017fa431a97e482f7de08a9e2a2600)).
Instead, this pr updated github actions to spin up a job for each language to build its respective language-locale pairs. a new `build_i18n.yml` workflow is added, which will create a language matrix based on `ALL_LANGUAGES` variable from `conf/i18n/constants.js` file. Each value in matrix will spawn a new github job/runner for each language. 

The language value is also stored in an environment variable so gulpfile can access it. Gulp templates and library tasks are modified to accept a list of languages to build locale assets for (if not specified, build for ALL_LANGUAGES by default).

previous npm scripts `build-locales` and `build-search-bar-only` also invoke `size-limit` command. This target the default english assets which may not exist for jobs that generate non-english assets. As such, updated npm scripts to have a version without size-limit command and a `build-blah-**dev**` for the build command version with size-limit command that should be use locally.

TEST=manual&auto

- ran `LANGUAGE=ja npm run build-locales` and `LANGUAGE=ja npm run build-search-bar-only`. See that the build script works as expected and only the relevant assets to ja language-locale pairs were created.
- in commit [8d665e0](https://github.com/yext/answers-search-ui/pull/1744/commits/8d665e05639f4abc249537b67241a4d74dc36a0b) of this pr, ran build_and_deploy_i18n.yml workflow. See that all 16 jobs were spawn for each language. See that the time for each job is 1-2 mins and executed successfully. Scan through s3 folder for this pr, and see that the assets were built and deployed as expected.


.github/workflows/build_i18n.yml
```diff
@@ -0,0 +1,54 @@
...
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Use Node.js 16
+      uses: actions/setup-node@v2
+      with:
+        node-version: 16
+        cache: 'npm'
+    - run: npm ci
+    - run: echo ${{ matrix.language }}
```
oshi97: Instead of this echo, maybe we cuold give the step that runs the build a name instead?

conf/gulp-tasks/templates.gulpfile.js
```diff
@@ -106,15 +106,17 @@ async function createTemplatesForLanguages (languages, isSearchBarOnly = false)
  * Creates a template bundle for each supported language, locale pair.
  *
  * @param {boolean} isSearchBarOnly If only templates related to the SearchBar should be included.
+ * @param {string[]} languages a list of languages to build template bundle for,
+ *                             with the corresponding language-locale pairs.
  * @returns {Promise<Function>}
  */
-function allLocaleTemplates (isSearchBarOnly = false) {
+function allLocaleTemplates (isSearchBarOnly = false, languages = ALL_LANGUAGES) {
```
oshi97: is it necessary to define a default here if the functions that call this all do the same defaulting?

.github/workflows/build_i18n.yml
yen-tt: oops, this was for testing. It's not needed, I will delete

conf/gulp-tasks/templates.gulpfile.js
yen-tt: will update to remove the default here

oshi97 (approved):

PR #1747, opened 2022-05-16
keep results on search failure

yen-tt:
Rendering failures leading to the catch statements in Universal and Vertical search would clear all results. User would still expects other results from the response to render on page. This pr updates the `storage.set` calls to only update the searchState, instead of storing a new empty results, so the non-problematic result sets can still be render.
 
The catch statements in core.js will terminate any further rendering of sub components once it encounter a problematic sub component. User expects sdk to render as much as it can. To ensure non-problematic sub-components can still render, this pr added more catch statements in component.js to gracefully handle errors coming from individual child component's data transformation, initialization and mount stage.

J=none
TEST=manual

use `dev-keep-data-on-search-failure` for SDK version ([from this test branch](https://github.com/yext/answers-search-ui/pull/1745)) in the config file of the site with the issue. Searched 'Listings', and see that the same error appeared in console but the other cards still render as normal.
use `dev-dont-clear-data-on-failure` to test changes in this pr through local html pages that use SDK directly. See that errors are logged but the loading icon is reset to complete during error handling process

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -433,12 +433,13 @@ export default class Core {
    * @param {Searcher} searcherType
    */
   _markSearchComplete (searcherType) {
-    const results = searcherType === Searcher.UNIVERSAL
-      ? this.storage.get(StorageKeys.UNIVERSAL_RESULTS)
-      : this.storage.get(StorageKeys.VERTICAL_RESULTS);
+    const resultStorageKey = searcherType === Searcher.UNIVERSAL
+      ? StorageKeys.UNIVERSAL_RESULTS
+      : StorageKeys.VERTICAL_RESULTS;
+    const results = this.storage.get(resultStorageKey);
     if (results) {
       results.searchState = SearchStates.SEARCH_COMPLETE;
-      this.storage.set(StorageKeys.UNIVERSAL_RESULTS, results);
+      this.storage.set(resultStorageKey, results);
     }
     const directanswer = this.storage.get(StorageKeys.DIRECT_ANSWER);
     if (directanswer) {
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -273,13 +273,6 @@ export default class Core {
         querySource: this.storage.get(StorageKeys.QUERY_SOURCE),
         additionalHttpHeaders: this._additionalHttpHeaders
       })
-      .catch(error => {
-        this._markSearchComplete(Searcher.VERTICAL);
-        throw error;
-      })
-      .catch(error => {
-        console.error('The following problem was encountered while executing a vertical search: ' + error);
-      })
       .then(response => SearchDataTransformer.transformVertical(response, this._fieldFormatters, verticalKey))
       .then(data => {
         this._persistFacets();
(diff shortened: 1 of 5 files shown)


src/core/core.js
```diff
@@ -323,9 +323,6 @@ export default class Core {
         window.performance.mark('yext.answers.verticalQueryResponseRendered');
       }).catch(error => {
         console.error('Vertical search failed with the following error: ' + error);
```
tmeyer2115: Since we catch more than API failures, could we update this copy? If a card fails to render, the search didn't fail. Maybe something like "The following problem was encountered during Vertical search: "

src/core/core.js
```diff
@@ -323,9 +323,6 @@ export default class Core {
         window.performance.mark('yext.answers.verticalQueryResponseRendered');
       }).catch(error => {
```
oshi97: Would it make sense to move this catch immediately after coreLibrary.verticalSearch is called, so that it only catches errors thrown by answers-core? We could then have another catch here at the bottom for all other types of errors

src/core/core.js
yen-tt: sure thing

oshi97: +1!

src/core/core.js
yen-tt: yea that works, I will update

src/core/core.js
```diff
@@ -273,6 +273,9 @@ export default class Core {
         querySource: this.storage.get(StorageKeys.QUERY_SOURCE),
         additionalHttpHeaders: this._additionalHttpHeaders
       })
+      .catch(error => {
+        console.error('The following problem was encountered while executing a vertical search: ' + error);
```
tmeyer2115: Would we want to have `SEARCH_COMPLETE` here as well? Otherwise, would the loading spinner just keep going?

src/core/core.js
```diff
@@ -416,13 +420,37 @@ export default class Core {
...
       });
   }
 
+  /**
+   * Update the search state of the results in storage
+   *
+   * @param {Searcher} searcherType
+   * @param {SearchStates} searchState
+   */
+  _updateResultsSearchState (searcherType, searchState) {
```
tmeyer2115: This method is somewhat confusing. Part of that is probably because `SearchStates` isn't the greatest name. Since we always call this with `SearchStates.SEARCH_COMPLETE`, could we remove the second param and just call this `_markSearchComplete` or something?

yen-tt: yup, will update

src/core/core.js
yen-tt: yes, I will add it there too! I just pushed some wip changes to deploy so I can test something with the new result update code in the site

src/core/core.js
yen-tt: per discussion with oshi offline, the throw error should already be clear on where the issue is from. Adding additional catch statements here would require handling breaking the then/catch chain mid way or add logic to avoid bleeding into the later catch statements (potentially end up with duplicate update loading state calls and console errors)

tmeyer2115 (approved):

PR #1760, opened 2022-06-21
install `serve` before starting acceptance tests to reduce spurious failures

oshi97:
In our testcafe's appCommand we run `npx serve` to serve our html fixtures.
But we don't have serve as a dep, so npx automatically installs it.
Unfortunately that takes time which eats into our "appInitDelay" of 4000ms.
We have appInitDelay set to 4000 and pageRequestTimeout set to 8000, which gives
testcafe what I believe to be 12seconds to install and start up serve.
From my testing these are the only two config options that matter here.

If we install serve before starting testcafe things go much more smoothly.

Originally I was concerned about the `-l tcp://0.0.0.0:9999` port we use as opposed to just specifying
the port with `-p 9999`, but there seems to be no issue there upon further investigation.

I also removed testcafe-browser-provider-browserstack as a general dependency, and only install
it before running the browserstack acceptance tests. This is a huge dependency and includes
random things browserstack uses like jimp, which is an image processing library.
Removing this dependency seems to speed up npm ci by about 20 seconds, from 50->30.

J=SLAP-2100
TEST=manual

used this branch to test my hypotheses
https://github.com/yext/answers-search-ui/pull/1759

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,12 +103,12 @@
     "rollup-plugin-node-resolve": "^4.0.1",
     "rollup-plugin-svg": "^2.0.0",
     "sass": "^1.34.0",
+    "serve": "^13.0.2",
     "serve-handler": "^6.1.3",
     "size-limit": "^4.9.0",
     "stylelint": "^14.5.3",
     "stylelint-config-standard-scss": "^3.0.0",
     "testcafe": "^1.18.4",
-    "testcafe-browser-provider-browserstack": "^1.13.1",
     "vinyl-buffer": "^1.0.1",
     "vinyl-source-stream": "^2.0.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -103,12 +103,12 @@
     "rollup-plugin-node-resolve": "^4.0.1",
     "rollup-plugin-svg": "^2.0.0",
     "sass": "^1.34.0",
+    "serve": "^13.0.2",
     "serve-handler": "^6.1.3",
     "size-limit": "^4.9.0",
     "stylelint": "^14.5.3",
     "stylelint-config-standard-scss": "^3.0.0",
     "testcafe": "^1.18.4",
-    "testcafe-browser-provider-browserstack": "^1.13.1",
```
yen-tt, line 111: Does `npm run acceptance` depends on this?

oshi97: no

oshi97: this is only used by browserstack, though if you want to run tests in browserstack from your local machines I believe you will need it

yen-tt (approved):
nice!!

PR #1762, opened 2022-06-22
add facets fixtures for acceptance tests

oshi97:
Adds vertical search response fixtures for facets acceptance tests.
Also moves the and breaks up a number of tests from acceptancesuite.js into their own files.
Fixes bugs in the experience links tests where the links were not actually being tested.
Removes safari from the `npm run acceptance` script, since during development we typically only use chrome.

Static filters to follow in a separate item.

J=SLAP-2151
TEST=auto


tests/acceptance/acceptancesuites/acceptancesuite.js
```diff
@@ -139,68 +139,12 @@ test('navigating pages and hitting the browser back button lands you on the righ
...
-
-  // Reset the filters, and check that the number of results
-  // is the same as the initial amount
-  await filterBox.reset();
-  await filterBox.applyFilters();
-  actualResultsCount = await verticalResultsComponent.getResultsCountTotal();
-  await t.expect(actualResultsCount).eql(initialResultsCount);
-});
-
 test('selecting a sort option and refreshing maintains that sort selection', async t => {
```
yen-tt: this should also be move to facetssuite.js yea?

yen-tt (commented):
can we also update fixture that uses FACETS_PAGE to include the mock responses too? some of them includes: `Experience links work as expected`, `Performance marks on search`, and `W3C Accessibility standards are met`

tests/acceptance/acceptancesuites/acceptancesuite.js
oshi97: I can move to sortoptionssuite.js! This one is technically not testing facets but sorts which is why I didn't wanna move it. I was going to move more stuff out of acceptancesuite.js in a future PR but I can do that here too

package.json
```diff
@@ -123,7 +123,7 @@
     "dev": "gulp dev",
     "lint": "eslint .",
     "test": "eslint . && stylelint src/**/*.scss && cross-env NODE_ICU_DATA=node_modules/full-icu jest",
-    "acceptance": "testcafe safari,chrome --config-file ./.github/testcafe.json",
+    "acceptance": "testcafe chrome --config-file ./.github/testcafe.json",
```
tmeyer2115: Is there harm in leaving Safari here? I'd argue that because we almost exclusively do local dev in Chrome, we should test Safari here.

tests/acceptance/acceptancesuites/experiencelinkssuite.js
```diff
@@ -0,0 +1,84 @@
...
+import FacetsPage from '../pageobjects/facetspage';
+import { Selector } from 'testcafe';
+import { MockedVerticalSearchRequest } from '../fixtures/responses/vertical/search';
+import StorageKeys from '../../../src/core/storage/storagekeys';
+
+fixture`Experience links work as expected`
+  .requestHooks(MockedVerticalSearchRequest)
+  .page`${FACETS_PAGE}`;
+
+test('experience links are clean', async t => {
```
tmeyer2115: Could we add a TODO here? This test is massive. Having a massive test with a ton of assertions is always dangerous. When it fails, it's not exactly clear which piece of the myriad functionality was tested. I'm think we split it up something like: "Test that pagination is not included on nav links", "Test that sort options are not included on nav links", etc. I think that will make this a lot more manageable.

tests/acceptance/acceptancesuites/facetssuite.js
```diff
@@ -0,0 +1,73 @@
...
+  VERTICAL_SEARCH_URL_REGEX
+} from '../constants';
+import FacetsPage from '../pageobjects/facetspage';
+import { MockedVerticalSearchRequest } from '../fixtures/responses/vertical/search';
+import {
+  registerIE11NoCacheHook
+} from '../utils';
+import SearchRequestLogger from '../searchrequestlogger';
+
+fixture`Facets page`
```
tmeyer2115, line 12: If possible, I think we should split these tests up as well. Is there a cost associated with that? We don't set up the page for each test right, we just run the `beforeEach` hook?

tests/acceptance/acceptancesuites/experiencelinkssuite.js
oshi97: I can break it up right now, it don't see any complications with doing so and this test in general is bothering me. The `verifyCleanLinks` helper function doesn't take in parameters and relies on a `universalUrl` defined later in the code, but also later in the code it is directly passed in a `universalUrl` parameter despite it not taking in any params.

package.json
oshi97: eh not really I can add it back

tests/acceptance/acceptancesuites/facetssuite.js
oshi97: The test currently goes
1. select a facet
2. reset the facets
3. select 3 facets
4. reset the facets

I think we can cut this test in half in after the first reset. I think I might have been worried in the past about the SDK being able to select facets the first time but not the second time, but realistically speaking that shouldn't be an issue.

tests/acceptance/acceptancesuites/experiencelinkssuite.js
oshi97: Later on this test does `await verifyCleanLink(changeFiltersLink)` which ends up checking the `universalUrl` not `changeFiltersLink`

tmeyer2115 (approved):

PR #1766, opened 2022-06-24
use testcafe Smart Assertions with PaginationComponentBlock

oshi97:
This PR also fixes incorrect usages of promises with regards to testcafe's [Smart Assertion Query Mechanism](https://testcafe.io/documentation/402837/guides/basic-guides/assertions#smart-assertion-query-mechanism).
We need to be careful not to arbitrarily await different promises for testcafe browser things, like a Selector's textContent. Awaiting too early prevents testcafe from doing its Smart Assertions and can result in spurious acceptance tests.

This was likely the cause of the pretty common "Page 1" not equal to "Page 2"  acceptance test error we would use to see.
Letting testcafe do smart assertions lets us remove the arbitrary waits we have in our browser refresh/page nav utils.
We can now just wait for the results to appear on the page.
Note that before making this change, removing the t.waits in the utils would cause in very consistent test failures,
even if we added the waitForResults call.

J=SLAP-2177
TEST=auto


nmanu1 (approved):
🔥

PR #1771, opened 2022-08-30
switch throw to console error in analytics reporter

yen-tt:
Throw error from not having `ytag` script tag for conversion tracking opt in would prevent a search fired on load. This is an aggressive behavior as analytics shouldn't affect search behavior. This pr updated the two throw error into console error instead.

TEST=manual

Added `ANSWERS.setConversionsOptIn(true);` to a local index page with Answers setup. see that the previously uncaught throw error is now a console error and a search on load is no longer blocked.

Screen Shot 2022-08-30 at 5 06 28 PM

### DIFF
--- a/src/core/analytics/analyticsreporter.js
+++ b/src/core/analytics/analyticsreporter.js
@@ -1,7 +1,6 @@
 /** @module AnalyticsReporter */
 
 import AnalyticsEvent from './analyticsevent';
-import { AnswersAnalyticsError } from '../errors/errors';
 import { PRODUCTION } from '../constants';
 import HttpRequester from '../http/httprequester';
 import { getAnalyticsUrl } from '../utils/urlutils';
--- a/src/core/analytics/analyticsreporter.js
+++ b/src/core/analytics/analyticsreporter.js
@@ -109,11 +109,13 @@ export default class AnalyticsReporter {
       ytag('optin', true);
       cookieData = ytag('yfpc', null);
     } else if (this._conversionTrackingEnabled) {
-      throw new AnswersAnalyticsError('Tried to enable conversion tracking without including ytag');
+      console.error('Tried to enable conversion tracking without including ytag');
+      return false;
     }
 
     if (!(event instanceof AnalyticsEvent)) {
-      throw new AnswersAnalyticsError('Tried to send invalid analytics event', event);
+      console.error('Tried to send invalid analytics event', event);
+      return false;
     }
 
     if (includeQueryId) {
--- a/tests/core/analytics/analyticsreporter.js
+++ b/tests/core/analytics/analyticsreporter.js
@@ -21,7 +21,7 @@ describe('reporting events', () => {
   });
 
   it('throws an error if given a non-AnalyticsEvent', () => {
-    const consoleErrorSpy = jest.spyOn(global.console, 'error').mockImplementation();
+    const consoleErrorSpy = jest.spyOn(console, 'error');
     expect(analyticsReporter.report({ event_type: 'fake event' })).toBeFalsy();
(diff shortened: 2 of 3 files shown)


oshi97:
do we need to update support branches? (ideally we don't but wanted to check)

src/core/analytics/analyticsreporter.js
```diff
@@ -109,11 +109,13 @@ export default class AnalyticsReporter {
       ytag('optin', true);
       cookieData = ytag('yfpc', null);
     } else if (this._conversionTrackingEnabled) {
-      throw new AnswersAnalyticsError('Tried to enable conversion tracking without including ytag');
```
nmanu1, line 112: nit: the import for this can be removed now. looks like we don't have the GH check for linting

yen-tt:
> do we need to update support branches? (ideally we don't but wanted to check)

This hofix branch v1.13.1 will go into support branch v1.13. I will also make a hotfix branch for v1.14.2

nmanu1 (approved):

nmanu1 (dismissed):
looks like tests need to be updated

tests/core/analytics/analyticsreporter.js
```diff
@@ -22,9 +21,12 @@ describe('reporting events', () => {
   });
 
   it('throws an error if given a non-AnalyticsEvent', () => {
-    expect(() => {
-      analyticsReporter.report({ event_type: 'fake event' });
-    }).toThrow(AnswersAnalyticsError);
+    const consoleErrorSpy = jest.spyOn(global.console, 'error').mockImplementation();
```
nmanu1: can this be simplified to `jest.spyOn(console, 'error');`?

yen-tt: I can remove `global.` but I kind of prefer to mock implementation of console so it doesn't actually log these errors in the terminal

yen-tt: I usually prefer to mock implementation of console so it doesn't actually console log the error messages in terminal. But it doesn't look like it does here in SDK so I can remove

tests/core/analytics/analyticsreporter.js
```diff
@@ -22,9 +21,12 @@ describe('reporting events', () => {
   });
 
   it('throws an error if given a non-AnalyticsEvent', () => {
```
nmanu1: nit: can the test names be updated too?

tests/core/analytics/analyticsreporter.js
```diff
@@ -68,11 +70,11 @@ describe('reporting events', () => {
       expect.anything());
   });
 
-  it('throws error if opted in and ytag missing', () => {
+  it('provide a console error if opted in and ytag missing', () => {
```
nmanu1: super nit: missing plural. and actually, can this be "logs a console error" instead?

nmanu1 (commented):
looks like acceptance tests are failing?

yen-tt:
> looks like acceptance tests are failing?

looks unrelated, I will run again to see if it's spurious

nmanu1:
> looks unrelated, I will run again to see if it's spurious

I think v1.13 is from before we mocked the acceptance test fixtures, so it might just be that the response changed and needs to be updated

yen-tt:
> > looks unrelated, I will run again to see if it's spurious
> 
> I think v1.13 is from before we mocked the acceptance test fixtures, so it might just be that the response changed and needs to be updated

yea, was seeing if the answers experience match the new expected before making changes. It seems like we turned on facets for c_cellPhone and removed some entities data for c_popularities so those caused the acceptance test to fail. I was fiddling with the experience config but thought I should just update the response here anyway like you said.

tmeyer2115 (approved):

PR #1770, opened 2022-08-15
deploy to gcp bucket

yen-tt:
Update all deploy workflow to upload to GCP EU bucket, in addition to the AWS S3 bucket. A GCP service account key is added to the repo with upload permission for answers* objects to the gcp bucket.

publicRead access is granted through IAM configuration in terraform file by infra. Cache control is set through the gcp github action.

Note: we still can NOT use environment and reusable workflow together so I can't reuse deploy workflow directly. there would be some duplication between deploy.yml workflow and deploy_hold.yml workflow for now. (for more context from [previous pr](https://github.com/yext/answers-search-ui/pull/1661)). Also tried in github IDE:
Screen Shot 2022-08-16 at 3 22 10 PM

some deployment note "error" within this pr is because I cancelled them to run with new changes

J=SLAP-2144
TEST=manual

- see that files deploy.yml workflow got deployed successfully to GCP bucket in answers/ folder. See that they are in GCP cloud storage site, and that I can access them through the [URL](https://storage.googleapis.com/assets-eu.sitescdn.net/answers/dev/dev-deploy-gcp/answers-modern.js).
- see that cache control is properly set if it's provided to the deploy workflow
- see that i18n files got uploaded successfully
- see that files get upload to answers-search-bar successfully. See that they are in GCP cloud storage site, and that I can access them through the provided URL.
- see that on hold functionality works for gcp deployment too

PR #1787, opened 2022-11-15
Version 1.15

tmeyer2115:
### Changes
- Improvements to organization, reliability, and speed of acceptance tests.
- SDK releases are now deployed to GCP EU as well as AWS. (#1770)
- CI/CD pipeline was sped up. (#1743, #1744)
- The SDK now requires Node v14 or newer (previously the minimum version was v12). (#1751)

### Fixes
- The `searcher` attribute is now correctly passed in Direct Answer Analytics Events. (#1786)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@yext/answers-search-ui",
   "version": "1.14.3",
```
nmanu1: package version needs to be bumped

package.json
```diff
@@ -173,7 +181,7 @@
   "author": "Yext",
   "license": "BSD-3-Clause",
   "engines": {
-    "node": ">=12"
+    "node": ">=14"
```
nmanu1, line 184: do we want to explicitly mention this change in the release notes?

nmanu1 (approved):

PR #1789, opened 2022-11-18
Update GH workflows to remove deprecation warnings

nmanu1:
Upgrade GitHub actions (`actions/checkout`, `actions/setup-node`, `actions/upload-artifact`, `actions/download-artifact`, and `aws-actions/configure-aws-credentials`) and update `set-output` calls to remove all deprecation warnings from the workflows. These warnings were for the deprecation of [Node 12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) and for the `set-output` and `save-state` [commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

J=SLAP-2427
TEST=manual

Test in this PR as well as on `dev/test-output-changes`. See that the warnings are no longer present (except for the exception mentioned in [this PR](https://github.com/yext/slapshot-reusable-workflows/pull/25)).

Note: The Browserstack acceptance test is failing because the Google Maps JS API dropped support for IE11 recently (see Oliver's [post](https://yext.slack.com/archives/C016ZKY42CF/p1668808136926099) on Slack).

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -8,15 +8,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Use Node.js 16
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: 16
           cache: 'npm'
       - run: npm ci
       - name: Download build-output artifact
-        uses: actions/download-artifact@v2
+        uses: actions/download-artifact@v3
         with:
           name: build-output
           path: dist/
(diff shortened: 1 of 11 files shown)


tatimblin (approved):

.github/workflows/unit_test.yml
```diff
@@ -8,16 +8,16 @@ jobs:
     runs-on: ubuntu-latest
```
yen-tt, line 9: can we use the reusable test workflow instead?

nmanu1: this workflow has a step to download the build output artifact, which isn't in the reusable workflow. we need the `dist` directory for the tests to pass and it's nice if we can just download the artifact instead of running the `build` command again like the reusable workflow would do. so, I think it would be better to keep this workflow separate

yen-tt (approved):

PR #1791, opened 2022-11-29
Remove Browserstack acceptance testing on IE11

nmanu1:
Now that the Google Maps JS API has dropped support for IE11, Browserstack acceptance tests on IE11 fail. This PR updates the Browserstack acceptance test script to only run on Safari, and not IE11 anymore.

J=SLAP-2502
TEST=auto

See that the Browserstack check on this PR only runs on Safari and there are no more Google Maps API errors.


oshi97 (approved):
👋

oshi97 (approved):

PR #1802, opened 2023-02-21
Navigation: Open Original Url On Click

oshi97:
Now, when a Navigation component's parentUrl is set, clicking on the link will still
open the original url with target="_self" instead of target="_parent".

This ensures that when a link is clicked on from within an iframe, the link will only cause the iframe
to refresh, and not the parent frame.

J=SLAP-2638
TEST=manual

hook up my local SDK to a local copy of the ufc answers experience 
See that clicking navigation links works without triggering full page reloads
see that a new tab is still opened when I cmd + click
see that middle click still opens a new tab


src/ui/components/navigation/navigationcomponent.js
```diff
@@ -237,6 +237,19 @@ export default class NavigationComponent extends Component {
       this.refitNav();
       DOM.on(DOM.query(this._container, '.yxt-Nav-more'), 'click', this.toggleMoreDropdown.bind(this));
     }
+    const navLinks = DOM.queryAll(this._container, '[data-originalurl]');
+    navLinks.forEach(link => {
+      const originalUrl = link.dataset.originalurl;
+      if (originalUrl) {
+        DOM.on(link, 'click', e => {
+          if (e.metaKey) {
```
nmanu1: I thought `metaKey` would only watch for `Cmd` on Mac. Does this also catch middle click, `Ctrl-click` on Windows, and opening a new tab through the context menu?

oshi97: Ah yeah I tested and on windows it needs to be ctrlKey. For the others, middle click doesn't trigger onclick handlers and the new tab through context menu behavior shouldn't be changed

tmeyer2115 (approved):

src/ui/components/navigation/navigationcomponent.js
nmanu1: if we don't explicitly check here for the new tab being opened through the context menu, wouldn't the new tab open the iFrame directly instead of having it within the parent domain?

oshi97: The new tab will use the parent url from setParentUrl instead of the original url. This is because the tab's href is set to the parent url

nmanu1 (approved):

PR #1804, opened 2023-02-22
Version 1.15.3

oshi97:
## Fixes
* Navigation Component: after setParentUrl is called, clicking on links will still open the original url (#1802)


nmanu1 (commented):
looks like the package version needs to be updated

nmanu1 (approved):

PR #1809, opened 2023-03-24
Fix `near me` searches sometimes not triggering HTML5 geolocation

oshi97:
This PR updates the SDK to the latest version of core, which has been updated to include `inputIntents`
on search-core's `AutocompleteResult` that were expected by the SDK but did not exist before.

search-core has been updated from v1.6 to v2.2, however none of the breaking changes affect the SDK's
external interfaces. The SDK's internal filter registry has been updated to support the breaking static filter changes from v2.0 of search-core.

T=TECHOPS-8462
TEST=auto,manual

hook local SDK up to affected techops site. see that geolocation now fires for edge case where a user
clicks on an autocomplete option for a "near me" search as opposed to typing it out manually.

see that selecting a single static filter works as expected
multiple static filters from the same field works as expected (are $or-ed together)
and multiple static filters from different fields works as expected (are $and-ed together)


src/answers-umd.js
```diff
@@ -731,6 +730,7 @@ class Answers {
       this.core.init({ visitor: visitor });
     } else {
       console.error(`Invalid visitor. Visitor was not set because "${JSON.stringify(visitor)}" does not have an id.`);
+      this.core.init();
```
oshi97, line 733: Needed this change to fix a test. I asked Nidhi (she originally implemented this visitor feature) and she expected it to init here. Without the init the SDK will just error out anyways so I there's no way this could break anything.

tmeyer2115 (approved):

PR #1814, opened 2023-04-20
wcag: use appropriate aria roles for search input

benmcginnis:
A customer noticed that we had an aria-expanded attribute on the search input, which they flagged as incorrect.  Based on https://w3c.github.io/aria/#combobox, it appears we need to use the combobox role on the input element and search on the form element.

J=TECHOPS-8758
TEST=auto

Reran acceptance tests

### DIFF
--- a/src/ui/templates/search/search.hbs
+++ b/src/ui/templates/search/search.hbs
@@ -6,8 +6,8 @@
 <div class="yxt-SearchBar">
   <div class="yxt-SearchBar-container">
     {{#if useForm}}
-      <form class="yxt-SearchBar-form">
-        <label class="sr-only" 
+      <form class="yxt-SearchBar-form" role="search">
+        <label class="sr-only"
                for="{{inputIdName}}"
                id="{{inputLabelIdName}}"
         >
@@ -16,6 +16,7 @@
         <input class="js-yext-query yxt-SearchBar-input"
           id="{{inputIdName}}"
           type="text"
+          role="combobox"
           name="query"
           value="{{query}}"
           {{#if _config.placeholderText}}placeholder="{{_config.placeholderText}}"{{/if}}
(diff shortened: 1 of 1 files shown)


src/ui/templates/search/search.hbs
```diff
@@ -100,6 +102,10 @@
         </button>
       </div>
     {{/if}}
-    <div id="{{autocompleteContainerIdName}}" class="yxt-SearchBar-autocomplete"></div>
+    <div
+      id="{{autocompleteContainerIdName}}"
+      class="yxt-SearchBar-autocomplete"
+      role="listbox">
```
tmeyer2115: Will having a role here conflict with this? https://github.com/yext/answers-search-ui/blob/782a818271c800792c56901136b28d59992080fe/src/ui/templates/search/autocomplete.hbs#L42
We have a `role="listbox"` already within the Autocomplete Component itself.

benmcginnis: ah, this might be part of the problem. The aria-controls attribute's ID needs to match with the ID of the actual listbox control

benmcginnis: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

benmcginnis: Do you know how I might refactor to all the id to be injected to the autocomplete?

benmcginnis: actually it looks like I can hoist `listbox` and change the role on the autocomplete `ul` to `group`, as the current code appears to support grouping of options: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role

benmcginnis: Ok, I _think_ this should work

EmilyZhang777 (dismissed):

PR #1819, opened 2023-05-03
Update Github Actions to sync a separate EU build

cea2aj:
Previously the GH actions were syncing the US
assets to GCP. This change updates the actions so
that a seprate EU build is ran, and only those assets are synced to GCP. Also, remove the syncing of assets from dev branches.

J=BACK-2268
TEST=none

I will test this with a test version (e.g. 0.0.10) and confirm the build works as expected

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -16,7 +16,7 @@ jobs:
   call_unit_test:
     uses: ./.github/workflows/unit_test.yml
     needs: call_build
-  
+
   call_misc_tests:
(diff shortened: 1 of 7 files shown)


EmilyZhang777 (approved):
LGTM! But would like to have mcgin take a look

benmcginnis (approved):
I assume we don't need to do anything for lower environments?

cea2aj:
> I assume we don't need to do anything for lower environments?

yes, I confirmed with Tom that we only need to host assets in the EU for our full releases, and not for dev builds

EmilyZhang777 (approved):

PR #1820, opened 2023-05-05
Pass CloudRegion to Search Core

cea2aj:
This change uses the CloudRegion injected during the build
process and passes it to the Search Core config. It also
passes the environment using the environment provided from
the config rather than manually constructing the endpoints.

J=BACK-2269
TEST=manual

Manually tested the eu assets with an eu test site
and confirmed the requests were being made correctly. Also
tested the search-bar only integration in eu. Confirmed the
us build still worked as expected.

### DIFF
--- a/tests/acceptance/constants.js
+++ b/tests/acceptance/constants.js
@@ -1,5 +1,5 @@
-export const VERTICAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/answers\/vertical\/query/;
-export const UNIVERSAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/answers\/query/;
+export const VERTICAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/search\/vertical\/query/;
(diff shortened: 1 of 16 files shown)


benmcginnis (approved):

EmilyZhang777 (approved):

tests/acceptance/constants.js
```diff
@@ -1,5 +1,5 @@
-export const VERTICAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/answers\/vertical\/query/;
-export const UNIVERSAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/answers\/query/;
+export const VERTICAL_SEARCH_URL_REGEX = /v2\/accounts\/me\/search\/vertical\/query/;
```
EmilyZhang777, line 1: would these 2 endpoints work the same?

cea2aj: yes!

EmilyZhang777 (approved):

PR #1830, opened 2023-06-28
fix: correct implementation of accesssible text for vertical result counts

cea2aj:
The current implementation of verticalresultcount.hbs unnecessarily uses an aria-label and aria-hidden property to handle text that is already accessible for a non-interactive element.

J=TECHOPS-9030


conf/i18n/translations/messages.pot
```diff
@@ -360,14 +359,14 @@ msgctxt "Example: 1-10 of 50"
...
 msgid ""
 "<span class=\"yxt-VerticalResultsCount-start\">[[start]]</span>\n"
-"<span class=\"yxt-VerticalResultsCount-dash\">-</span>\n"
-"<span class=\"yxt-VerticalResultsCount-end\">[[end]]</span>\n"
-"<span class=\"yxt-VerticalResultsCount-of\">of</span>\n"
-"<span class=\"yxt-VerticalResultsCount-total\">[[resultsCount]]</span>"
+"    <span class=\"yxt-VerticalResultsCount-dash\">-</span>\n"
+"    <span class=\"yxt-VerticalResultsCount-end\">[[end]]</span>\n"
+"    <span class=\"yxt-VerticalResultsCount-of\">of</span>\n"
+"    <span class=\"yxt-VerticalResultsCount-total\">[[resultsCount]]</span>"
```
cea2aj: Is it necessary to add a tab to these rows? Since that changed the msgid, that requires us to update all of the .po files to match it. I believe that's what caused the translation test to fail

cea2aj: I just removed this extra spacing because it didn't look like it was required in order to fix the issue, and because it messed up the translation ids

PR #1837, opened 2023-10-03
techops10133: fix multiple search bars on a page

jesuyedavid:


benmcginnis (approved):
lgtm, are there any other ids that we need to scope like this?  also have you tried running your test page through any accessibility tools like Chrome Lighthouse?

src/ui/templates/search/search.hbs
```diff
@@ -23,10 +23,10 @@
           aria-label="{{labelText}}"
           aria-autocomplete="list"
           aria-controls="{{autocompleteContainerIdName}}"
-          aria-describedby="instructions"
+          aria-describedby="{{searchBarContainerId}}-instructions"
           aria-haspopup="listbox"
         >
-        <div id="instructions" class="yxt-SearchBar-instructions sr-only">
+        <div id="{{searchBarContainerId}}-instructions" class="yxt-SearchBar-instructions sr-only">
           {{translate
```
cea2aj, line 30: does this solution use the autocompleteContainerIdName at all?

jesuyedavid: yes, i see it on line 25. if that's what you are asking?

cea2aj: ah that makes sense

PR #1839, opened 2023-10-04
fix breaking unit test and upgdated package version

jesuyedavid:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-search-ui",
-  "version": "1.16.1",
+  "version": "1.16.2",
```
cea2aj, line 3: can you update the package-lock.json as well?

src/ui/components/search/autocompletecomponent.js
```diff
@@ -214,7 +214,7 @@ export default class AutoCompleteComponent extends Component {
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
       listLabelIdName: this.listLabelIdName,
-      autocompleteContainerIdName: `yxt-AutoComplete-container-${this._config.name.replaceAll('.', '-')}`,
+      autocompleteContainerIdName: `yxt-AutoComplete-container-${this._config.name ? this._config.name.replaceAll('.', '-') : ''}`,
```
cea2aj: does `yxt-AutoComplete-container-${this._config.name?.replaceAll('.', '-')}` work instead?

src/ui/rendering/defaulttemplatesloader.js
```diff
@@ -44,7 +44,7 @@ export default class DefaultTemplatesLoader {
         onload: resolve,
         onerror: reject,
         async: true,
-        src: COMPILED_TEMPLATES_URL
+        src: 'answerstemplates.compiled.min.js'
```
cea2aj: What's the reasoning behind this change? I don't think it should be needed

src/ui/components/search/autocompletecomponent.js
```diff
@@ -214,7 +214,7 @@ export default class AutoCompleteComponent extends Component {
       resultIndex: this._resultIndex,
       promptHeader: this._originalQuery.length === 0 ? this.promptHeader : null,
       listLabelIdName: this.listLabelIdName,
-      autocompleteContainerIdName: `yxt-AutoComplete-container-${this._config.name.replaceAll('.', '-')}`,
+      autocompleteContainerIdName: `yxt-AutoComplete-container-${this._config.name?.replaceAll('.', '-')}`,
```
cea2aj: I think you need a question mark after `replaceAll` as well like like `this._config.name?.replaceAll?('.', '-')`

src/ui/components/search/searchcomponent.js
```diff
@@ -678,7 +678,7 @@ export default class SearchComponent extends Component {
       labelText: this.labelText,
       inputLabelIdName: this.inputLabelIdName,
       submitIcon: this.submitIcon,
-      searchBarContainerId: this._container.id,
+      searchBarContainerId: this._container.id || '',
```
cea2aj: I think that `??` is preferable to `||` here because if the id is 0 and `||` us used, that will fall back to the empty string whereas `??` won't fall back if the value is 0

src/ui/components/search/autocompletecomponent.js
jesuyedavid: i thought if this._config.name is null or undefined, it will just return `yxt-AutoComplete-container-`

cea2aj: oh, you're correct. The extra ? after `replaceAll` would only be useful if name is present but for some reason it doesn't have the replaceAll method

cea2aj:
It looks like there is a similar testing error in the code coverage checks which I think are worth fixing

cea2aj (approved):
LTGM

PR #1841, opened 2023-10-05
Fix null analytics reporter

jesuyedavid:


src/core/core.js
```diff
@@ -94,7 +95,7 @@ export default class Core {
      * A local reference to the analytics reporter, used to report events for this component
      * @type {AnalyticsReporter}
      */
-    this._analyticsReporter = config.analyticsReporter;
+    this._analyticsReporter = config.analyticsReporter || new AnalyticsReporter();
```
cea2aj: What do you think about falling back to the `NoopAnalyticsReporter` instead? If we were to create the `AnalyticsReporter` like this, it wouldn't be configured properly because no params are passed to the constructor

jesuyedavid: i thought `NoopAnalyticsReporter` does nothing when you call report(event); it simply returns true. so, if we fall back on it, and then try to report an event as we are doing on line 184, it will just report true, which is not the behavior i believe we want.

cea2aj: My thought process is that if an `analyticsReporter` isn't present on the config, then we should use the `NoopAnalyticsReporter` in order to not send any analytics events. That way we won't run into an error about the analytics reporter being null. If we do want the analytics reporter to actually send the event, we should look into why the `config.analyticsReporter` isn't set

jesuyedavid: that makes sense! i will use `NoopAnalyticsReporter` then. thanks!

benmcginnis: do we have any idea _why analytics reporter was missing_ when we do local development?

jesuyedavid: i traced it to `answers-search-ui/src/answers-umd.js` looks like it `_analyticsReporterService` was initialized as null. and then on lines 257 to 260, it didn't end up with a value probably because it didn't meet the conditions in those lines. i think it makes sense to not send any analytics in core.js if somehow, the value being passed through is null. the interface allows for that.

benmcginnis: yeah so I think we want to understand _why it was initialized as null_.  It appears to work in the platform, so there is something different when you use the code editor vs when you do local development.  It seems like if we can determine that then it might work.  @cea2aj do you think it might be related to the api domain logic?  I suspect that most developers would expect to be able to test analytics locally, which is why I would prefer to not just use the noop reporter.

cea2aj: I think it's more likely something other than the api domain logic. Have we double checked that the businessId is set in the config? That's the only thing that affects the `_eligibleForAnalytics` variable in answers-umd.js. I think it's still worth using the `NoopAnalyticsReporter` in answers-umd if `_eligibleForAnalytics` is false

jesuyedavid: looks like this. _eligibleForAnalytics is the culprit. core config needs businessId, but it looks like nationwide's `global_config.json` for that site didn't have businessId set. once i set it and did another run, the original issue disappeared. i will abandon this cr. thanks everyone.

PR #1843, opened 2023-11-30
Bump Core to v2.4.0

nmanu1:
This PR bumps the `search-core` version to v2.4.0 to get support for new Rich Text & Markdown field types returned as
raw html in direct answers (see [release notes](https://github.com/yext/search-core/releases/tag/v2.4.0)). We had already added [support](https://github.com/yext/answers-hitchhiker-theme/pull/1138) to render these fields in the Theme, but without the Core update, a search that returns an `html` field type gives an error and doesn't actually get rendered.

J=TECHOPS-10638
TEST=manual

Point the site from the techops to a locally served version of the SDK with this version bump and see that an HTML field type for a featured snippet direct answer is properly rendered instead of throwing an error.

Screenshot 2023-11-30 at 4 55 46 PM

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -487,7 +487,7 @@ CDN
 
 The following NPM package may be included in this product:
 
- - @yext/search-core@2.3.0
+ - @yext/search-core@2.4.0
 
 This package contains the following license and notice below:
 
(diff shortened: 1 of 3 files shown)


benmcginnis (approved):
LGTM!

PR #1848, opened 2023-12-01
Version 1.16.4

nmanu1:
### Fixes
- Updated `@yext/search-core` to v2.4.0 to properly support new Rich Text & Markdown field types returned as raw html in direct answers (#1847)


nmanu1 (commented):
This change is the same as the one approved in #1844, except with the correct version (v1.16.4 instead of v1.16.3)

PR #1851, opened 2023-12-01
Upgrade @babel/traverse

nmanu1:
Upgrade `@babel/traverse` from v7.17.10 to 7.23.5 to address a critical vulnerability in the package.

J=VULN-38480
TEST=auto

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -1,8 +1,8 @@
 The following NPM packages may be included in this product:
 
- - @babel/code-frame@7.16.7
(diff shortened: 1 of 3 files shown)


benmcginnis (approved):
lgtm, but do we need to set the version in the package.json or are we just relying on the package lock?

nmanu1:
> lgtm, but do we need to set the version in the package.json or are we just relying on the package lock?

it's a transitive dependency, so we don't currently specify a version in our package.json

benmcginnis:
Ah ok great

PR #1855, opened 2024-01-30
Resolve Vulnerabilities

EmilyZhang777:
This PR resolves the following vulnerabilities:
- bump `search-core`'s version to 2.5.1 to include the latest vulnerabilities fix
- remove usages of insecure document method
- bump the version of `http-cache-semantics` (to 4.1.1), `get-func-name` (to 2.0.2) and `terser` (to 5.14.2) to resolve vulnerabilities to uncontrolled resource consumption and inefficient regular expression complexity

J=VULN-37771, VULN-37772, VULN-37773, VULN-38373, VULN-38391, VULN-38401
TEST=auto

Added new tests for the chang. Ran `npm run test` and `npm run acceptance`.


package.json
```diff
@@ -77,6 +77,7 @@
     "full-icu": "^1.3.4",
     "generate-license-file": "^1.3.0",
     "gettext-extractor": "^3.5.2",
+    "get-func-name": "^v2.0.1",
```
nmanu1: why are we adding this as a dev dependency when we didn't need it before? same with `http-cache-semantics`?

EmilyZhang777: They get added as dependencies from other libraries. But the version defined on the other libraries are still lower than what we want so I specified the versions like this.

nmanu1: I see. I don't think we need to add transitive deps to the package.json. Now that we've updated the package-lock, the repo no longer has the vulnerability. And, if a user installs the repo, they'll get the latest versions of all transitive deps anyways. If one of our deps pins to a specific version of the transitive dep with the issue, then we'd have to add the transitive dep to the package.json overrides section. But, it doesn't look like that's the case here

EmilyZhang777: ah I see. Didn't know I can do that. Updated!

nmanu1: I actually don't think we'd want to use the overrides section for these transitive deps because it doesn't look like we're blocked from upgrading them using `npm update`. Adding deps to the overrides section unnecessarily can cause bloat and can cause issues if the libraries that actually use them change their dependency requirements. We don't want to force libraries to use this version unless we're not able to upgrade the deps without it

EmilyZhang777: I was trying to have something that specifically says what's the min version we want for the transitive deps. But I agree with your concern about any future dependency requirements change. Seems like `ansi-regex` is also a transitive dep. Maybe it's fine to keep them there?

nmanu1: the reason the other deps (like ansi-regex) are there is because we're not able to update those via `npm update`. some of our other deps require a version of ansi-regex that still has the vulnerability. we have to force them to use a different version, by putting them in the overrides

I agree that it would be nice to have a place where we specify the min version, but I'm not too concerned because this won't be an issue for users setting up a new repo, and especially because these particular transitive deps are all dev dependencies and wouldn't be installed by a user anyways

tests/setup/enzymeadapter.js
```diff
@@ -110,7 +110,16 @@ export default class AnswersAdapter extends EnzymeAdapter {
           composed: args.composed,
           cancelable: args.cancelable
         });
-        Object.assign(event, args);
+
+        if (Array.isArray(...args)) {
```
nmanu1, line 114: what vulnerability is this change related to?

package.json
EmilyZhang777: ok sounds good. Going to remove them!

package.json
```diff
@@ -89,7 +89,7 @@
     "gulp-rollup-lightweight": "^1.0.10",
     "gulp-rtlcss": "^1.4.2",
     "gulp-sass": "^5.1.0",
-    "gulp-uglify-es": "^2.0.0",
+    "gulp-uglify-es": "^3.0.0",
```
nmanu1: since we use this dep for bundling, just wanted to double check that you tried running a build and using the bundle locally to test that it still works? wasn't sure how big of a difference there was in this library since we're bumping to a new major version

tests/setup/enzymeadapter.js
EmilyZhang777: This never worked and there were no usage of this case so it didn't surface. But the test I added need it.

package-lock.json
nmanu1, line 1: looks like we've reverted to lockfile v1

EmilyZhang777: reverted

package.json
nmanu1: actually, I'm not seeing any of the vulnerability items pointing specifically to gulp-uglify-es. if this is for the terser vulnerability, can we just `npm update terser`?

EmilyZhang777: updated

tests/ui/components/filters/filteroptionscomponent.js
```diff
@@ -116,6 +116,48 @@ describe('filter options component', () => {
...
+      control: 'multioption',
+      searchable: true
+    };
+    const component = COMPONENT_MANAGER.create('FilterOptions', config);
+    const wrapper = mount(component);
+    expect(options).toHaveLength(6);
+    const searchInputEl = wrapper.find('.js-yxt-FilterOptions-filter');
+    expect(searchInputEl).toHaveLength(1);
+
+    // empty input
```
nmanu1, line 131: can we split each of these cases into separate tests? right now, if one case fails, we won't run any cases after it. it would also make it easier to debug if it fails in the future

src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -448,21 +448,21 @@ export default class FilterOptionsComponent extends Component {
             if (!filter) {
               filterOption.classList.remove('hiddenSearch');
               filterOption.classList.remove('displaySearch');
-              labelEl.innerHTML = labelText;
+              labelEl.textContent = labelText;
```
nmanu1, line 451: in case you haven't already, it would be good to spin up this component locally and test that the text still appears as expected. you should be able to just serve and navigate to the acceptance test page for it and try things out

EmilyZhang777: The jest test I added already covers what I changed. Ack

tests/ui/components/filters/filteroptionscomponent.js
EmilyZhang777: I think they fit better in a single test as the initial config settings are the same only the input in the UI changed. Ack

nmanu1: if you want to reuse the config settings, you could pull out the config variable. or you could put all the tests into a describe block and do all the setup in the beforeEach

PR #1857, opened 2024-02-05
Change icons from divs to spans

likimmy:
To address the first issue in [this](https://yexttest.atlassian.net/browse/TECHOPS-11042) techops, the icons for answers-search-ui should be wrapped in spans, not divs.


EmilyZhang777 (approved):

nmanu1 (commented):
just to double check, does all the styling in the test site and acceptance test pages look the same after this change? `div`s and `span`s are sometimes styled differently

also, it's probably not a big deal, but we may want to call this change out in the release documentation because any changes to the DOM like this could technically break the styling of an existing implementation if they specifically targeted `div`s that we're changing to `span`s

benmcginnis:
> just to double check, does all the styling in the test site and acceptance test pages look the same after this change? `div`s and `span`s are sometimes styled differently
> 
> also, it's probably not a big deal, but we may want to call this change out in the release documentation because any changes to the DOM like this could technically break the styling of an existing implementation if they specifically targeted `div`s that we're changing to `span`s

The Icon class adds a `display: inline-block;` to the element it's added to so this shouldn't change anything.

src/ui/templates/icons/searchBarIcon.hbs
```diff
@@ -1,41 +1,41 @@
 {{#if showLoadingIndicator}}
   {{#if customLoadingIconUrl}}
-    <div class="yxt-SearchBar-CustomLoadingImage js-yxt-SearchBar-LoadingIndicator js-yxt-SearchBar-Icon yxt-SearchBar-Icon--inactive">
+    <span class="yxt-SearchBar-CustomLoadingImage js-yxt-SearchBar-LoadingIndicator js-yxt-SearchBar-Icon yxt-SearchBar-Icon--inactive">
```
benmcginnis, line 3: on this one we might need to check on what if any css this class adds as it doesn't use the `Icon` class

src/ui/templates/icons/voiceSearchIcon.hbs
```diff
@@ -1,17 +1,17 @@
-<div class="yxt-SearchBar-voiceIconWrapper Icon--lg">
-  <div class="yxt-SearchBar-micIconWrapper {{#if customMicIconUrl}}yxt-SearchBar-CustomMicIcon{{/if}}">
+<span class="yxt-SearchBar-voiceIconWrapper Icon--lg">
```
benmcginnis: need to double check these too

src/ui/templates/icons/searchBarIcon.hbs
likimmy: doesn't look like this one adds any css

src/ui/templates/icons/voiceSearchIcon.hbs
likimmy: the only css I see is for micIconWrapper which is just `position: absolute`

likimmy: tested changes by pointing HH theme at my local answers-search-ui and the icons seem to be looking as expected. do we want to call this change out in the release documentation like nidhi mentioned?

benmcginnis: We should probably add the Icon class where it's missing just to be sure. Previously this would have been block display and making it a span would change it to inline. Inline-block is probably safest.

benmcginnis:
How do we do the callout in the release documentation, Nidhi?

nmanu1:
> How do we do the callout in the release documentation, Nidhi?

On our side, this would just mean adding an extra note in our GH release notes explaining possible ramifications of the change. But, it would be up to Baigel if he wants it to be documented more widely, such as in a HH post. The likelihood of this breaking someone's styling is small, but technically possible, so I'm not sure how cautious we want to be about it.

As it is right now, if we release this change in a hotfix, all users on v1.16 will get the change automatically without any action on their part (since most implementations pin to a minor version rather than a patch). This could technically mean that styling may suddenly change for them with no action on their part. If we're worried about this possibility, it may be better to release this in a new minor version so implementations wouldn't get the change without deliberately upgrading. If not, then it's probably fine as is

src/ui/templates/icons/voiceSearchIcon.hbs
likimmy: i am not super familiar with handlebars js, so my understanding may be incorrect, but for voiceSearchIcon and SearchBarIcon, the children in the spans seems to be either iconPartial or builtinIcon, which use the Icon class. would this cover the display styling to be inline-block or would we need to set this explicitly again somehow? if it is the former, then i think the only instance of not using the Icon class is on line 10 of `searchBarIcon.hbs`, in which case should i just add an Icon class span wrapped right on the circle element?

likimmy: discussed offline and concluded that this PR will just address the icon classes directly and any other elements like voiceSearchIcon can be addressed separately in the future.

nmanu1 (commented):
for this techops, would we need to address all instances of divs being used within a button anywhere in answers-search-ui and the HH theme? or did they specifically only care about answers-search-ui icons?

likimmy:
> for this techops, would we need to address all instances of divs being used within a button anywhere in answers-search-ui and the HH theme? or did they specifically only care about answers-search-ui icons?

I'm honestly not sure, I was going off of a code link that was included in the techops that was from answers-search-ui. I'm assuming they only care about the icons from that sdk since they didn't link any other relevant code.

benmcginnis:
> > for this techops, would we need to address all instances of divs being used within a button anywhere in answers-search-ui and the HH theme? or did they specifically only care about answers-search-ui icons?
> 
> I'm honestly not sure, I was going off of a code link that was included in the techops that was from answers-search-ui. I'm assuming they only care about the icons from that sdk since they didn't link any other relevant code.

I think what's called out in the techops is specifically riven by the the icon component

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -280,6 +280,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
   &-buttonImage .Icon svg, .Icon-image
   {
+    display: inline-block;
```
nmanu1: I think `div` styling adds `display: block` rather than `inline-block`

src/ui/templates/icons/searchBarIcon.hbs
nmanu1, line 3: do we also need to update this element to have `display: block` styling?

src/ui/templates/icons/voiceSearchIcon.hbs
nmanu1: I think all the new spans in this file would also need the styling to be added. or we could revert the change for now if the techops didn't need this icon and if we're making a follow-up item to address any remaining similar violations

likimmy: i think the former is the approach @benmcginnis was wanting for this specific PR, I'll make a follow-up item for addressing the remaining violations.

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -194,6 +194,7 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
 
   &-LoadingIndicator
   {
+    display: block;
```
nmanu1: I don't think we need to add this style on this class because it's only used for elements that already have `yxt-SearchBar-AnimatedIcon` which has display: flex

likimmy: done

likimmy:
to-do item address voiceSearchIcon [here](https://yexttest.atlassian.net/browse/BACK-2975)

src/ui/sass/modules/_SearchBar.scss
```diff
@@ -280,13 +280,18 @@ $searchbar-button-text-color-active: var(--yxt-searchbar-button-text-color-base)
...
     width: 2em;
     height: 2em;
   }
 
   &-instructions {
     display: none;
   }
+
+  &-customLoadingImage {
```
nmanu1: the class name on the span is capitalized, so this should be `&-CustomLoadingImage` if you're trying to target `yxt-SearchBar-CustomLoadingImage`

src/ui/templates/icons/voiceSearchIcon.hbs
```diff
@@ -13,5 +13,5 @@
       {{> icons/builtInIcon iconName='voice_search_dots'}}
     {{/if}}
   </div>
-  <span class="yxt-SearchBar-voiceSearchText sr-only"></span>
-</div>
\ No newline at end of file
+  <div class="yxt-SearchBar-voiceSearchText sr-only"></div>
```
nmanu1: looks like this was originally a span, so we should keep it that way

nmanu1:
> to-do item address voiceSearchIcon [here](https://yexttest.atlassian.net/browse/BACK-2975)

can we update this item to also include looking into any other usages of a div within a button? if this is not valid html, then we'll want to address all of them

nmanu1 (approved):

PR #1869, opened 2024-03-05
Bump browserify-sign to v4.2.3

EmilyZhang777:
Bump browserify-sign to v4.2.3 to resolve a vulnerability.

J=VULN-38744
TEST=auto,manual

Ran `npm run test` and `npm run acceptance`. Ran `npx serve -p 9999` and clicked around.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -5322,7 +5322,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The following NPM package may be included in this product:
 
- - readable-stream@2.3.7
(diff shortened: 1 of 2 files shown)


nmanu1 (commented):
did you also check that running the build and using the generated bundles in the Theme test site was fine? it looks like browserify-sign is dep of a library we use for bundling, so it would affect our build step

nmanu1 (commented):
this change should point to a hotfix branch if we're making a new version, or develop if we're not

EmilyZhang777:
> did you also check that running the build and using the generated bundles in the Theme test site was fine? it looks like browserify-sign is dep of a library we use for bundling, so it would affect our build step

yep! Tested HH theme still works with a local build.

nmanu1 (approved):

PR #1871, opened 2024-03-08
Fix Filter Options Highlighting

EmilyZhang777:
Previously, when we moved from `innerHTML` to `textContent`, it broke the filter options highlighting since the `<strong>` tag is no longer applied properly. This PR fixes the issue while still avoid insecure document method.

J=TECHOPS-11319
TEST=auto, manual

Ran `npm run test`. Serve answers-search-ui locally and checked out hh theme. Made sure the highlighting are working.
https://drive.google.com/file/d/1lBVOA759BPwsqmENIv6kFe2FasR25EKO/view?usp=sharing

### DIFF
--- a/src/core/models/highlightedvalue.js
+++ b/src/core/models/highlightedvalue.js
@@ -19,6 +19,15 @@ export default class HighlightedValue {
   }
 
   /**
+   * get highlighted value elements
+   * @returns {Array<Element|string>}
+   */
+  getElements () {
+    this._sortMatchedSubstrings();
+    return this.buildHighlightedElements(this.value, this.matchedSubstrings);
+  }
+
+  /**
    * get highlighted value string
    * @param {Function} transformFunction takes a string and returns the transformed string
    * @returns {string} The value interpolated with highlighting markup and transformed in between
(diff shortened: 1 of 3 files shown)


src/ui/components/filters/filteroptionscomponent.js
```diff
@@ -455,10 +455,13 @@ export default class FilterOptionsComponent extends Component {
...
                 filterOption.classList.add('displaySearch');
                 filterOption.classList.remove('hiddenSearch');
-                labelEl.textContent = new HighlightedValue({
+                const highlightedElements = new HighlightedValue({
                   value: labelText,
                   matchedSubstrings: [matchedSubstring]
-                }).get();
+                }).getElements();
+
+                labelEl.textContent = '';
```
nmanu1, line 463: do we need to set the textContent to an empty string? or is it sufficient to just append the elements in the next line?

src/core/models/highlightedvalue.js
```diff
@@ -135,6 +144,56 @@ export default class HighlightedValue {
...
+   * @param {Object} highlightedSubstrings highlighting specifiers to apply to input object
+   * @param {Function} transformFunction function to apply to strings in between highlighting markup
+   * @returns {Array<Element|string>} a list of elements/strings representing the val with
+   * highlighting applied. Example returned value :
+   *   [
+   *     'Save time & bank on your terms at over 1,800',
+   *     <strong>ATM</strong>,
+   *   ]
+   */
+  buildHighlightedElements (
```
nmanu1: nit: if this is just an internal helper function, can the name begin with an underscore?

src/core/models/highlightedvalue.js
```diff
@@ -135,6 +144,56 @@ export default class HighlightedValue {
...
+    val,
+    highlightedSubstrings,
+    transformFunction = function (x) { return x; }
```
nmanu1: it doesn't look like we use this in getElements. can we remove it?

src/core/models/highlightedvalue.js
nmanu1: this function looks similar to buildHighlightedValue above. what's the difference between the two? just the return type? is it possible to reuse some of the logic between them?

nmanu1 (commented):
can you add some screenshots/recording of what it looks like? also, can you provide a little more detail in the PR description?

src/ui/components/filters/filteroptionscomponent.js
EmilyZhang777: it's used to clear the previous value

src/core/models/highlightedvalue.js
EmilyZhang777: are you talking about buildHighlightedValue? AppliedHighlightedFields still uses it.

nmanu1: I meant the transformFunction param for buildHighlightedElements. It looks like getElements is the only function that calls buildHighlightedElements and it doesn't pass in a transform function

src/core/models/highlightedvalue.js
EmilyZhang777: Yeah. I'd prefer to keep them separate since the return type and how to construct it is quite different. Keeping them separate can be clearer.

EmilyZhang777: updated!

EmilyZhang777:
> can you add some screenshots/recording of what it looks like? also, can you provide a little more detail in the PR description?

Added more details and a screenshot.

src/core/models/highlightedvalue.js
EmilyZhang777: yeah, that's fair! Updated.

nmanu1 (approved):
nit: the PR description styling seems a little off because the editor accepts html. you'll need to put the strong tag in a code style (e.g. `<strong>`) or escape it (e.g. \<strong>)

besides that, lgtm!

PR #1872, opened 2024-03-12
Version 1.17.1

EmilyZhang777:
### Fixes
- Bump browserify-sign to v4.2.3 to resolve a vulnerability (#1869)
- Fix filter options highlighting (#1871)


nmanu1 (commented):
It looks like the third party notices check is failing. Can you run the generate-notices command locally to make sure there are no changes that need to be committed? We may need an item to investigate why the check is failing, but it doesn't look related to your changes

nmanu1 (approved):

PR #1877, opened 2024-05-01
Update Workflow Versions to Remove node16 Deprecation Warnings

jesuyedavid:
Notable Changes:

1. We added two new jobs: `merge_multiple_artifacts` and `overwrite_artifact` in `.github/workflows/build_i18n.yml` to handle merging multiple locale files into the `build-output-US` and `build-output-EU` artifacts. This was necessary since we have to handle no longer being able to have multiple jobs in the same workflow write to one artifact name and mutate it, [a breaking change](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact) in `action/upload-artifact@v4`. Since `build_and_deploy_i18n.yml` calls both `build.yml` and `build_i18n.yml`, we can't have both write an artifact called `build-output-US` unless we pass the `overwrite: true` input in the latter. The `action/upload-artifact/merge@v4` doesn't allow overwriting, so we need an additional step of merging and uploading with a temporary name and then downloading the merged artifact and re-uploading to overwrite the correct name.

2. We bumped `google-github-actions/auth` and `google-github-actions/upload-cloud-storage` from v0 to v2

3. Updated  `actions/checkout`, `actions/setup-node`, `actions/upload-artifact`, `actions/download-artifact`, and `aws-actions/configure-aws-credentials` actions to v4

[J=WAT-4020](https://yexttest.atlassian.net/browse/WAT-4020)
TEST=manual

Testing was done in a forked repo. View [actions](https://github.com/jesuyedavid/answers-search-ui/actions) to see testing details and test commits. We also checked the US/EU build artifact unzipped files to ensure they are uploaded to AWS/GCP properly and that they look similar to build files prior to this update. Finally, we did local testing from the Answers-Hitchhiker-Theme repo's test site that pointed to our dev assets in AWS. The `dev` bucket in GCP did not have the correct permissions to allow local testing, so this local testing will be done after merging this PR and the changes from develop are pushed to `canary` in GCP.


.github/workflows/deploy_hold.yml
```diff
@@ -27,14 +27,14 @@ jobs:
     runs-on: ubuntu-latest
     environment: production # sets in github repo with reviewer requirement protection rule
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Download build-output-US artifact
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
```
nmanu1, line 32: have you checked that the newly uploaded artifacts in aws and gcp are the same format as before?

.github/workflows/build_and_deploy_hold.yml
```diff
@@ -16,16 +16,42 @@ jobs:
     with:
       cloud_region: EU
 
+  merge_multiple_us_artifacts:
```
nmanu1: have you tried testing the other build and deploy workflows, particularly build_and_deploy_i18n and build_and_deploy_search_bar? I think they'll all run into the same issue, which is why I suggested having these merge jobs in build_i18n.yml so it can be reused and downstream workflows don't have to worry about it

.github/workflows/deploy_hold.yml
```diff
@@ -50,19 +50,19 @@ jobs:
...
           path: dist/
       - name: Authenticate to Google Cloud
         id: auth
-        uses: google-github-actions/auth@v0
+        uses: google-github-actions/auth@v2
         with:
           credentials_json: '${{ secrets.GCP_SA_KEY }}'
       - name: Deploy to GCP Bucket
-        uses: google-github-actions/upload-cloud-storage@v0
+        uses: google-github-actions/upload-cloud-storage@v2
```
nmanu1, line 65: have you tested anywhere that this action and the auth action above complete successfully? I know it's not possible from the forked repo since we don't have the creds there. if you need help with testing it, let me know!

.github/workflows/build_and_deploy_hold.yml
```diff
@@ -16,16 +16,42 @@ jobs:
...
+    needs: call_build
+    steps:
+      - name: Merge into build-output-US Artifact
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: build-output-US
+          pattern: build-output-US-*
+          merge-multiple: true
+
+  merge_multiple_eu_artifacts:
```
nmanu1: seeing some warnings [here](https://github.com/jesuyedavid/answers-search-ui/actions/runs/8993739075) from these merge jobs when they ran in the forked repo

jesuyedavid: looks like the warning is [expected](https://github.com/octokit/request-action/issues/26#issuecomment-661936263). i can make a follow up item to address this warning specifically: a potential solution is to use github secrets instead.

.github/workflows/deploy_hold.yml
jesuyedavid: for aws, seems like although [the new file size](https://github.com/yext/answers-search-ui/actions/runs/9024459183) of `build-output-US` is ~8mb, the files in its unzipped folder accumulate to ~39mb, and look same as they are in the current [aws bucket ](https://us-east-1.console.aws.amazon.com/s3/buckets/assets.sitescdn.net?region=us-east-1&bucketType=general&prefix=answers/dev/dev-update-workflow/&showversions=false) (after logging in). not sure why the artifact sizes in the workflow screens are [different](https://github.com/yext/answers-search-ui/actions/runs/8694652127). my guess is that github is now using some new kind of zipping algorithm? 


Screenshot 2024-05-09 at 11 33 12 PM

nmanu1: ok, so after unzipping both the old and new artifacts, the folders have the same size? if so, that seems fine. also, have you tested by spinning up the Theme test site and pointing to the new assets you deployed to aws? you'll want to spot check to make sure everything still works as expected using them

.github/workflows/build_and_deploy_search_bar.yml
```diff
@@ -18,12 +18,28 @@ jobs:
...
+      - call_build_for_eu
+    with:
+      cloud_region: EU
 
   call_acceptance_search_bar:
     uses: ./.github/workflows/acceptance_search_bar.yml
-    needs: call_build
+    needs: 
+      - merge_multiple_us_artifacts
+      - merge_multiple_eu_artifacts
```
nmanu1: we don't need any of the changes to this file, right? and same with build_and_deploy_hold?

.github/workflows/build_and_deploy.yml
```diff
@@ -13,6 +13,11 @@ jobs:
   call_build:
     uses: ./.github/workflows/build.yml
 
+  call_build_eu_locales:
```
nmanu1: we don't need the eu stuff here anymore since this was just for testing

.github/workflows/build_and_deploy_i18n.yml
```diff
@@ -31,6 +31,11 @@ jobs:
   call_build_locales:
     uses: ./.github/workflows/build_i18n.yml
 
+  call_build_for_eu_locales:
```
nmanu1: similarly, we don't need eu stuff here anymore either

.github/workflows/build_i18n.yml
```diff
@@ -51,7 +51,34 @@ jobs:
...
       with:
-        name: build-output-${{ inputs.cloud_region }}
+        name: build-output-${{ inputs.cloud_region }}-${{ matrix.language }}
         path: dist/
+        
+  merge_multiple_artifacts:
+      needs: build
+      runs-on: ubuntu-latest
+      steps:
+        - name: Merge into build-output-${{ inputs.cloud_region }} Artifact
```
nmanu1: nit: update to have the `-temp` prefix for the name

.github/workflows/deploy.yml
```diff
@@ -42,3 +44,27 @@ jobs:
           --acl public-read \
           --recursive \
           --cache-control ${{ inputs.cache-control }}
+
+  deploy-gcp:
```
nmanu1: we don't need any gcp stuff in this workflow. that was just for testing

.github/workflows/build_and_deploy.yml
```diff
@@ -39,6 +39,7 @@ jobs:
     secrets:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      GCP_SA_KEY: ${{ secrets.GCP_SA_KEY }}
```
nmanu1: can remove gcp secret

.github/workflows/build_and_deploy_i18n.yml
```diff
@@ -45,6 +45,7 @@ jobs:
     secrets:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      GCP_SA_KEY: ${{ secrets.GCP_SA_KEY }}
```
nmanu1: can remove all the gcp secrets

.github/workflows/deploy.yml
```diff
@@ -19,19 +19,21 @@ on:
         required: true
       AWS_SECRET_ACCESS_KEY:
         required: true
+      GCP_SA_KEY:
+        required: true
```
nmanu1: can remove gcp input

.github/workflows/deploy.yml
```diff
@@ -41,4 +43,4 @@ jobs:
           aws s3 cp ./dist/ s3://assets.sitescdn.net/${{ inputs.bucket }}/${{ inputs.directory }} \
           --acl public-read \
           --recursive \
-          --cache-control ${{ inputs.cache-control }}
+          --cache-control ${{ inputs.cache-control }}
```
nmanu1: nit: add back newline

nmanu1 (approved):

PR #1883, opened 2024-06-28
Upgrade testcafe to remove dependency vuln

Fondryext:
Upgrade testcafe from 1.19.0 to 3.6.1. In the process, this removes the dependency on jsonwebtoken, which had a vuln in the version we were using.
https://nvd.nist.gov/vuln/detail/CVE-2022-23539#vulnCurrentDescriptionTitle J=VULN-39101
TEST=auto


vijay267 (approved):
LGTM

PR #1884, opened 2024-07-05
Update package version to 1.17.2

Fondryext:
Since we're incrementing the version, we need to reflect this
in the package.json file
J=VULN-39101
TEST=auto


vijay267 (approved):

vijay267 (commented):
Although do you know what's going on with the failed WCAG check?

Fondryext:
It looks like some issues with making some api calls, and also a variety of html issues, as WCAG violations tend to be. I haven't looked into it any further.

PR #1896, opened 2024-09-11
Answers SDK: add cloudChoice support

Fondryext:
Now that cloudChoice is a param inside of search-core, this CR adds support for setting it in the SDK, in the init() function, including setting the old behavior as the default. It is not needed elsewhere like with the CLOUD_REGION setting, as that is used for things like analytics tracking.

J=WAT-4375
TEST=auto

Ran test suite

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -487,7 +487,7 @@ CDN
 
 The following NPM package may be included in this product:
 
- - @yext/search-core@2.5.1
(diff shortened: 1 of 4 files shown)


Fondryext:
Honestly not too familiar with manual testing of just the SDK. Willing to look into it, but figured I'd put this up since it's smaller and also I'm heading out early today.

vijay267 (approved):
Looks good!

PR #1911, opened 2025-01-21
Add GDA Object Model

k-gerner:
Similar to the existing `DirectAnswer` model, we will take in an object from `search-core` (`GenerativeDirectAnswerResponse` in this case) and use it to construct a model.

J=WAT-4592
TEST=auto
Ran unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-search-ui",
-  "version": "1.17.8",
+  "version": "1.18.0",
   "description": "Javascript Search Programming Interface",
   "main": "dist/answers-umd.js",
   "repository": {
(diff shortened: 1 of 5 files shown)


Fondryext (changes requested):
Normally on the develop branch we'll make a version like '2.5.0-beta.5' to make it clear that the version isn't the real one.
In this case, though, I think it actually makes more sense to have a branch, not develop, where we push most of the changes to until all the sdk work for this portion is done - since this change by itself isn't really a full change. Up to you on that branch naming.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-search-ui",
-  "version": "1.17.8",
+  "version": "1.18.0",
```
Fondryext: thanks for making the other branch! My previous message wasn't worded too well - we'll still want this to be a beta version

k-gerner: Ah okay, I've updated it now

Fondryext (approved):
LGTM! Thanks

PR #1912, opened 2025-01-23
Add flow for automatically calling into the GDA endpoint

k-gerner:
In this change I added the following functionality:
- at the global config level, there will be a new property named
"useGenerativeDirectAnswers" that must be set to true in order
for GDA to work on the site
- When the universal results or vertical results change, that will
trigger a call into the GDA endpoint.

To accomplish the above, I have added two new storage keys - one
for the GDA itself, and one for the Search ID, which we need for
the request into GDA.

J=WAT-4593
TEST=auto, manual
Auto: Wrote new tests
Manual: Spun up local instance of the answers-search-ui + HH theme
and changed the vars to use a Prod Search config with GDA
configured. Saw the request being fired in the network tab when
the prop was set to true, and not fired when the prop was set to
false


package.json
```diff
@@ -63,6 +63,7 @@
     "@size-limit/file": "^4.9.0",
     "@types/jest": "^24.0.15",
     "autoprefixer": "^10.4.4",
+    "cheerio": "1.0.0-rc.12",
```
k-gerner, line 66: needed due to an `npm update` call causing this to be updated to an unsupported version for tests

vijay267: Ok sounds good

vijay267 (approved):
Anyways we answered this on slack about handling empty answers.

PR #1913, opened 2025-01-30
Add Result Component for Generative Direct Answers

k-gerner:
This change includes the logic necessary to visualize a GDA response in the UI via the answers-search-ui package. The component styled here (styled using the Handlebars template and custom CSS) will be used if there is not a custom card for GDA set (in the theme, we will create a nearly identical, but still different, card visual).

The actual *logic* for getting the data to the component is entirely present in this PR. This includes handling the different GDA search states (pre-search, in progress, complete) and  analytics events (citation clicks and rich text link clicks).

Here is a demo of it in action: https://drive.google.com/file/d/1KIxTRGBPCDvcUoIICT-jeF-swdXd7Pl6/view?usp=sharing

Some key functionalities to note from the demo video:
- An in-progress GDA call will show just the banner, with text "AI Generating Answer..."
- The GDA answer will be formatted into an HTML string to be used in the snippet (the response from the endpoint gives a markdown representation)
- If the answer has links, they will be formatted properly and should fire analytics events in the same way that regular direct answers do
- Citations will display at the bottom of the component in a scrollable list
  - Citations have fields name (req), description (opt), and link (opt).
  - If a citation has a description, a truncated version will be displayed in the citation card, below the name
  - If a citation has a link, the citation card will become clickable
  - Clicking on a citation will fire a CITATION_CLICK analytics event

J=[WAT-4594](https://yexttest.atlassian.net/browse/WAT-4594)
TEST=auto, manual
Auto: Wrote new tests
Manual: Spun up test-site with theme pointing to Red Dog Prod experience, and verified various states and edge cases


package.json
```diff
@@ -83,7 +83,7 @@
     "gulp-declare": "^0.3.0",
     "gulp-filter": "^7.0.0",
     "gulp-handlebars": "^5.0.2",
-    "gulp-postcss": "^8.0.0",
+    "gulp-postcss": "^9.0.0",
```
k-gerner, line 86: We needed this update in order to actually build the CSS in the compiled JavaScript

src/ui/components/results/generativedirectanswercomponent.js
```diff
@@ -0,0 +1,186 @@
...
+      customCard: this._getCard()
+    }));
+  }
+
+  /**
+   * Determines the card that should be used for GDA
+   * @returns {string}
+   */
+  _getCard () {
+    return this._customCard;
```
k-gerner, line 188: If, down the line, we add more fine-tuned card styling (e.g. "use X card for Y vertical, and B card for C vertical) then that logic will go here

src/ui/sass/modules/_GenerativeDirectAnswer.scss
```diff
@@ -0,0 +1,198 @@
...
+  {
+    @include Text(
+      $size: var(--yxt-gda-citation-name-font-size),
+      $line-height: var(--yxt-gda-citation-name-line-height),
+      $weight: var(--yxt-gda-citation-name-font-weight),
+      $color: var(--yxt-gda-citation-name-color)
+    );
+  }
+
+  &-citationDescription
```
k-gerner, line 165: Hypothetically to match the React SDK behavior (include 2 lines of description instead of 1), we could add this:
```
-webkit-line-clamp: 2;
-webkit-box-orient: vertical;
 display: -webkit-box;
```
However, in my testing, this didn't seem to work. Furthermore, it seems like this may not be supported in Firefox

src/ui/components/results/generativedirectanswercomponent.js
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @param {HTMLElement} citationElement The citation element that was clicked.
+   */
+  _handleCitationClickAnalytics (citationElement) {
+    if (!citationElement.hasAttribute('entityId')) {
+      console.error('No entityId found on citation element for analytics.');
+      return;
+    }
+    const entityId = citationElement.getAttribute('entityId');
+    const analyticsOptions = {
```
k-gerner, line 111: Also, not sure what options we want for this event type, but I figured this would be useful. Open to changing them if this is incorrect

Fondryext: yup, so this is item https://yexttest.atlassian.net/browse/WAT-4602. If you wanna go ahead and just do that fully here, that's fine by me. So basically add everything that makes sense referencing the directanswercomponent, but also add a new boolean property marking it as GDA. We can iron out exact phrasing in slack.

Fondryext (commented):
Looks good to me! All your comments make sense, not marking them as resolved though just so it's easy for other reviewers to find. LMK what you think about the analytics bit, otherwise I think this is GTG

src/ui/components/results/generativedirectanswercomponent.js
k-gerner: Added both to this PR. Also updated the way that the click type / entity id attributes are set on the HTML elements to more closely align with how it's done for other components (using `event.target.data`). This required a minor tweak in the `_handleCitationClickAnalytics()` method so that we can get the parent element if applicable.

tests/ui/components/results/generativedirectanswercomponent.js
```diff
@@ -0,0 +1,146 @@
...
+
+    // Citation count is correct
+    const citationsHeaderText = wrapper.find('.yxt-GenerativeDirectAnswer-citationsHeader').text().trim();
+    expect(citationsHeaderText).toEqual('Sources (2)');
+    const numCitations = wrapper.find('.yxt-GenerativeDirectAnswer-citationCard').length;
+    expect(numCitations).toEqual(2);
+
+    // Citation data is correct
+    const firstCitation = wrapper.find('.yxt-GenerativeDirectAnswer-citationCard').at(0);
+    console.log(firstCitation.debug());
```
Fondryext: nit: remove this log

PR #1916, opened 2025-02-19
Version 1.18.0

Fondryext:
### Changes

- GDA support (objects and analytics)
- VULN fix


vijay267 (approved):
Looks good to me. As we discussed, let's make a maintenance item for fixing the acceptance tests.

PR #1923, opened 2025-03-05
Include full entity data in GDA citations

k-gerner:
In this change we have the following changes:
- Pass raw data from entity result to GDA component (this will allow customization of what data is shown or used in the GDA component)
- No longer include links on the citations in the default GDA template. This should not be an issue as no one should be using the default template from this repo
J=[WAT-4680](https://yexttest.atlassian.net/browse/WAT-4680?atlOrigin=eyJpIjoiMmE2Y2M2MzU5MTY1NDcxZThkYWIyY2I1YzVmMjdkMzMiLCJwIjoiaiJ9)

### DIFF
--- a/src/core/models/generativedirectanswer.js
+++ b/src/core/models/generativedirectanswer.js
@@ -33,7 +33,7 @@ export default class GenerativeDirectAnswer {
           id: result.id,
           name: result.name,
           description: result.description,
-          link: result.link
+          rawData: result.rawData
         };
       });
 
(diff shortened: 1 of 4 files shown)


Fondryext (approved):
Remind me, we're removing the link because not all results will have the relevant link as the .link field?

k-gerner:
> Remind me, we're removing the link because not all results will have the relevant link as the .link field?

Yes, and the `link` field is not used in the theme anymore, so it's only purpose would be for this default `answers-search-ui` version of the component, which should never get used anyways.

PR #1924, opened 2025-03-17
Remove run shell injection vulnerability

mkouzel-yext:
Prevents attackers from injecting their own code into the github actions runner using variable interpolation to steal screts and code. We now use an intermediate environment variable to store input data.


Fondryext (approved):

k-gerner (changes requested):
I just merged develop into master, would you mind pulling and re uploading?

PR #1932, opened 2025-05-19
Fix search result vertical list reversal before GDA call

mkouzel-yext:
The list of each vertical of search results was being reversed before being sent to the backend for generative answers computation. This patch fixes an array reversal that was mutating the list of verticals when it should not have been.

J=TECHOPS-14115
TEST=manual

Served answers-search-ui bundle locally and spun up answers-hitchhiker-theme. Confirmed that search results render verticals in the correct order still and that the results we send to generative answers are also in the correct order.

### DIFF
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -482,8 +482,8 @@ export default class Component {
       return;
     }
 
-    childData.reverse();
-    childData.forEach(data => {
+    const reversedChildData = childData.slice().reverse();
+    reversedChildData.forEach(data => {
       this.addChild(data, type, opts);
     });
   }
--- a/src/ui/components/component.js
+++ b/src/ui/components/component.js
@@ -482,8 +482,7 @@ export default class Component {
       return;
     }
 
-    const reversedChildData = childData.slice().reverse();
-    reversedChildData.forEach(data => {
+    childData.slice().reverse().forEach(data => {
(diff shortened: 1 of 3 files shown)


anguyen-yext2 (commented):
Can you add a test description? I saw that the Component class is imported and used in other classes, so it would be good to verify that this change doesn't cause other issues.

src/ui/components/component.js
```diff
@@ -482,8 +482,8 @@ export default class Component {
       return;
     }
 
-    childData.reverse();
-    childData.forEach(data => {
+    const reversedChildData = childData.slice().reverse();
```
k-gerner: Out of curiosity why are we reversing this in the first place?

mkouzel-yext: Empirically, if we don't include this reversal then the result verticals will be rendered last -> first instead of first -> last. I expected that its due to the order child components get rendered in, corroborated by the description on [this old PR](https://github.com/yext/answers-search-ui/pull/183):
> Components render top-down recursively, so parents don't complete mounting until
all children finish

mkouzel-yext:
> Can you add a test description? I saw that the Component class is imported and used in other classes, so it would be good to verify that this change doesn't cause other issues.

Added testing info to original comment

anguyen-yext2 (approved):

PR #1935, opened 2025-06-06
ksearch: use linkTarget in gda cards

QuinnCiccoretti:
J=WAT-4879
TEST=manual

Spun up HH theme locally to use answers-search-ui, set the test-site/config/index.json not to override answers-search-ui. Added a custom linkTarget to same config, verified rendered html shows custom target attribute.
Removed linkTarget config and verified default behaviour.


src/ui/templates/results/generativedirectanswer.hbs
```diff
@@ -29,6 +29,7 @@
             data-entityid={{this.id}}
             data-eventtype="CITATION_CLICK"
             data-is-analytics-attached="true"
+            target="{{#if ../_config.linkTarget}}{{../_config.linkTarget}}{{else}}_blank{{/if}}"
```
k-gerner, line 32: Just curious why it needs the `../` before it

QuinnCiccoretti: handlebars #each creates a child scope, need _config on parent scope hence ../

QuinnCiccoretti: not necessary in directanswer because directanswers does not have #each

k-gerner: Ah TIL

k-gerner (approved):

PR #1952, opened 2025-08-11
Update click event listeners v1.15

mkouzel-yext:
Updates click event listeners in minor version 1.15


.github/workflows/deploy_hold.yml
```diff
@@ -28,8 +28,13 @@ jobs:
     environment: production # sets in github repo with reviewer requirement protection rule
     steps:
       - uses: actions/checkout@v3
+<<<<<<< HEAD
```
vijay267: This looks like an unresolved merge conflict?

mkouzel-yext: Fixed

vijay267 (approved):

mkouzel-yext:
Every change in this PR is present in the master branch. There is one functional change in this PR: updating the click events to work correctly. The other changes either (1) fix github workflow actions to be compatible with the modern version because the old one was deprecated or (2) fix unit and acceptance tests.

PR #1949, opened 2025-08-11
Update click event listeners v1.16

mkouzel-yext:
Update click event listeners to minor version 1.16

PR #1958, opened 2025-08-12
Fix i18n build artifact process for v1.16

mkouzel-yext:
This code was present in the most recent version, but I didn't realize it was necessary to include when I updated the script in #1949. 

Ultimately the reason we have to make this change is because the version of 'actions/[upload/download]-artifact' that we use in our GitHub workflow script was deprecated since the last time we had to deploy a new bundle for this minor version.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -15,7 +15,7 @@ jobs:
           cache: 'npm'
       - run: npm ci
       - name: Download build-output-US artifact
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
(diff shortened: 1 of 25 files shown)


mkouzel-yext:
Annoyingly, since I squashed and merged #1949, GitHub thinks that all the changes in that PR did not take effect and includes them in the diff here. The "[Include merge and overwrite steps when creating build output artifact](https://github.com/yext/answers-search-ui/pull/1958/commits/61b9d14811392698a9f7acd89732c8a024abda03)" commit is the only actual change.

k-gerner (approved):

PR #1960, opened 2025-08-14
Update click event listeners in v1.14

mkouzel-yext:
PR #1942 took a stab at this already, but ran into some issues in the acceptance tests. I decided to revert those changes and follow the same pattern I took when updating the listeners in versions 1.15, 1.16, and 1.17.


mkouzel-yext:
Given that this version is already 3 years old at time of writing, it's not worth the effort right now to resolve these acceptance test and CI/CD problems just to get this click event update out. Going to close this PR.

PR #1971, opened 2025-11-03
feat: support publishing with OIDC

Fondryext:
Adjusts the repo to use the OIDC-backed npm publish as defined in slapshot resuable workflows
Also removes outdated package upgrade script which we no longer use.

J=WAT-5096
TEST=manual

Ran with the dryRun flag, and it ran all tests and did the version selection flow as expected.


Fondryext:
I know I haven't tested yet, and happy to go ahead and do that. I think the process was making a forked repo and then running e.x. the dry-run to see if it worked as expected?
I also recall a thread of people asking Bowman to manually adding 'Trusted Publisher' for repos. Do we think that's still part of the flow/I should go hunt down that thread?

anguyen-yext2:
> I know I haven't tested yet, and happy to go ahead and do that. I think the process was making a forked repo and then running e.x. the dry-run to see if it worked as expected? 

If you run with the dryRun flag, i think you should be good to run it on this repo. Forking repo is more to test the actual release without the dryRun flag. 

> I also recall a thread of people asking Bowman to manually adding 'Trusted Publisher' for repos. Do we think that's still part of the flow/I should go hunt down that thread?

Here is the thread: https://yext.slack.com/archives/C09GQV5BXDY/p1760385166821179
I think the plan is to ask Bowman to set it up for the rest of our repos all at once when they're all done. I was only asking for modern search-sdks separately because we wanted to confirm the scripts work, plus we actually needed to publish new features. So if this repo needs to be published before we finish the oidc epic, feel free to ask Bowman in that thread. Otherwise I can ask him later. 

Also while you're at it, could you remove the version-update workflow? Thank you!

anguyen-yext2 (approved):

PR #1973, opened 2026-01-06
feat: support autocompleteLimit (v1.19.0)

k-gerner:
Supports the `limit` param for `/autocomplete` endpoint.

J=[WAT-5264](https://yext.atlassian.net/browse/WAT-5264?atlOrigin=eyJpIjoiOWZlZDdmZDAxNzFjNGJlMzgwM2Y0NWM1YjE4MzdmZmEiLCJwIjoiaiJ9)
TEST=manual
Tested with new HH theme and search-core


src/core/models/searchconfig.js
```diff
@@ -17,6 +17,12 @@ export default class SearchConfig {
      */
     this.universalLimit = config.universalLimit;
 
+    /**
+     * The max autocomplete results per search.
+     * @type {number}
+     */
+    this.autocompleteLimit = config.autocompleteLimit || 10;
```
Fondryext: Since the serving code handles if limit isn't passed-in I figure we can keep it optional here as well and remove the || 10, if we wanna.

k-gerner: Done!

Fondryext (approved):

PR #1976, opened 2026-01-08
chore: update package version to v1.19.0

k-gerner:
`slap trusted-publish` command can't push directly to `master` so doing this


k-gerner:
Was able to get it to work by unchecking the `Do not allow bypassing the above settings` setting for the `master` branch in the repo settings

PR #1978, opened 2026-01-14
ability to disable autocomplete (v1.19.1)

anguyen-yext2:
J=WAT-5279
TEST=manual

tested on HH Theme test site


src/ui/components/search/searchcomponent.js
```diff
@@ -584,7 +596,7 @@ export default class SearchComponent extends Component {
    * @returns {Promise} A promise that will perform the query and update storage accordingly.
    */
   promptForLocation (query) {
-    if (this._promptForLocation) {
+    if (!this._autocompleteDisabled && this._promptForLocation) {
```
Fondryext, line 599: Why're we not prompting for location if they've disabled autocomplete?

anguyen-yext2: because the only thing done in `promptForLocation` is `fetchQueryIntents`, which is computed based on autocomplete results. But please let me know if I'm misunderstanding something

Fondryext: Thanks, I did need to double-check the current logic, because it wasn't implemented how I expected. It surprised me that query intents in this repo is built such that we always check the autocomplete data to see what the user had queried. I know this wasn't a design choice any of us made, but it's a little unfortunate we'd have to give up the 'prompt for location' logic if someone turns off autocomplete (queryIntents are only used for promptForLocation, as far as I can tell)

I'm not saying we should necessarily rewrite it here, but it might at least be worth talking with Deepak to make sure it's alright to have disabling autocomplete also disable that functionality.

anguyen-yext2: Confirmed Deepak is ok with this ([thread](https://yext.slack.com/archives/GTTQTJH9B/p1769105802588019?thread_ts=1769098389.544329&cid=GTTQTJH9B))

k-gerner (dismissed):

k-gerner (approved):

PR #1981, opened 2026-01-23
Fix various workflows when releasing search bar

anguyen-yext2:
- Remove the extra `}` that cause mismatching tag name
- Skip publish on search bar tags


Fondryext (approved):
ty!

PR #1983, opened 2026-02-23
fix: remove autocomplete-related html when disabled

Fondryext:
Some parts of the hbs file should be removed if
autocomplete is disabled. This PR does that.

J=WAT-5402
TEST=manual

Ran both normal and search bar only test site, and saw UI updated for both


mkouzel-yext (dismissed):
LGTM

k-gerner (dismissed):

src/ui/templates/search/search.hbs
```diff
@@ -16,21 +16,25 @@
         <input class="js-yext-query yxt-SearchBar-input"
           id="{{inputIdName}}"
           type="text"
-          role="combobox"
+          {{#unless autocompleteDisabled}}
+            role="combobox"
```
Fondryext, line 20: It is true I fully removed role here. I did this because without autocomplete this is more of a textbox, and the [dev docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/textbox_role) say:

> ARIA: textbox role
The textbox role is used to identify an element that allows the input of free-form text. Whenever possible, rather than using this role, use an [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input) element with [type="text"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/text), for single-line input, or a [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea) element for multi-line input.

anguyen-yext2 (dismissed):

k-gerner (approved):

PR #1989, opened 2026-04-28
v2.0.0

Fondryext:
### Support Events API

This PR adds Events API (also known as Analytics v1.0) support to the SDK. 
The SDK now beacons its Analytics payloads to the new endpoint, and does stricter parsing of the payload
to reflect the validation performed in the new pipeline.
Users now need to set an eventsApiKey param on their config, similar to the apiKey, which has Events API
permissions, so this is a major change.
Conversion tracking is not supported in the Events API, and is therefore removed as part of this change.
J=WAT-4646
TEST=manual

Served locally and saw events fire as intended


src/core/analytics/analyticsreporter.js
```diff
@@ -111,36 +127,42 @@ export default class AnalyticsReporter {
...
-    }
+    event.addOptions(this._globalOptions);
+
+    const apiEvent = event.toApiEvent();
+    const eventPayload = {
+      authorization: 'KEY ' + this._eventsApiKey,
+      action: apiEvent.action,
+      ...(event.entityId !== undefined && { entity: event.entityId }),
+      ...(event.destinationUrl !== undefined && { destinationUrl: event.destinationUrl }),
+      ...(event.label !== undefined && { label: event.label }),
```
mkouzel-yext, line 148: In search-ui-react, we set `locale` [as well](https://github.com/yext/search-ui-react/blob/main/src/models/SearchAnalyticsEventServiceImpl.ts#L45), but we don't set `label`. Let's set both.

Also, can we alphabetize the order of the fields, at least on the top level event payload? Makes it easier to read

Fondryext: sure thing, done

src/answers-umd.js
```diff
@@ -272,6 +273,18 @@ class Answers {
         storageKey: StorageKeys.QUERY_ID,
         callback: id => this._analyticsReporterService.setQueryId(id)
       });
+      // listen to search id updates
+      storage.registerListener({
```
mkouzel-yext, line 277: Did we just not previously set the search ID/term on analytics events?

mkouzel-yext (approved):
LGTM, one small clarifying question unrelated to functionality. Would be nice to get a second set of eyes since it's a new major version, though.

src/answers-umd.js
Fondryext: Search id is relatively new, and yeah search term I just figured since it also changes by state we could stick a listener here.

src/core/analytics/analyticsreporter.js
```diff
@@ -62,11 +63,10 @@ export default class AnalyticsReporter {
     this._baseUrl = getAnalyticsUrl(this._environment, this._cloudChoice);
 
     /**
-     * Boolean indicating if opted in or out of conversion tracking
-     * @type {boolean}
-     * @private
+     * The key authorized for the Events API
+     * @type {string}
      */
-    this._conversionTrackingEnabled = false;
```
vijay267, line 69: I get that conversion tracking is not supported in the Events API, but can we also mention that in the commit message? Otherwise it's probably not obvious to folks involved as to why this is being removed.

src/core/analytics/analyticsreporter.js
```diff
@@ -111,36 +127,43 @@ export default class AnalyticsReporter {
...
+      ...(event.destinationUrl !== undefined && { destinationUrl: event.destinationUrl }),
+      ...(event.entityId !== undefined && { entity: event.entityId }),
+      ...(event.label !== undefined && { label: event.label }),
+      ...(event.locale !== undefined && { locale: event.locale }),
+      search: {
+        ...(event.directAnswer !== undefined && { isDirectAnswer: event.directAnswer }),
+        experienceKey: this._globalOptions.experienceKey,
+        ...(event.generativeDirectAnswer !== undefined &&
+            { isGenerativeDirectAnswer: event.generativeDirectAnswer }),
+        ...(includeQueryId && { queryId: this.getQueryId() }),
```
vijay267, line 155: When is "includeQueryId" false again? Is it when they don't want any tracking? Or is it when it's not correlated with a search?

src/ui/components/results/directanswercomponent.js
```diff
@@ -182,10 +182,8 @@ export default class DirectAnswerComponent extends Component {
     const analyticsOptions = {
       verticalKey: relatedItem.verticalConfigId,
       directAnswer: true,
-      fieldName: this.getState('answer').fieldApiName,
```
vijay267, line 185: Is there a reason we're no longer sending the api name? Feels useful to know which field this direct answer is being returned on.

src/core/analytics/analyticsreporter.js
Fondryext: Sure, done.

src/ui/components/results/directanswercomponent.js
Fondryext: It is no longer a field in the EventPayload supported by Events API: https://github.com/yext/analytics/blob/main/src/EventPayload.ts

src/core/analytics/analyticsreporter.js
Fondryext: Yeah, the latter. So like built-in, we use it for VOICE_START/STOP for example.

vijay267: Ok makes sense!

src/ui/components/results/directanswercomponent.js
vijay267: I wonder if we should readd "field id" as part of the event payload long term, but given we already updated search-ui-react like this, sounds good.

vijay267 (approved):

PR #1993, opened 2026-04-30
v2.0.1

Fondryext:
Fixes a small bug with analytics reporting: Events API does not accept empty searchTerms, so only set if there is a non-empty string saved.
J=WAT-4646


k-gerner (dismissed):

src/core/analytics/analyticsreporter.js
```diff
@@ -157,7 +158,7 @@ export default class AnalyticsReporter {
         versionLabel: this._globalOptions.experienceVersion,
         ...(event.verticalKey !== undefined && { verticalKey: event.verticalKey })
       },
-      searchTerm: this.getSearchTerm(),
+      ...(!!searchTerm && { searchTerm: searchTerm } ),
```
mkouzel-yext: For my own understanding, what was wrong with the first version?

mkouzel-yext (dismissed):
LGTM, small question

src/core/analytics/analyticsreporter.js
Fondryext: The Events API schema rejects any payloads where searchTerm is the empty string, something I didn't realize. This change only sends along searchTerm if it's not empty.

k-gerner (approved):

